PR #9, opened 2026-04-13
feat(ontology): add ontology models, loader, and tests

libei:
## Summary
- Add ontology Pydantic models and YAML loader under `src/ontology/`
- Add validation covering inheritance, key constraints, relationship endpoints, covariant narrowing, cardinality inheritance, and ontology-wide name uniqueness
- Add 23 unit tests under `tests/ontology/`

## Test plan
- [x] `pytest tests/ontology/ -v` — 23 passed


caohy1988:
fix the cla check and cli format check

src/ontology/models.py
```diff
@@ -0,0 +1,140 @@
...
+  keys: Optional[Keys] = None
+  from_: str = Field(alias="from")
+  to: str
+  cardinality: Optional[Cardinality] = None
+  properties: list[Property] = Field(default_factory=list)
+  description: Optional[str] = None
+  synonyms: Optional[list[str]] = None
+  annotations: Optional[dict[str, AnnotationValue]] = None
+
+  model_config = ConfigDict(extra="forbid", populate_by_name=True)
```
caohy1988: Bug: `model_config` is defined twice on `Relationship` — once here (with `populate_by_name=True`) and once at line 111 (without it). Python silently uses the last assignment, so the first one at line 111 is dead code.

Remove the first `model_config = ConfigDict(extra="forbid")` declaration at line 111.

src/ontology/loader.py
```diff
@@ -0,0 +1,331 @@
...
+    _check_relationship_endpoints(rel, entity_map)
+    _check_covariant_narrowing(rel, rel_map, entity_map)
+
+
+# --------------------------------------------------------------------- #
+# Individual checks                                                      #
+# --------------------------------------------------------------------- #
+
+
+def _check_unique_names(items, kind: str) -> dict:
```
caohy1988: Nit: Several internal helpers use bare `-> dict` or untyped `items` params. For a public SDK, these would benefit from proper annotations like `dict[str, Entity]`, `dict[str, Relationship]`, or a generic/union. Same applies to `_ancestors`, `_effective_properties`, etc.

src/ontology/loader.py
```diff
@@ -0,0 +1,331 @@
...
+  """Rules 7, 8, 9 for relationships."""
+  keys = _effective_keys(rel, rel_map)
+  if keys is None:
+    return  # no uniqueness constraint; multi-edges permitted.
+  if keys.primary and keys.additional:
+    raise ValueError(
+        f"Relationship {rel.name!r}: primary and additional are "
+        f"mutually exclusive."
+    )
+  if keys.additional is not None and keys.alternate:
```
caohy1988, line 316: This check is partially redundant with `_check_alternate_keys` (which already raises "alternate keys require a primary key" when `primary is None`). The only scenario this catches independently is `additional` present + `alternate` present + `primary` absent — which `_check_alternate_keys` also catches with a different message.

Not blocking, but a brief comment explaining this is an intentional "better error message for relationships" would help future readers.

caohy1988 (commented):
## PR Review

Well-structured implementation of the ontology spec. The code is clean, the validation logic maps clearly to the spec's §10 rules, and the test suite has solid coverage. A few issues to address before merge.

### Build / Packaging Issues

**`src/ontology` is not included in the wheel build**

`pyproject.toml` specifies:
```toml
[tool.hatch.build.targets.wheel]
packages = ["src/bigquery_agent_analytics"]
```

The new `src/ontology/` package won't be distributed. Either add `"src/ontology"` to the `packages` list, or move the ontology code under `src/bigquery_agent_analytics/ontology/`.

**Tests can't resolve the `ontology` import without `pythonpath` config**

The test imports `from ontology import load_ontology_from_string`, but `[tool.pytest.ini_options]` has no `pythonpath = ["src"]`. This means `pytest tests/ontology/ -v` would fail unless `PYTHONPATH` is set manually or the package is installed in editable mode. Add:

```toml
[tool.pytest.ini_options]
pythonpath = ["src"]
```

**Missing `tests/ontology/__init__.py`**

The repo has `tests/__init__.py` but this PR doesn't add `tests/ontology/__init__.py`. For consistency with the existing structure, it should be included.

### Spec Compliance Nit

**Spec says entity `keys` is "required" (§3), model says `Optional`**

The spec table in §3 marks `keys` as required on entities. The Pydantic model declares `keys: Optional[Keys] = None` on `Entity`, deferring enforcement to the loader's `_check_entity_keys`. This works because child entities inherit keys — but the existing comment on the field could call this out more explicitly to save readers a trip to the loader.

### What Looks Good

- Clean separation between shape validation (Pydantic `extra="forbid"`) and semantic validation (loader)
- Cycle detection is correct and handles transitive chains
- Covariant narrowing for relationship endpoints is well-implemented
- `coerce_numbers_to_str` on `Ontology` is a nice UX touch for `version: 0.1`
- The `from_` / `Field(alias="from")` pattern correctly handles Python's reserved keyword
- Full-text round-trip testing style is thorough (just be aware it's brittle to future serialization changes)

src/ontology/loader.py
```diff
@@ -0,0 +1,331 @@
...
+    has_inherited_keys = any(
+        _has_keys(a) for a in _ancestors(item.name, item_map)
+    )
+    if has_inherited_keys and _has_keys(item):
+      raise ValueError(
+          f"{kind.capitalize()} {item.name!r} redeclares inherited keys."
+      )
+
+
+def _has_keys(item) -> bool:
```
caohy1988: Bug: `_has_keys()` treats empty key lists as if the key block were absent because it relies on truthiness. That means invalid inputs like `keys: {primary: []}` or `keys: {additional: []}` load successfully, and a child can even bypass the inherited-key redeclaration check with `keys.primary: []`.

The spec models these as non-empty lists (`[<string>, ...]`), so this should fail validation rather than being treated as "no keys".

src/ontology/loader.py
```diff
@@ -0,0 +1,331 @@
...
+  """True if ``child`` equals ``parent`` or transitively extends it."""
+  cur: str | None = child
+  while cur is not None:
+    if cur == parent:
+      return True
+    cur = entity_map[cur].extends
+  return False
+
+
+def _check_covariant_narrowing(
```
caohy1988, line 355: Bug: relationship inheritance validation never enforces the spec's cardinality rule. §9 says relationship cardinality is "inherited unchanged," but this code only checks endpoint narrowing, so a child can extend a `many_to_many` parent and declare `one_to_one` here without any error.

Either reject cardinality changes during validation or explicitly inherit the parent cardinality when the child omits it; the current behavior accepts mismatches silently.

src/ontology/models.py
libei: Good catch!

libei:
Thanks for the thorough review — addressed in 90c970f and 9273fca. Summary:

**Build / packaging**
- Added `src/ontology` to `[tool.hatch.build.targets.wheel] packages`.
- Added `pythonpath = ["src"]` to `[tool.pytest.ini_options]`.
- Added `tests/ontology/__init__.py`.

**Inline comments**
- `models.py:124` — removed the duplicate `model_config` on `Relationship`; kept the one with `populate_by_name=True` and added a comment explaining why.
- `loader.py:102` — typed all the previously bare `dict` annotations: `_check_unique_names`, `_ancestors`, `_effective_properties`, `_effective_keys`, `_check_*_keys`, `_check_relationship_endpoints`, `_check_covariant_narrowing`, `_is_entity_subtype`, etc. Introduced a `GraphElement = Union[Entity, Relationship]` alias and a `GraphElementT` TypeVar so per-kind helpers preserve the concrete type.
- `loader.py:185` — fixed the empty-list bug by enforcing `min_length=1` on `Keys.primary` / `Keys.alternate` / `Keys.additional` at the Pydantic layer. `keys: {primary: []}` now fails at parse time. New test: `test_empty_primary_key_list_is_error`.
- `loader.py:281` — added a comment noting the relationship-specific `additional+alternate` check overlaps with `_check_alternate_keys` but emits a clearer message.
- `loader.py:315` — added cardinality-inheritance enforcement in `_check_covariant_narrowing`: a child relationship may omit cardinality (inherits silently) or restate the parent's value, but cannot override it. Two new tests: `test_relationship_cardinality_must_match_parent`, `test_relationship_child_may_omit_cardinality`.

**Spec compliance nit**
- The existing comment on `Entity.keys` already explains that the field is Optional at the model level because children inherit; reads more cleanly after the chapter-reference cleanup pass. Happy to expand further if you'd still like it more explicit.

**Other**
- Removed the accidental `uv.lock` commit and added it to `.gitignore` (9273fca).
- Renamed the internal type alias `Declarable` → `GraphElement` for clarity.
- Stripped section/rule references (§N, "Rule N:") from docstrings to avoid stale links if the spec is renumbered.
- Ran `isort` + `pyink` for the format check.

All 22 tests pass locally. CLA is the remaining external blocker — working on signing it.

caohy1988 (commented):
## Follow-up Review (post `90c970f7`, `9273fcad`)

All issues from my first review have been addressed. Nice work on the turnaround.

### Resolved

| Issue | Status |
|---|---|
| Duplicate `model_config` on `Relationship` | Fixed — single declaration with `populate_by_name=True` |
| `src/ontology` missing from wheel build | Fixed — added to `packages` in `pyproject.toml` |
| Missing `pythonpath = ["src"]` in pytest config | Fixed |
| Missing `tests/ontology/__init__.py` | Fixed — empty file added |
| Loose type annotations on internal helpers | Fixed — `GraphElement`/`GraphElementT` type system, fully typed dicts |
| Redundant check missing explanation | Fixed — comment added at the `additional+alternate` check |
| Empty key lists silently accepted | Fixed — `min_length=1` on `Keys` fields, new test |
| Cardinality inheritance not enforced | Fixed — new validation in `_check_covariant_narrowing`, two new tests |

### New Minor Items

**1. Cardinality check edge case: child declares cardinality when parent has `None`**

The new check rejects a child that declares *any* cardinality if the parent omitted it (since `one_to_one != None`). Whether this is correct depends on spec interpretation — "inherited unchanged" could mean "don't change what's there" (allow adding when parent has none) or "the value is frozen from the parent" (reject any deviation). The current strict interpretation is defensible, but worth a sentence in the docstring confirming this is intentional.

**2. PR description says 19 tests — actual count is 22**

Three tests were added (`test_empty_primary_key_list_is_error`, `test_relationship_cardinality_must_match_parent`, `test_relationship_child_may_omit_cardinality`). Minor — just update the description.

### Verdict

Looks good to merge once the minor items above are addressed (or explicitly waved off).

src/ontology/loader.py
```diff
@@ -0,0 +1,372 @@
...
+) -> None:
+  """Endpoints must reference declared entities."""
+  if rel.from_ not in entity_map:
+    raise ValueError(
+        f"Relationship {rel.name!r}: from {rel.from_!r} is not a "
+        f"declared entity."
+    )
+  if rel.to not in entity_map:
+    raise ValueError(
+        f"Relationship {rel.name!r}: to {rel.to!r} is not a declared entity."
```
caohy1988, line 339: Edge case to consider: if the parent has `cardinality: None` (omitted) and the child declares `cardinality: one_to_one`, this check rejects it because `one_to_one != None`. That's the strict reading of "inherited unchanged," but there's a reasonable argument that *adding* a cardinality when the parent had none isn't *changing* an inherited value — it's establishing one.

If the strict behavior is intentional, a brief comment here would help (`# A child may not introduce a cardinality that the parent did not declare`).

src/ontology/loader.py
```diff
@@ -0,0 +1,372 @@
...
+  _validate_ontology(ontology)
+  return ontology
+
+
+# --------------------------------------------------------------------- #
+# Validation                                                             #
+# --------------------------------------------------------------------- #
+
+
+def _validate_ontology(ont: Ontology) -> None:
```
caohy1988, line 77: Bug: name uniqueness is still enforced only within each kind, not across the whole ontology. Because entities and relationships are checked in two separate `_check_unique_names(...)` calls, an ontology with an entity `X` and a relationship `X` currently loads successfully.

That seems to conflict with the spec text in `ontology.md` (`name` is "unique in ontology" for both entities and relationships, and validation rule 1 says "Every entity and relationship name is unique within the ontology"). If cross-kind collisions are meant to be legal, the spec should be narrowed; otherwise this validator should reject them.

libei:
Follow-up review addressed in 3e8d97a:

- **`loader.py:330` — cardinality edge case**: kept the strict reading (a child cannot introduce a cardinality the parent didn't declare) and added a comment that says so explicitly, with a note on how to loosen later if needed.
- **`loader.py:77` — cross-kind name collision bug**: spec says names are "unique within the ontology" (rule 1), but the validator was only checking per-kind. Added a cross-kind collision check after the per-kind ones, plus a new test (`test_entity_and_relationship_cannot_share_name`).
- **PR description test count**: updated to 23.

All 23 tests pass.

PR #11, opened 2026-04-13
feat(ontology): add gm CLI with validate command

libei:
## Summary
- Add `src/ontology/cli.py` implementing the `gm` CLI per `docs/ontology/cli.md`
- Implements `gm validate` for ontology YAML files (ontology vs binding detection, `--json` structured errors, exit codes per spec §3)
- Wires up `gm` console script and includes `src/ontology` in the wheel build
- Adds `.gitignore` entry for `uv.lock`
- Adds 7 full-text CLI tests under `tests/ontology/test_cli.py`

Depends on GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK#9 (ontology models/loader). Merge that first.

## Test plan
- [x] `pytest tests/ontology/ -v` — 26 passed (19 models + 7 CLI)


haiyuan-eng-google:
fix the CI failure

caohy1988 (commented):
## PR Review

Clean CLI implementation that follows the spec closely. The error-reporting structure, exit codes, and file-kind detection all match `docs/ontology/cli.md`. A few issues to address.

### Carried Over from PR #9

This PR includes the full `models.py` / `loader.py` diff from PR #9. The issues I raised there still apply:
- **Duplicate `model_config`** on `Relationship` class (dead first declaration)
- **Missing `tests/ontology/__init__.py`**
- **Missing `pythonpath = ["src"]`** in `[tool.pytest.ini_options]`

Please address those in whichever PR lands first.

### CLI-Specific Issues

**`uv.lock` in `.gitignore` — intentional?**

Many projects commit lock files for reproducible installs (like `poetry.lock`). If this project uses `uv`, ignoring the lock file means CI and contributors won't get pinned dependency versions. Worth a deliberate decision rather than a drive-by `.gitignore` addition.

**Error line/col are always `0:0` except for YAML parse errors**

The spec (§3) defines the format `<file>:<line>:<col>: <rule> — <message>`, but `_collect_errors` sets `line: 0, col: 0` for all `ValueError` and `ValidationError` cases. This is a known limitation since the loader validates after parsing and discards position info — but it's worth a comment in the code or a note in the PR so future readers know this is expected, not a bug.

### What Looks Good

- Exit code mapping matches the spec (0/1/2/3) precisely
- `_detect_kind` cleanly separates ontology vs binding detection by top-level key
- Reserved commands (`compile`, `import-owl`) documented as exit-2 stubs — good for forward compatibility
- Typer `no_args_is_help=True` + `@app.callback()` for multi-command mode is the correct pattern
- Test coverage hits all exit code paths: success, validation error, malformed YAML, binding, unknown kind, missing file
- JSON structured output test verifies the full error shape

src/ontology/cli.py
```diff
@@ -0,0 +1,224 @@
...
+  2 — usage error (bad flag, missing file, unimplemented command)
+  3 — internal error
+"""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+import sys
+from typing import List, Optional
```
caohy1988: Nit: `Optional` is imported but unused in this file. Only `List` is used (in `_emit_errors` and `_collect_errors`). Also, since the project targets Python >=3.10, you could use the built-in `list[dict]` instead of `List[dict]` and drop the `typing` import entirely.

src/bigquery_ontology/cli.py
```diff
@@ -0,0 +1,224 @@
...
+# --------------------------------------------------------------------- #
+
+
+def _detect_kind(text: str) -> str:
+  """Return ``'ontology'``, ``'binding'``, or ``'unknown'``.
+
+  Raises ``yaml.YAMLError`` on parse failure so the caller can route it
+  through the ``yaml-parse`` error path.
+  """
+  data = yaml.safe_load(text)
```
caohy1988, line 144: The YAML gets parsed twice: once here in `_detect_kind` via `yaml.safe_load`, and again inside `load_ontology_from_string` (loader.py line 52). For small files this is negligible, but you could avoid it by having `_detect_kind` return `(kind, parsed_data)` and passing the dict directly to a variant of the loader that skips re-parsing.

Not blocking for v0, but worth a `# TODO` if you plan to support larger ontologies.

tests/bigquery_ontology/test_cli.py
```diff
@@ -0,0 +1,204 @@
...
+
+from pathlib import Path
+import textwrap
+
+from typer.testing import CliRunner
+
+from ontology.cli import app
+
+
+_RUNNER = CliRunner()
```
caohy1988, line 40: The CLI writes errors to stderr (`typer.echo(..., err=True)`), but the tests assert on `result.output`. This works because Click's `CliRunner` defaults to `mix_stderr=True`, which merges stderr into `result.output`.

This coupling is fragile — if someone later initializes CliRunner with `mix_stderr=False`, all error-assertion tests would break silently (they'd pass with empty output). Consider either:
- Adding a comment here documenting the `mix_stderr=True` dependency, or
- Using `CliRunner(mix_stderr=False)` and asserting on `result.stderr` for error tests

pyproject.toml
```diff
@@ -52,9 +52,10 @@ all = [
 
 [project.scripts]
 bq-agent-sdk = "bigquery_agent_analytics.cli:main"
+gm = "ontology.cli:main"
```
caohy1988: The package name `ontology` is very generic and likely to conflict with other PyPI packages. Since this is a console script entry point that gets installed globally (or in a venv), users could hit import collisions if another package also provides an `ontology` module.

Consider namespacing under the existing package: `gm = "bigquery_agent_analytics.ontology.cli:main"` (with the code moved accordingly), or at minimum `gm_ontology` as the package name.

src/ontology/cli.py
```diff
@@ -0,0 +1,224 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""``gm`` command-line interface (v0).
+
+Implements ``docs/ontology/cli.md``. Only ``gm validate`` is functional in
```
caohy1988: Bug: this docstring promises `gm compile` and `gm import-owl` as reserved subcommands that exit 2 until implemented, but this module never registers either command. Right now `gm compile ...` and `gm import-owl ...` fall through to Typer's generic "No such command" path instead of a deliberate stub, and they do not appear in `gm --help` despite being part of the documented v0 surface.

Either add explicit placeholder commands that raise `typer.Exit(2)` with the intended message, or scale back the docs/help text for this revision.

src/bigquery_ontology/cli.py
```diff
@@ -0,0 +1,224 @@
...
+) -> None:
+  """Validate a single ontology or binding YAML file."""
+  text = file.read_text(encoding="utf-8")
+  try:
+    kind = _detect_kind(text)
+  except yaml.YAMLError as exc:
+    _emit_errors(_collect_errors(str(file), exc), as_json=json_output)
+    raise typer.Exit(code=1)
+
+  if kind == "binding":
```
caohy1988, line 198: Bug: this turns every binding file into exit code 2 / `cli-unimplemented`, but `docs/ontology/cli.md` says `gm validate <binding>` is part of both documented workflows and should validate a binding spec (with companion-ontology lookup for cross-validation when available).

If binding validation is intentionally deferred for now, the command help and spec need to say that explicitly. As written, the implementation contradicts the documented contract for `gm validate`.

src/bigquery_ontology/cli.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+# --------------------------------------------------------------------- #
+# gm validate                                                            #
+# --------------------------------------------------------------------- #
+
+
+@app.command("validate")
+def validate(
+    file: Path = typer.Argument(
```
caohy1988, line 164: Bug: `exists=True` makes Click reject missing paths before `validate()` runs, so `gm validate missing.yaml --json` still emits Typer's human usage text instead of the structured JSON error format promised by the CLI spec for diagnostics.

If `--json` is supposed to apply uniformly to validation/usage failures, this argument needs to be validated inside the command instead of in the Typer parameter declaration.

libei:
Thanks for the review. Rebased onto the updated `feat/ontology-models` so the "carried over from PR #9" comments are no longer in this diff (duplicate `model_config`, `tests/ontology/__init__.py`, `pythonpath`, and the wheel-packages addition all live in PR #9 now).

CLI-specific comments addressed in f0ade34:

- **`cli.py:35`** — dropped unused `Optional`; replaced `List[dict]` with built-in `list[dict]` and removed the `typing` import entirely.
- **`cli.py:17`** — rewrote the docstring and Typer help to scope this revision to `validate` only; no longer claims `compile`/`import-owl` as reserved subcommands.
- **`cli.py:136`** — added a `TODO` next to `_detect_kind`'s `yaml.safe_load` noting the double-parse and the `load_ontology_from_dict` variant as the fix.
- **`cli.py:153`** — removed `exists=True` from the Typer argument and validate inside the command instead, so `gm validate missing.yaml --json` now emits a structured `cli-missing-file` JSON object on stderr (added a test for it).
- **`cli.py:174`** — renamed the rule from `cli-unimplemented` to `cli-binding-deferred` and rewrote the message to "not yet implemented; only ontology files are supported in this revision," with a comment pointing to `cli.md` and `binding.md` so the next person knows to update both when binding lands.
- **`test_cli.py:35`** — added an explanatory comment above the `CliRunner()` construction documenting the `mix_stderr=True` dependency and what to switch to if it ever flips.
- **`pyproject.toml:55`** — package rename (`ontology` → something namespaced) deferred for a separate change; flagged but skipping in this PR.

**Open question on `uv.lock`** — currently in `.gitignore`. We added it as a drive-by because the file was generated by a one-off test run, not by an authoritative `uv` workflow. Happy to commit it instead if the project does want pinned lockfiles in CI; let me know which way you want it.

All 30 tests pass locally (22 models + 8 CLI). CLA still pending.

caohy1988 (commented):
## Follow-up Review (post `375abfab`)

All issues from my first review have been addressed. The PR is also much cleaner now — properly scoped to just the CLI delta (3 files) on top of PR #9.

### Resolved

| Issue | Status |
|---|---|
| Unused `Optional` import | Fixed — now uses built-in `list[dict]`, no `typing` import |
| Double YAML parse | Addressed — `TODO` comment added in `_detect_kind` (acceptable for v0) |
| CliRunner `mix_stderr` coupling | Fixed — detailed comment added above `_RUNNER` |
| Binding exit as "unimplemented" | Fixed — changed to `cli-binding-deferred` with clear message and comment explaining when it will be wired up |
| `exists=True` blocking `--json` on missing files | Fixed — validation moved inside command body; two new tests cover human and JSON output for missing files |
| Reserved stub commands undeclared | Fixed — docstring and help text now accurately scoped to "Currently supports: validate" |
| `uv.lock` gitignore | Moved to PR #9 where it belongs |

### Still Outstanding

**Generic `ontology` package name** — `gm = "ontology.cli:main"` still uses the bare `ontology` namespace. This is the one item from my first review that wasn't addressed. If this is intentional (e.g., the package will be renamed or always installed in project-specific venvs), a brief comment in the PR would close the loop.

### One Minor Nit

**PR description says 7 tests — actual count is 9.** Two tests were added for missing-file structured errors. Minor — just update the description.

### Verdict

Looks good to merge (after PR #9 lands). The package-name question is the only open design item — not blocking if it's a deliberate choice.

libei:
Renamed the package from `ontology` → `bigquery_ontology` in 8fc82a1.

**Why not `bigquery_agent_analytics.ontology`** (the namespacing option you suggested): the ontology spec/loader is intended to be reusable beyond the agent-analytics SDK and may be extracted into its own project later. Embedding it under `bigquery_agent_analytics` would force every downstream user to rewrite imports on extraction day. Keeping it as a standalone top-level package preserves the import path across an eventual split — `from bigquery_ontology import …` works the same whether it ships in this wheel or its own.

**Why `bigquery_ontology`** specifically: distinctive enough to avoid the PyPI collision risk you raised, while signaling the BigQuery-first focus today. The ontology layer itself is backend-neutral (the spec also targets Spanner) and we may broaden the name later if more backends land — but for v0 this matches the project branding without locking us out of the future split.

Changes:
- `src/ontology/` → `src/bigquery_ontology/`
- `tests/ontology/` → `tests/bigquery_ontology/`
- `pyproject.toml`: `gm` script entry → `bigquery_ontology.cli:main`; wheel `packages` updated
- Test imports updated to `from bigquery_ontology import …`

All 31 tests pass.

PR #17, opened 2026-04-14
feat(ontology): add binding loader with cross-ontology validation

libei:
Stacks on #16. Review this PR's diff against \`feat/binding-models\` to isolate the loader changes.

## Summary
Adds \`load_binding\` and \`load_binding_from_string\`, paralleling the ontology loader. Pairs a binding YAML with an injected \`Ontology\` (or auto-discovers a companion \`<name>.ontology.yaml\` next to the binding file) and enforces the cross-ontology rules the model layer cannot see.

The governing mental model documented across the module/class/test docstrings is **partial at the ontology level, total within each element**:

- You may leave whole entities or relationships out of the binding; absent elements are simply not realized on this target.
- Whatever you *do* include must bind every one of its non-derived properties (inherited ones included) and must not bind any derived (\`expr:\`) property.

## Rules enforced
- Binding's declared \`ontology:\` name matches the paired \`Ontology\` object.
- Entity and relationship binding names are unique within the binding and each resolves to a declared ontology element.
- Total coverage within each included element (catches: missing non-derived property, derived property bound, duplicate binding, unknown property name).
- Relationship \`from_columns\` / \`to_columns\` arity matches the endpoint entity's primary-key arity.
- Each included relationship has ≥1 bound descendant at each endpoint (prevents dangling edges into unbound entity trees).

## Also
- Tightens \`binding_models.py\` with \`min_length=1\` on \`binding\`, \`ontology\`, \`PropertyBinding.name\`, and \`PropertyBinding.column\` (rule 1, and the "non-empty column" half of rule 7 — pure shape).
- Reuses \`_effective_properties\`, \`_effective_keys\`, \`_is_entity_subtype\` from the ontology loader via package-private imports.

## Intentionally out of scope
- Spanner-specific type-compat rules (this loader is BigQuery-only).
- CLI wiring — the deferred \`gm validate <binding>\` branch will be a follow-up PR so the CLI diff stays reviewable on its own.

## Test plan
- [x] \`uv run --with pytest python -m pytest tests/bigquery_ontology -q\` — 62 passed
- [x] Full happy path (\`finance\` example with inheritance, derived property, single-column edge)
- [x] Partial binding: omit unrelated entities and relationships
- [x] Abstract-parent endpoint satisfied by a bound concrete child
- [x] Ontology-name mismatch between binding and paired \`Ontology\`
- [x] Duplicate entity / relationship binding names
- [x] Binding names that don't resolve in the ontology
- [x] Coverage failures: missing non-derived property, derived property bound, unknown property, double-binding
- [x] Arity mismatch on both \`from_columns\` and \`to_columns\`
- [x] Relationship included without a bound endpoint descendant
- [x] File-based: companion-ontology auto-discovery, missing-companion \`FileNotFoundError\`, explicit \`ontology=\` override

### DIFF
--- a/src/bigquery_ontology/binding_loader.py
+++ b/src/bigquery_ontology/binding_loader.py
@@ -1,0 +1,379 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Loader and validator for binding YAML.
+
+Shape validation (required fields, unknown keys, enum membership, list
+min-length) lives in ``binding_models`` and runs at pydantic parse
+time. Semantic validation — everything that requires consulting the
+referenced ontology — lives here. The two halves together cover the
+binding spec end-to-end.
+
+The governing mental model is **partial at the ontology level, total
+within each element**:
+
+  - You may leave whole entities or whole relationships out of the
+    binding. Anything absent is simply not realized on this target.
+  - But once you include an entity or relationship, you must bind
+    every one of its non-derived properties (including inherited
+    ones) — no cherry-picking. Derived (``expr:``) properties are the
+    mirror image and must *never* appear.
+
+Rules enforced by ``_validate_binding``:
+
+  - The binding's declared ontology name matches the injected
+    ``Ontology`` object's name.
+  - Entity and relationship binding names are unique within the
+    binding and each resolves to a declared element in the ontology.
+  - Total coverage within each included entity/relationship, per the
+    model above.
+  - Each included relationship's ``from_columns`` / ``to_columns``
+    arity matches the endpoint entity's primary-key arity.
+  - Each included relationship's endpoints each have at least one
+    bound descendant in the binding — an edge that points at an
+    entity tree with no bound node would dangle.
+
+Spanner-specific type checks from the spec are intentionally skipped —
+this loader is BigQuery-only, which supports every logical type.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Iterable, Optional
+
+import yaml
+
+from .binding_models import Binding
+from .binding_models import EntityBinding
+from .binding_models import PropertyBinding
+from .binding_models import RelationshipBinding
+from .loader import _effective_keys
+from .loader import _effective_properties
+from .loader import _is_entity_subtype
+from .loader import load_ontology
+from .ontology_models import Entity
+from .ontology_models import Ontology
+from .ontology_models import Property
+from .ontology_models import Relationship
+
+# --------------------------------------------------------------------- #
+# Public entry points                                                    #
+# --------------------------------------------------------------------- #
+
+
+def load_binding(
+    path: str | Path, *, ontology: Optional[Ontology] = None
+) -> Binding:
+  """Load and validate a binding from a YAML file.
+
+  If ``ontology`` is not supplied, the loader reads the binding's
+  top-level ``ontology:`` key and looks for ``<name>.ontology.yaml`` in
+  the same directory as the binding file. Supply ``ontology``
+  explicitly to override that lookup, or to share a single parsed
+  ontology across many bindings.
+
+  Raises:
+      FileNotFoundError: The binding file, or its auto-discovered
+          companion ontology file, does not exist.
+      ValueError: Any semantic validation failure.
+      pydantic.ValidationError: Shape failures (unknown keys, bad
+          enums, missing required fields).
+      yaml.YAMLError: Malformed YAML in either file.
+  """
+  binding_path = Path(path)
+  text = binding_path.read_text(encoding="utf-8")
+
+  if ontology is None:
+    ontology = _discover_ontology(text, binding_path)
+
+  return load_binding_from_string(text, ontology=ontology)
+
+
+def load_binding_from_string(
+    yaml_string: str, *, ontology: Ontology
+) -> Binding:
+  """Parse and validate a binding from a YAML string.
+
+  Unlike :func:`load_binding`, the ontology must be supplied
+  explicitly here — there is no file context from which to discover
+  one.
+  """
+  data = yaml.safe_load(yaml_string)
+  if not isinstance(data, dict):
+    raise ValueError("Binding document must be a YAML mapping.")
+  binding = Binding(**data)
+  _validate_binding(binding, ontology)
+  return binding
+
+
+# --------------------------------------------------------------------- #
+# Companion-ontology discovery                                           #
+# --------------------------------------------------------------------- #
+
+
+def _discover_ontology(binding_text: str, binding_path: Path) -> Ontology:
+  """Locate and load ``<ontology>.ontology.yaml`` next to the binding.
+
+  Peeks at the binding YAML purely to pull out the ``ontology:`` name;
+  any structural errors here are swallowed so that the richer pydantic
+  error from ``Binding(**data)`` surfaces instead.
+  """
+  try:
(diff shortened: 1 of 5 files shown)


src/bigquery_ontology/binding_loader.py
```diff
@@ -0,0 +1,379 @@
...
+
+import yaml
+
+from .binding_models import Binding
+from .binding_models import EntityBinding
+from .binding_models import PropertyBinding
+from .binding_models import RelationshipBinding
+from .loader import _effective_keys
+from .loader import _effective_properties
+from .loader import _is_entity_subtype
```
caohy1988, line 64: These three functions are private-by-convention (`_`-prefixed) in `loader.py`, meaning the ontology loader considers them an internal implementation detail. Importing them cross-module creates a coupling that the naming contract doesn't advertise.

For v0 this is fine — the alternative is duplicating non-trivial graph-walking logic. But if a third consumer appears (e.g., the compiler), consider promoting these to a shared internal module (e.g., `_graph_helpers.py`) or dropping the underscore to signal they're internal-but-stable.

src/bigquery_ontology/binding_loader.py
```diff
@@ -0,0 +1,379 @@
...
+def _discover_ontology(binding_text: str, binding_path: Path) -> Ontology:
+  """Locate and load ``<ontology>.ontology.yaml`` next to the binding.
+
+  Peeks at the binding YAML purely to pull out the ``ontology:`` name;
+  any structural errors here are swallowed so that the richer pydantic
+  error from ``Binding(**data)`` surfaces instead.
+  """
+  try:
+    data = yaml.safe_load(binding_text)
+  except yaml.YAMLError:
```
caohy1988: Nit: `try: yaml.safe_load(binding_text) except yaml.YAMLError: raise` is functionally a no-op — the exception propagates identically without the try/except. The comment says "let `load_binding_from_string` surface the YAML error," but the error actually surfaces here in `_discover_ontology` (before `load_binding_from_string` runs), since the call path is `load_binding` → `_discover_ontology` → raises → caller.

The comment is useful as documentation of intent. The try/except/raise just adds ceremony around it.

caohy1988 (commented):
## PR Review

Excellent work. The binding loader is clean, correct, and thoroughly tested. The "partial at the ontology level, total within each element" mental model is clearly articulated across docstrings and test comments — this is some of the best-documented validation code I've seen in these PRs.

### Issues

**1. Cross-module import of private (`_`-prefixed) functions**

The loader imports `_effective_keys`, `_effective_properties`, and `_is_entity_subtype` from `loader.py`. These are private by Python convention, meaning the ontology loader is free to rename or refactor them without considering downstream callers. Since these are now part of a cross-module contract, consider either:
- Dropping the underscore prefix to signal they're internal-but-stable, or
- Extracting them into a shared `_graph_helpers.py` module

Not blocking — the current approach works and avoids duplication. But if more loaders land (e.g., compilation), this coupling will multiply.

### Non-blocking Observations

**2. `_discover_ontology` try/except/raise is a no-op**

The `try: yaml.safe_load(...) except yaml.YAMLError: raise` block is functionally identical to calling `yaml.safe_load` without the try/except. The comment explains the intent ("let `load_binding_from_string` surface the error"), but actually the error surfaces right here in `_discover_ontology` before `load_binding_from_string` ever runs. Consider removing the try/except and keeping just the comment, or leaving it as forward-defensive documentation.

**3. Minor style: missing blank line in test file**

Only one blank line between `_load` (line ~110) and `_assert_value_error` (line ~111) — PEP 8 wants two between top-level function definitions. May or may not be caught by pyink depending on config.

### What Looks Good

- `_check_property_coverage` is elegant — all four property-level failure modes (unknown, derived, duplicate, missing) checked in a single pass with clear error messages
- `_has_bound_descendant` correctly handles abstract-parent endpoints via `_is_entity_subtype` traversal
- `_primary_key_len` has a defensive invariant check for ontologies that shouldn't have passed validation — belt and suspenders
- `_validate_binding` runs checks in a logical dependency order (name match → resolution → coverage → arity → closure)
- Test fixture (`ONTOLOGY`) is deliberately shaped to cover all dimensions the loader cares about, and the docstring explains why each entity/relationship exists
- 18 tests cover every validation rule plus file-based entry points (auto-discovery, missing companion, explicit override)
- Test comments explain *why* each test exists, not just *what* it tests
- `min_length=1` tightening on `binding_models.py` is well-placed in this PR since the loader depends on it

### Verdict

Looks good to merge (after PR #16 lands). The private-function import is the only item worth discussing — everything else is clean.

src/bigquery_ontology/binding_models.py
```diff
@@ -99,17 +107,22 @@ class EntityBinding(BaseModel):
...
-  # An empty list is accepted so this shape model stays ontology-free;
-  # whether an empty list is *correct* depends on how many non-derived
-  # properties the bound entity declares, which only the loader can
-  # answer. A strict non-empty check here would false-positive on
-  # entities that legitimately have no non-derived properties.
+  # Whether a ``properties: []`` list is *correct* depends on how many
+  # non-derived properties the entity declares — a question this shape
+  # model cannot answer. Accept any list length here and let the loader
+  # enforce all-or-nothing coverage once it has the ontology in hand.
   properties: list[PropertyBinding] = Field(default_factory=list)
```
caohy1988: Bug: the binding spec still marks `EntityBinding.properties` as required, but this model makes the field optional via `default_factory=list`. On the current PR head, an entity binding with just `{name, source}` loads successfully and round-trips without any `properties` field at all. Allowing an explicit empty list for ontology-free shape validation is reasonable, but that is different from allowing the field to be omitted entirely.

src/bigquery_ontology/binding_loader.py
```diff
@@ -0,0 +1,379 @@
...
+
+def _check_unique_binding_names(binding: Binding) -> None:
+  """Entity and relationship binding names must each be unique.
+
+  An entity and a relationship sharing a binding-level name cannot
+  collide — the ontology already rejects that case across its own
+  namespace, and we validate against the ontology below — so we only
+  check within each kind.
+  """
+  _assert_unique((eb.name for eb in binding.entities), "entity binding")
```
caohy1988, line 199: Bug: this only enforces uniqueness within each kind. The binding specs rule 6 says "No duplicate entity or relationship names within the binding," which reads as a single binding-wide namespace. As written, a binding can include an entity binding `X` and a relationship binding `X` with no error. If cross-kind collisions are meant to be legal, the spec needs to be narrowed; otherwise this check should reject them explicitly.

caohy1988:
Review finding against latest `main` that is still present on this PR head but is not commentable inline here because `binding_models.py` is outside the visible diff of the stacked PR:\n\nMedium: the binding shape is still hard-coded to BigQuery-only in `src/bigquery_ontology/binding_models.py` even though `docs/ontology/binding.md` still models `target.backend` as `bigquery | spanner`. A valid Spanner-shaped binding fails at parse time instead of making it through shape validation. If Spanner targets are intentionally deferred, the spec needs to be narrowed in the same change.

src/bigquery_ontology/binding_loader.py
libei: Agreed — leaving as-is for v0 since it avoids duplicating the graph-walking logic. Will extract to a shared `_graph_helpers.py` if a third consumer (e.g. compiler) appears.

src/bigquery_ontology/binding_loader.py
libei: Good catch — removed the try/except, just calling `yaml.safe_load` directly now. The docstring already explains the intent.

src/bigquery_ontology/binding_models.py
libei: Fixed — restored `properties` to required (no default). Also restored the `test_entity_missing_properties_is_error` test that was dropped when the field was made optional.

src/bigquery_ontology/binding_loader.py
libei: Added a defensive cross-kind check after the within-kind checks. The ontology loader already prevents this (loader.py:81–89), so it's belt-and-suspenders — but the error message is clearer than the downstream name-resolution failure. Added a unit test that calls `_check_unique_binding_names` directly with a hand-crafted `Binding` to exercise it.

libei:
Re: Spanner in the spec — the Spanner additions in `binding.md` were reverted. The spec now matches the code (BigQuery-only), consistent with the narrowing from PR #16.

PR #18, opened 2026-04-14
feat(cli): wire gm validate for binding files

libei:
Stacks on #17 (which stacks on #16). Review this PR's diff against \`feat/binding-loader\` to isolate the CLI changes.

## Summary
Replaces the "binding validation deferred" stub with a real dispatch to the binding loader. Users can now run \`gm validate path/to/foo.binding.yaml\` and get actionable errors end-to-end.

## How companion-ontology resolution works
Two paths, both landing at the same validation:
- **Auto-discovery (default):** peek the binding for \`ontology: <name>\`, look for \`<name>.ontology.yaml\` next to the binding file.
- **Explicit override:** \`--ontology PATH\` — skips discovery entirely, useful for CI or cross-directory layouts.

## Error routing
Every error names the file that actually contains the mistake:

| Scenario                               | file                 | rule                    | exit |
|----------------------------------------|----------------------|-------------------------|------|
| Auto-discovery: companion missing      | binding              | \`cli-missing-ontology\`  | 2    |
| \`--ontology\`: path missing             | ontology path        | \`cli-missing-ontology\`  | 2    |
| Ontology file present but invalid      | **ontology path**    | **\`ontology-validation\`** | 1    |
| Binding has its own semantic error     | binding              | \`binding-validation\`    | 1    |
| Binding is malformed YAML              | binding              | \`yaml-parse\`            | 1    |

The third row is the critical one — it required refactoring the CLI to load the ontology itself rather than letting \`load_binding\` do discovery internally. Otherwise an ontology-side error would bubble through \`load_binding\` as a generic \`ValueError\` and get tagged \`binding-validation\` with \`file=<binding>\`, sending users grep'ing the wrong file.

## Also
- Generalized \`_collect_errors\` to accept a \`kind\` parameter so shape and semantic error rules carry an \`ontology\` or \`binding\` prefix. \`yaml-parse\` stays kind-agnostic.
- Added \`_peek_companion_path\` / \`_peek_ontology_name\` helpers for the CLI-side ontology-name discovery.

## Test plan
- [x] \`uv run --with pytest python -m pytest tests/bigquery_ontology -q\` — 68 passed
- [x] Auto-discovery happy path
- [x] Explicit \`--ontology\` happy path (ontology deliberately stashed where auto-discovery can't find it)
- [x] Auto-discovery missing-companion error
- [x] Explicit flag pointing at a nonexistent ontology file
- [x] Binding with a semantic error (coverage rule violation)
- [x] **Regression guard**: binding is valid, ontology is broken — error must land on the ontology file with \`rule=ontology-validation\`
- [x] JSON output preserves \`binding-validation\` rule prefix

### DIFF
--- a/src/bigquery_ontology/cli.py
+++ b/src/bigquery_ontology/cli.py
@@ -12,18 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""``gm`` command-line interface (v0).
+"""``gm`` command-line interface.
 
-Implements the surface described in ``docs/ontology/cli.md``. Only
-``gm validate`` is wired up in this revision; the ``gm compile`` and
-``gm import-owl`` commands referenced by the spec will be added when
-the binding and OWL-import modules land.
+``gm validate`` accepts either an ontology YAML or a binding YAML and
+dispatches to the matching loader. For binding files, the companion
+ontology is auto-discovered as ``<name>.ontology.yaml`` next to the
+binding by default; pass ``--ontology PATH`` to point at a specific
+ontology file instead. The ``gm compile`` and ``gm import-owl`` commands
+referenced elsewhere will be wired up when their modules land.
 
 Exit codes:
 
   0 — success
   1 — validation / compilation error
-  2 — usage error (bad flag, missing file)
+  2 — usage error (bad flag, missing file, missing companion ontology)
   3 — internal error
 """
 
@@ -37,6 +39,9 @@ from pydantic import ValidationError
 import typer
 import yaml
 
+from .binding_loader import load_binding
+from .binding_loader import load_binding_from_string
+from .loader import load_ontology
 from .loader import load_ontology_from_string
 
 app = typer.Typer(
@@ -78,8 +83,16 @@ def _emit_errors(
 def _collect_errors(
     file: str,
     exc: BaseException,
+    *,
+    kind: str,
 ) -> list[dict]:
-  """Convert an exception raised during loading into structured errors."""
+  """Convert an exception raised during loading into structured errors.
+
+  ``kind`` is either ``"ontology"`` or ``"binding"`` and is used purely
+  to tag the ``rule`` field on shape and semantic errors so downstream
+  tooling can tell which validator produced them. YAML-parse errors
+  share a single ``yaml-parse`` rule regardless of kind.
+  """
   if isinstance(exc, ValidationError):
     out: list[dict] = []
     for err in exc.errors():
@@ -89,7 +102,7 @@ def _collect_errors(
               "file": file,
               "line": 0,
               "col": 0,
-              "rule": f"ontology-shape:{err.get('type', 'invalid')}",
+              "rule": f"{kind}-shape:{err.get('type', 'invalid')}",
               "severity": "error",
               "message": f"{loc}: {err.get('msg', '')}",
           }
@@ -119,7 +132,7 @@ def _collect_errors(
           "file": file,
           "line": 0,
           "col": 0,
-          "rule": "ontology-validation",
+          "rule": f"{kind}-validation",
           "severity": "error",
           "message": str(exc),
       }
@@ -170,6 +183,14 @@ def validate(
         "--json",
         help="Emit structured JSON errors on stderr.",
     ),
+    ontology_path: Path = typer.Option(
+        None,
+        "--ontology",
+        help=(
+            "For binding files: path to the companion ontology YAML. "
+            "Defaults to <ontology>.ontology.yaml next to the binding."
+        ),
+    ),
 ) -> None:
   """Validate a single ontology or binding YAML file."""
   if not file.exists() or not file.is_file():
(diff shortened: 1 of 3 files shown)


caohy1988 (commented):
## PR Review

This is the best-architected PR in the stack. The key insight — having the CLI resolve the ontology itself rather than delegating to `load_binding`'s auto-discovery — solves a real usability problem: without it, an ontology-side error would be misattributed to the binding file, sending users grep'ing the wrong file. The error routing table in the PR description is excellent documentation.

### Non-blocking Observations

**1. YAML parsed up to 4 times for binding files**

A binding validation passes through: `_detect_kind` (parse 1), `_peek_ontology_name` via `_peek_companion_path` (parse 2), the error message path calls `_peek_ontology_name(text)` again (parse 3), and finally `load_binding_from_string` → `yaml.safe_load` (parse 4). For hand-authored YAML files this is negligible, but the third parse is particularly easy to avoid — cache the result of `_peek_ontology_name` in a local variable and reuse it for the error message.

**2. `kind="ontology"` on pre-detection YAML error is harmless but misleading**

When `_detect_kind` raises `yaml.YAMLError`, the CLI passes `kind="ontology"` to `_collect_errors`. The file could be a malformed binding. This is harmless because `_collect_errors` uses the hard-coded `"yaml-parse"` rule for YAML errors (ignoring `kind`), so the output is correct. Just noting that the `kind` argument is a lie in this code path — a comment would clarify.

### What Looks Good

- **Error routing is correct across all paths:** ontology errors → `file=<ontology>`, `rule=ontology-*`; binding errors → `file=<binding>`, `rule=binding-*`; missing files → `rule=cli-missing-ontology`, exit 2
- **`_validate_binding_file` docstring** clearly explains the resolution order and error routing, including *why* the CLI does its own ontology loading
- **Regression guard test** (`test_validate_binding_with_broken_companion_ontology_reports_ontology_file`) directly tests the misattribution bug that motivated the split — this is the most important test in the file
- **Auto-discovery vs explicit flag** are tested with a deliberate sibling-directory layout that proves `--ontology` actually bypasses discovery
- **Graceful fallback** when peek fails: defers to `load_binding(file)` so the real parse/shape error surfaces naturally
- **`file` field attribution** is carefully different for auto-discovery (points at binding) vs explicit flag (points at missing ontology path)
- **Test comments** continue to be excellent — each test explains its purpose, what it proves, and why the fixture is shaped the way it is
- **`_collect_errors` generalization** with `kind` is clean — one parameter, two prefixes, YAML-parse stays agnostic

### Verdict

Looks good to merge (after PR #17 lands). No blocking issues. The YAML re-parsing and kind-on-yaml-error nits are both cosmetic.

src/bigquery_ontology/cli.py
```diff
@@ -192,61 +213,192 @@ def validate(
...
+      ontology_path = peeked
+      discovered_via_peek = True
+
+  ontology = None
+  if ontology_path is not None:
+    if not ontology_path.exists() or not ontology_path.is_file():
+      # Auto-discovery and explicit-flag paths get distinct messages —
+      # the former explains *why* we looked where we did, the latter
+      # simply reports what the user asked us to open.
+      if discovered_via_peek:
```
caohy1988, line 323: Contract mismatch: `docs/ontology/cli.md` still says a missing companion ontology should fall back to syntax-only binding validation and emit a warning, and its flag table still documents only `--json`. This branch changes that contract in two user-visible ways: missing companion is now an exit-2 `cli-missing-ontology` error, and `--ontology PATH` is part of the supported surface. If that new behavior is intentional, the CLI docs need to be updated in the same change so users are not told the opposite.

src/bigquery_ontology/cli.py
```diff
@@ -170,6 +183,14 @@ def validate(
         "--json",
         help="Emit structured JSON errors on stderr.",
     ),
+    ontology_path: Path = typer.Option(
```
caohy1988: Bug: this new `Path` option is still pre-validated by Click/Typer for readability before `_validate_binding_file()` runs, so `--json` is not uniform here. Repro on this branch with `CliRunner`: create a real ontology file, chmod it to `000`, then run `gm validate good.binding.yaml --ontology unreadable.ontology.yaml --json`. The command exits 2 with Typers human usage text (`Invalid value for --ontology: ... is not readable`) instead of the structured error format this CLI is trying to guarantee.

src/bigquery_ontology/cli.py
libei: Fixed in 94c6ec4. Updated cli.md §4: missing companion is now documented as exit-2 `cli-missing-ontology` (not a warning fallback), and the `--ontology <path>` flag is added to the flag table.

src/bigquery_ontology/cli.py
libei: Fixed in 94c6ec4. Changed `ontology_path` from `Path` to `str | None` so Typer no longer maps it to `TyperPath(readable=True)`. The command body converts to `Path` after entry, and an `os.access()` readability check catches unreadable files with a structured error. Added a test that creates a `chmod 0o000` ontology file and asserts `--json` produces `cli-missing-ontology` (not Typer's human usage text).

PR #19, opened 2026-04-14
feat(compile): add gm compile pipeline for BigQuery property graphs

libei:
Stacks on #18 (which stacks on #17, which stacks on #16). Review this PR's diff against \`feat/binding-cli\` to isolate the compile changes.

## Summary
Adds \`compile_graph(ontology, binding) -> str\`, which turns a validated \`Ontology\` + \`Binding\` pair into one deterministic \`CREATE PROPERTY GRAPH\` statement for BigQuery. Closes the happy-path arc: YAML → validated objects → DDL.

## Pipeline
\`\`\`
  ontology.yaml ─┐
                 ├── Resolver ── ResolvedGraph ── Emitter ── DDL
  binding.yaml ──┘
\`\`\`

**Resolver** (\`compiler.py\`) cross-references the two inputs, builds the intermediate, and catches compile-time-only rules.

**Emitter** (\`compiler.py\`) walks the \`ResolvedGraph\` and produces text. Mechanical — all interesting decisions happen during resolution so emission is byte-deterministic.

**Intermediate** (\`resolved_models.py\`): frozen dataclasses — \`ResolvedProperty\`, \`ResolvedLabelAndProperties\`, \`ResolvedNodeTable\`, \`ResolvedEdgeTable\`, \`ResolvedGraph\`. Shape mirrors the GCP grammar:
- \`LabelAndProperties\` is a first-class bundle (label + its properties), not two peer fields
- \`label_and_properties: tuple[...]\` on node/edge tables, always length-1 in v0 but reserves the shape for future multi-label work

## Compile-time rules beyond the loaders
- **No \`extends\`**. v0 rejects hierarchical ontologies; inheritance lowering is a separate future design.
- **Derived-expression substitution**. Word-boundary regex, recursive, cycle-detected. Documented limitation: doesn't inspect SQL string literals (pragmatic for v0).
- **Edge keys always emitted**. Three resolution cases — \`keys.primary\` → bound columns; \`keys.additional\` → endpoint columns + additional; none → endpoint columns as safe default.

## Emitted DDL shape
- One \`CREATE PROPERTY GRAPH <name>\`.
- Node and edge tables alphabetized by alias (= label = entity/relationship name).
- Aliases are the ontology labels verbatim (\`raw.accounts AS Account\`), so DDL reads by logical type. Collision-free by construction since the ontology loader already enforces unique names.
- \`LABEL X PROPERTIES (...)\` bundled per the GCP grammar's \`LabelAndProperties\` production. Inline when short, wrapped across lines when long (80-col threshold).

## Spec alignment
\`docs/ontology/compilation.md\` updated so the spec describes what we ship:
- §2a Type overview: \`Resolved*\` prefix on all types, \`label_and_properties\` field, edge \`key_columns\` first-class.
- §3 Resolution: new subsections for alias derivation and edge-key resolution (three cases).
- §4 Worked example: expanded to the complete three-node, one-edge form the emitter actually produces.
- §4 Grammar-mapping table: reflects the new model shape, including the \`LabelAndPropertiesList\` footnote.

## Test plan
- [x] \`uv run --with pytest python -m pytest tests/bigquery_ontology -q\` — 80 passed
- [x] Byte-exact golden of the spec's finance worked example
- [x] Determinism: compile twice, assert byte-identical
- [x] NODE TABLES without EDGE TABLES (clean terminator placement)
- [x] Passthrough: \`id\` column bound to \`id\` property emits bare, not \`id AS id\`
- [x] Nested derived substitution: \`(col_a || col_b)\` spliced, outer derived wraps inner in parens
- [x] Edge KEY: primary → standalone; additional → endpoints + additional; no keys → endpoint-only fallback
- [x] Compile-time errors: entity \`extends\`, relationship \`extends\`, derived cycle
- [x] Regression: two entities with the same source basename compile cleanly (label-based aliasing)

## Intentionally out of scope
- \`gm compile\` CLI wiring — next PR, so this one stays focused on the compilation core.
- Spanner emission, inheritance lowering, measures — per spec §9.

### DIFF
--- a/src/bigquery_ontology/compiler.py
+++ b/src/bigquery_ontology/compiler.py
@@ -1,0 +1,619 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Compile an ontology + binding into BigQuery ``CREATE PROPERTY GRAPH`` DDL.
+
+The compiler is a single pass split into two stages:
+
+  1. **Resolve.** Cross-reference the ontology and the binding to
+     produce an in-memory ``ResolvedGraph``. This is where derived
+     expressions get rewritten to reference physical columns, edge
+     endpoints pick up their node-table aliases and key columns, and
+     the list is sorted alphabetically for deterministic output.
+  2. **Emit.** Walk the ``ResolvedGraph`` and produce the DDL text.
+     The emitter is intentionally mechanical — all interesting
+     decisions happen in stage 1 so that stage 2 is just rendering.
+
+The compiler trusts that the loaders already ran: it does *not*
+re-validate shape, name uniqueness, or property coverage. It only
+enforces the rules that are specifically compilation-level and that
+the loaders don't know about:
+
+  - No ``extends`` anywhere in the ontology (v0 rejects hierarchies;
+    inheritance lowering is a separate, future design).
+  - No cycles among derived properties.
+  - Node-table aliases are unique.
+
+Determinism contract: same ``Ontology`` + ``Binding`` → byte-identical
+output. Node and edge tables are emitted in alphabetical order by
+label; properties within each table are emitted in ontology
+declaration order.
+
+Known limitation (documented, not fixed in v0): derived-expression
+substitution uses word-boundary regex rather than a SQL lexer. This
+is sufficient for expressions of the form
+``first_name || ' ' || last_name`` but will misfire if a property
+name happens to appear as a substring inside a single-quoted string
+literal (e.g. ``'name'``). Upgrading to a real SQL lexer is a
+follow-up when a real user expression exercises the gap.
+"""
+
+from __future__ import annotations
+
+import re
+
+from .binding_models import Binding
+from .binding_models import EntityBinding
+from .binding_models import PropertyBinding
+from .binding_models import RelationshipBinding
+from .ontology_models import Entity
+from .ontology_models import Ontology
+from .ontology_models import Property
+from .ontology_models import Relationship
+from .resolved_models import ResolvedEdgeTable
+from .resolved_models import ResolvedGraph
+from .resolved_models import ResolvedLabelAndProperties
+from .resolved_models import ResolvedNodeTable
+from .resolved_models import ResolvedProperty
+
+
+# --------------------------------------------------------------------- #
+# Public entry point                                                    #
+# --------------------------------------------------------------------- #
+
+
+def compile_graph(ontology: Ontology, binding: Binding) -> str:
+  """Compile an ontology + binding pair into BigQuery DDL.
+
+  Returns a single ``CREATE PROPERTY GRAPH`` statement terminated by a
+  semicolon and a trailing newline. The output is deterministic:
+  re-running with the same inputs produces byte-identical text, and
+  unrelated ontology changes (e.g. adding a new entity) only show up
+  as an additive diff in the emitted DDL.
+
+  Raises:
+      ValueError: Any compile-time rule is violated (``extends`` used,
+          alias collision, derived-expression cycle).
+  """
+  graph = _resolve(ontology, binding)
+  return _emit_bigquery(graph)
+
+
+# --------------------------------------------------------------------- #
+# Resolution                                                             #
+# --------------------------------------------------------------------- #
+
+
+def _resolve(ontology: Ontology, binding: Binding) -> ResolvedGraph:
+  """Produce a ``ResolvedGraph`` from validated ontology+binding inputs."""
+  _reject_extends(ontology)
+
+  # Index ontology and binding by name so the per-entity / per-edge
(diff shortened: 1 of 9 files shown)


caohy1988 (commented):
## PR Review

This is the most substantial PR in the stack and the quality matches. The resolver/emitter split is clean, the frozen intermediate model mirrors the GCP grammar, derived-expression substitution handles recursion and cycles correctly, and the byte-exact golden test anchors everything. This closes the happy-path arc from YAML to DDL.

### Non-blocking Observations

**1. `assert` for invariant check in `_resolve_node_table`**

```python
assert entity.keys is not None and entity.keys.primary is not None
```

`assert` is disabled under `python -O`, which some production deployments use. Since this guards an invariant the loaders guarantee, it should never fire — but if it does, a silent `None` propagation would be harder to debug than a clear `ValueError`. The docstring acknowledges this as intentional ("let it blow up loudly"), so not blocking — just flagging for awareness.

**2. Longest-first sorting in `_resolve_sql` is redundant with `\b` word boundaries**

The comment says: "Longer names first so a property called `full_name` isn't shadowed by a substitution of `name`." But `\bname\b` already won't match inside `full_name` because `_` is a word character — there's no word boundary between `full_` and `name`. The sorting adds defense-in-depth but the stated rationale doesn't hold. Consider updating the comment to reflect the actual (weaker) motivation, or dropping the sort since `\b` handles it.

**3. Spec update and implementation are well-synchronized**

The compilation.md updates (alias derivation, edge-key resolution, formatting rules, worked example, grammar mapping table) all match the code. The worked example in the spec now byte-matches the golden test output. This is how spec-implementation parity should look.

### What Looks Good

- **Resolver/emitter split** — all interesting decisions happen in resolution; emission is mechanical rendering. This means a future Spanner emitter only has to replace `_emit_bigquery`, not redo the cross-referencing
- **Edge-key resolution** covers all three cases (primary, additional, endpoint-fallback) with clear documentation of why BigQuery needs a KEY even when the ontology doesn't declare one
- **Derived-expression substitution** is recursive with per-call cycle detection via `resolving` set. The `finally: resolving.discard(prop.name)` correctly cleans up even on error
- **`ResolvedLabelAndProperties` as a first-class type** mirrors the GCP grammar's `LabelAndProperties` production. The `tuple[...]` (always length-1 in v0) reserves the shape for multi-label without overengineering
- **Frozen dataclasses** for the intermediate — no mutation, no persistence, produce-consume-discard lifecycle. Right tool for the job
- **Label-based aliasing** (`raw.accounts AS Account`) is collision-free by construction (ontology loader guarantees unique names), eliminating the need for runtime alias dedup
- **80-column line-wrap** on `LABEL X PROPERTIES(...)` keeps output grep-friendly and diff-friendly
- **Golden test** (`test_compiles_finance_worked_example_to_exact_ddl`) is the most valuable test in the file — it locks the full DDL shape byte-for-byte
- **Determinism test** guards against accidental dict/set iteration order leaks
- **Colliding-source-basenames test** is a great regression guard for the aliasing approach
- **Test comments** continue the pattern of explaining *why*, not just *what*

### Verdict

Looks good to merge (after PR #18 lands). No blocking issues. The `assert` vs `ValueError` and the redundant sort comment are cosmetic.

src/bigquery_ontology/compiler.py
```diff
@@ -0,0 +1,619 @@
...
+          resolving=resolving,
+          owner=owner,
+      )
+      # If the nested result is itself a derived expression, wrap it
+      # in parentheses before splicing so operator precedence in the
+      # outer expression stays correct regardless of what the user
+      # wrote. Column names don't need this.
+      if property_map[name].expr is not None:
+        nested = f"({nested})"
+      substituted = re.sub(pattern, nested, substituted)
```
caohy1988: Bug: this loop rewrites against the already-substituted string, so a replacement can get re-matched as if it were another logical property name. Repro on this branch: declare stored properties `a` and `x`, bind them to columns `x` and `y`, then add derived `d: expr: "a"`. The compiler emits `(y) AS d` instead of `(x) AS d`, because substituting `a -> x` makes the later `x` pass fire on the inserted column token. The rewrite needs to distinguish original logical-property references from text introduced by earlier substitutions.

src/bigquery_ontology/compiler.py
```diff
@@ -0,0 +1,619 @@
...
+    result = substituted
```
caohy1988: Bug: unresolved names in `expr:` just survive into emitted SQL instead of failing compilation, even though `docs/ontology/compilation.md` §6 says every name in a derived expression must resolve to a property on the same element. Repro on this branch: `expr: "missing_prop || id"` compiles to `(missing_prop || row_id) AS d` with no error. The compiler needs a post-check here (or a token-validation pass up front) so typos in derived properties are caught as compile-time validation errors rather than leaking into broken DDL.

src/bigquery_ontology/compiler.py
libei: Fixed in 1d94bed. Replaced the sequential per-property rewrite loop with a single-pass `re.sub` using a combined alternation pattern. The replacement callback resolves each matched name on the fly, but substitutions happen against the original `prop.expr` — so a column name introduced by one replacement can never be re-matched as a different property name. Added a regression test with the exact repro (properties `a`→column `x`, `x`→column `y`, derived `d: expr: "a"`) that asserts `(x) AS d` appears and `(y) AS d` does not.

src/bigquery_ontology/compiler.py
libei: Fixed in 1d94bed. Added `_reject_unresolved_names` which scans the expression for `\b[a-zA-Z_]\w*\b` identifiers and raises `ValueError` if any are not in the property map (excluding self-references). Runs before substitution so typos are caught as compile-time validation errors. Added a test with `expr: "missing_prop || id"` that asserts `ValueError` matching `missing_prop`.

PR #20, opened 2026-04-14
feat(cli): add gm compile command

libei:
Stacks on #19 (which stacks on #18 → #17 → #16). Review this PR's diff against \`feat/compile\` to isolate the CLI changes.

## Summary
Wires \`compile_graph\` into the CLI. Closes the happy-path arc: YAML source → validated objects → emitted DDL, all reachable from one binary.

\`\`\`
gm compile <binding.yaml> [--ontology PATH] [-o OUT_PATH] [--json]
\`\`\`

On success the DDL goes to stdout (or to \`-o PATH\`) and stderr is empty. On failure, structured errors land on stderr in the same shape \`gm validate\` already uses.

## What's new

- **Shared loader extracted.** \`_load_ontology_and_binding(file, ontology_path, json_output) -> (Ontology, Binding)\` now owns the companion-ontology resolution, ontology-side error routing, and binding-side error routing. Both \`gm validate\` and \`gm compile\` call through it. Validate becomes a thin load-and-discard wrapper; compile takes the pair and calls \`compile_graph\`.
- **New rule prefix \`compile-validation\`.** \`compile_graph\`'s \`ValueError\`s (inheritance rejection, derived cycles) surface with this prefix, distinct from \`binding-validation\` / \`ontology-validation\` so tooling can tell compile-time rule violations from loader violations.
- **New usage error \`cli-wrong-kind\` (exit 2).** \`gm compile\` invoked on an ontology file or unknown-kind YAML gets rejected up front with a clear message — you can't compile an ontology alone because there's no physical mapping to emit.

## Error routing summary (compile)

| Scenario                                       | rule                    | exit |
|------------------------------------------------|-------------------------|------|
| Binding file missing                           | \`cli-missing-file\`      | 2    |
| Input is an ontology or unknown-kind YAML      | \`cli-wrong-kind\`        | 2    |
| Companion ontology missing (auto-discovery)    | \`cli-missing-ontology\`  | 2    |
| \`--ontology PATH\` points at missing file       | \`cli-missing-ontology\`  | 2    |
| Ontology file has validation errors            | \`ontology-validation\`   | 1    |
| Binding has shape / semantic errors            | \`binding-validation\`    | 1    |
| Compile-time rule violation (extends, cycles)  | \`compile-validation\`    | 1    |
| Malformed YAML                                 | \`yaml-parse\`            | 1    |

## Test plan
- [x] \`uv run --with pytest python -m pytest tests/bigquery_ontology -q\` — 88 passed
- [x] Happy path: DDL written to stdout matches the compiler's exact output
- [x] \`-o PATH\`: DDL written to file, stdout empty
- [x] \`--ontology PATH\`: override works when auto-discovery can't find the companion
- [x] Compile-time error (entity \`extends\`) routed with \`compile-validation\` rule, plain output
- [x] Compile-time error with \`--json\` keeps the \`compile-validation\` prefix in the JSON payload
- [x] Ontology file passed to \`gm compile\` rejected with \`cli-wrong-kind\`
- [x] Missing binding file surfaces \`cli-missing-file\`
- [x] Missing companion ontology surfaces \`cli-missing-ontology\` (regression guard for the shared loader)
- [x] All 11 pre-existing \`gm validate\` tests still pass after the refactor

### DIFF
--- a/src/bigquery_ontology/cli.py
+++ b/src/bigquery_ontology/cli.py
@@ -15,17 +15,18 @@
 """``gm`` command-line interface.
 
 ``gm validate`` accepts either an ontology YAML or a binding YAML and
-dispatches to the matching loader. For binding files, the companion
-ontology is auto-discovered as ``<name>.ontology.yaml`` next to the
-binding by default; pass ``--ontology PATH`` to point at a specific
-ontology file instead. The ``gm compile`` and ``gm import-owl`` commands
-referenced elsewhere will be wired up when their modules land.
+dispatches to the matching loader. ``gm compile`` takes a binding YAML
+and emits the corresponding ``CREATE PROPERTY GRAPH`` DDL on stdout
+(or to ``-o PATH``). Both commands resolve a binding's companion
+ontology by auto-discovering ``<name>.ontology.yaml`` next to the
+binding; ``--ontology PATH`` overrides that lookup.
 
 Exit codes:
 
   0 — success
   1 — validation / compilation error
-  2 — usage error (bad flag, missing file, missing companion ontology)
+  2 — usage error (bad flag, missing file, missing companion ontology,
+      compile invoked on a non-binding file)
   3 — internal error
 """
 
@@ -42,12 +43,15 @@ import yaml
 
 from .binding_loader import load_binding
 from .binding_loader import load_binding_from_string
+from .binding_models import Binding
+from .compiler import compile_graph
 from .loader import load_ontology
 from .loader import load_ontology_from_string
+from .ontology_models import Ontology
 
 app = typer.Typer(
     name="gm",
-    help="Graph-model CLI. Currently supports: validate.",
+    help="Graph-model CLI. Commands: validate, compile.",
     add_completion=False,
     no_args_is_help=True,
 )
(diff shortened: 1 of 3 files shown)


caohy1988 (commented):
## PR Review

Clean incremental PR that closes the happy-path arc from YAML to DDL in the CLI. The key contribution is extracting `_load_ontology_and_binding` as a shared helper — both `gm validate` and `gm compile` get the same error-routing behavior for free, and `_validate_binding_file` reduces to a one-liner.

### No Blocking Issues

Two minor observations, neither worth changing:

**1. Duplicate file-existence check pattern**

Both `validate` and `compile_command` start with identical `if not file.exists() or not file.is_file(): ...` blocks. A shared `_check_file_exists(file, json_output)` helper would remove the duplication, but with only two call sites and each having slightly different context (validate accepts both ontology and binding; compile only binding), the duplication is tolerable.

**2. Defensive `typer.Exit(code=3)` in fallback path**

The `# pragma: no cover` path after `load_binding(file)` succeeds without a peek path is a correct invariant guard. It should genuinely be unreachable — if peek failed, either `load_binding` fails too (caught above) or something very wrong happened. Good defensive code.

### What Looks Good

- **`_load_ontology_and_binding` extraction** — shared by validate and compile, owns all companion-ontology resolution and error routing. Validate becomes `load_and_discard`; compile takes the pair and calls `compile_graph`. Clean separation
- **Control flow is more linear** than PR #18's version — the `None`-peek case is handled first (early exit), then the main path is a straight sequence: check existence → load ontology → load binding → return
- **`compile-validation` rule prefix** correctly distinguishes compile-time errors (extends, cycles) from loader-level errors. Downstream tooling can filter by prefix without parsing the message
- **`cli-wrong-kind` for ontology-on-compile** catches the usage error early with a clear message, rather than letting it fail deep in the loader
- **`-o PATH` writes file and leaves stdout empty** — important for pipeline composability (`gm compile foo.yaml | bq query` vs `gm compile foo.yaml -o ddl.sql && bq < ddl.sql`)
- **`typer.echo(ddl, nl=False)`** correctly avoids double-newline since the emitter already terminates with `\n`
- **8 new tests** cover stdout output, file output, `--ontology` flag, compile-time errors (plain + JSON), wrong-kind rejection, missing file, and missing companion
- **Test fixtures at module level** (`_COMPILE_ONTOLOGY`, `_COMPILE_BINDING`, `_COMPILE_EXPECTED_DDL`) with a note that the DDL must stay in sync with the emitter — changes fail both this and the golden test simultaneously, which is deliberate
- **Pre-existing validate tests pass** after the refactor — confirmed by PR description (88 total = 11 validate + 13 binding-shape + 18 binding-loader + 8 compile-cli + ... ontology + compiler tests)

### Verdict

Looks good to merge (after PR #19 lands). No changes needed.

src/bigquery_ontology/cli.py
```diff
@@ -258,20 +262,160 @@ def validate(
...
 
 
+# --------------------------------------------------------------------- #
+# gm compile                                                             #
+# --------------------------------------------------------------------- #
+
+
+@app.command("compile")
+def compile_command(
+    file: Path = typer.Argument(
```
caohy1988: Bug: the new `compile` command still relies on Typer/Click `Path` validation for readability, so `--json` is not uniform here either. Repro on this branch with `CliRunner`: chmod the binding file to `000` and run `gm compile unreadable.binding.yaml --json`, or pass an unreadable `--ontology` path. Both cases exit 2 with Typers human usage text (`Invalid value for ... is not readable`) instead of the structured error format this CLI is trying to guarantee.

src/bigquery_ontology/cli.py
```diff
@@ -258,20 +262,160 @@ def validate(
...
+    )
+    raise typer.Exit(code=1)
+  except Exception as exc:  # pragma: no cover - defensive
+    typer.echo(f"internal error: {exc}", err=True)
+    raise typer.Exit(code=3)
+
+  # Write the DDL. The string already ends in ``\n`` so neither branch
+  # adds another one.
+  if output_path is not None:
+    output_path.write_text(ddl, encoding="utf-8")
```
caohy1988: Bug: output-write failures are not routed through the CLI error path. Repro on this branch: point `-o` at a file in a nonexistent directory. `output_path.write_text(...)` raises `FileNotFoundError`, `CliRunner` sees an uncaught exception, and there is no structured stderr output even though the command docstring says failures should land as CLI errors. This needs the same kind of explicit handling the input-path checks have, otherwise `gm compile ... -o ...` is less robust than the rest of the command surface.

caohy1988:
Review finding against latest `main` that is still present on this PR head but is not commentable inline here because `docs/ontology/cli.md` is outside the visible diff of the stacked PR:

Medium: the CLI docs are now materially out of sync with the shipped behavior. `docs/ontology/cli.md` still says `gm validate` only warns and continues when the companion ontology is missing, documents only `--json` for validate, and documents `gm compile <binding> [-o <out>]` / `--out` without the new `--ontology` override or the `cli-wrong-kind` behavior. Since this PR makes `compile` part of the user-facing surface, the docs need to be updated in the same stack so users are not told the opposite of what the binary does.

libei:
Rebased onto `main` (after `feat/compile` merged via #24) and addressed all three review items in `0eaab6d`:

**1. Docs out of sync** — Updated `docs/ontology/cli.md` §5: added `--ontology <path>` flag, fixed `--out` → `--output` spelling, added `cli-wrong-kind` note. §4 was already correct after the rebase.

**2. Unreadable files bypass `--json`** — All CLI path parameters (`file`, `--ontology`, `-o`) now use `str` (not `Path`) so Typer/Click does not pre-validate readability. Each command casts to `Path` internally and checks `os.access(path, os.R_OK)` alongside the existence checks, routing through `_emit_errors` with `cli-missing-file` rule. Added 3 new tests: unreadable binding in compile, unreadable `--ontology` in compile, unreadable main file in validate.

**3. Output write failure unhandled** — Wrapped `write_text` in `try/except (FileNotFoundError, PermissionError)`, routing through `cli-output-error` rule (exit 1). Added 2 new tests: nonexistent output dir in plain and `--json` modes.

All 98 tests pass.

src/bigquery_ontology/cli.py
libei: Fixed. All CLI path parameters (`file`, `--ontology`, `-o`/`--output`) now use `str` instead of `Path` so Typer/Click never pre-validates readability. Each command casts to `Path` internally and checks `os.access(path, os.R_OK)` alongside the existence check, routing failures through `_emit_errors` with the `cli-missing-file` rule. Three new tests cover the unreadable-file scenarios for both commands.

src/bigquery_ontology/cli.py
libei: Fixed. The `write_text` call is now wrapped in `try/except (FileNotFoundError, PermissionError)`, routing failures through `_emit_errors` with a `cli-output-error` rule (exit 1). Two new tests cover the nonexistent-parent-dir scenario in both plain and `--json` modes.

libei:
@caohy1988 Re: docs out of sync — fixed. After rebasing onto `main`, §4 was already correct (the rebase preserved the companion-ontology/`--ontology` docs from PR #18). §5 needed updating: added `--ontology <path>` to the usage line and flags table, fixed `--out` → `--output` to match the actual flag name, and added the `cli-wrong-kind` rejection note.

PR #21, opened 2026-04-14
docs(scaffold): add gm scaffold design doc

libei:
## Summary
- Adds `docs/ontology/scaffold.md` — design for `gm scaffold`, a one-shot bootstrapper that emits `table_ddl.sql` plus a matching `binding.yaml` stub from an ontology. Intended for greenfield BigQuery projects; outputs are user-owned after generation (no idempotent regeneration).
- Node/edge DDL is fully derived from the ontology: `keys.primary` drives `PRIMARY KEY ... NOT ENFORCED` and `NOT NULL` columns (compound-capable); relationship `from`/`to` drive `FOREIGN KEY ... REFERENCES ... NOT ENFORCED` endpoints (also compound-capable). Scaffold emits no `OPTIONS` clauses — semantic metadata belongs on `CREATE PROPERTY GRAPH`, which is `gm compile`'s job.
- Updates `docs/ontology/cli.md` to document the `graph_ddl.sql` filename convention on `gm compile -o`, pairing with scaffold's `table_ddl.sql`.

## Test plan
- [ ] Review scaffold.md sections 3–4 for correctness against the ontology key/endpoint model
- [ ] Confirm scope boundaries in §5a (no `OPTIONS`, no `PARTITION BY`, no `CREATE PROPERTY GRAPH`) match intent
- [ ] Confirm the `scaffold → bq query → compile → bq query` flow in §2 reads cleanly


caohy1988 (commented):
## PR Review — `docs/ontology/scaffold.md`

Focusing on the scaffold design doc and the cli.md filename-convention addition (the other 15 files are carry-overs from PRs #16–#20 already reviewed).

Overall this is a well-thought-out design. The non-idempotent "generator, not compiler" contract is the right call — it avoids the complexity of merge/diff that tools like Alembic struggle with, and "user-owned after generation" is a clear mental model. The `table_ddl.sql` / `graph_ddl.sql` naming convention makes the directory self-describing.

### Issues

**1. Derived properties are not addressed in §3 or §4**

The doc says columns come from "the entity's declared properties" but doesn't say whether derived properties (those with `expr:`) get a physical column. They almost certainly should NOT — the binding loader forbids binding derived properties, `gm compile` substitutes the expression at DDL-emission time, and a physical column for a derived value would go unused.

The omission creates ambiguity: a reader implementing scaffold wouldn't know whether to emit a column for `full_name`. Add a sentence to §3 (and §4 for relationship properties): "Derived properties — those with `expr:` in the ontology — are excluded from the DDL. They are computed by the `CREATE PROPERTY GRAPH` layer, not stored."

**2. §7 uses `src_id` / `dst_id` naming but §4 uses `from_*` / `to_*`**

§7 says the binding stub carries `id, src_id, dst_id, snake-cased property columns`. But §4 established the convention as `from_party_id` / `to_party_id`. The `src_id` / `dst_id` in §7 appears to be a leftover from an earlier naming iteration. Update §7 to use `from_*` / `to_*` to match §4.

**3. Potential column-name collision between endpoint columns and relationship properties**

If a relationship has a property named `from_party_id`, it would collide with the generated `from_*` endpoint column for an entity whose primary key is `party_id`. The spec should either:
- Declare this an error ("relationship property names must not collide with generated endpoint column names"), or
- Document a disambiguation strategy

This is an edge case but one that would produce confusing DDL if hit.

### Non-blocking Observations

**4. `--naming=snake` algorithm is unspecified**

Examples cover `Person → person` and `firstName → first_name`, but the exact conversion algorithm isn't stated. Edge cases like `HTTPRequest` or `XMLParser` can produce different results depending on the library (`http_request` vs `h_t_t_p_request`). Worth a sentence: "CamelCase is split at uppercase-to-lowercase boundaries" or naming the library.

**5. §7 would benefit from a concrete binding YAML example**

The section is thin compared to the DDL examples in §3–§4. A complete binding stub for the `Person` + `Follows` examples would help reviewers verify the shape matches the DDL.

### What Looks Good

- **§1's anti-goals** are as valuable as the goals — "does not introspect BigQuery", "does not perform schema evolution", "does not participate in the compile pipeline" — these prevent scope creep
- **`PRIMARY KEY ... NOT ENFORCED`** is the correct BigQuery pattern; the optimizer uses it for join planning even though it's not enforced at write time
- **`FOREIGN KEY ... NOT ENFORCED`** on edge tables documents the graph topology at the DDL layer and feeds the property-graph tooling
- **Three edge-key modes** (no keys, primary, additional) are well-specified with clear examples including the compound-endpoint case
- **Refusing to write to non-empty `--out`** is good safety for a one-shot tool
- **§5a's exclusion list** is well-reasoned — especially the point that `OPTIONS` metadata belongs on `CREATE PROPERTY GRAPH`, not physical tables
- **§10 open questions** are honest and the leanings are all sound
- **cli.md filename convention** (`graph_ddl.sql` for compile, `table_ddl.sql` for scaffold) is a nice touch that makes the directory self-documenting

### Verdict

Fix items **1** (derived properties) and **2** (src/dst naming inconsistency) before merge. Item **3** (endpoint collision) is worth a sentence. Items **4** and **5** are nice-to-have.

src/bigquery_ontology/cli.py
```diff
@@ -192,61 +217,347 @@ def validate(
...
+    )
+    raise typer.Exit(code=1)
+  except Exception as exc:  # pragma: no cover - defensive
+    typer.echo(f"internal error: {exc}", err=True)
+    raise typer.Exit(code=3)
+
+  # Write the DDL. The string already ends in ``\n`` so neither branch
+  # adds another one.
+  if output_path is not None:
+    output_path.write_text(ddl, encoding="utf-8")
```
caohy1988, line 388: Bug: output-write failures are not routed through the CLI error path. Repro on this branch: point `-o` at a file in a nonexistent directory. `output_path.write_text(...)` raises `FileNotFoundError`, `CliRunner` sees an uncaught exception, and there is no structured stderr output even though the command docstring says failures should land as CLI errors. This needs explicit handling, otherwise `gm compile ... -o ...` is less robust than the rest of the command surface.

src/bigquery_ontology/compiler.py
```diff
@@ -0,0 +1,619 @@
...
+          resolving=resolving,
+          owner=owner,
+      )
+      # If the nested result is itself a derived expression, wrap it
+      # in parentheses before splicing so operator precedence in the
+      # outer expression stays correct regardless of what the user
+      # wrote. Column names don't need this.
+      if property_map[name].expr is not None:
+        nested = f"({nested})"
+      substituted = re.sub(pattern, nested, substituted)
```
caohy1988, line 417: Bug: this loop rewrites against the already-substituted string, so a replacement can get re-matched as if it were another logical property name. Repro on this branch: declare stored properties `a` and `x`, bind them to columns `x` and `y`, then add derived `d: expr: "a"`. The compiler emits `(y) AS d` instead of `(x) AS d`, because substituting `a -> x` makes the later `x` pass fire on the inserted column token. The rewrite needs to distinguish original logical-property references from text introduced by earlier substitutions.

src/bigquery_ontology/cli.py
```diff
@@ -192,61 +217,347 @@ def validate(
...
+
+
+# --------------------------------------------------------------------- #
+# gm compile                                                             #
+# --------------------------------------------------------------------- #
+
+
+@app.command("compile")
+def compile_command(
+    file: Path = typer.Argument(
```
caohy1988, line 272: Bug: `gm compile` still relies on Typer/Click `Path` validation for readability, so `--json` is not uniform here. Repro on this branch with `CliRunner`: chmod the binding file to `000` and run `gm compile unreadable.binding.yaml --json`, or pass an unreadable `--ontology` path. Both cases exit 2 with Typers human usage text (`Invalid value for ... is not readable`) instead of the structured error format the CLI claims to provide.

src/bigquery_ontology/compiler.py
```diff
@@ -0,0 +1,619 @@
...
+    result = substituted
```
caohy1988, line 418: Bug: unresolved names in `expr:` just survive into emitted SQL instead of failing compilation, even though `docs/ontology/compilation.md` says every name in a derived expression must resolve to a property on the same element. Repro on this branch: `expr: "missing_prop || id"` compiles to `(missing_prop || row_id) AS d` with no error. The compiler needs a post-check here (or a token-validation pass up front) so typos in derived properties are caught as compile-time validation errors rather than leaking into broken DDL.

haiyuan-eng-google:
fix the format check

libei:
Superseded by #25, which is rebased cleanly on main without the stale code carry-overs.

PR #25, opened 2026-04-14
docs(scaffold): add gm scaffold design doc

libei:
## Summary
- Adds `docs/ontology/scaffold.md` — design for `gm scaffold`, a one-shot bootstrapper that emits `table_ddl.sql` plus a matching `binding.yaml` stub from an ontology.
- Updates `docs/ontology/cli.md` to document the `graph_ddl.sql` filename convention on `gm compile -o`.
- Incorporates all review feedback from PR #21 (which carried stale code commits and is superseded by this PR):
  - Derived properties (`expr:`) excluded from DDL (§3, §4)
  - Fixed `src_id`/`dst_id` → `from_*`/`to_*` naming in §7
  - Added endpoint-column / property-name collision error rule (§4)
  - Clarified snake-case conversion algorithm (§6)
  - Added concrete binding YAML example (§7)

Supersedes #21.

## Test plan
- [ ] Review scaffold.md §3–§4 for correctness against the ontology key/endpoint model
- [ ] Confirm scope boundaries in §5a match intent
- [ ] Confirm the `scaffold → bq query → compile → bq query` flow in §2 reads cleanly
- [ ] Verify §7 binding example is consistent with §3–§4 DDL examples

### DIFF
--- a/docs/ontology/scaffold.md
+++ b/docs/ontology/scaffold.md
@@ -1,0 +1,340 @@
+# Scaffold — Core Design (v0)
+
+Status: draft
+Scope: `gm scaffold` — a one-shot bootstrapper that emits a starter DDL
+file and a matching binding stub from an ontology. Intended for greenfield
+projects with no existing BigQuery tables.
+
+## 1. What scaffold is, and what it is not
+
+Scaffold is a **generator**, not a compiler. It runs once at the start of
+a project. Its outputs are then owned by the user, committed to version
+control, and hand-edited freely. Scaffold is never re-run against an
+existing output directory, and nothing downstream assumes its outputs are
+still in sync with the ontology after the user has edited them.
+
+In particular:
+
+- Scaffold does **not** participate in the compile pipeline. `gm compile`
+  treats the emitted binding as any other user-authored binding.
+- Scaffold does **not** execute DDL against BigQuery. It emits `.sql`
+  text; the user runs it with `bq query`, dbt, Terraform, or whatever
+  they already use.
+- Scaffold does **not** perform schema evolution. Ontology changes after
+  scaffold are handled by editing the binding and writing migrations —
+  not by regenerating.
+- Scaffold does **not** introspect BigQuery. It has no notion of "some
+  tables already exist."
+
+The binding schema stays pure-logical. Physical concerns (partitioning,
+clustering, labels, table options) live in the DDL file only, where the
+user edits them directly.
+
+## 2. Command shape
+
+```
+gm scaffold \
+  --ontology path/to/ontology.yaml \
+  --dataset my_dataset \
+  --out ./graph/
+```
+
+Required: `--ontology`, `--dataset`.
+Optional: `--naming {snake|preserve}` (default `snake`),
+`--project <id>` (omit to emit dataset-qualified names without a project
+prefix).
+
+Output directory layout:
+
+```
+graph/
+  table_ddl.sql    # CREATE TABLE statements, one per entity and relationship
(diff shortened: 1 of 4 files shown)


caohy1988 (commented):
## PR Review — `docs/ontology/scaffold.md` (v2, supersedes #21)

Clean docs-only PR with just 2 files. All five items from my PR #21 review are addressed:

| PR #21 Feedback | Status |
|---|---|
| Derived properties unaddressed in §3/§4 | Fixed — explicit exclusion sentence in both sections |
| `src_id`/`dst_id` → `from_*`/`to_*` in §7 | Fixed — uses `from_*`/`to_*` consistently |
| Endpoint-column / property-name collision | Fixed — error rule added in §4 |
| `--naming=snake` algorithm unspecified | Fixed — §6 specifies uppercase-to-lowercase boundary splitting with acronym handling |
| §7 needs concrete binding YAML example | Fixed — full Person + Follows example added |

### One Minor Observation

**§7 binding example omits `project` from the target block**

The example shows `target: {backend: bigquery, dataset: my_dataset}` with no `project` field. The current `BigQueryTarget` model (from PR #16) requires `project: str`. This is fine for a design doc — the scaffold implementation will need to decide whether `--project` is required for a valid binding, or whether `BigQueryTarget.project` becomes optional. Just flagging the gap so it doesn't surprise the implementer.

### What Looks Good

- §3/§4 derived-property exclusion aligns with the binding loader's rule (derived properties must not appear in bindings) and the compiler's substitution behavior
- §4 endpoint-collision error rule is the right call — failing early is better than emitting ambiguous DDL
- §6 snake-case spec (`HTTPRequest` → `http_request`) removes implementation ambiguity
- §7 binding example is consistent with the DDL examples — same column names, same naming convention, derived properties excluded
- The binding example comment (`# Generated by gm scaffold. This file is user-owned — edit freely.`) reinforces the non-idempotent contract from §1

### Verdict

Looks good to merge. The `project` field observation is a heads-up for the implementer, not a blocker for the design doc.

docs/ontology/scaffold.md
```diff
@@ -0,0 +1,340 @@
...
+    from_columns: [from_party_id]
+    to_columns: [to_party_id]
+    properties:
+      - {name: since, column: since}
+```
+
+## 8. Determinism
+
+Given the same ontology, flags, and version of `gm`, scaffold produces
+byte-identical `ddl.sql` and `binding.yaml`. Ordering follows ontology
```
caohy1988, line 313: Docs nit: §8 says scaffold produces byte-identical `ddl.sql` and `binding.yaml`, but §2 defines the actual output filename as `table_ddl.sql` and the rest of the doc leans on that distinction versus `graph_ddl.sql`. This looks like a stale filename from an earlier draft and is easy to misread as a third artifact name.

caohy1988:
Review finding against latest `main` that is still present on this PR head but is not commentable inline here because the relevant `src/bigquery_ontology/cli.py` lines are outside the tiny visible diff of this docs-forward PR:

Medium: unreadable companion ontologies are now reported as if they were missing. In this branch `_load_ontology_and_binding()` treats `not os.access(ontology_path, os.R_OK)` the same as `not exists()` / `not is_file()`, so `gm compile binding.yaml --ontology locked.ontology.yaml --json` returns `rule=cli-missing-ontology` with `"Ontology file not found"` even though the file exists and is just unreadable. The same happens on auto-discovery, which says no companion was found at a path that is actually present. That gives the user the wrong remediation and hides the actual permission problem.

PR #26, opened 2026-04-14
feat: Add Streamlit webapp for agent analytics metrics & visuals

Saherpathan:
This PR implements a dashboard for monitoring agent execution logs in Google BigQuery. It transitions the analytics from static cells to an interactive, multi-tab interface while resolving all previously identified security and logic issues.

There are 6 dedicated tabs for Usage (tokens), Performance (latency), Reliability (errors), Multimodal content, HITL activity, and a Session Trace Explorer.

Fixes:
-  Fully resolved SQL Injection risks by implementing parameterized BigQuery queries.
- Fixed duplicate session entries in the Trace Explorer using unique `(Time | Agent | ID)` labels.
- Refined HITL classification to explicitly target `HITL_` event types, preventing platform noise.
- Implemented `UNNEST` logic to correctly display Multimodal MIME types.
- Extracted shared event classification into a single `ENRICHED_EVENTS_CTE` to ensure consistency across the application.
- Provided a streamlined  Cloudflare Tunnel launch script for Google Colab and a standard local deployment workflow.

Prerequisites
* **Schema**: Data must follow the [ADK BigQuery Schema](https://adk.dev/integrations/bigquery-agent-analytics/#schema-reference).
* **Setup**: Placeholder credentials in `app.py` or the dashboard sidebar must be updated with your specific GCP Project, Dataset, and Table IDs.

### Verification Checklist
* [x] SQL parameterized.
* [x] Shared CTE implemented.
* [x] Notebook outputs cleared for privacy.

**Files Included**: `app.py`, `webapp_dashboard.ipynb`, `requirements.txt`, `README.md`


caohy1988 (changes requested):
## PR Review

This adds a Streamlit dashboard notebook for visualizing agent analytics from BigQuery. The idea is useful, but the PR has security issues and several quality gaps that should be fixed before merge.

### Blocking: SQL Injection

**1. User input interpolated directly into SQL strings**

Multiple queries use f-string interpolation of unvalidated user input:

```python
agent_name = st.text_input("Filter by Agent Name", "")
where_base += f" AND agent = '{agent_name}'"
```

and:

```python
selected_sid = st.selectbox("Select Session ID", sessions['session_id'])
trace_detail = query(f"... WHERE session_id = '{selected_sid}' ...")
```

A user typing `'; DROP TABLE foo; --` into the agent name field would inject SQL. The BigQuery client supports parameterized queries — use them:

```python
job_config = bigquery.QueryJobConfig(
    query_parameters=[
        bigquery.ScalarQueryParameter("agent", "STRING", agent_name)
    ]
)
client.query(sql, job_config=job_config)
```

Same for `selected_sid`, `time_hours`, and any other user-provided values that land in SQL. The `table_ref` (project/dataset/table from text inputs) is also injectable — validate these identifiers against a safe pattern (e.g., `^[\w.-]+$`).

### Should Fix

**2. Notebook outputs contain environment-specific data**

The committed notebook includes execution output with localtunnel URLs (`https://tame-guests-fetch.loca.lt`) and local/external IP addresses (`172.28.0.12`, `35.237.66.184`). Clear all outputs before committing — they leak network info and will confuse users who think the URLs are still live.

**3. `google.colab.auth` imported unconditionally**

`import google.colab.auth` and `google.colab.auth.authenticate_user()` run at app startup. This crashes outside Colab. Guard with:
```python
try:
    import google.colab.auth
    google.colab.auth.authenticate_user()
except ImportError:
    pass  # Running outside Colab; assume ADC
```

**4. Expected table schema is undocumented**

The queries assume specific columns (`session_id`, `trace_id`, `invocation_id`, `event_type`, `status`, `error_message`, `user_id`, `timestamp`, `content`, `latency_ms`, `attributes`, `content_parts`, `agent`). Without documentation of what table/schema this dashboard expects, users will get cryptic SQL errors. Add a markdown cell at the top listing the expected columns, or link to the schema definition.

**5. No PR description**

The PR body is empty. Add context: what the dashboard visualizes, what table schema it expects, and how to run it (Colab-only? local Streamlit?).

**6. Dependencies incomplete**

The notebook installs `streamlit` but also needs `plotly`, `pandas`, `google-cloud-bigquery`. Either `pip install` them all or note that they're pre-installed in the expected environment.

### Non-blocking Observations

**7. `%%writefile app.py` pattern**

Writing the app into a Python file from a notebook cell, then running it with localtunnel, is a Colab-specific pattern. Consider shipping `dashboard/app.py` as a standalone file with a README for running it locally or in Colab. This makes the code greppable, lintable, and directly runnable.

**8. No error handling on BigQuery queries**

If a query fails (wrong table, missing columns, permissions), the app crashes with an unhandled exception. A `try/except` wrapper around `query()` with `st.error()` would make the dashboard resilient to partial schema mismatches.

**9. `enriched_events` CTE duplicated**

The same CTE (classifying events into user/llm/tool/other families) appears in both the Performance and HITL tabs. Extract it into a shared query fragment or BigQuery view.

**10. Missing trailing newline**

The JSON file doesn't end with a newline character.

### Verdict

Fix the SQL injection (#1) before merge — it's the one thing that's genuinely unsafe. Items #2–#6 should also be addressed for a mergeable PR. The rest are improvements for a follow-up.

dashboard/dashboard_webapp.ipynb
```diff
@@ -0,0 +1,293 @@
...
+        "                    WHEN event_type = 'USER_MESSAGE_RECEIVED' THEN 'user'\n",
+        "                    WHEN event_type IN ('LLM_REQUEST','LLM_RESPONSE','LLM_ERROR') THEN 'llm'\n",
+        "                    WHEN event_type IN ('TOOL_STARTING','TOOL_COMPLETED','TOOL_ERROR') THEN 'tool'\n",
+        "                    ELSE 'other'\n",
+        "                END AS event_family\n",
+        "            FROM {table_ref}\n",
+        "            WHERE {where_base}\n",
+        "        )\n",
+        "        SELECT event_type, COUNT(*) as count\n",
+        "        FROM enriched_events WHERE event_family = 'other' # Assuming 'hitl' events fall under 'other' category as per Flask app logic.\n",
```
caohy1988: Logic bug: this labels every event family not recognized as `user` / `llm` / `tool` as HITL. That will sweep in any miscellaneous platform events, future event types, or malformed rows and present them as human-in-the-loop activity even when no human handoff happened. The inline comment already hints this is only an assumption. If the table has explicit HITL event types, filter for those directly; otherwise this tab should be named more neutrally than `Human-in-the-Loop Orchestration`.

dashboard/dashboard_webapp.ipynb
```diff
@@ -0,0 +1,293 @@
...
+        "        GROUP BY 1\n",
+        "    \"\"\")\n",
+        "    st.table(hitl_data)\n",
+        "\n",
+        "# --- TAB: TRACE DRILL-DOWN ---\n",
+        "with tab_trace:\n",
+        "    st.subheader(\"Session Explorer\")\n",
+        "    sessions = query(f\"SELECT session_id, agent, MIN(timestamp) as start_ts FROM {table_ref} WHERE {where_base} GROUP BY 1, 2 ORDER BY 3 DESC LIMIT 50\")\n",
```
caohy1988: Logic bug: the session picker is built from a query grouped by `(session_id, agent)` but the UI only displays `session_id`. If a single session spans multiple agents, this produces duplicate indistinguishable options in the selectbox, and `trace_detail` then ignores `agent` anyway and reloads the whole session. Either group just by `session_id`, or include enough label text in the selectbox (agent / start time) to make each option unique.

caohy1988 (approved):
## Follow-up Review (post-refactor commit b69b635)

The refactored version addresses most of the original review items. Here's what's fixed and what remains.

### Fixed (from original review)

| # | Original Issue | Status |
|---|----------------|--------|
| 1 | SQL injection via f-string interpolation | **Fixed** — `agent_name` and `time_hours` use `ScalarQueryParameter`. `table_ref` identifiers validated via `re.match(r'^[\w.-]+$')`. Session ID in trace explorer uses `@sid` parameter. |
| 2 | Notebook outputs with leaked URLs/IPs | **Fixed** — notebook outputs cleared. |
| 3 | `google.colab.auth` unconditional import | **Fixed** — notebook still has the import but it's in a Colab-specific notebook cell, not in `app.py`. The standalone `app.py` has no Colab dependency. |
| 4 | Expected schema undocumented | **Fixed** — README links to ADK BigQuery Schema Reference. |
| 5 | No PR description | **Fixed** — comprehensive PR body with tabs, fixes, prerequisites, checklist. |
| 6 | Dependencies incomplete | **Fixed** — `requirements.txt` includes streamlit, plotly, pandas, google-cloud-bigquery, db-dtypes. |
| 8 | No error handling on queries | **Fixed** — `safe_query()` wrapper with try/except and `st.error()`. |
| 9 | Duplicated CTE | **Fixed** — shared `ENRICHED_EVENTS_CTE` used across Performance, Reliability, and HITL tabs. |
| HITL classification | Logic bug: non-HITL events swept in | **Fixed** — now uses `WHEN event_type LIKE 'HITL_%'` with explicit `ELSE 'other'`. |
| Session picker duplicates | Duplicate session entries | **Fixed** — label uses `FORMAT_TIMESTAMP + agent + session_id[:8]`. |

### Remaining Issues

**1. (Medium) `table_ref` still interpolated via string replacement, not parameterized.**

`safe_query()` does `formatted_sql = sql.replace("{table_ref}", table_ref)` (line 111 of app.py). While identifiers are regex-validated (`^[\w.-]+$`), the table reference is still injected via string formatting, not a BQ parameter. This is acceptable because BigQuery doesn't support parameterized table names — but the comment should explain that the regex validation is the security boundary here, not parameterization.

**2. (Medium) `agent_name` filter uses string concatenation in some tabs.**

The Performance, Reliability, and HITL tabs append the agent filter via Python string conditional:
```python
{" AND agent = @agent_name" if agent_name else ""}
```
This is safe because `agent_name` is parameterized via `@agent_name` in `global_params`. But the pattern is fragile — if someone adds a new tab and copies this pattern but forgets to include `agent_name` in `global_params`, the query will fail at runtime. Consider making the agent filter part of the shared CTE instead of conditional per-tab.

**3. (Low) `requirements.txt` has trailing whitespace.**

Line 5 has trailing spaces after `db-dtypes`. Minor but will fail strict linting.

**4. (Low) `app.py` is duplicated — exists both as a standalone file and inside the notebook `%%writefile` cell.**

The notebook's `%%writefile app.py` cell contains the full `app.py` content. If someone edits one but not the other, they'll diverge. Consider either:
- Making the notebook the source of truth (remove standalone `app.py`, generate it from notebook)
- Or making `app.py` the source of truth (notebook cell just runs it without `%%writefile`)

### Verdict

The PR is now in good shape. The SQL injection issue is resolved, the shared CTE is clean, the HITL classification is correct, and the session picker deduplication works. Items #1-#4 above are improvements, not blockers. **Approve with minor suggestions.**

caohy1988:
Review finding:

Medium: the top-level Error Rate KPI still undercounts canonical errors compared with the rest of the dashboard. In `dashboard/app.py`, `ENRICHED_EVENTS_CTE` defines `is_canonical_error` as `(status = ERROR OR error_message IS NOT NULL OR event_type LIKE %_ERROR)`, but the Executive Summary query computes `error_rate` using only `COUNTIF(status = ERROR OR error_message IS NOT NULL)`. That means rows like `TOOL_ERROR` or `LLM_ERROR` with no populated `status` / `error_message` will appear in the Reliability tab but not in the headline error-rate metric.\n\nRecommended fix: make the KPI use the same canonical-error definition as the Reliability tab, ideally by reusing `ENRICHED_EVENTS_CTE` and `COUNTIF(is_canonical_error)` so the dashboard has one error contract everywhere.

caohy1988 (approved):
## Follow-up Review v3 (post commits 9440d76 + 20895b2)

### Changes since last review

| Change | Details |
|--------|---------|
| **Agent filter unified into CTE** | `ENRICHED_EVENTS_CTE` now includes `AND (@agent_name IS NULL OR agent = @agent_name)` — no more per-tab conditional string append |
| **Error rate unified** | Executive Summary now uses the shared CTE's `is_canonical_error` instead of a separate inline `COUNTIF(status = 'ERROR')` — consistent with the Reliability tab |
| **`agent_name` always in params** | `global_params` always includes `agent_name` (as `None` when empty) — fixes the BQ 400 error when tabs reference `@agent_name` but it wasn't in params |
| **`table_ref` comment added** | Explains why string replacement is used (BQ doesn't support parameterized table names) with note that regex validation is the security boundary |
| **Notebook simplified** | No longer embeds app.py via `%%writefile` — just checks if `app.py` exists and launches it |
| **Notebook renamed** | `webapp_dashboard.ipynb` → `agent_analytics_dashboard.ipynb` |
| **`requirements.txt` cleaned** | Trailing whitespace removed |

### Previous review items — status

| # | Issue | Status |
|---|-------|--------|
| 1 | `table_ref` comment explaining security boundary | **Fixed** — comment at line 219-222 |
| 2 | Agent filter per-tab fragility | **Fixed** — agent filter now in shared CTE via `@agent_name IS NULL OR agent = @agent_name` |
| 3 | `requirements.txt` trailing whitespace | **Fixed** |
| 4 | `app.py` duplicated in notebook | **Fixed** — notebook no longer contains `%%writefile`, just checks for `app.py` |
| New | Error rate inconsistency (Executive Summary vs Reliability) | **Fixed** — both now use `is_canonical_error` from the shared CTE |

### Remaining items

**1. (Low) Notebook auth cell still crashes outside Colab.**

The notebook has a bare `from google.colab import auth` (line 115-116) with no try/except guard. This crashes when running the notebook locally or in Vertex AI Workbench. The standalone `app.py` is fine — only the notebook is affected.

**2. (Low) Multimodal and Usage tabs don't use the shared CTE.**

The Usage tab (Token Consumption) and Multimodal tab query `{table_ref}` directly with `{where_clause}` instead of going through `ENRICHED_EVENTS_CTE`. This means the agent filter for those tabs uses the `where_clause` string (which conditionally appends `AND agent = @agent_name`) rather than the CTE's `@agent_name IS NULL OR agent = @agent_name` pattern. Both work correctly, but the two different agent-filter mechanisms could diverge if someone edits one and not the other. Not blocking.

**3. (Low) Missing trailing newline on notebook JSON.**

Minor — `agent_analytics_dashboard.ipynb` doesn't end with a newline.

### Verdict

All substantive issues from the first and second reviews are resolved. The error rate is now consistent across Summary and Reliability. The agent filter is centralized in the CTE. The `app.py` is no longer duplicated. **Approve.**

haiyuan-eng-google (approved):

PR #28, opened 2026-04-14
refactor: rename internal modules for clarity

libei:
## Summary
- Rename `loader.py` → `ontology_loader.py`
- Rename `compiler.py` → `graph_ddl_compiler.py`
- Rename `resolved_models.py` → `graph_ddl_models.py`
- Rename `test_compiler.py` → `test_graph_ddl_compiler.py`
- Update all internal imports (16 statements across 5 files)
- Public API unchanged

## Test plan
- [x] All 1525 tests pass
- [x] No stale references to old module names


caohy1988 (commented):
## PR Review

Clean mechanical rename. Three files renamed, 16 import paths updated across 5 files, no behavioral changes. The naming improvements are sensible:

| Old | New | Rationale |
|---|---|---|
| `loader.py` | `ontology_loader.py` | Disambiguates from `binding_loader.py` |
| `compiler.py` | `graph_ddl_compiler.py` | Distinguishes graph DDL (CREATE PROPERTY GRAPH) from table DDL (scaffold) |
| `resolved_models.py` | `graph_ddl_models.py` | Pairs with `graph_ddl_compiler.py` |

Git tracks all three as `similarity index 100%` (or 98% for the compiler, due to its own import updates). Public API via `__init__.py` is unchanged.

### One Nit

`test_scaffold.py` imports directly from the internal module:
```python
from bigquery_ontology.ontology_loader import load_ontology_from_string
```

This couples the test to the internal module name (which this PR just renamed). The public API already exports this:
```python
from bigquery_ontology import load_ontology_from_string
```

Pre-existing issue, not introduced here — but worth fixing while you're touching the import.

### Note on base branch

This PR targets `feat/scaffold-impl` (PR #27), which is closed. The merge path for this PR depends on how the scaffold work re-lands.

### Verdict

Looks good. No blocking issues.

src/bigquery_ontology/scaffold.py
```diff
@@ -0,0 +1,354 @@
...
+class NamingMode(str, Enum):
+    """Column and table naming convention applied to ontology identifiers."""
+
+    SNAKE = "snake"
+    PRESERVE = "preserve"
+
+
+def _to_snake(name: str) -> str:
+    """Convert PascalCase or camelCase to snake_case."""
+    s = re.sub(r"([a-z0-9])([A-Z])", r"\1_\2", name)
```
caohy1988: Bug: this implementation does not match the acronym handling the scaffold doc now promises. With the current regex, `HTTPRequest` becomes `httprequest`, not `http_request`. Either tighten the docs back to the simpler lower/digit-to-upper split, or update `_to_snake()` and add a regression test covering acronym runs.

src/bigquery_ontology/scaffold.py
```diff
@@ -0,0 +1,354 @@
...
+        ref = _table_ref(table_name, dataset, project)
+
+        from_entity = entity_map[rel.from_]
+        to_entity = entity_map[rel.to]
+        from_pk = from_entity.keys.primary
+        to_pk = to_entity.keys.primary
+
+        from_ep_props = {p.name: p for p in from_entity.properties}
+        to_ep_props = {p.name: p for p in to_entity.properties}
+
```
caohy1988: Bug: the docs now say scaffold should error if a relationship property collides with a generated endpoint column, but there is no collision check here. As written, a relationship property with the same name as a generated `from_*` or `to_*` column will produce duplicate column names in both the DDL and the binding stub. For example, a relationship property named `from_party_id` will collide with the generated endpoint column for a `party_id` key. This needs an explicit validation pass plus a regression test.

libei (commented):
Thanks for the reviews! Both comments (acronym handling in `_to_snake()` and endpoint column collision check) are valid bugs, but they're on `scaffold.py` which isn't part of this rename PR — it showed up from the old branch base. I'll address both on `feat/scaffold-impl` where scaffold actually lives.

PR #29, opened 2026-04-14
docs: add ontology user manual

libei:
## Summary
- Add a comprehensive user manual for the ontology system at `docs/ontology/user_manual.md`
- Covers the ontology YAML spec, binding YAML spec, `gm` CLI reference, compilation pipeline, and an end-to-end finance domain walkthrough
- All content is grounded in the actual source code (`bigquery_ontology` module), not derived from other docs

## Test plan
- [ ] Verify all YAML field names match Pydantic models in `ontology_models.py` and `binding_models.py`
- [ ] Verify CLI flags and exit codes match `cli.py` Typer definitions
- [ ] Verify validation rules match `ontology_loader.py` and `binding_loader.py`
- [ ] Verify DDL output format matches `graph_ddl_compiler.py` emission logic
- [ ] Read through end-to-end example for self-consistency

### DIFF
--- a/docs/ontology/user_manual.md
+++ b/docs/ontology/user_manual.md
@@ -1,6 +1,6 @@
 # Ontology User Manual
 
-## Overview
+## Part 1: Guide
 
 ### What Is an Ontology?
 
@@ -14,8 +14,6 @@ An **ontology** is a formal, explicit specification of a shared conceptualizatio
 The key distinction from a database schema is that an ontology captures **meaning**, not just structure. It answers "what does this data represent in the real world?" rather than "what columns does this table have?" This makes ontologies portable across different storage backends — the same ontology can be realized on BigQuery, Spanner, or any other data warehouse.
 
 ### Why Use an Ontology?
-
-Ontologies provide several benefits for data-driven systems:
 
 - **Shared vocabulary** — teams agree on what "Account", "Transaction", or "Security" means, preventing ambiguity across systems.
 - **Backend independence** — the logical model is defined once and can be mapped to different physical storage systems without changing the model itself.
@@ -36,8 +34,6 @@ ontology  +  binding  →  CREATE PROPERTY GRAPH DDL
 
 This separation means the same ontology can have multiple bindings — one per environment, backend, or data source — without duplicating the logical model.
 
-The workflow uses three artifacts:
-
 | Artifact | File convention | Purpose |
 |----------|----------------|---------|
 | Ontology | `*.ontology.yaml` | Logical graph schema (what exists) |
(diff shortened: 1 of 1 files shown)


docs/ontology/user_manual.md
```diff
@@ -0,0 +1,762 @@
...
+```yaml
+binding: finance-bq-prod
+ontology: finance
+target:
+  backend: bigquery
+  project: my-project
+  dataset: finance
+```
+
+The `ontology` field is the ontology's name (not a file path). When loading, the system looks for `finance.ontology.yaml` in the same directory.
```
libei: This is about how CLI works. Please leave this part out

caohy1988 (commented):
## PR Review

This is an excellent user manual. The guide-first/reference-second structure is well-chosen — a new user can read Part 1 sequentially to get productive, then use Part 2 as a lookup. The inheritance section is particularly thorough, including the validate-succeeds/compile-fails distinction that trips people up.

### Issues to Fix

**1. Ontology validation rule 3 claims cross-kind name disjointness that isn't enforced**

The manual states (§Ontology Validation Rules, rule 3): "Entity and relationship names are disjoint — no name can be used by both." This is also stated in the Entity fields table: "Unique within the ontology (disjoint from relationship names)."

The ontology loader checks entity names unique within entities and relationship names unique within relationships, but does NOT cross-validate between kinds. `_check_unique_names` is called separately per kind. If an entity and relationship share a name (e.g., both named `HOLDS`), the loader accepts it — only the compiler would eventually produce a DDL alias collision.

Either remove the cross-kind claim from the manual, or flag this as a gap that the implementation should fix.

**2. Binding validation rule 2 claims cross-kind uniqueness that isn't enforced**

The manual states: "Entity and relationship binding names are unique within the binding and across kinds." The binding loader's `_check_unique_binding_names` only checks within each kind (entities vs relationships), not across. Its docstring explains: "An entity and a relationship sharing a binding-level name cannot collide — the ontology already rejects that case across its own namespace." But per issue #1, the ontology doesn't enforce this either.

Drop "and across kinds" to match the implementation.

**3. `cli-output-error` rule code doesn't exist in the implementation**

The Error Rule Codes table lists `cli-output-error` ("Cannot write output file"). The compile command's output writing is a bare `output_path.write_text(ddl)` with no error handling — a write failure produces an uncaught exception, not a structured error. Either remove this row or implement the handling.

**4. "Golden test" developer jargon in user-facing docs**

Line ~517: "Properties where the column matches the logical name (like `person_id` in the golden test) render as bare names." The term "golden test" is developer jargon that belongs in test comments, not a user manual. Replace with a concrete reference from the walkthrough itself.

### Non-blocking Observations

**5. Cardinality default "unconstrained" is an editorial choice**

The reference table shows `Default: unconstrained` for cardinality. The spec simply marks it as optional with no semantic default. "Unconstrained" is a reasonable interpretation of "absent" (and correct for compiled DDL), but it's not sourced from the spec.

**6. Manual documents behavior from unmerged PRs**

The PR description says content is "grounded in the actual source code," but that code (PRs #16–#20) hasn't merged to main. The manual is accurate for the implementation as reviewed, but the test plan's ask to "verify against source code" can't be completed against main today. This resolves itself once the code PRs merge.

### What Looks Good

- **Guide section** teaches concepts progressively with minimal forward references — a reader can follow it linearly
- **Inheritance documentation** is the most thorough in the entire doc set — covers what children can/cannot do, covariant narrowing with examples, cardinality inheritance, binding implications, and the v0 compiler limitation
- **Relationship keys section** clearly explains the three cases (primary, additional, no keys) with YAML snippets that match the DDL walkthrough
- **End-to-end walkthrough** is complete and self-consistent: ontology → binding → validate → compile → DDL, with callouts explaining each notable feature in the output
- **Error rule codes table** is a useful reference for CI/tooling integration
- **Reference tables** are exhaustive with correct types, defaults, and context columns
- **Derived expression substitution** section clearly shows the before/after with arrows and explains nested parenthesization

### Verdict

Fix items 1–4 before merge. The manual is otherwise publication-ready.

docs/ontology/user_manual.md
```diff
@@ -0,0 +1,851 @@
...
+gm validate <file> [--json] [--ontology PATH]
+```
+
+| Argument/Option | Required | Default | Description |
+|-----------------|----------|---------|-------------|
+| `file` | yes | — | Path to an ontology or binding YAML file. Kind is auto-detected. |
+| `--json` | no | false | Emit structured JSON errors on stderr. |
+| `--ontology PATH` | no | auto | Companion ontology path. Default: `<ontology_name>.ontology.yaml` next to the binding. |
+
+Success: no output, exit 0. Failure: errors on stderr, exit 1 or 2.
```
caohy1988: This sentence contradicts the exit-code table below. `gm validate` can also exit 3 on the defensive internal-error path in `src/bigquery_ontology/cli.py` (`except Exception ... raise typer.Exit(code=3)`). If the manual is going to mention failure codes here, it should either include 3 as well or defer entirely to the dedicated exit-code table.

docs/ontology/user_manual.md
```diff
@@ -0,0 +1,851 @@
...
+| Rule | Source |
+|------|--------|
+| `ontology-shape:<type>` | Pydantic shape validation on ontology |
+| `ontology-validation` | Semantic validation on ontology |
+| `binding-shape:<type>` | Pydantic shape validation on binding |
+| `binding-validation` | Semantic validation on binding |
+| `compile-validation` | Compile-time validation (e.g. `extends` not supported in v0) |
+| `yaml-parse` | YAML syntax error |
+| `cli-missing-file` | File not found or not readable |
+| `cli-missing-ontology` | Companion ontology not found |
```
caohy1988: The rule description is narrower than the current implementation. In `src/bigquery_ontology/cli.py`, `cli-missing-ontology` is also used when the companion ontology path exists but is unreadable (`not os.access(...)`), and the CLI tests cover that path (`test_validate_binding_unreadable_ontology_flag_emits_structured_json`, `test_compile_unreadable_ontology_with_json_emits_structured_json`). If this manual is meant to mirror the shipped behavior, this row should say something like "missing or unreadable companion ontology" unless the code stack is going to split those cases.

docs/ontology/user_manual.md
```diff
@@ -0,0 +1,851 @@
...
+
+**Cycle detection:** The loader walks `extends` chains from every element and rejects any chain that revisits a name.
+
+**Compiler support:** The v0 compiler rejects all uses of `extends`. Validation accepts it; compilation does not. The error rule code is `compile-validation`.
+
+---
+
+### Ontology Validation Rules
+
+The ontology loader enforces 13 cross-element semantic rules beyond per-field shape validation:
```
caohy1988: The manual says this section is an exhaustive summary of the ontology loader's semantic rules, but it currently omits one rule the implementation does enforce: inherited relationship cardinality may not be changed by a child. `src/bigquery_ontology/ontology_loader.py` checks this in `_check_covariant_narrowing()` after the endpoint narrowing checks. Since the guide already explains cardinality inheritance earlier, the reference section should include it here too; otherwise the rule count / summary is incomplete.

docs/ontology/user_manual.md
libei: Done — removed the auto-discovery detail. The binding guide now just states that `ontology` is a name, not a file path.

docs/ontology/user_manual.md
libei: Good catch. Replaced the inline exit codes with a link to the dedicated exit codes table so there's a single source of truth.

docs/ontology/user_manual.md
libei: Fixed — updated to "Companion ontology not found or not readable" to match the implementation.

docs/ontology/user_manual.md
libei: Added rule 14 (cardinality inheritance). Count updated to 14. Note: the full inheritance section has been removed from the guide since `extends` doesn't compile yet — tracked in #30.

PR #31, opened 2026-04-14
docs: move ontology user manual into package

libei:
## Summary
- Move `docs/ontology/user_manual.md` → `src/bigquery_ontology/docs/user_manual.md`
- The `bigquery_ontology` module is heading toward becoming a standalone package, so the manual should travel with it

## Test plan
- [ ] Verify the file renders correctly at the new path on GitHub
- [ ] Confirm no broken links reference the old path


caohy1988 (commented):
## PR Review

Pure file move: `docs/ontology/user_manual.md` → `src/bigquery_ontology/docs/user_manual.md`. Similarity index 100%, no content changes. No references to `user_manual` exist elsewhere in the repo, so nothing breaks.

### Two observations worth considering

**1. The docs directory will ship inside the wheel**

`pyproject.toml` has `packages = ["src/bigquery_ontology"]`, so `src/bigquery_ontology/docs/user_manual.md` will be included in the distributed wheel. This is fine if intentional (some packages ship docs for `help()` or offline reference), but it adds ~40KB to every install. If the goal is just co-location in the git repo, a `tool.hatch.build.targets.wheel.exclude` pattern for `docs/` would keep it out of the distribution while keeping it in the source tree.

**2. Design docs stay in `docs/ontology/` while the user manual moves**

After this PR, the ontology docs are split across two locations:
- `docs/ontology/` — spec/design docs (`ontology.md`, `binding.md`, `cli.md`, `compilation.md`, `scaffold.md`)
- `src/bigquery_ontology/docs/` — user manual only

If the package is heading toward standalone status, the design docs would eventually need to move too (or be replaced by the user manual as the canonical reference). The split is fine for now but worth being deliberate about — a contributor looking for ontology docs might not think to check inside `src/`.

### Verdict

Looks good to merge. The wheel-inclusion question is the only thing worth a quick decision on.

libei:
Addressed both observations:

1. **Wheel exclusion** — added `exclude = ["src/bigquery_ontology/docs"]` to `[tool.hatch.build.targets.wheel]` in pyproject.toml. The docs stay in the source tree for co-location but won't ship in installed wheels.

2. **Split docs locations** — agreed this is a temporary state. When the package goes standalone, the design docs in `docs/ontology/` will either move into the package or be superseded by the user manual. Leaving that for the package split work.

PR #33, opened 2026-04-14
feat(scaffold): add scaffold generator for table DDL and binding stubs

libei:
## Summary

- Add `scaffold()` function that generates BigQuery `CREATE TABLE` DDL and a matching `binding.yaml` stub from an ontology — for greenfield projects with no existing tables.
- Supports all three relationship key modes (no keys, `keys.primary`, `keys.additional`), compound entity keys, derived property exclusion, `--naming` (snake/preserve), and `--project` qualification.
- Keyless edge tables emit a commented-out `PRIMARY KEY` with a TODO, per [BQ best practices](https://docs.cloud.google.com/bigquery/docs/graph-schema-best-practices), so users can opt into the query optimizer hint if their data is unique.
- Column definitions are padded for vertical alignment, matching BQ documentation style.
- CLI wiring (`gm scaffold` command) is deferred to a follow-up PR.

## Test plan

- [x] 36 unit tests covering snake-case conversion, type mapping, entity DDL, relationship DDL (all three key modes), endpoint-column collision detection, binding YAML generation, round-trip validation, determinism, `extends` rejection, suggested PK comments
- [x] Round-trip test: emitted binding parses successfully via `load_binding_from_string`
- [x] Full existing test suite (134 tests) passes with zero regressions

### DIFF
--- a/src/bigquery_ontology/scaffold.py
+++ b/src/bigquery_ontology/scaffold.py
@@ -40,8 +40,8 @@ after generation.
 
 from __future__ import annotations
 
-import re
 from dataclasses import dataclass
+import re
 
 from .ontology_models import Entity
 from .ontology_models import Keys
@@ -53,9 +53,7 @@ from .ontology_models import Relationship
 # Naming
 # ---------------------------------------------------------------------------
 
-_SNAKE_RE = re.compile(
-    r"(?<=[a-z0-9])([A-Z])|(?<=[A-Z])([A-Z][a-z])"
-)
+_SNAKE_RE = re.compile(r"(?<=[a-z0-9])([A-Z])|(?<=[A-Z])([A-Z][a-z])")
 
 
 def _to_snake_case(name: str) -> str:
@@ -167,20 +165,24 @@ def _resolve_entity_table(
 
   for pk_name in entity.keys.primary:
     prop = prop_map[pk_name]
-    pk_columns.append(_ScaffoldColumn(
-        name=_apply_naming(prop.name, naming),
-        bq_type=_ONTOLOGY_TO_BQ_TYPE[prop.type],
-        not_null=True,
-    ))
+    pk_columns.append(
+        _ScaffoldColumn(
+            name=_apply_naming(prop.name, naming),
+            bq_type=_ONTOLOGY_TO_BQ_TYPE[prop.type],
+            not_null=True,
+        )
+    )
 
   for prop in entity.properties:
     if prop.name in pk_set or prop.expr is not None:
       continue
-    other_columns.append(_ScaffoldColumn(
-        name=_apply_naming(prop.name, naming),
-        bq_type=_ONTOLOGY_TO_BQ_TYPE[prop.type],
-        not_null=False,
-    ))
+    other_columns.append(
+        _ScaffoldColumn(
+            name=_apply_naming(prop.name, naming),
+            bq_type=_ONTOLOGY_TO_BQ_TYPE[prop.type],
+            not_null=False,
+        )
+    )
 
   table_name = _qualify(dataset, _apply_naming(entity.name, naming), project)
   columns = tuple(pk_columns + other_columns)
@@ -209,11 +211,13 @@ def _endpoint_columns(
   for pk_name in entity.keys.primary:
     prop = prop_map[pk_name]
     col_name = f"{prefix}_{_apply_naming(prop.name, naming)}"
-    cols.append(_ScaffoldColumn(
-        name=col_name,
-        bq_type=_ONTOLOGY_TO_BQ_TYPE[prop.type],
-        not_null=True,
-    ))
+    cols.append(
+        _ScaffoldColumn(
+            name=col_name,
+            bq_type=_ONTOLOGY_TO_BQ_TYPE[prop.type],
+            not_null=True,
+        )
+    )
   return cols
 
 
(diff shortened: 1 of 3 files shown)


src/bigquery_ontology/scaffold.py
```diff
@@ -0,0 +1,497 @@
...
+  lines: list[str] = []
+  lines.append(
+      "# Generated by gm scaffold. "
+      "This file is user-owned \u2014 edit freely."
+  )
+  lines.append(f"binding: {dataset}")
+  lines.append(f"ontology: {ontology.ontology}")
+  lines.append("target:")
+  lines.append("  backend: bigquery")
+  if project:
```
caohy1988, line 421: Bug: when `project` is falsy, the emitted binding has no `project:` field. But `BigQueryTarget` on main requires `project: str` — the binding will fail `load_binding_from_string` with a Pydantic validation error. The round-trip test always passes `project="my_project"` so this path is never validated.

Either make `BigQueryTarget.project` optional, or make `project` required on `scaffold()`.

caohy1988 (commented):
## PR Review

Well-implemented scaffold that follows the resolve→emit pattern established by the compiler. The three relationship key modes, compound entity keys, column padding, and the suggested-PK comment for keyless edges are all handled correctly. Test coverage is thorough with 36 tests covering every dimension.

### Bug

**1. No-project scaffold produces a binding that fails `load_binding_from_string`**

When `--project` is omitted, the binding emitter skips the `project:` field entirely. But `BigQueryTarget` (on main, from PR #16) has `project: str` as a required field — Pydantic will reject the binding at parse time.

The round-trip test always passes `project="my_project"`, so it never exercises the no-project path:

```python
def test_binding_round_trip(self):
    _, binding_text = scaffold(ontology, dataset="my_dataset", project="my_project")
    binding = load_binding_from_string(binding_text, ontology=ontology)  # passes
```

But `test_person_follows_binding` asserts exact output without `project` and never round-trips:
```python
def test_person_follows_binding(self):
    _, binding = scaffold(ontology, dataset="my_dataset")  # no project
    assert binding == expected  # passes, but binding is invalid
```

Fix options:
- Make `BigQueryTarget.project` optional (schema change), or
- Make `--project` required on `scaffold()`, or
- Emit a placeholder `project: ""` (ugly), or
- Add a round-trip test without `project` — it will fail and surface the issue

### Non-blocking Observations

**2. Suggested PK comment for keyless edges is a nice UX touch**

The `-- TODO: uncomment if (...) is unique per row` pattern gives users the BQ optimizer hint without implying false uniqueness. The three tests (one positive, two negative) verify it appears only for the no-keys case. Well done.

**3. Column padding for vertical alignment**

Matching BQ documentation style. `_pad_columns` is clean and handles varying name lengths correctly.

**4. `assert` for invariant checks**

Same pattern as the compiler — consistent across the codebase. Noted but not blocking.

### What Looks Good

- **Snake-case converter** handles all edge cases correctly: `HTTPRequest→http_request`, `XMLParser→xml_parser`, `parseJSON→parse_json`, `myURL→my_url`. 11 parametrized test cases.
- **Type mapping** covers all 11 ontology types with a test that asserts complete coverage via `for pt in PropertyType`
- **Derived property exclusion** from both DDL and binding, tested for both entities and relationships
- **Endpoint-column collision detection** with a clear error message naming the conflicting property and column
- **Binding YAML reads endpoint column names from resolved `_ScaffoldFK` objects** (not recalculated), eliminating divergence risk between DDL and binding
- **Round-trip test** validates the binding parses successfully via `load_binding_from_string`
- **Determinism test** guards against ordering instability
- **Extends rejection** tested for both entities and relationships
- **Comment header** in generated binding reinforces the non-idempotent "user-owned" contract

### Verdict

Fix the no-project binding validity issue (#1) before merge. Everything else is clean.

src/bigquery_ontology/scaffold.py
```diff
@@ -0,0 +1,497 @@
...
+) -> _ScaffoldEntityTable:
+  assert entity.keys is not None and entity.keys.primary is not None
+  pk_set = set(entity.keys.primary)
+
+  pk_columns: list[_ScaffoldColumn] = []
+  other_columns: list[_ScaffoldColumn] = []
+
+  prop_map = {p.name: p for p in entity.properties}
+
+  for pk_name in entity.keys.primary:
```
caohy1988, line 167: Bug: scaffold still accepts key properties that are derived (`expr:`), but the binding emitter later excludes all derived properties from `binding.yaml`. That creates an invalid round-trip: an entity `keys.primary: [full_name]` (where `full_name` is derived) will scaffold successfully here, but `load_binding_from_string()` will reject the emitted binding because derived properties must not appear in bindings. `keys.additional` has a related issue below: if the additional-key property is derived, it never reaches `prop_columns`, so the generated PK tuple silently drops part of the intended key. This needs an explicit scaffold-time rejection of derived key properties (or a loader-level rule that keys must reference stored properties), plus a regression test.

src/bigquery_ontology/scaffold.py
libei: Fixed — made `project` a required parameter on `scaffold()` (and all internal helpers: `_qualify`, `_resolve_entity_table`, `_resolve_rel_table`, `_emit_binding_yaml`). The `if project:` guard is removed; the binding YAML now always emits `project:` and all table names use 3-part qualification. Updated all tests to pass `project=`.

src/bigquery_ontology/scaffold.py
libei: Fixed — added `_reject_derived_keys()` that checks all primary and additional key columns are stored (not derived). Called from both `_resolve_entity_table` and `_resolve_rel_table` at resolve time, before any column generation. Raises `ValueError` with a clear message if a key column has `expr`.

Added three regression tests:
- entity with derived primary key → rejected
- relationship with derived `keys.primary` → rejected  
- relationship with derived `keys.additional` → rejected

PR #34, opened 2026-04-14
feat(scaffold): add gm scaffold CLI command and user guide

libei:
## Summary

- Add `gm scaffold` CLI command with `--ontology`, `--dataset`, `--out`, `--naming`, `--project`, and `--json` flags
- Non-empty output directory is refused with `cli-non-empty-dir` error; file write failures produce structured `cli-output-error`
- Add scaffold section to user manual: guide (naming table, greenfield workflow), CLI reference, and 3 new error rule codes

Depends on #33 (`feat/scaffold` — scaffold core logic).

## Test plan

- [x] 10 CLI tests: happy path (files written, nested dirs, --project, --naming preserve), error paths (non-empty dir, missing ontology, invalid naming, extends, --json), and scaffold-then-compile round-trip integration
- [x] Full test suite (144 tests) passes with zero regressions

### DIFF
--- a/src/bigquery_ontology/cli.py
+++ b/src/bigquery_ontology/cli.py
@@ -17,9 +17,11 @@
 ``gm validate`` accepts either an ontology YAML or a binding YAML and
 dispatches to the matching loader. ``gm compile`` takes a binding YAML
 and emits the corresponding ``CREATE PROPERTY GRAPH`` DDL on stdout
-(or to ``-o PATH``). Both commands resolve a binding's companion
-ontology by auto-discovering ``<name>.ontology.yaml`` next to the
-binding; ``--ontology PATH`` overrides that lookup.
+(or to ``-o PATH``). ``gm scaffold`` generates starter ``CREATE TABLE``
+DDL and a matching binding stub from an ontology. Both ``validate`` and
+``compile`` resolve a binding's companion ontology by auto-discovering
+``<name>.ontology.yaml`` next to the binding; ``--ontology PATH``
+overrides that lookup.
 
 Exit codes:
 
@@ -48,10 +50,11 @@ from .graph_ddl_compiler import compile_graph
 from .ontology_loader import load_ontology
 from .ontology_loader import load_ontology_from_string
 from .ontology_models import Ontology
+from .scaffold import scaffold
 
 app = typer.Typer(
     name="gm",
-    help="Graph-model CLI. Commands: validate, compile.",
+    help="Graph-model CLI. Commands: validate, compile, scaffold.",
     add_completion=False,
     no_args_is_help=True,
 )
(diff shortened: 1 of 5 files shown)


caohy1988 (commented):
## PR Review

Clean CLI wiring that follows the established patterns from `gm validate` and `gm compile`. The error routing, input validation sequence, and user manual additions are all well-structured. The scaffold-then-compile round-trip integration test is the highlight.

### No Blocking Issues

Two minor observations:

**1. `cli-output-error` uses exit code 1 — consider exit code 2**

The file-write error handler exits with code 1 (validation/compilation error). But a filesystem write failure is arguably a usage-level error (exit 2) — the user's filesystem state is wrong, not the YAML content. This is a judgment call; either is defensible. Just noting the inconsistency with `cli-missing-file` (exit 2) and `cli-non-empty-dir` (exit 2) which are similar "your filesystem is wrong" errors.

**2. Round-trip integration test uses `project="p"`, avoiding the no-project bug from PR #33**

The `test_scaffold_then_compile_succeeds` test always passes `--project p`, which means the generated binding includes a `project:` field and round-trips correctly through `gm compile`. If someone removes the `--project` flag, the test would still pass (scaffold succeeds) but the binding would fail `gm compile` due to the missing required `project` in `BigQueryTarget`. This is PR #33's issue, not this PR's — just noting the coupling.

### What Looks Good

- **Input validation sequence** is thorough and ordered correctly: naming → file existence → readability → directory emptiness → ontology load → scaffold → file write. Each step has its own rule code.
- **`scaffold-validation` rule prefix** correctly distinguishes scaffold-time errors (extends, endpoint collision) from loader and compiler errors. Consistent with the `ontology-validation` / `binding-validation` / `compile-validation` pattern.
- **`cli-output-error` handler** wraps `mkdir` + file writes in a try/except — this was flagged as missing in my PR #29 user manual review. Now implemented.
- **`cli-non-empty-dir`** refuses to overwrite with a clear message telling users to delete/move contents. Safety first for a one-shot generator.
- **`cli-usage`** for invalid `--naming` values — validates against a set rather than relying on Typer's `click.Choice`, which gives a structured JSON-compatible error.
- **Readability check** (`os.access`) is more defensive than validate/compile — a nice addition.
- **Round-trip integration test** (`test_scaffold_then_compile_succeeds`) validates the full arc: ontology → scaffold → binding → compile → DDL. This is the most valuable test in the file.
- **User manual additions** are well-placed: guide section with naming table and numbered workflow, CLI reference with flag table, 3 new error rule codes (`cli-usage`, `cli-non-empty-dir`, `scaffold-validation`).
- **10 tests** cover all happy paths (basic, nested dirs, project, preserve naming) and error paths (non-empty dir, missing ontology, invalid naming, extends, JSON output).

### Verdict

Looks good to merge (after PR #33 lands). No changes needed.

src/bigquery_ontology/cli.py
```diff
@@ -438,6 +441,167 @@ def compile_command(
...
+                "severity": "error",
+                "message": f"File not readable: {ontology_path}",
+            }
+        ],
+        as_json=json_output,
+    )
+    raise typer.Exit(code=2)
+
+  out_path = Path(out)
+  if out_path.exists() and any(out_path.iterdir()):
```
caohy1988: Bug: this assumes `--out` is either a directory or absent. If the path already exists as a regular file, `out_path.iterdir()` raises `NotADirectoryError` before the command reaches the structured `cli-output-error` path below, so the user gets an unhandled internal failure instead of a normal CLI diagnostic. This needs an `is_dir()` guard (or a dedicated pre-check for `exists() and not is_dir()`) before calling `iterdir()`, plus a regression test for `--out existing-file`.

libei: Fixed — added an `is_dir()` pre-check before `iterdir()`. If `--out` exists and is not a directory, the CLI now emits a structured `cli-output-error` diagnostic and exits with code 2, matching the other error paths. Added a regression test (`test_scaffold_out_is_file_is_usage_error`).

PR #35, opened 2026-04-15
feat(owl-import): add OWL-to-ontology importer core logic

libei:
## Summary

- Add `import_owl()` function that converts OWL sources (Turtle/RDF-XML) into ontology YAML, implementing the four-stage pipeline from `owl-import.md`: parse, filter, map, emit
- Covers all design doc mapping rules: entity/relationship extraction, XSD type mapping, `owl:hasKey` → primary keys, single/multi-parent inheritance, `FILL_IN` placeholders for ambiguities, drop annotations for unsupported OWL features
- Robust YAML emission: handles booleans, nulls, numbers, special characters, quote/backslash escaping
- Validates: name collisions (entity-entity, entity-relationship, property duplicates), missing domain/range → `FILL_IN`
- English label preference per §9, namespace filtering with exclusion reporting per §4
- `rdflib>=7.0` added as optional dependency (`pip install bigquery-agent-analytics[owl]`)
- CLI wiring (`gm import-owl`) deferred to follow-up PR

## Test plan

- [x] 39 tests across 3 classes: `yamo_sample.ttl` fixture, design doc §18 worked example, and edge cases
- [x] Edge cases: namespace filtering, multi-parent FILL_IN, functional property cardinality, missing domain/range, relationship extends, name collisions, duplicate properties, YAML special value quoting, English label preference
- [x] Full test suite (137 tests) passes with zero regressions

### DIFF
--- a/src/bigquery_ontology/owl_importer.py
+++ b/src/bigquery_ontology/owl_importer.py
@@ -433,6 +433,13 @@ def _extract_datatype_properties(
 
     for domain_name in domains:
       if domain_name in entities:
+        existing = {p.name for p in entities[domain_name].properties}
+        if prop_name in existing:
+          raise ValueError(
+              f"Duplicate property {prop_name!r} on entity "
+              f"{domain_name!r}. Two OWL properties with the same "
+              "local name share a domain."
+          )
         entities[domain_name].properties.append(imported_prop)
 
 
@@ -571,11 +578,52 @@ def _resolve_keys(entities: dict[str, _ImportedEntity]) -> None:
 # ---------------------------------------------------------------------------
 
 
+_YAML_BOOL_NULL = frozenset(
+    {
+        "true",
+        "false",
+        "yes",
+        "no",
+        "on",
+        "off",
+        "True",
+        "False",
+        "Yes",
+        "No",
+        "On",
+        "Off",
+        "TRUE",
+        "FALSE",
+        "YES",
+        "NO",
+        "ON",
+        "OFF",
+        "null",
+        "Null",
+        "NULL",
+        "~",
+    }
+)
+
+_YAML_NEEDS_QUOTE_CHARS = frozenset(":{}[],\"'#\n\\*&!%@`?|>-")
+
+
 def _yaml_scalar(value: str) -> str:
-  if any(
-      c in value
-      for c in (":", "{", "}", "[", "]", ",", "#", "'", '"', "\n", "\\")
-  ):
+  if not value:
+    return '""'
+  needs_quote = (
+      value in _YAML_BOOL_NULL
+      or any(c in _YAML_NEEDS_QUOTE_CHARS for c in value)
+      or value[0] in (" ", "\t")
+      or value[-1] in (" ", "\t")
+  )
+  if not needs_quote:
+    try:
+      parsed = float(value)
+      needs_quote = True
+    except ValueError:
+      pass
+  if needs_quote:
     escaped = value.replace("\\", "\\\\").replace('"', '\\"')
     return f'"{escaped}"'
   return value
@@ -777,6 +825,14 @@ def import_owl(
   _extract_datatype_properties(g, entities, include_namespaces, drops)
   relationships = _extract_relationships(g, entities, include_namespaces, drops)
   _resolve_keys(entities)
+
+  overlap = set(entities) & set(relationships)
+  if overlap:
+    names = ", ".join(sorted(overlap))
+    raise ValueError(
+        f"Name collision between entities and relationships: {names}. "
+        "Entity and relationship names must be disjoint."
+    )
 
   if ontology_name is None:
     ns = include_namespaces[0].rstrip("#/")
--- a/src/bigquery_ontology/owl_importer.py
+++ b/src/bigquery_ontology/owl_importer.py
@@ -1,0 +1,735 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


caohy1988 (commented):
## PR Review

Substantial and well-implemented OWL importer. The four-stage pipeline (parse → filter → map → emit) is clean, the FILL_IN placeholder system handles all the OWL-to-ontology ambiguities correctly, and the YAML scalar quoting is hardened against booleans, nulls, numbers, and special characters. The 39 tests cover the fixture, design doc example, and a thorough set of edge cases.

### Issues

**1. `import_owl` not exported from `__init__.py`**

The PR adds `owl_importer.py` but doesn't update `__init__.py` to export `import_owl`. Tests import directly from the module (`from bigquery_ontology.owl_importer import import_owl`). Every other public function in the package (`scaffold`, `compile_graph`, `load_ontology`, etc.) is exported from the package root. Either:
- Add the export now (consistent with the rest of the package), or
- Defer to the CLI-wiring PR (but note this in the PR description)

**2. `xsd:dateTime` maps to `timestamp` — potentially lossy**

The mapping sends both `xsd:dateTime` and `xsd:dateTimeStamp` to `timestamp`. OWL's `xsd:dateTime` allows but doesn't require timezone info — it could map to `datetime` (timezone-less) instead. The current choice is pragmatic (over-specifying is safer than losing timezone data), but a user with genuinely timezone-free datetime values will get a mismatch. Worth a YAML comment or `xsd_type` annotation on the imported property so the user knows to check.

### Non-blocking Observations

**3. Multiline descriptions will produce ugly YAML**

`_extract_labels_and_description` concatenates labels and comments with `\n\n`, then `_yaml_scalar` double-quotes the result (because `\n` is in `_YAML_NEEDS_QUOTE_CHARS`). This produces `"line1\n\nline2"` which is technically valid YAML but renders as a single-line string with literal `\n` characters. YAML's block scalar style (`|`) would be more readable. Not blocking for v0 since most OWL labels are single-line, but worth a TODO.

**4. Entity-relationship name disjointness enforced here but not in the ontology loader**

The importer correctly checks `set(entities) & set(relationships)` and rejects collisions. This is a compile-time concern the ontology loader doesn't enforce (as flagged in my PR #29 review). The importer catching it at import time is good — just noting the gap remains in the loader for hand-authored ontologies.

### What Looks Good

- **rdflib lazy import** with a clear error message pointing to the `[owl]` extra. Exactly the right pattern for an optional dependency.
- **XSD type mapping** is comprehensive: 27 XSD types → 11 ontology types, with `anyURI` getting a `xsd_type: anyURI` annotation so the user knows it was mapped to `string`
- **FILL_IN placeholders** cover all the right ambiguities: multi-parent (`rdfs:subClassOf` with 2+ parents), missing `owl:hasKey`, multi-range datatype properties, missing/multi domain/range on relationships
- **Drop annotations** preserve `owl:disjointWith`, `owl:equivalentClass`, `owl:inverseOf`, property characteristics (Transitive, Symmetric, etc.), and restrictions — all as structured annotations or YAML comments
- **Functional property → `many_to_one` cardinality** is the correct OWL mapping
- **YAML scalar quoting** is thorough: booleans (`true/false/yes/no/on/off`), nulls (`null/~/Null`), numbers, special leading characters, empty strings, quote/backslash escaping. 7 dedicated tests.
- **Name collision detection** across three dimensions: entity-entity (same local name from different namespaces), relationship-relationship, entity-relationship overlap
- **English label preference** (`@en*` first, then untagged, alphabetical tiebreak) per design doc §9
- **Canonical alphabetical ordering** in emitted YAML for entities, relationships, properties, annotations, synonyms — deterministic output
- **`_extract_parents`** handles excluded-by-namespace parents as drop annotations rather than silently ignoring them
- **Design doc worked example test** validates the finance OWL → YAML conversion end-to-end including inheritance, drop annotations, and determinism
- **`_resolve_keys`** correctly skips entities with `extends` (they inherit keys) and only adds FILL_IN to root entities without `owl:hasKey`

### Verdict

Fix item #1 (`__init__.py` export) for consistency. Item #2 is worth a comment in the code. Everything else is clean — looks good to merge.

src/bigquery_ontology/owl_importer.py
```diff
@@ -0,0 +1,844 @@
...
+  elif len(parents) > 1:
+    parents.sort()
+    return None, True, parents, annotations
+  return None, False, [], annotations
+
+
+def _extract_keys(
+    g: Graph, class_iri: URIRef, namespaces: list[str]
+) -> tuple[list[str] | None, bool, list[str]]:
+  for key_list in g.objects(class_iri, OWL.hasKey):
```
caohy1988, line 259: One more edge case worth surfacing: this silently drops `owl:hasKey` members that fall outside the namespace filter. If a class *does* declare `owl:hasKey`, but every key property is excluded by namespace, `_extract_keys()` returns `(None, False, [])` and `_resolve_keys()` later rewrites that into the generic `# no owl:hasKey in OWL source` / `primary: [FILL_IN]` path. That loses the real reason for the ambiguity and points the user at the wrong remediation. This seems like it should mirror the parent/domain/range handling: preserve an annotation or comment that the key existed but was excluded by namespace, and ideally add a regression test for that filtered-key case.

haiyuan-eng-google:
fix the CI failure

src/bigquery_ontology/owl_importer.py
libei: Good catch — fixed. `_extract_keys()` now tracks excluded key IRIs and surfaces them via:

1. A distinct YAML comment: `# owl:hasKey declared but key properties excluded by namespace filter` (instead of the generic "no owl:hasKey in OWL source")
2. An `owl:hasKey_excluded` annotation listing the filtered IRIs

Added a regression test (`test_haskey_excluded_by_namespace`) that declares `owl:hasKey` pointing to a property in an external namespace and verifies both the comment and annotation appear.

PR #36, opened 2026-04-15
feat(owl-import): OWL-to-ontology importer with CLI and user guide

libei:
## Summary

- Add `owl_importer.py` module that converts OWL ontologies (Turtle/RDF-XML) into `ontology.yaml` format via a 4-stage pipeline (parse, filter, map, emit)
- Wire up `gm import-owl` CLI command with `--include-namespace`, `--format`, `-o`/`--out`, and `--json` flags
- Add "Importing from OWL" guide section and `gm import-owl` CLI reference to the user manual
- Handle namespace filtering, FILL_IN placeholders for ambiguities, and drop annotations for unsupported OWL features
- rdflib is an optional dependency (`pip install 'bigquery-agent-analytics[owl]'`)

## Test plan

- [x] 38 CLI tests pass (10 new import-owl tests covering golden path, output routing, drop summary, error formatting, missing dependency)
- [x] 110 owl_importer unit tests pass (24 test classes covering mapping, edge cases, determinism)
- [x] Manual verification: `gm import-owl tests/fixtures/yamo_sample.ttl --include-namespace "https://example.com/yamo#"` produces valid YAML
- [x] Verify `gm import-owl --help` shows correct usage
- [x] Verify `gm validate` fails on output with FILL_IN placeholders (expected)

### DIFF
--- a/src/bigquery_ontology/cli.py
+++ b/src/bigquery_ontology/cli.py
@@ -777,7 +777,7 @@ def import_owl_command(
     resolved_output = Path(output_path)
     try:
       resolved_output.write_text(yaml_text, encoding="utf-8")
-    except (FileNotFoundError, PermissionError) as exc:
+    except OSError as exc:
       _emit_errors(
           [
               {
--- a/src/bigquery_ontology/cli.py
+++ b/src/bigquery_ontology/cli.py
@@ -17,16 +17,18 @@
 ``gm validate`` accepts either an ontology YAML or a binding YAML and
 dispatches to the matching loader. ``gm compile`` takes a binding YAML
 and emits the corresponding ``CREATE PROPERTY GRAPH`` DDL on stdout
-(or to ``-o PATH``). Both commands resolve a binding's companion
-ontology by auto-discovering ``<name>.ontology.yaml`` next to the
-binding; ``--ontology PATH`` overrides that lookup.
+(or to ``-o PATH``). ``gm import-owl`` reads OWL source files and
+emits ``ontology.yaml``. Both ``validate`` and ``compile`` resolve a
+binding's companion ontology by auto-discovering
+``<name>.ontology.yaml`` next to the binding; ``--ontology PATH``
+overrides that lookup.
 
 Exit codes:
 
   0 — success
   1 — validation / compilation error
   2 — usage error (bad flag, missing file, missing companion ontology,
-      compile invoked on a non-binding file)
+      compile invoked on a non-binding file, missing dependency)
   3 — internal error
 """
 
@@ -51,7 +53,7 @@ from .ontology_models import Ontology
 
 app = typer.Typer(
     name="gm",
-    help="Graph-model CLI. Commands: validate, compile.",
+    help="Graph-model CLI. Commands: validate, compile, import-owl.",
     add_completion=False,
     no_args_is_help=True,
 )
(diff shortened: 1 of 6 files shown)


caohy1988 (commented):
## PR Review

Clean CLI wiring for `gm import-owl` that follows all established patterns, plus an excellent user manual guide section. The `owl_importer.py` delta from PR #35 improves excluded-key handling. I've already reviewed the core importer logic in PR #35 — this review focuses on the CLI, user manual, and delta.

### One Item Carried Forward from PR #35

**`import_owl` still not exported from `__init__.py`**

Every other public function (`scaffold`, `compile_graph`, `load_ontology`, `load_binding`) is exported from the package root. `import_owl` is only accessible via `from bigquery_ontology.owl_importer import import_owl`. The CLI uses a lazy `from .owl_importer import import_owl` which is correct for the optional-dependency pattern — but programmatic users shouldn't need to know the internal module name. Either add a lazy export to `__init__.py` (guarded by try/except) or document that it's import-from-module only.

### Non-blocking Observations

**1. Lazy import catch assumes rdflib is the only possible failure**

The CLI catches `ImportError` from `from .owl_importer import import_owl` and tells the user to install rdflib. If the import fails for any other reason (e.g., a bug in owl_importer.py), the user gets a misleading error. Unlikely in practice, but a more precise check (e.g., catching and inspecting the exception message) would be more robust.

**2. `owl_importer.py` delta: `keys_excluded` handling is a good improvement**

When `owl:hasKey` references properties outside the namespace filter, the importer now:
- Adds an `owl:hasKey_excluded` annotation listing what was excluded
- Uses a more informative FILL_IN comment ("declared but key properties excluded by namespace filter" vs generic "no owl:hasKey")
- New test `test_haskey_excluded_by_namespace` covers the edge case

This is strictly better than PR #35's version.

### What Looks Good

- **Lazy import pattern** — `from .owl_importer import import_owl` inside the command handler means the CLI module loads without rdflib, and ImportError only fires when `gm import-owl` is actually invoked. Correct for an optional dependency.
- **`_FORMAT_MAP`** abstracts user-facing format names ("ttl", "rdfxml") from rdflib format strings ("turtle", "xml"). Clean.
- **`import-validation` rule** correctly routes importer ValueErrors (name collisions, etc.) with the established `*-validation` pattern
- **Drop summary goes to stderr** so stdout can be piped to a file cleanly
- **`--include-namespace` is required and repeatable** — enforced by Typer. Matches the design doc.
- **`test_import_owl_missing_rdflib_exits_2`** uses `sys.modules` patching to simulate missing rdflib without actually uninstalling it. Clever and correct.
- **`_requires_rdflib` skip marker** cleanly handles CI environments where `[owl]` extras aren't installed
- **User manual guide section** is the best part of this PR:
  - "What maps to what" table gives immediate OWL→ontology correspondence
  - Namespace filtering section explains WHY it matters with a practical FIBO example
  - FILL_IN placeholder section shows exactly what the user will see (3 concrete YAML examples) and what to do
  - Drop annotations section distinguishes machine-readable annotations from human-readable comments
- **CLI reference table** is complete: required/optional, defaults, descriptions
- **3 new error rule codes** (`cli-missing-dependency`, `import-validation`, updated `cli-usage`) are documented in the manual
- **10 CLI tests** cover golden path (stdout + file output), drop summary, format flag, multiple namespaces, error paths (missing file, bad format, output error, missing rdflib), and JSON error output

### Verdict

Looks good to merge. The `__init__.py` export is the only item to address — and it's the same one from PR #35. Everything else is clean.

src/bigquery_ontology/cli.py
```diff
@@ -619,6 +620,182 @@ def scaffold_command(
...
+    typer.echo(f"internal error: {exc}", err=True)
+    raise typer.Exit(code=3)
+
+  if drop_summary:
+    typer.echo(drop_summary, err=True)
+
+  if output_path is not None:
+    resolved_output = Path(output_path)
+    try:
+      resolved_output.write_text(yaml_text, encoding="utf-8")
```
caohy1988, line 779: Bug: this only catches `FileNotFoundError` and `PermissionError`. If `-o/--out` points at an existing directory, `write_text()` raises `IsADirectoryError` (an `OSError`) and the command falls through to the generic internal-error path instead of returning a structured `cli-output-error`. The same thing can happen for other write-time `OSError`s. This should either pre-check `resolved_output.is_dir()` or widen the handler to include `OSError`, plus add a regression test for `-o existing-directory`.

libei: Good catch. Widened the handler from `(FileNotFoundError, PermissionError)` to `OSError` and added a regression test for `-o existing-directory`. Fixed in 546025f.

PR #37, opened 2026-04-15
docs(user-manual): add CLI installation section

libei:
## Summary
- The user manual described `gm` commands but never explained how to install them
- Added an "Installing the CLI" section before "Validating and Compiling" with `pip install` instructions, verification step, editable-mode for development, and the OWL extra
- Added a TOC entry for the new section

## Test plan
- [x] Verify the markdown renders correctly
- [x] Confirm TOC anchor link resolves to the new section

### DIFF
--- a/src/bigquery_ontology/docs/user_manual.md
+++ b/src/bigquery_ontology/docs/user_manual.md
@@ -18,6 +18,7 @@
   - [Structure](#structure)
   - [Mapping entities](#mapping-entities)
   - [Mapping relationships](#mapping-relationships)
+- [Installing the CLI](#installing-the-cli)
 - [Validating and Compiling](#validating-and-compiling)
   - [Validate](#validate)
   - [Compile](#compile)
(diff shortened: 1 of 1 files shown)


caohy1988 (commented):
## PR Review

Good catch — the manual described CLI commands without ever explaining how to install them. The section is well-placed (before "Validating and Compiling") and covers the four cases a user would hit: standard install, verification, dev mode, and the OWL extra.

### No blocking issues.

One minor suggestion: the `pip install bigquery-agent-analytics[owl]` line should be quoted in bash since `[` and `]` are shell glob characters. Most shells handle this fine, but zsh users will get a `no matches found` error without quotes:

```bash
pip install 'bigquery-agent-analytics[owl]'
```

The unquoted form appears here and also in the "Importing from OWL > Prerequisites" section added by PR #36 (which *does* use quotes). Matching the quoting style would be consistent.

### Verdict

Looks good to merge.

src/bigquery_ontology/docs/user_manual.md
```diff
@@ -266,6 +267,34 @@ The arity (number of columns) of `from_columns` must match the source entity's p
...
+pip install bigquery-agent-analytics
+```
+
+To verify the installation:
+
+```bash
+gm --help
+```
+
+For local development, install in editable mode from the repo root:
```
caohy1988, line 284: One small follow-up for developer ergonomics: this editable install example gives you `gm`, but not the optional `owl` extra. A reader following the section literally can land in a slightly confusing state where local development works for `validate`/`compile`/`scaffold`, but `gm import-owl` still fails on missing `rdflib`. It may be worth adding the editable-mode equivalent here as well, e.g. `pip install -e '.[owl]'`, or a short note that OWL support still needs the extra in dev installs too.

libei: Good catch — added `pip install -e '.[owl]'` right next to the OWL extra so dev installs don't silently miss rdflib.

PR #42, opened 2026-04-15
fix: extract A2A_INTERACTION content in categorical evaluation transcripts

### LINKED ISSUE #40: Categorical evaluation fails for A2A sessions — transcript builder missing A2A_INTERACTION content extraction
### Description

When using `evaluate_categorical()` on sessions that involve remote A2A agents (via ADK's `TransferToAgentTool` with `RemoteA2aAgent`), all such sessions receive `parse_error=True` with `raw_response=None` because `AI.GENERATE` returns NULL.

### Root cause

The transcript-building SQL in `categorical_evaluator.py` uses this COALESCE to extract text from event content:

```sql
COALESCE(
  JSON_VALUE(content, '$.text_summary'),
  JSON_VALUE(content, '$.response'),
  JSON_VALUE(content, '$.tool'),
  ''
)
```

However, `A2A_INTERACTION` events (introduced in [google/adk-python#5325](https://github.com/google/adk-python/pull/5325)) store the remote agent's response under a different JSON path:

```json
{
  "artifacts": [{"parts": [{"kind": "text", "text": "The actual agent response..."}]}],
  "contextId": "...",
  "history": [...]
}
```

None of the existing JSON paths (`$.text_summary`, `$.response`, `$.tool`) match this structure. The transcript for A2A sessions looks like:

```
USER_MESSAGE_RECEIVED [knowledge_supervisor]: How many PTO days...
LLM_RESPONSE [knowledge_supervisor]: call: transfer_to_agent
TOOL_STARTING [knowledge_supervisor]: transfer_to_agent
A2A_INTERACTION [knowledge_supervisor]:              <-- EMPTY, 3000+ bytes of content but nothing extracted
AGENT_COMPLETED [pto_agent]:
```

The evaluation model sees a question with no answer, cannot classify, and returns NULL.

### Impact

- 100% of A2A sessions get `parse_error=True` and `category=None`
- In our testing: 38-42 out of 100 sessions were A2A, so ~40% of all evaluations failed
- These show up as "UNKNOWN" in category distributions, inflating parse error rate

### Reproduction

```python
from bigquery_agent_analytics import Client, CategoricalEvaluationConfig, TraceFilter

client = Client(project_id=..., dataset_id=..., table_id=...)
# Configure metrics...
report = client.evaluate_categorical(config=cat_config, filters=TraceFilter(limit=100))

# Check parse errors - all A2A sessions will have raw_response=None
for sr in report.session_results:
    for mr in sr.metrics:
        if mr.parse_error and mr.raw_response is None:
            print(f"NULL result: {sr.session_id}")
```

### Proposed fix

Add `JSON_VALUE(content, '$.artifacts[0].parts[0].text')` to the COALESCE chain in all 4 transcript-building locations in `categorical_evaluator.py`:

```sql
COALESCE(
  JSON_VALUE(content, '$.text_summary'),
  JSON_VALUE(content, '$.response'),
  JSON_VALUE(content, '$.artifacts[0].parts[0].text'),   -- NEW: A2A_INTERACTION events
  JSON_VALUE(content, '$.tool'),
  ''
)
```

Affected locations in `categorical_evaluator.py`:
1. `CATEGORICAL_TRANSCRIPT_QUERY` (line ~230)
2. `CATEGORICAL_AI_GENERATE_QUERY` (line ~256)
3. `build_ai_classify_query()` (line ~376)
4. `build_ai_generate_query()` (line ~445)

evekhm:
## Summary

Adds `JSON_VALUE(content, '$.artifacts[0].parts[0].text')` to the COALESCE chain in all 4 transcript-building SQL queries in `categorical_evaluator.py`, fixing categorical evaluation for sessions that use remote A2A agents.

Fixes #40

## Background: How Categorical Evaluation Works

The SDK's `evaluate_categorical()` classifies agent sessions by:

1. **Building a transcript** — SQL query groups events by `session_id`, extracts text from each event's `content` JSON column via COALESCE, and concatenates into one string per session
2. **Sending to AI.GENERATE** — BigQuery's `AI.GENERATE` calls an LLM (e.g. gemini-2.5-flash) inside BigQuery with: `classification_prompt + transcript`
3. **Parsing the result** — the LLM returns a JSON array of `{metric_name, category, justification}` per session

The transcript-building COALESCE currently extracts text from these JSON paths:
```sql
COALESCE(
  JSON_VALUE(content, '$.text_summary'),
  JSON_VALUE(content, '$.response'),
  JSON_VALUE(content, '$.tool'),
  ''
)
```

## Problem

`A2A_INTERACTION` events (logged by ADK when using `RemoteA2aAgent` via `TransferToAgentTool`) store the remote agent's response in a nested structure that none of the existing paths match:

```
JSON_VALUE(content, '$.text_summary')  => null
JSON_VALUE(content, '$.response')      => null
JSON_VALUE(content, '$.tool')          => null
```

The transcript for A2A sessions shows the user's question but not the agent's answer:

```
USER_MESSAGE_RECEIVED [knowledge_supervisor]: How many PTO days do I have?
LLM_RESPONSE [knowledge_supervisor]: call: transfer_to_agent
TOOL_STARTING [knowledge_supervisor]: transfer_to_agent
A2A_INTERACTION [knowledge_supervisor]:              <-- EMPTY, content not extracted
AGENT_COMPLETED [pto_agent]:
```

AI.GENERATE sees a question with no answer, cannot classify, and returns NULL. The session gets `parse_error=True` with `category=None`.

## Real A2A_INTERACTION content payload from BigQuery

The `content` column for an `A2A_INTERACTION` event is the full A2A Task response object. Here is a real row from our dataset:

```json
{
  "artifacts": [
    {
      "artifactId": "<id>",
      "parts": [
        {
          "kind": "text",
          "text": "Alright, let's see what the PTO gods have divined for you!..."
        }
      ]
    }
  ],
  "contextId": "<id>",
  "history": [ ... ],
  "id": "<id>",
  "kind": "task",
  "metadata": {
    "adk_app_name": "pto_agent",
    "adk_author": "pto_agent",
    ...
  },
  "status": { "state": "completed" }
}
```

The response text lives at `artifacts[0].parts[0].text`. This path comes from the [A2A protocol's Task schema](https://google.github.io/A2A/#/documentation?id=task), where `artifacts` contains the agent's output parts — this is the standard A2A response structure, not an observed shape.

**Extraction proof from BigQuery:**
```
JSON_VALUE(content, '$.text_summary')              => null
JSON_VALUE(content, '$.response')                   => null
JSON_VALUE(content, '$.tool')                       => null
JSON_VALUE(content, '$.artifacts[0].parts[0].text') => "Alright, let's see what the PTO gods..."
```

## Impact

From our production dataset:

| Metric | Value |
|---|---|
| Total sessions | 770 |
| Sessions with A2A_INTERACTION | 329 (42.7%) |
| All A2A sessions unclassifiable | Yes — 100% get `parse_error=True` |

## Fix

Add `JSON_VALUE(content, '$.artifacts[0].parts[0].text')` to the COALESCE chain in all 4 transcript-building locations:

```sql
COALESCE(
  JSON_VALUE(content, '$.text_summary'),
  JSON_VALUE(content, '$.response'),
  JSON_VALUE(content, '$.artifacts[0].parts[0].text'),   -- NEW: A2A_INTERACTION
  JSON_VALUE(content, '$.tool'),
  ''
)
```

**Changed locations in `categorical_evaluator.py`:**
1. `CATEGORICAL_TRANSCRIPT_QUERY` (line ~233)
2. `CATEGORICAL_AI_GENERATE_QUERY` (line ~259)
3. `build_ai_classify_query()` (line ~380)
4. `build_ai_generate_query()` (line ~449)

**With the fix**, the transcript becomes complete:
```
USER_MESSAGE_RECEIVED [knowledge_supervisor]: How many PTO days do I have?
LLM_RESPONSE [knowledge_supervisor]: call: transfer_to_agent
TOOL_STARTING [knowledge_supervisor]: transfer_to_agent
A2A_INTERACTION [knowledge_supervisor]: Alright, let's see what the PTO gods...
AGENT_COMPLETED [pto_agent]:
```

AI.GENERATE can now see both the question and the answer, and classifies successfully.

## Scope Note

This PR fixes only the 4 categorical evaluation SQL sites. As noted in the [issue discussion](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/40), the same missing-A2A extraction pattern exists in other modules (`insights.py`, `views.py`, `trace.py`, etc.) and in the shared Python extractor (`udf_kernels.py`). Those are a broader fix and should be addressed in a follow-up.

## Test plan

- [ ] Run `evaluate_categorical()` on a dataset containing A2A sessions (using `RemoteA2aAgent`)
- [ ] Verify A2A session transcripts include the remote agent's response text
- [ ] Confirm parse error rate drops from ~40% to near 0% for A2A sessions
- [ ] Verify non-A2A sessions are unaffected (COALESCE falls through as before)

### DIFF
--- a/src/bigquery_agent_analytics/categorical_evaluator.py
+++ b/src/bigquery_agent_analytics/categorical_evaluator.py
@@ -230,6 +230,7 @@ SELECT
       COALESCE(
         JSON_VALUE(content, '$.text_summary'),
         JSON_VALUE(content, '$.response'),
+        JSON_VALUE(content, '$.artifacts[0].parts[0].text'),
         JSON_VALUE(content, '$.tool'),
         ''
       )
@@ -255,6 +256,7 @@ WITH session_transcripts AS (
         COALESCE(
           JSON_VALUE(content, '$.text_summary'),
           JSON_VALUE(content, '$.response'),
+          JSON_VALUE(content, '$.artifacts[0].parts[0].text'),
           JSON_VALUE(content, '$.tool'),
           ''
         )
@@ -375,6 +377,7 @@ WITH session_transcripts AS (
         COALESCE(
           JSON_VALUE(content, '$.text_summary'),
           JSON_VALUE(content, '$.response'),
+          JSON_VALUE(content, '$.artifacts[0].parts[0].text'),
           JSON_VALUE(content, '$.tool'),
           ''
         )
@@ -443,6 +446,7 @@ WITH session_transcripts AS (
         COALESCE(
           JSON_VALUE(content, '$.text_summary'),
           JSON_VALUE(content, '$.response'),
+          JSON_VALUE(content, '$.artifacts[0].parts[0].text'),
           JSON_VALUE(content, '$.tool'),
           ''
         )


caohy1988 (approved):
## Review: PR #42 — A2A_INTERACTION content extraction in categorical evaluation

### Verdict: **Approve with scope note**

The fix is correct, minimal, and addresses the exact root cause described in issue #40.

### What's correct

1. **JSON path is right.** `$.artifacts[0].parts[0].text` matches the A2A Task response schema ([A2A spec](https://google.github.io/A2A/#/documentation?id=task)). Verified against the PR's included real payload example.

2. **COALESCE position is right.** Placing the A2A path before `$.tool` ensures A2A responses are extracted before falling through to tool content — correct precedence since A2A_INTERACTION events carry the remote agent's full response, not a tool invocation name.

3. **All 4 sites patched.** The categorical evaluator has 4 transcript-building SQL sites. All 4 are updated consistently:
   - `CATEGORICAL_TRANSCRIPT_QUERY` (line ~233)
   - `CATEGORICAL_AI_GENERATE_QUERY` (line ~259)
   - `build_ai_classify_query()` (line ~380)
   - `build_ai_generate_query()` (line ~449)

4. **No behavioral change for non-A2A sessions.** The COALESCE chain falls through — if `$.artifacts[0].parts[0].text` is NULL (which it will be for all non-A2A events), the existing `$.tool` path still works.

### Scope gap (expected, documented in PR)

The same missing-A2A extraction exists in **6+ other locations** across the SDK that the PR explicitly does not fix:

| File | Sites | Event types affected |
|------|-------|---------------------|
| `insights.py` | 3 COALESCE sites (lines 429, 459, 516) | Session insights, drill-down |
| `evaluators.py` | 2 COALESCE sites (lines 735, 779) | Streaming evaluation |
| `views.py` | 1 site (line 131) | Materialized views |
| `ontology_graph.py` | 2 sites (lines 83-85, 130-132) | Ontology graph extraction (uses `JSON_EXTRACT_SCALAR` not `JSON_VALUE`, same gap) |

The PR correctly scopes to categorical evaluation only and notes the broader fix as a follow-up. That's the right call — fixing all 10+ sites in one PR would make it harder to review and test.

### Missing: no test coverage for A2A extraction

There are no tests asserting that `$.artifacts[0].parts[0].text` appears in the generated SQL. The existing `test_categorical_evaluator.py` has no A2A references. A unit test like this would prevent regression:

```python
def test_transcript_query_extracts_a2a_content():
    """A2A artifact path must be in the COALESCE chain."""
    from bigquery_agent_analytics.categorical_evaluator import (
        CATEGORICAL_TRANSCRIPT_QUERY,
    )
    assert "artifacts[0].parts[0].text" in CATEGORICAL_TRANSCRIPT_QUERY
```

Not blocking for this fix, but should be added in the follow-up.

### One observation on the ADK plugin side

The BQ Agent Analytics plugin in ADK (`bigquery_agent_analytics_plugin.py`) serializes A2A content as the raw Task response object. The content JSON structure (`artifacts[0].parts[0].text`) is the **A2A protocol's standard Task schema**, not something the plugin invented. This means the fix is protocol-aligned, not a workaround for a serialization choice.

However, the plugin's `_format_content()` function (line 139) could optionally extract and flatten A2A text at write time, which would make downstream SQL simpler. That's a separate plugin-side improvement, not a blocker for this PR.

### Summary

| Check | Status |
|-------|--------|
| Fix addresses root cause | Yes |
| All categorical SQL sites patched | Yes (4/4) |
| Non-A2A sessions unaffected | Yes |
| JSON path matches A2A spec | Yes |
| Test coverage | Missing (not blocking) |
| Other modules need same fix | Yes (follow-up) |

caohy1988:
Follow-up note (non-blocking):

The A2A fix is applied consistently in all four categorical transcript SQL sites, but this PR also highlights a maintainability gap: the same `COALESCE(JSON_VALUE(...))` transcript-extraction chain is still duplicated four times inside `categorical_evaluator.py` (`CATEGORICAL_TRANSCRIPT_QUERY`, `CATEGORICAL_AI_GENERATE_QUERY`, `build_ai_classify_query()`, and `build_ai_generate_query()`). That makes future format additions easy to miss again — exactly what happened here.

Suggested follow-up: factor the shared transcript payload expression into one constant/helper (for example a `TRANSCRIPT_CONTENT_SQL` fragment) and interpolate that into all four query templates. That would make future schema-path fixes single-site changes instead of a 4-site grep/update.

caohy1988:
Follow-up scope note (non-blocking, based on current `upstream/main`):

The earlier “same gap exists elsewhere” summary is directionally right, but on latest `main` the remaining cases are not all the same shape. `context_graph.py`, `insights.py`, and `ontology_graph.py` still use the same transcript payload fallback pattern and can likely take the same A2A addition. `evaluators.py` is a little different: its LLM-judge trace text currently uses only `$.text_summary`, so that follow-up is broader than just appending `$.artifacts[0].parts[0].text` to an existing chain. And `views.py` mostly exposes structured per-event fields like `$.tool` / `$.result`, not session transcript aggregation, so it probably should not be grouped into the same transcript-extraction follow-up bucket.

Suggested follow-up framing: split the remaining work into (1) transcript aggregation sites that can take the same A2A fallback update, and (2) evaluator/query paths that need a broader transcript normalization pass.

haiyuan-eng-google (approved):

PR #46, opened 2026-04-16
fix: strip markdown code fences from LLM classification responses

### LINKED ISSUE #43: parse_classifications fails when Gemini wraps JSON response in markdown code blocks
### Description

`parse_classifications()` in `categorical_evaluator.py` fails to parse valid classification responses from Gemini because the model wraps its JSON output in markdown code blocks.

### Root cause

When `classify_sessions_via_api()` calls Gemini to classify a session, the model returns:

~~~
```json
[
  {"metric_name": "response_usefulness", "category": "meaningful", "justification": "..."},
  {"metric_name": "task_grounding", "category": "grounded", "justification": "..."}
]
```
~~~

The `parse_classifications()` function passes this directly to `json.loads()`, which fails because of the surrounding `` ```json `` and `` ``` `` markers. Every metric for that session gets `parse_error=True` and `category=None`.

### Impact

- Affects the Gemini API fallback path (`classify_sessions_via_api`) and any retry logic that uses it
- In testing with `gemini-2.5-flash`: 100% of API-classified sessions had parse errors before fixing this
- Combined with the A2A transcript issue (#40), this meant the A2A retry path also failed, leaving all A2A sessions permanently unclassified

### Reproduction

```python
from bigquery_agent_analytics.categorical_evaluator import parse_classifications

# Simulate what Gemini actually returns
raw = '```json\n[{"metric_name": "tone", "category": "positive", "justification": "test"}]\n```'

results = parse_classifications(raw, config)
for r in results:
    print(f"{r.metric_name}: parse_error={r.parse_error}")
    # Output: tone: parse_error=True
```

### Proposed fix

Add markdown code block stripping at the top of `parse_classifications()`, before `json.loads()`:

```python
# In parse_classifications(), after the empty-check and before json.loads():

# Strip markdown code blocks (```json ... ```) that models often add
text = raw_json.strip()
if text.startswith("```"):
    # Remove opening ```json or ``` line
    text = text.split("\n", 1)[-1] if "\n" in text else text[3:]
    if text.endswith("```"):
        text = text[:-3]
    text = text.strip()

try:
    parsed = json.loads(text)  # was: json.loads(raw_json)
```

**Affected location:** `parse_classifications()` in `categorical_evaluator.py` (line ~658)

evekhm:
## Summary

- Extract shared `_strip_markdown_fences()` helper in `evaluators.py` using regex (matches the pattern already used in SQL `REGEXP_REPLACE` in `ontology_graph.py` and `context_graph.py`)
- Apply it in `parse_classifications()` before `json.loads()` — fixes 100% parse errors when Gemini wraps JSON in `` ```json `` fences
- Refactor `_parse_json_from_text()` and `trace_evaluator.py` to use the shared helper instead of inline `split()` logic

## Test plan

- [x] 3 new tests for `parse_classifications()` with markdown-fenced input (` ```json `, plain ` ``` `, no-newline edge case)
- [x] 8 new tests for `_strip_markdown_fences()` directly (JSON fence, plain fence, SQL fence, no fence, empty, None, no-newline, whitespace)
- [x] All 90 categorical evaluator tests pass
- [x] All 52 evaluator tests pass
- [x] Verified against real Gemini API responses that previously caused parse errors

Fixes #43


caohy1988 (commented):
## Code review

Nice catch on the 100% parse-error bug. The refactor to a shared helper is the right consolidation. A few polish items below, none are blockers.

### Findings

**1. Type annotation lies about `None` handling** — `src/bigquery_agent_analytics/evaluators.py:821`

Signature says `def _strip_markdown_fences(text: str) -> str:` but the function returns `None` when `text` is `None` (and `tests/test_categorical_evaluator.py` explicitly asserts this). Fix:

```python
def _strip_markdown_fences(text: Optional[str]) -> Optional[str]:
```

**2. Private name imported across modules** — `categorical_evaluator.py:69`, `trace_evaluator.py:56`

`_strip_markdown_fences` has a leading underscore (Python's "module-private" convention) but is imported into two other modules. Trips `pylint W0212`. Either rename to `strip_markdown_fences` (it's a legit public helper now, three callsites across three modules) or relocate to a `_helpers.py`.

**3. `import re` inside function body** — `evaluators.py:836`

`re` is stdlib, called on every invocation. Move to module-top imports alongside `json`.

**4. Regex tag list is narrow** — `evaluators.py:838`

Current: `^```(?:json|sql)?\s*\n?` — matches only `json`, `sql`, or untagged. Models produce `` ```JSON `` (uppercase), `` ```javascript ``, `` ```python ``, `` ```text `` in the wild. These bypass the strip and rely on brace-matching fallback — which works in `_parse_json_from_text` but fails cleanly in `parse_classifications` (no fallback — goes straight to `parse_error` at `categorical_evaluator.py:667`).

Suggestion:

```python
text = re.sub(r"^```[a-zA-Z]*\s*\n?", "", text)
```

Language-agnostic, same shape. Add `re.IGNORECASE` if you want to keep the narrow allowlist but handle case.

**5. Fragile "no fences found" heuristic** — `trace_evaluator.py:822`

```python
json_str = _strip_markdown_fences(response_text)
if not json_str or json_str == response_text:
    # fall through to brace matching
```

This works today only because `response_text = response.text.strip()` was called upstream AND the helper also strips internally, so the equality holds. If either strip disappears in a future refactor, this silently stops triggering the brace-matching fallback. Safer:

```python
if not response_text.startswith("```"):
    # brace matching path
```

Checks the actual condition directly.

**6. Test coverage gaps**

- No test for uppercase/unknown language tags (`` ```JSON ``, `` ```python ``).
- No test for truncated response with opening fence but no closing fence.

Cheap to add alongside the existing fence tests.

### Verdict

**Approve with minor nits.** Findings #1 (type annotation) and #4 (regex breadth) are the two that materially change behavior under real-world LLM output. The rest can ship as-is.

caohy1988 (commented):
## Re-review (head `ea3495a`)

Thanks for the fast turnaround. All six findings from the prior pass are addressed cleanly:

| # | Prior finding | Status |
|---|---|---|
| 1 | Type annotation should be `Optional[str]` | ✅ `evaluators.py:822` |
| 2 | Private name imported across modules | ✅ Renamed to `strip_markdown_fences` + backwards-compat alias |
| 3 | `import re` inside function body | ✅ Moved to module top |
| 4 | Regex tag list narrow | ✅ Now `[a-zA-Z]*` |
| 5 | Fragile `== response_text` heuristic | ✅ Uses `startswith("```")` directly |
| 6 | Missing test coverage | ✅ 3 new tests (uppercase, unknown lang, truncated) |

### New findings

**1. Closing-fence regex can't handle trailing content after fence** — `evaluators.py:839`

Pattern `\n?\s*```\s*$` requires `` ``` `` at end of string. For Gemini judge output like:

````
```json
{"score": 1}
```
Here's my analysis of the trajectory...
````

…the closing fence isn't stripped, and `json.loads` fails on the whole blob.

Impact by callsite:
- `_parse_json_from_text` — recovers via brace-matching fallback. Fine.
- `parse_classifications` — no fallback, goes straight to `parse_error`.
- `trace_evaluator._llm_judge_evaluate` — enters the `startswith("```")` branch, skips brace matching, falls through to the regex score-scraper. **Mild regression** vs. the old `split("```")[0]`, which handled trailing commentary.

This is a real pattern in judge-style prompts where models sometimes append commentary. Suggested fix:

```python
text = re.sub(r"\n?\s*```[\s\S]*$", "", text)
```

Plus a test for `` "```json\n{...}\n```\ntrailing commentary" ``.

**2. Language tag regex drops tags with digits** — `evaluators.py:837`

`[a-zA-Z]*` partial-matches `json5` or `es6`, leaving the digit(s) as a prefix. Low probability, trivial fix: `[a-zA-Z0-9]*` or `[\w-]*`.

**3. Backwards-compat alias is probably dead weight** — `evaluators.py:845`

The helper is new in this PR. Nothing in the repo imports `_strip_markdown_fences` except the test that exists to verify the alias. Consider deleting both the alias and `test_backwards_compat_alias` — no real callers to compat against. Taste call; keep if you prefer belt-and-suspenders.

**4. `trace_evaluator.py:822` condition is readable but awkward**

`if not json_str or not response_text.startswith("```"):` mixes two signals. Behaviorally equivalent but easier to audit as two branches:

```python
if response_text.startswith("```"):
    pass  # json_str already set by strip_markdown_fences
elif "{" in response_text:
    # brace-matching fallback
    ...
```

Not a bug. Optional.

### Verdict

**Approve.** Finding #1 is the only one I'd want addressed before merge (real regression for `trace_evaluator` under trailing-commentary output, three-char regex fix). The rest are polish and can ship as-is.

caohy1988 (commented):
## Re-review v3 (head `07a6f45b`)

Thanks for the quick turn on the v2 feedback. All four findings addressed cleanly in this commit.

### Findings from v2: status

|---|---|---|---|
| 1 | Closing-fence regex failed on trailing content | ✅ Fixed | `evaluators.py:838` — `\n?\s*```[\s\S]*$` |
| 2 | `[a-zA-Z]*` missed numeric tags (`json5`, `es6`) | ✅ Fixed | `evaluators.py:837` — `[a-zA-Z0-9]*` |
| 3 | Backwards-compat alias was dead weight | ✅ Fixed | `_strip_markdown_fences` alias removed, along with its test |
| 4 | `trace_evaluator` condition was awkwardly phrased | ✅ Fixed | `trace_evaluator.py:822-828` — clean `if startswith("```"): ... else: ...` split |

Plus two new tests:

- `test_trailing_content_after_fence` pins the regression case exactly (`"```json\n{...}\n```\nHere's my analysis..."` → `{...}`)
- `test_language_tag_with_digits` pins `json5`

### Fresh pass — no new issues

Ran through the adversarial cases for the new regex `\n?\s*```[\s\S]*$`:

- Multiple fences in input — `re.sub` takes the leftmost match, trailing content after the first close fence is correctly stripped ✅
- Empty fence `"```\n```"` → `""`, falls through to `return {}, response_text` ✅
- Fence with non-JSON content → strips fences, `json.loads` fails, fallback regex scorer still works ✅
- Raw brace-only input (no fence) → bypasses strip, hits brace-matching fallback ✅

Previously-covered critical categories (SQL safety, race conditions, LLM trust boundary, shell injection) unchanged. No new surface area introduced.

### Verdict

**Approve.** Ship it. Clean iteration, complete test coverage, no loose ends.

haiyuan-eng-google (approved):

PR #48, opened 2026-04-16
fix: increase max_output_tokens and retry NULL AI.GENERATE rows

### LINKED ISSUE #44: AI.GENERATE returns NULL for individual rows; Gemini API fallback truncated by max_output_tokens=1024
### Description

When `_categorical_ai_generate()` runs the batch `AI.GENERATE` query in BigQuery, some rows (~15-20%) return NULL in the `classifications` column. These are silently marked as `parse_error=True` with no retry or fallback.

When retrying these NULL sessions via the Gemini API fallback (`classify_sessions_via_api`), the retries also fail — but **not** due to transient errors. The root cause is that `max_output_tokens` is hardcoded to **1024** in three places, which is too small for classification responses that include justifications for multiple metrics.

### Root cause

There are **two issues** compounding:

**Issue A: AI.GENERATE returns NULL for individual rows (no per-row fallback)**

BigQuery's `AI.GENERATE` processes all sessions in a single SQL query. For some rows, the underlying model call fails (likely per-row rate limiting / 429 or transient timeouts). BigQuery does not surface per-row errors — it just returns NULL for the `classifications` column.

The existing fallback cascade only triggers when the **entire query** fails:
```
AI.CLASSIFY (whole query) → AI.GENERATE (whole query) → Gemini API (whole query)
```
There is no per-row retry.

**Issue B: `max_output_tokens=1024` truncates Gemini API responses (the real blocker)**

Even after adding per-row retry via the Gemini API, most sessions **still fail** because the API response is truncated. The model needs to output a JSON array with `metric_name`, `category`, and `justification` for each metric. With 2+ metrics and detailed justifications, this regularly exceeds 1024 tokens.

Debug logging reveals every failing retry has `finish_reason=FinishReason.MAX_TOKENS`:

```
API returned response but parse failed for session 63ffb968:
  finish_reason=FinishReason.MAX_TOKENS
  raw_text_len=122
  raw_text='```json\n[\n  {\n    "metric_name": "response_usefulness",
    "category": "false_positive", "justification": "The agent'
    <-- TRUNCATED HERE, invalid JSON
```

The truncated JSON can never be parsed, so retrying produces the exact same result every time. This is **not transient** — 50 retries would all fail identically.

`max_output_tokens=1024` is hardcoded in three places:

1. `categorical_evaluator.py` line ~281 — `AI.GENERATE` SQL: `"maxOutputTokens": 1024`
2. `categorical_evaluator.py` line ~471 — `build_ai_generate_query()` SQL: `"maxOutputTokens": 1024`
3. `categorical_evaluator.py` line ~866 — `classify_sessions_via_api()`: `max_output_tokens=1024`

### Evidence

We debugged this by adding `finish_reason` and `raw_response` logging to the retry loop:

**With `max_output_tokens=1024` (current):**
```
AI.GENERATE returned NULL for 8 sessions, retrying via Gemini API
Retry attempt 1: finish_reason=MAX_TOKENS, raw_text_len=122  (session 63ffb968)
Retry attempt 1: finish_reason=MAX_TOKENS, raw_text_len=134  (session a950b6a8)
Retry attempt 1: finish_reason=MAX_TOKENS, raw_text_len=115  (session f97c7376)
Retry attempt 1: finish_reason=MAX_TOKENS, raw_text_len=944  (session 051acc13)
... identical output on every retry — not transient
6 sessions still unresolved after 10 retries
```

**With `max_output_tokens=8192`:**
```
AI.GENERATE returned NULL for 8 sessions, retrying via Gemini API
Gemini API retry resolved 8/8 NULL sessions
parse_errors: 0
parse_error_rate: 0.0
```

All 8 sessions resolved on the **first** retry. Zero parse errors.

### Impact

- ~15-20% of sessions get NULL from AI.GENERATE in a typical run
- With `max_output_tokens=1024`, the Gemini API retry path also fails for most sessions
- Net effect: 10-12% permanent parse error rate even with retry logic
- Increasing to 8192 eliminates parse errors entirely (0%)

### Proposed fix

1. **Add per-row NULL retry via Gemini API** — after `AI.GENERATE`, detect rows where `classifications` is NULL but `transcript` exists, and retry via `classify_sessions_via_api()` up to 3 times.

2. **Increase `max_output_tokens` from 1024 to 8192** in all three locations:

```python
# categorical_evaluator.py, classify_sessions_via_api():
config=types.GenerateContentConfig(
    temperature=config.temperature,
    max_output_tokens=8192,  # was: 1024
)

# categorical_evaluator.py, AI.GENERATE SQL (2 locations):
model_params => JSON '{"generationConfig": {"temperature": ..., "maxOutputTokens": 8192}}'
```

3. Consider making `max_output_tokens` configurable via `CategoricalEvaluationConfig` rather than hardcoded.

### Affected locations

- `_categorical_ai_generate()` in `client.py` (line ~1350) — needs per-row retry
- `classify_sessions_via_api()` in `categorical_evaluator.py` (line ~866) — `max_output_tokens`
- `CATEGORICAL_AI_GENERATE_QUERY` in `categorical_evaluator.py` (line ~281) — `maxOutputTokens`
- `build_ai_generate_query()` in `categorical_evaluator.py` (line ~471) — `maxOutputTokens`

evekhm:
## Summary
- **max_output_tokens**: Replaced hardcoded `1024` with configurable `max_output_tokens` field on `CategoricalEvaluationConfig` (default `8192`). The old value truncated multi-metric classification responses, causing parse errors.
- **NULL retry**: When `AI.GENERATE` returns NULL classifications (due to rate limiting or transient errors), those sessions are now retried via the Gemini API up to 3 times.
- **Observability**: Added `finish_reason` logging on parse errors and safe handling of `response.text = None`.

## Changes
- `categorical_evaluator.py`: Added `max_output_tokens` config field, parameter to `build_ai_generate_query()`, updated SQL templates and `classify_sessions_via_api()` 
- `client.py`: Added `_retry_null_sessions()` method, NULL detection in `_categorical_ai_generate()`, threaded `max_output_tokens` through call sites
- `tests/test_categorical_evaluator.py`: 15 new tests (max_output_tokens propagation, finish_reason logging, NULL retry logic, boundary conditions)

## Test plan
- [x] All 1633 tests pass locally
- [x] `bash autoformat.sh` clean (pyink/isort)
- [x] 13 of 15 new tests fail if the fix is reverted (2 are regression guards for boundary conditions)

Fixes #44


caohy1988 (commented):
## Code review

Good fix. Both problems (token truncation + NULL AI.GENERATE) are real, both are addressed, and the test depth (15 new tests covering propagation, logging, retry success/exhaust/partial/exception) is genuinely solid. A few things worth a second look before merge.

### Findings

**1. No bounds validation on `max_output_tokens`** — `src/bigquery_agent_analytics/categorical_evaluator.py:130` (confidence 7/10)

```python
max_output_tokens: int = Field(
    default=8192,
    description="Max output tokens for classification response.",
)
```

Pydantic validates the type but not the range. A caller passing `max_output_tokens=0` or `max_output_tokens=-1` produces invalid SQL like `"maxOutputTokens": -1` and invalid Gemini API params. Fix:

```python
max_output_tokens: int = Field(
    default=8192,
    ge=1,
    le=32768,  # or whatever the hard model ceiling is
    description="Max output tokens for classification response.",
)
```

Catches invalid values at config time with a clear error instead of an opaque BQ/Gemini failure downstream.

**2. No backoff between retry attempts** — `src/bigquery_agent_analytics/client.py:1468` (confidence 8/10)

```python
for attempt in range(1, max_retries + 1):
    if not remaining:
      break
    try:
      results = _run_sync(
          classify_sessions_via_api(remaining, config, endpoint)
      )
```

The PR description explicitly names **rate limiting** as the root cause of NULL `AI.GENERATE` results. Retrying immediately after a rate-limit hit is the textbook case for re-hitting the same limit. Recommend exponential backoff:

```python
import time
...
for attempt in range(1, max_retries + 1):
    if not remaining:
      break
    if attempt > 1:
      time.sleep(2 ** (attempt - 2))  # 1s, 2s, 4s
    try:
      ...
```

Same tree-of-life, an order of magnitude better behavior under sustained 429s.

**3. Bare `except Exception` swallows programming errors** — `src/bigquery_agent_analytics/client.py:1487` (confidence 7/10)

```python
except Exception as e:
    logger.warning(
        "Gemini API retry attempt %d failed: %s",
        attempt,
        e,
    )
```

A `NameError` or `TypeError` in `classify_sessions_via_api` (or its call graph) is logged as a warning and silently retried 3 times. Consider catching `google.api_core.exceptions.GoogleAPIError` (or whichever concrete exception types you expect from the genai client) so that programming bugs surface instead of hiding.

**4. Tight coupling to sync context via `_run_sync`** — `src/bigquery_agent_analytics/client.py:1470` (confidence 6/10)

`_retry_null_sessions` is sync and calls `_run_sync(classify_sessions_via_api(...))`. This is fine because `_categorical_ai_generate` is also sync — but if anyone ever calls the client from an async context, `_run_sync` will blow up with "This event loop is already running." Not a bug today; worth a one-line comment noting the sync-only assumption, or wrap `_run_sync` to detect an existing loop and fall back to `asyncio.run_coroutine_threadsafe`.

**5. Default bump 1024 → 8192 is an ~8x per-call cost increase** — `src/bigquery_agent_analytics/categorical_evaluator.py:130` (confidence 8/10)

Not a code issue, but a real operator surprise for anyone on a fixed Gemini budget who hasn't set `max_output_tokens` explicitly. This deserves a callout in the CHANGELOG / release notes: *"Default max_output_tokens increased 1024 → 8192 to prevent truncation of multi-metric classifications. Users with cost-sensitive workloads can override via `CategoricalEvaluationConfig(max_output_tokens=...)`."*

**6. Adjacent-string concatenation artifact in log message** — `src/bigquery_agent_analytics/categorical_evaluator.py:886` (confidence 9/10)

```python
"Categorical API classification EXCEPTION for %s: %s " "(type=%s)",
```

Two adjacent string literals. Functionally fine (Python concatenates at compile time), but looks like a formatter hiccup. Cleaner:

```python
"Categorical API classification EXCEPTION for %s: %s (type=%s)",
```

### What's good

- **`finish_reason` logging on parse error** (`categorical_evaluator.py:865`) — this is the single most useful addition. Distinguishes MAX_TOKENS (→ bump the limit) from SAFETY/STOP/OTHER (→ prompt problem) without having to dig through API responses by hand.
- **`response.text if response.text else ""`** (`categorical_evaluator.py:862`) — defensive null guard. Gemini does occasionally return responses with no text when blocked or finish_reason == SAFETY.
- **NULL-without-transcript is correctly gated from retry** (`client.py:1397`, test `test_ai_generate_null_without_transcript_skips_retry`). No point retrying sessions whose source transcript is missing.
- **Retry loop terminates cleanly** — `remaining` narrows each attempt, succeeded sessions move to `all_results` and are never retried again. No infinite loops, no duplicate work.
- **`max_retries` default = 3** matches the PR description. Consistent.

### Scope check: CLEAN

Two related fixes shipped together (token bump + NULL retry), both justified as different manifestations of the same root cause (insufficient output budget + transient BQ AI failures). 15 tests, 13 of which fail if the fix is reverted per your PR description. 

### Verdict

**Approve with recommendations.** Findings #1 (bounds) and #2 (backoff) are the two that materially change behavior under adversarial conditions. The rest are polish or release-note items. None are merge blockers.

caohy1988:
Review findings:

1. Medium: the new per-row retry only covers `classifications IS NULL`, but the parser still marks several non-NULL payloads as `parse_error=True` (`json.loads` failure, malformed/truncated JSON, invalid category payloads, missing required metrics). In `client.py`, `_categorical_ai_generate()` only adds a session to `null_sessions` when `r.get("classifications") is None`, so a non-NULL but broken `AI.GENERATE` payload still never reaches the retry path.

Recommended fix: build the retry candidate set from retryable parse-error rows with a transcript present, not just literal NULL classifications. At minimum, consider retrying rows where `classifications` is NULL or the parsed result is an all-metrics parse error due to malformed JSON.

2. Medium: the new null-retry observability only exists in logs, not in the returned report. `_categorical_ai_generate()` now detects NULL rows and logs retry success/failure counts, but `evaluate_categorical()` still sets only `report.details["execution_mode"] = "ai_generate"` for this path. That means callers cannot tell from the returned report how many rows were NULL, how many were retried, or how many were recovered.

Recommended fix: have `_categorical_ai_generate()` return both `session_results` and a small metadata struct/counters (`null_count`, `retry_attempted`, `retry_resolved`, `retry_unresolved`) and attach those to `report.details`, similar to `classify_null_count` on the AI.CLASSIFY path.

caohy1988 (commented):
## Re-review (head `5998e37c`)

Five commits since my prior review — genuinely good iteration on bounds, backoff, retry metadata, and observability. But the narrowing of the exception catch went in the wrong direction for this use case, and the scope expansion to retry all parse errors introduces a new class of cost/loop risk. Details below.

### Prior findings: status

| # | Prior finding | Status |
|---|---|---|
| 1 | No bounds validation on `max_output_tokens` | ✅ `ge=1, le=65536` |
| 2 | No backoff between retry attempts | ✅ Exponential 1s/2s/4s |
| 3 | Bare `except Exception` too wide | ⚠️ Narrowed too far — see finding #1 below |
| 4 | Sync/async coupling call-out | ✅ Docstring note |
| 5 | Default bump cost surprise | ✅ Called out in PR description |
| 6 | Adjacent-string concatenation artifact | ✅ Combined |

---

### New findings

**1. [P1] Exception catch was narrowed past the actual retryable errors — `client.py:1497` (confidence 9/10)**

```python
except (RuntimeError, ValueError, OSError) as e:
    logger.warning("Gemini API retry attempt %d failed: %s (type=%s)", ...)
```

The `google-genai` SDK raises `google.api_core.exceptions.ResourceExhausted` for 429 rate limits, `ServiceUnavailable` for 503s, `DeadlineExceeded` for timeouts. **None of these inherit from `RuntimeError`, `ValueError`, or `OSError`.** They inherit from `GoogleAPIError → Exception`.

So under sustained rate limiting — the exact scenario named in the PR description as the root cause of NULL rows — the retry loop now propagates the first 429 as an unhandled exception straight through `_retry_null_sessions` → `_categorical_ai_generate` → caller. The exponential-backoff code becomes dead code for the one thing it was added to handle.

Fix:

```python
from google.api_core.exceptions import GoogleAPIError
...
except (GoogleAPIError, RuntimeError, ValueError, OSError) as e:
```

Or, given this is a retry-and-continue loop, a broader catch is defensible here:

```python
except Exception as e:  # noqa: BLE001 — retry loop, log + continue
```

At minimum, `GoogleAPIError` must be in the tuple. Consider adding a test that injects a `ResourceExhausted` and asserts the retry loop logs + continues instead of crashing.

**2. [P2] Retry scope now covers all parse errors, including deterministic ones — `client.py:1410` (confidence 8/10)**

Commit `6d4addb4` broadened the retry criterion from "row has NULL classifications" to "row has any `parse_error=True`". That's an improvement for truly transient failures, but it also means that prompts producing persistently malformed output — model returns prose, finish_reason=SAFETY, finish_reason=MAX_TOKENS at the current config — now get retried 3× at 8192 tokens each, with 1s+2s+4s backoff, for every affected session, every evaluation run.

At the default 8K tokens × 3 retries × N sessions, the cost for a batch where the prompt is just wrong can be ~12× what it was pre-PR. The whole batch is retried per run.

Suggestion: gate retry on signals that suggest transience, not just any parse error. Roughly:

```python
RETRYABLE_FINISH_REASONS = {None, "FINISH_REASON_UNSPECIFIED", "OTHER"}
# Retry NULL (no finish_reason) and unknown/other terminations.
# Don't retry STOP-with-malformed-JSON (prompt issue),
# SAFETY (not a transient condition),
# MAX_TOKENS (same config will produce same result — or bump tokens and retry).
```

Even a simple "don't retry `finish_reason == STOP`" guard would prevent the worst case.

**3. [P2] Return type annotation is now a lie — `client.py:1348` (confidence 9/10)**

```python
def _categorical_ai_generate(
    self,
    ...
) -> list:
    ...
    return session_results, retry_meta
```

Annotation still says `-> list` but the function now returns a tuple. Callers at `client.py:1253` unpack it correctly, but mypy/pyright will flag this. Fix:

```python
) -> tuple[list, dict]:
```

**4. [P3] `null_sessions` / `null_count` naming is stale — `client.py:1394`, `:1434` (confidence 7/10)**

After commit `6d4addb4`, these variables contain sessions with any `parse_error=True`, not just NULL rows. The `retry_meta["null_count"]` key exposed in `report.details["retry"]` is misleading for operators reading reports. Rename to `failed_sessions` / `failed_count` (or `retried_count`). Public API consideration if anyone's already consuming the report shape — pick a name and pin it.

**5. [P3] Subtle logic gap in retry dedup — `client.py:1478-1485` (confidence 6/10)**

```python
for r in results:
    has_error = any(m.parse_error for m in r.metrics)
    if has_error and r.session_id in remaining:
        still_failed[r.session_id] = remaining[r.session_id]
        for m in r.metrics:
            if m.parse_error:
                logger.warning(...)
                break
    else:
        all_results[r.session_id] = r
```

If `has_error` is True but `r.session_id` is not in `remaining` (unlikely in practice but possible under shared state or API quirk), the result falls into `else` and gets recorded as success despite the parse error. Cleaner as two nested conditions:

```python
for r in results:
    has_error = any(m.parse_error for m in r.metrics)
    if has_error:
        if r.session_id in remaining:
            still_failed[r.session_id] = remaining[r.session_id]
            # log ...
    else:
        all_results[r.session_id] = r
```

**6. [P3] `import time` still inline — `client.py:1467` (confidence 8/10)**

```python
for attempt in range(1, max_retries + 1):
    ...
    if attempt > 1:
        import time   # ← module-level import inside a loop
        ...
        time.sleep(backoff)
```

Same kind of style issue as #3 from the PR #46 review. Move to module-top imports alongside the other stdlib imports.

**7. [P3] Missing test for non-NULL parse-error retry — `tests/test_categorical_evaluator.py` (confidence 7/10)**

Commit `6d4addb4` broadens retry scope, but the existing tests only cover NULL classifications (`test_ai_generate_detects_null_classifications`) and NULL-without-transcript skip. There's no test that asserts a non-NULL row with invalid JSON triggers retry. Easy add, would prevent a silent regression.

### What got better since v1

- `retry_meta` piped into `report.details["retry"]` — great observability. `null_count` / `retry_attempted` / `retry_resolved` / `retry_unresolved` is exactly what operators want to see without re-running.
- `finish_reason` warnings on parse error (`categorical_evaluator.py:870`) — the single most useful addition in this PR.
- `response.text if response.text else ""` — defensive null guard, well-tested.
- Retry logging now captures the metric name and truncated raw response per attempt. Good forensics.

### Verdict

**Approve conditional on finding #1 — this one I'd want fixed before merge.** Narrowing the exception set past `GoogleAPIError` makes the retry loop no-op for the 429 case that motivated it.

Finding #2 (retry-scope cost risk) is the next-most-important one — not a blocker, but worth a `finish_reason` gate before the next release if multi-metric prompts are in use at scale.

The rest are polish.

caohy1988:
Follow-up re-review on head `5998e37c`:

The two substantive findings from the latest review are still present:

1. **P1**: `_retry_null_sessions()` still catches only `(RuntimeError, ValueError, OSError)` in `client.py`, so it still will not catch the `google.api_core` exceptions that matter for 429/503/timeout behavior. That means the new backoff path is still bypassed for the exact retryable failures this PR is meant to handle.

2. **P2**: `_categorical_ai_generate()` still retries any parsed session with `parse_error=True`, not just literal NULL rows, and there is still no `finish_reason` / failure-type gate. Deterministic malformed-output cases will therefore still be retried 3 times at the higher token budget.

One additional low-severity follow-up from the scope expansion:

3. **P3**: the docs/naming are now stale in a few places. `_categorical_ai_generate()` and `_retry_null_sessions()` still describe “NULL sessions”, and `retry_meta["null_count"]` / the “resolved %d/%d NULL sessions” log message no longer match actual behavior now that the retry scope includes general parse-error rows too. Not a correctness bug, but it will confuse operators reading logs and report metadata.

I reran `pytest tests/test_categorical_evaluator.py tests/test_sdk_client.py -q` on this head; result was `137 passed, 1 warning`.

caohy1988 (commented):
## Re-review v3 (head `fd2dc5ea`)

Commit `fd2dc5ea` cleanly addresses six of the seven v2 findings. Finding #2 (retry-scope cost risk) is unaddressed but was called out as a recommendation rather than a blocker in the v2 review — noting it below so it's tracked, not blocking approval.

### v2 findings: status

|---|---|---|---|
| 1 | Exception catch narrowed past `GoogleAPIError` | ✅ Fixed | `client.py:1496` — broad `except Exception as e:  # Broad catch: retry loop logs + continues`, with comment justifying the choice |
| 2 | Retry scope covers deterministic parse errors | ⚠️ Deferred | Still retries all parse errors — no `finish_reason` gate. See below. |
| 3 | Return type annotation lied (`-> list` vs tuple) | ✅ Fixed | `client.py:1361` — `-> tuple[list, dict]:` |
| 4 | Stale `null_sessions` / `null_count` naming | ✅ Fixed | Renamed throughout: variable `failed_sessions`, meta key `failed_count`, method `_retry_failed_sessions`, test class `TestRetryFailedSessions` |
| 5 | Subtle logic gap in retry dedup | ✅ Fixed | `client.py:1473-1486` — nested `if has_error: ...` / `else: all_results[r.session_id] = r`, parse-error results no longer leak into `all_results` |
| 6 | `import time` still inline | ✅ Fixed | `client.py:52` — `import time` at module top |
| 7 | Missing test for non-NULL parse-error retry | ✅ Fixed | `test_ai_generate_detects_parse_error_classifications` — asserts a non-NULL row with `"not valid json"` triggers retry |

### Finding #2 re-note

Retry still fires on any `parse_error=True`, including deterministic failures (model returns prose, `finish_reason=SAFETY`, same-config `MAX_TOKENS`). This is a cost-explosion risk for batches with persistent prompt-shape issues — 3× retries at 8K tokens with 1s+2s+4s backoff, every run, per affected session.

Not a blocker for this PR, but worth a follow-up. Simplest gate:

```python
RETRYABLE_FINISH_REASONS = {None, "OTHER", "FINISH_REASON_UNSPECIFIED"}
# Plus: MAX_TOKENS with bumped max_output_tokens on retry
```

Even a "don't retry `finish_reason=STOP`-with-malformed-JSON" guard would prevent the worst case. File it as a follow-up issue if you want to defer.

### Fresh adversarial pass — no new issues

Walked through the retry loop's edge cases:

- Batch-level exception mid-iteration → `remaining` unchanged, next attempt retries cleanly ✅
- Partial success inside `classify_sessions_via_api` → per-session `parse_error=True` entries are correctly identified and requeued ✅
- All 3 attempts fail → `remaining` stays full, `all_results` empty, `list(all_results.values())` returns `[]`, `retry_meta.retry_resolved=0` ✅
- `KeyboardInterrupt` propagates (not caught by `Exception`) ✅
- `max_output_tokens=65536` ceiling matches the top of the Gemini 2.5 Pro window ✅
- `retry_resolved` counter matches `all_results` success criterion (both use `not any(m.parse_error ...)`) ✅

### Verdict

**Approve.** Ship it. The important v2 regression (finding #1) is cleanly fixed with an explicit comment documenting why the broad catch is correct for a retry loop. The naming rename, dedup fix, and new test round out the cleanup. Finding #2 is a legitimate follow-up, not a blocker.

PR #54, opened 2026-04-17
Add quality evaluation report script

evekhm:
## Summary
- Adds a self-contained quality evaluation script (`scripts/quality_report.py`) that runs LLM-as-a-judge categorical evaluation over agent sessions stored in BigQuery using the SDK
- Includes shell wrapper (`scripts/quality_report.sh`) with env var validation and automatic log capture
- Adds `scripts/README.md` with usage documentation, env var reference, and metric descriptions
- Supports A2A response resolution, markdown report generation, and result persistence to BigQuery

## Features
- **Two evaluation metrics**: `response_usefulness` (meaningful/partial/unhelpful) and `task_grounding` (grounded/ungrounded/no_tool_needed)
- **Console output**: per-session details, per-agent quality table, unhelpful contribution ranking, category distributions
- **Markdown reports**: `--report` flag generates structured reports in `scripts/reports/`
- **Persistence**: `--persist` flag saves evaluation results back to BigQuery
- **A2A support**: automatically resolves responses from remote A2A agents via `A2A_INTERACTION` events
- **Flexible filtering**: `--limit`, `--time_period`, `--model` options

## Test plan
- [x] `python quality_report.py --help` works without env vars
- [x] `./scripts/quality_report.sh --limit 10` runs evaluation successfully
- [x] `./scripts/quality_report.sh --no-eval --limit 5` browses Q&A without evaluation
- [x] `./scripts/quality_report.sh --report --limit 10` generates markdown report
- [x] Verify A2A response resolution works correctly


caohy1988 (commented):
## Code review

Substantial, useful addition — a CLI wrapper around the SDK's categorical evaluation with agent breakdown, A2A resolution, and markdown export is exactly the kind of thing operators ask for. The shape is right. A handful of concrete issues to fix before merge, and some polish that can come later.

### Findings

**1. [P1] `.gitignore` contradicts the committed sample report — `.gitignore:17`, `scripts/reports/quality_report_20260417_214854.md` (confidence 10/10)**

```
# Script outputs
scripts/reports/
```

…but the PR also commits `scripts/reports/quality_report_20260417_214854.md`. `gitignore` only affects *untracked* files, so the committed sample will stay in the repo forever, and any user who later regenerates a report with the same filename pattern will have their local copy tracked-but-modified vs the committed one. Pick one:

- **Option A (recommended):** Move the sample to a non-ignored path, e.g., `docs/sample_quality_report.md` or `scripts/sample_report.md`, and update the README link.
- **Option B:** Keep `scripts/reports/` ignored and add a path exception for the sample: `!scripts/reports/sample_*.md`, then rename the file accordingly.
- **Option C:** Drop the committed sample entirely and inline a short example in `scripts/README.md`.

Any of these is fine, but the current state is inconsistent.

**2. [P1] Sample report contains a real project/dataset identifier — `scripts/reports/quality_report_20260417_214854.md:154-155` (confidence 9/10)**

The header masks with `####`, but the **Execution Details** section doesn't:

```
- **project:** agent-operations-ek-05
- **dataset:** logging.agent_events_v011_v3
```

If `agent-operations-ek-05` is a personal sandbox, low stakes — still worth scrubbing for a sample that ships in a public repo. If it's anything shared, higher stakes. Either way:

- Scrub project/dataset/location from the committed sample, OR
- Have `_write_md_report` redact these fields when a `--redact` flag is set, and generate the committed sample with that flag.

Second option also helps users who want to share their own reports without leaking infra names.

**3. [P2] Cross-PR: sample report shows stale `null_count` key — report line 158 (confidence 8/10)**

```
- **retry:** {'null_count': 1, 'retry_attempted': True, 'retry_resolved': 1, 'retry_unresolved': 0}
```

This is the same `null_count` naming that's now stale after PR #48's commit `6d4addb4` broadened retry scope to all parse errors. If PR #48 lands a rename (`failed_count` or `retried_count`), regenerate this sample so the two PRs don't show contradictory keys to users.

**4. [P2] No input validation on `--limit` or `--samples` — `scripts/quality_report.py:830`, `:870` (confidence 8/10)**

```python
parser.add_argument("--limit", type=int, default=100, ...)
parser.add_argument("--samples", type=str, default=None, ...)
```

- `--limit -5` passes. Gets sent as `@trace_limit = -5` to BigQuery → query error or no rows, with no actionable error message.
- `--limit 10000000` passes. Evaluates 10M sessions against a BQ `AI.GENERATE` that isn't meant for that scale.
- `--samples abc` reaches `int(samples)` at `:659` → `ValueError` with raw stack trace.

Suggested:

```python
def _positive_int(v):
    n = int(v)
    if n < 1:
        raise argparse.ArgumentTypeError(f"must be >= 1, got {n}")
    return n

parser.add_argument("--limit", type=_positive_int, default=100, ...)

def _samples_arg(v):
    if v == "all":
        return "all"
    n = int(v)
    if n < 1:
        raise argparse.ArgumentTypeError(f"--samples must be 'all' or >= 1")
    return n
```

Fail fast at the CLI layer with a useful message.

**5. [P2] No tests — `scripts/` (confidence 9/10)**

This PR adds ~1000 lines of Python with deterministic pure helpers: `_extract_a2a_text`, `_is_single_word_routing`, `_group_by_category`, `_build_agent_stats`, `resolve_trace_responses`. All have branching logic that will silently break if trace/event schemas drift. Zero tests. At minimum:

- `_extract_a2a_text` with artifacts, history, metadata, missing fields, non-dict input.
- `_build_agent_stats` with mixed categories, `unclassified`, A2A flag, empty input.
- `_is_single_word_routing` boundary conditions (empty, 1-word, >=20 chars).

These are 10-line tests each. The script will be maintained for real — unit-test coverage now saves a real debugging session later.

**6. [P3] Global mutable module state — `scripts/quality_report.py:210-220` (confidence 8/10)**

```python
def _load_config():
    global PROJECT_ID, DATASET_ID, TABLE_ID, DATASET_LOCATION, EVAL_MODEL_ID
    PROJECT_ID = _require_env("PROJECT_ID")
    ...

PROJECT_ID = None
DATASET_ID = None
...
```

This pattern works for a one-off script, but `quality_report.py` is 1000 lines and imported in multiple places. Makes unit testing painful (need to patch the module), and any future import-before-`_load_config()` code path sees `None` values. Cleaner:

```python
@dataclass(frozen=True)
class Config:
    project_id: str
    dataset_id: str
    table_id: str
    dataset_location: str
    eval_model_id: str

def load_config() -> Config: ...
```

Pass `config` through the functions that need it. One-time refactor, pays forward.

**7. [P3] Module-scope side effect on `os.environ` — `scripts/quality_report.py:95-96` (confidence 8/10)**

```python
os.environ.setdefault("GOOGLE_GENAI_USE_VERTEXAI", "true")
os.environ.setdefault("GOOGLE_CLOUD_LOCATION", "global")
```

Executes at *import* time. If anything ever imports this module (tests, notebooks, another script), the process environment gets mutated. Move inside `main()` or `_load_config()`.

**8. [P3] Logger f-strings — throughout `quality_report.py` (confidence 7/10)**

```python
logger.info(f"Project: {PROJECT_ID}, Dataset: {DATASET_ID}, Table: {TABLE_ID}")
```

Python logging convention is lazy formatting:

```python
logger.info("Project: %s, Dataset: %s, Table: %s", PROJECT_ID, DATASET_ID, TABLE_ID)
```

f-strings evaluate even when the log level filters the message out. Trips `pylint W1203`. Cheap to fix with a find/replace; also makes the log lines structured-log-friendly.

**9. [P3] A2A fallback silently returns raw JSON on parse failure — `:395` (confidence 7/10)**

```python
except (json.JSONDecodeError, TypeError):
    return c, span.agent or "remote_agent"
```

If the A2A payload is a malformed JSON string, the helper returns the *raw malformed text* as the response. Downstream, `resolved_map[sid]["response"]` contains what looks like a valid A2A response but is actually JSON garbage. At minimum:

```python
except (json.JSONDecodeError, TypeError):
    logger.warning("Failed to parse A2A payload for session %s", ...)
    return None, None
```

Let the session fall through with no response, same as the "no A2A detected" path.

**10. [P3] Unused `good` / `bad` entries in `_category_label` — `:513` (confidence 9/10)**

```python
labels = {
    "meaningful": "✅ HELPFUL",
    ...
    "good": "✅ GOOD",
    "bad": "❌ BAD",
}
```

`get_eval_metrics()` only defines `meaningful/unhelpful/partial` and `grounded/ungrounded/no_tool_needed`. Drop `good`/`bad` or wire them to a metric that uses them.

**11. [P3] CLI arg naming inconsistency — `--time_period` vs `--no-eval` (confidence 7/10)**

argparse convention is hyphens (`--time-period`). argparse auto-maps to `args.time_period`, so the code reading it doesn't change. Renaming now (while this is new) is free; renaming later is a break.

**12. [P3] Broad `except Exception` loses stack trace — `:557` (confidence 7/10)**

```python
try:
    result = run_evaluation(...)
except Exception as e:
    logger.error(f"Evaluation failed: {e}")
    sys.exit(1)
```

Debugging this is rough without a traceback. Use:

```python
except Exception:
    logger.exception("Evaluation failed")
    sys.exit(1)
```

`logger.exception` prints the stack inside the log pipeline, which is almost always what you want.

**13. [P3] Markdown `Response: {r[:500]}` silently truncates without a marker — `:823`, `:856` (confidence 8/10)**

If the agent's response is >500 chars, the sample loses the tail with no indication. Users reading the report think the response ended there.

```python
r_display = (r[:500] + "…") if len(r) > 500 else r
w(f"- **Response:** {r_display}")
```

Tiny, makes the output trustworthy.

**14. [P3] Shell wrapper ANSI color breaks in non-TTY — `scripts/quality_report.sh:37` (confidence 7/10)**

```bash
echo -e "\033[0;32mLog: ${LOG_FILE}\033[0m"
```

CI logs, piped output, or anyone redirecting stdout gets `\033[0;32m` literal text. Guard:

```bash
if [ -t 1 ]; then
    echo -e "\033[0;32mLog: ${LOG_FILE}\033[0m"
else
    echo "Log: ${LOG_FILE}"
fi
```

**15. [P3] Browse mode doesn't log to `scripts/reports/` — `scripts/quality_report.sh:31` (confidence 6/10)**

```bash
if [[ " $* " != *" --no-eval "* ]]; then
    # tee to log file
else
    # no logging
fi
```

Intentional? Browse runs can still produce valuable diagnostic output (A2A resolution counts, sessions-without-response). Consider logging both modes and name them accordingly (`quality_browse_${TIMESTAMP}.log`).

### What's good

- **Clean separation of concerns**: config loading, metric definition, trace helpers, resolution, evaluation, printing, markdown — each in its own region. Easy to follow despite the file length.
- **A2A resolution logic** is thoughtful — handles artifacts → history fallback, metadata-for-agent-name, JSON-or-dict content. Corner cases are genuinely corner cases.
- **`--no-eval` browse mode** is a great addition. Lowers the cost of dogfooding the script without firing off real Gemini calls.
- **Status indicators (🟢🟡🔴)** in the per-agent table instantly communicate where to look. Good ops UX.
- **Elapsed time + execution mode in `report.details`** — operators will thank you for this.
- **Samples controls (`--samples 20` / `--samples all`)** — exactly the knob users need for report readability vs completeness.

### Scope check: CLEAN

Script is self-contained, doesn't touch SDK source. Intent (quality report CLI) matches delivery. One small scope expansion: the markdown report feature with templated sections is arguably a separate feature from the CLI, but it's tightly coupled enough to justify shipping together.

### Verdict

**Request changes.** Findings #1 and #2 (gitignore contradiction + possible data leakage in the sample report) are the two that should be fixed before merge. Finding #3 (stale `null_count`) is a coordination concern with PR #48.

The rest (validation, tests, global state, logger style) are the kind of polish that makes the difference between "a script that works" and "a script that's maintainable." Worth doing now while the code is fresh; can land in a follow-up if you want to merge fast.

caohy1988:
Follow-up review findings:

1. **P3: `--time_period` is ignored in browse mode.** `run_browse()` currently always calls `client.list_traces(TraceFilter(limit=args.limit))`, so `./scripts/quality_report.sh --no-eval --time_period 7d` silently ignores the user’s time filter. The CLI advertises `--time_period` generally, not as eval-only, so this is a real UX mismatch. Suggested fix: mirror the eval path and build the browse-mode `TraceFilter` from `args.time_period` too.

2. **P3: the per-agent `[A2A]` tag can be wrong for mixed local/remote traffic under the same agent name.** `_build_agent_stats()` stores `is_a2a` only when the agent key is first created and never updates it afterward. If the first session for `agent-x` is local and a later one is A2A, the row is never tagged; if the first is A2A, the whole row is tagged even if most sessions are local. Suggested fix: either track `a2a_sessions` as a count and render the count explicitly, or update the per-agent flag as `any(is_a2a)` across all sessions for that agent.

caohy1988 (commented):
## Re-review v2 (head `203bbe60`)

Solid iteration — 14 of 15 v1 findings fully addressed, one explicitly optional. One new concern introduced by the approach to testing that I want to flag before merge.

### v1 findings: status

|---|---|---|---|
| 1 | `.gitignore` contradicted the committed sample report | ✅ Fixed | Sample moved from `scripts/reports/…` to `scripts/sample_report.md`, README link updated |
| 2 | Sample report leaked `agent-operations-ek-05` / real dataset | ✅ Fixed | Scrubbed to `my-gcp-project` / `agent_logs.agent_events` throughout, including Execution Details |
| 3 | Stale `null_count` key in sample | ✅ Fixed | Sample shows `{'failed_count': 1, ...}` matching PR #48 rename |
| 4 | No input validation on `--limit` / `--samples` | ✅ Fixed | `_positive_int`, `_samples_arg` validators wired into argparse |
| 5 | No tests for ~1000 lines of logic | ✅ Partially Fixed | New `tests/test_quality_report_helpers.py` with 388 lines covering `_is_single_word_routing`, `_extract_a2a_text`, `_group_by_category`, `_build_agent_stats`. But see new finding below. |
| 6 | Global mutable module state | ⚠️ Deferred | Still `global PROJECT_ID, ...` — was explicitly optional in v1, fine to defer |
| 7 | Module-scope `os.environ.setdefault` | ✅ Fixed | Moved inside `_load_config()` |
| 8 | Logger f-strings | ✅ Fixed | All call sites switched to `%s` / `%d` lazy formatting |
| 9 | A2A fallback silently returned raw JSON | ✅ Fixed | Now logs warning and returns `None, None` |
| 10 | Unused `good`/`bad` labels | ✅ Fixed | Dropped from `_category_label` |
| 11 | `--time_period` underscore inconsistency | ✅ Fixed | Renamed to `--time-period` everywhere (argparse, README, epilog, docstring) |
| 12 | Broad `except Exception` lost stack trace | ✅ Fixed | `logger.exception("Evaluation failed")` |
| 13 | Response truncation at 500 chars silent | ✅ Fixed | `r_display = (r[:500] + "…") if len(r) > 500 else r` — both unhelpful and partial sections |
| 14 | Shell ANSI color broke non-TTY | ✅ Fixed | `if [ -t 1 ]; then ... else ... fi` guard on both branches |
| 15 | Browse mode didn't produce log file | ✅ Fixed | Now writes `quality_browse_${TIMESTAMP}.log` |

### New finding

**[P2] Tests reimplement the helpers instead of importing them — `tests/test_quality_report_helpers.py:27-105` (confidence 9/10)**

The test file literally redefines `_is_single_word_routing`, `_extract_a2a_text`, `_group_by_category`, and `_build_agent_stats` inline, with the comment:

```python
# We can't exec the whole file because of module-level imports and
# side effects.  Instead, extract just the pure functions we test.
# This is fragile but acceptable for a test file.
```

This is not acceptable for a test file — it defeats the purpose of having tests. The tests verify *copies* of the logic, not the actual code in `quality_report.py`. Someone changes the real helper, forgets to update the copy, tests still pass, script is broken in production. The "regression guard" becomes a "regression hider."

**Root cause:** `quality_report.py` has module-scope side effects (`logging.basicConfig`, dotenv loading) that run at import time.

**Fix:** make the script importable by moving those side effects into `main()` or `_load_config()`:

```python
# At top of file: declare logger but don't configure
logger = logging.getLogger("quality_report")

def _configure_logging():
    logging.basicConfig(
        level=logging.INFO,
        format="%(asctime)s [%(levelname)s] %(message)s",
        datefmt="%H:%M:%S",
    )

def _load_dotenv():
    try:
        from dotenv import load_dotenv
        for candidate in [...]:
            if os.path.isfile(candidate):
                load_dotenv(candidate, override=True)
                break
    except ImportError:
        pass

def main():
    _configure_logging()
    _load_dotenv()
    _load_config()
    ...
```

Then in the test file:

```python
import sys, os
sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", "scripts"))
from quality_report import (
    _is_single_word_routing,
    _extract_a2a_text,
    _group_by_category,
    _build_agent_stats,
)
```

The existing test cases stay the same — only the import changes. Once this lands, the `_load_helpers()` function in the test can be deleted entirely.

### Fresh pass — minor observations

**[P3] Shell wrapper eval/browse branches are near-duplicates — `scripts/quality_report.sh:30-55` (confidence 7/10)**

The two branches differ only by `LOG_PREFIX` (`quality_report` vs `quality_browse`). Could collapse to one path:

```bash
if [[ " $* " != *" --no-eval "* ]]; then
    LOG_PREFIX="quality_report"
else
    LOG_PREFIX="quality_browse"
fi
REPORTS_DIR="${SCRIPT_DIR}/reports"
mkdir -p "${REPORTS_DIR}"
TIMESTAMP=$(date +"%Y%m%d_%H%M%S")
LOG_FILE="${REPORTS_DIR}/${LOG_PREFIX}_${TIMESTAMP}.log"
if [ -t 1 ]; then
    echo -e "\033[0;32mLog: ${LOG_FILE}\033[0m"
else
    echo "Log: ${LOG_FILE}"
fi
python3 "${SCRIPT_DIR}/quality_report.py" "$@" 2>&1 | tee "${LOG_FILE}"
```

Taste; not a bug.

**[P3] Sample report has lingering `text: '...'` prefix in agent responses — `scripts/sample_report.md` unhelpful sections (confidence 6/10)**

Many responses start with literal `text: '...'` wrapper (e.g., `"text: 'The most frequently occurring error message..."`). This is upstream (the agent's content field probably serialized as `{"text": "..."}` and the string form leaks through). Not this PR's job to fix, but worth a follow-up to unwrap known wrapper shapes in `resolve_trace_responses`.

### Verdict

**Approve with one recommendation.** The test-duplicates finding is the one I'd want fixed before merge — the value of 388 lines of test code depends on it actually testing the real functions. The fix is ~10 minutes: add `_configure_logging()`, move `load_dotenv` into a function, change the test's import path.

Everything else landed cleanly. Nice work on the sanitization, validation, and the `a2a_count` refinement in `_build_agent_stats` — that per-agent partial-A2A ratio (`[A2A:3/12]`) is a meaningful improvement over the boolean flag.

PR #56, opened 2026-04-17
fix: add ORDER BY before LIMIT for deterministic session ordering

### LINKED ISSUE #55: fix: LIMIT without ORDER BY returns arbitrary sessions
## Use case

When running quality evaluation on a live agent system, the standard workflow is:

```bash
# "Evaluate the latest 100 sessions"
quality_report.sh --limit 100 --report
```

Users expect `--limit 100` to return the **100 most recent sessions** — the ones that reflect the current state of the system. Instead, BigQuery returns an arbitrary sample, so:

- Two consecutive runs (seconds apart, same data) can evaluate completely different sessions, producing different agent distributions and quality scores
- The report cannot be used for trend analysis or before/after comparisons
- Users cannot trust `--limit` to give a consistent view of recent agent performance

## Problem

The categorical evaluation queries (`AI.GENERATE`, `AI.CLASSIFY`, transcript fetch) and `list_traces` use `GROUP BY session_id ... LIMIT @trace_limit` without an `ORDER BY` clause. BigQuery does not guarantee row ordering without `ORDER BY`, so the `LIMIT` picks an arbitrary subset.

## Affected queries

- `CATEGORICAL_TRANSCRIPT_QUERY` in `categorical_evaluator.py`
- `CATEGORICAL_AI_GENERATE_QUERY` in `categorical_evaluator.py`
- `build_ai_classify_query()` in `categorical_evaluator.py`
- `build_ai_generate_query()` in `categorical_evaluator.py`
- `_LIST_TRACES_QUERY` in `client.py`

## Proposed fix

Add `ORDER BY MAX(timestamp) DESC` before the `LIMIT` clause in each affected query. This ensures `--limit N` consistently returns the N most recent sessions.

## Impact

- Makes `--limit` behave as users expect (most recent first)
- Results become reproducible across runs (same data → same sessions)
- No API or schema changes needed — purely internal query fix

evekhm:
## Summary
- Adds `ORDER BY MAX(timestamp) DESC` before `LIMIT` in all session-fetching queries so that `--limit N` consistently returns the N most recent sessions
- Without this fix, BigQuery returns an arbitrary subset, causing different results on consecutive runs with identical data

## Changed files
- `categorical_evaluator.py` — 4 query templates: `CATEGORICAL_TRANSCRIPT_QUERY`, `CATEGORICAL_AI_GENERATE_QUERY`, `build_ai_classify_query()`, `build_ai_generate_query()`
- `client.py` — `_LIST_TRACES_QUERY`

Fixes #55

## Test plan
- [x] Run `quality_report.sh --limit 100` twice consecutively — verify same agent distribution
- [x] Verify sessions are ordered by recency (most recent first)
- [ ] Run existing unit tests

### DIFF
--- a/src/bigquery_agent_analytics/categorical_evaluator.py
+++ b/src/bigquery_agent_analytics/categorical_evaluator.py
@@ -249,6 +249,7 @@ FROM `{project}.{dataset}.{table}`
 WHERE {where}
 GROUP BY session_id
 HAVING LENGTH(transcript) > 10
+ORDER BY MAX(timestamp) DESC, session_id
 LIMIT @trace_limit
 """
 
@@ -275,7 +276,7 @@ WITH session_transcripts AS (
   WHERE {where}
   GROUP BY session_id
   HAVING LENGTH(transcript) > 10
-  ORDER BY MAX(timestamp) DESC
+  ORDER BY MAX(timestamp) DESC, session_id
   LIMIT @trace_limit
 )
 SELECT
@@ -397,7 +398,7 @@ WITH session_transcripts AS (
   WHERE {where}
   GROUP BY session_id
   HAVING LENGTH(transcript) > 10
-  ORDER BY MAX(timestamp) DESC
+  ORDER BY MAX(timestamp) DESC, session_id
   LIMIT @trace_limit
 )
 SELECT
@@ -468,7 +469,7 @@ WITH session_transcripts AS (
   WHERE {where}
   GROUP BY session_id
   HAVING LENGTH(transcript) > 10
-  ORDER BY MAX(timestamp) DESC
+  ORDER BY MAX(timestamp) DESC, session_id
   LIMIT @trace_limit
 )
 SELECT
--- a/src/bigquery_agent_analytics/categorical_evaluator.py
+++ b/src/bigquery_agent_analytics/categorical_evaluator.py
@@ -275,6 +275,7 @@ WITH session_transcripts AS (
   WHERE {where}
   GROUP BY session_id
   HAVING LENGTH(transcript) > 10
+  ORDER BY MAX(timestamp) DESC
   LIMIT @trace_limit
 )
 SELECT
@@ -396,6 +397,7 @@ WITH session_transcripts AS (
   WHERE {where}
   GROUP BY session_id
   HAVING LENGTH(transcript) > 10
+  ORDER BY MAX(timestamp) DESC
   LIMIT @trace_limit
 )
 SELECT
@@ -466,6 +468,7 @@ WITH session_transcripts AS (
   WHERE {where}
   GROUP BY session_id
   HAVING LENGTH(transcript) > 10
+  ORDER BY MAX(timestamp) DESC
   LIMIT @trace_limit
 )
 SELECT
--- a/src/bigquery_agent_analytics/client.py
+++ b/src/bigquery_agent_analytics/client.py
@@ -135,9 +135,11 @@ ORDER BY timestamp ASC
 
 _LIST_TRACES_QUERY = """\
 WITH trace_sessions AS (
-  SELECT DISTINCT session_id
+  SELECT session_id, MAX(timestamp) AS latest_ts
   FROM `{project}.{dataset}.{table}`
   WHERE {where}
+  GROUP BY session_id
+  ORDER BY latest_ts DESC
   LIMIT @trace_limit
 )
 SELECT
--- a/src/bigquery_agent_analytics/client.py
+++ b/src/bigquery_agent_analytics/client.py
@@ -135,11 +135,11 @@ ORDER BY timestamp ASC
 
 _LIST_TRACES_QUERY = """\
 WITH trace_sessions AS (
-  SELECT session_id, MAX(timestamp) AS latest_ts
+  SELECT session_id
   FROM `{project}.{dataset}.{table}`
   WHERE {where}
   GROUP BY session_id
-  ORDER BY latest_ts DESC
+  ORDER BY MAX(timestamp) DESC, session_id
   LIMIT @trace_limit
 )
 SELECT
--- a/tests/test_categorical_evaluator.py
+++ b/tests/test_categorical_evaluator.py
@@ -1902,3 +1902,61 @@ class TestRetryFailedSessions:
     assert len(results) == 2
     assert retry_meta["failed_count"] == 1
     assert retry_meta["retry_attempted"] is True
+
+
+# ------------------------------------------------------------------ #
+# ORDER BY before LIMIT — regression tests (#55)                       #
+# ------------------------------------------------------------------ #
+
+
+class TestQueryOrderByBeforeLimit:
+  """Ensure all session-fetching query templates have ORDER BY before LIMIT."""
+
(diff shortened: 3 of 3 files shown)


caohy1988 (commented):
## Code review

Good catch on the root cause — non-deterministic `LIMIT` without `ORDER BY` is a real correctness bug for anyone using `--limit N` on repeat runs, and the fix is the right shape. One concrete miss on the scope, plus a few polish notes.

### Findings

**1. [P1] `CATEGORICAL_TRANSCRIPT_QUERY` is missing the `ORDER BY` fix — `categorical_evaluator.py:222-245` (confidence 9/10)**

The PR description claims all 4 session-fetching templates are covered, but the diff only touches 3: `CATEGORICAL_AI_GENERATE_QUERY`, `build_ai_classify_query()`, `build_ai_generate_query()`. `CATEGORICAL_TRANSCRIPT_QUERY` still ends with:

```sql
GROUP BY session_id
HAVING LENGTH(transcript) > 10
LIMIT @trace_limit
```

This template is used by `client._categorical_api_fallback` (`client.py:1408`) — the Python-side Gemini API fallback path that runs when BQ `AI.GENERATE` is unavailable or when the caller explicitly forces API mode. Any user who's on that code path (no `AI.GENERATE` connection configured, or `use_api=True`) still gets a non-deterministic subset under `--limit N`.

Fix:

```sql
GROUP BY session_id
HAVING LENGTH(transcript) > 10
ORDER BY MAX(timestamp) DESC
LIMIT @trace_limit
```

Same one-line add as the other three templates.

**2. [P2] No tiebreaker — sessions with identical `MAX(timestamp)` can still shuffle — all 4 templates (confidence 7/10)**

`ORDER BY MAX(timestamp) DESC` is deterministic *up to ties*. Two sessions whose most recent event shares a millisecond (rare but not impossible in batch-ingested event streams or high-QPS systems) can still swap positions across runs. Full-determinism form:

```sql
ORDER BY MAX(timestamp) DESC, session_id
```

`session_id` is unique per group, so the two-column ordering is a total order. Cheap.

**3. [P3] `latest_ts` is selected but never used — `client.py:137` (confidence 8/10)**

```sql
WITH trace_sessions AS (
  SELECT session_id, MAX(timestamp) AS latest_ts
  ...
  ORDER BY latest_ts DESC
  LIMIT @trace_limit
)
SELECT e.event_type, ...
FROM `...` e
JOIN trace_sessions ts ON e.session_id = ts.session_id
```

The outer SELECT only uses `ts.session_id`. `latest_ts` is computed for the CTE's own ORDER BY and then dropped. BigQuery's optimizer almost certainly prunes it, but if you want to be explicit:

```sql
WITH trace_sessions AS (
  SELECT session_id
  FROM `{project}.{dataset}.{table}`
  WHERE {where}
  GROUP BY session_id
  ORDER BY MAX(timestamp) DESC
  LIMIT @trace_limit
)
```

`ORDER BY MAX(timestamp) DESC` without aliasing works fine against a `GROUP BY session_id`. Matches the shape of the other 3 templates and drops one line.

**4. [P3] Test plan unchecked, no new tests — `## Test plan` (confidence 7/10)**

The PR description has `[ ] Run existing unit tests` unchecked, and no new test asserts the ORDER BY is actually present. A regression where someone edits the template and drops `ORDER BY` would not be caught. One-liner test per template:

```python
def test_transcript_query_orders_before_limit():
    sql = CATEGORICAL_TRANSCRIPT_QUERY
    order_pos = sql.index("ORDER BY MAX(timestamp)")
    limit_pos = sql.index("LIMIT @trace_limit")
    assert order_pos < limit_pos
```

Four such tests pin all four templates. Trivial to add.

### Scope check: mostly CLEAN, one miss

Intent: add `ORDER BY MAX(timestamp) DESC` before `LIMIT` in all session-fetching queries.
Delivered: 3 of 4 templates in `categorical_evaluator.py` + 1 in `client.py` — `CATEGORICAL_TRANSCRIPT_QUERY` missed.

### Verdict

**Request changes.** Finding #1 is the one that needs fixing before merge — the PR is advertised as comprehensive but leaves the API fallback path broken. Two-line fix. Findings #2-4 are polish and can land separately if you prefer.

caohy1988:
Follow-up re-review on current head `3870a336`:

The main requested-changes finding is still present as posted in the review:

- `CATEGORICAL_TRANSCRIPT_QUERY` in `categorical_evaluator.py` still ends with `GROUP BY session_id / HAVING LENGTH(transcript) > 10 / LIMIT @trace_limit` and still does **not** have the new `ORDER BY MAX(timestamp) DESC` before the limit.
- That query is still used by `client._categorical_api_fallback()` (`client.py:1408`), so the Python Gemini API fallback path still returns a non-deterministic subset under `--limit N`.

I re-ran `pytest tests/test_categorical_evaluator.py tests/test_sdk_client.py -q` on this head; result was `137 passed, 1 warning`. I don’t see a second blocker beyond the one already called out in the review, but the requested-changes verdict is still correct until the transcript-query path is fixed too.

caohy1988 (commented):
## Re-review v2 (head `a5e226b3`)

Clean sweep — all four v1 findings addressed in this commit. Nice turnaround.

### v1 findings: status

|---|---|---|---|
| 1 | `CATEGORICAL_TRANSCRIPT_QUERY` missed the ORDER BY fix | ✅ Fixed | `categorical_evaluator.py:244` now has `ORDER BY MAX(timestamp) DESC, session_id` before `LIMIT` |
| 2 | No tiebreaker — ties on `MAX(timestamp)` still non-deterministic | ✅ Fixed | All 5 templates use `ORDER BY MAX(timestamp) DESC, session_id` |
| 3 | Unused `latest_ts` column in `_LIST_TRACES_QUERY` | ✅ Fixed | `client.py:137-142` — simplified to `SELECT session_id ... GROUP BY session_id ORDER BY MAX(timestamp) DESC, session_id` |
| 4 | No tests asserting ORDER BY before LIMIT | ✅ Fixed | `TestQueryOrderByBeforeLimit` — 5 tests covering all 4 categorical templates, including a consolidated `test_all_queries_have_session_id_tiebreaker` |

### Fresh pass — no new issues

- All 5 session-fetching queries (4 in `categorical_evaluator.py`, 1 in `client.py`) now have `ORDER BY MAX(timestamp) DESC, session_id LIMIT @trace_limit`. Inventory matches the issue #55 affected-queries list exactly.
- `session_id` as secondary sort is ASC by default, giving a total order — deterministic across runs.
- No concerns on `MAX(timestamp)` returning NULL: `GROUP BY session_id` guarantees each group has ≥1 row, and event-stream rows always carry timestamps.
- BigQuery will push the sort into the partition scan for timestamp-partitioned tables; minimal perf cost.

### Tiny nit (non-blocking)

The `test_all_queries_have_session_id_tiebreaker` test covers the 4 queries in `categorical_evaluator.py`. `_LIST_TRACES_QUERY` in `client.py` isn't included in that test (since it lives in the client module). If someone later edits `_LIST_TRACES_QUERY` and drops the tiebreaker, the existing tests won't catch it. One extra assertion in a `test_client.py` file (or importing the constant into the existing test class) would close the loop. Optional.

### Verdict

**Approve.** Ship it. Complete fix, proper regression tests, full determinism.

PR #61, opened 2026-04-20
feat: Add agent improvement cycle demo

evekhm:
## Summary

- Adds an end-to-end demo showing a closed-loop agent improvement cycle powered by the BigQuery Agent Analytics SDK
- The demo runs a deliberately flawed Q&A agent, logs sessions to BigQuery via BigQueryAgentAnalyticsPlugin, evaluates quality with quality_report.py, and auto-improves the prompt using Gemini
- Quality climbs from ~30% to ~90%+ across 3 cycles
- Enhances scripts/quality_report.py with --app-name filtering, --output-json structured output, and --threshold configurable unhelpful rate warning

### Demo structure

- `agent/` - ADK agent with versioned prompts and intentional v1 flaws
- `eval/` - Test cases + InMemoryRunner-based eval that logs sessions to BigQuery
- `improver/` - Reads quality report JSON, calls Gemini to fix prompt and extend eval cases
- `run_cycle.sh` - Orchestrator with detailed step-by-step explanations
- `setup.sh` - One-time setup (APIs, deps, BigQuery dataset, .env)
- `DEMO_SCRIPT.md` - Presenter walkthrough with setup, talking points, and expected output

### SDK enhancements

- `quality_report.py --app-name` filters evaluation to a specific agent sessions
- `quality_report.py --output-json` produces structured JSON for programmatic consumption
- `quality_report.py --threshold` makes the unhelpful rate warning configurable (default: 10%)
- Fix `load_dotenv(override=True)` to `load_dotenv(override=False)` in `quality_report.py` -- the override caused repo-root .env values to overwrite caller-provided env vars, breaking any demo/script that sets its own DATASET_ID/TABLE_ID before invoking the report

## Test plan

- [x] Tested full 3-cycle run on Cloud Shell with a fresh GCP project
- [x] Verified setup.sh enables APIs, creates dataset, writes .env
- [x] Verified quality score progression from ~30% to ~90%+
- [x] Verified --app-name correctly filters to agent-specific sessions
- [x] Verified --output-json produces valid JSON consumed by the improver
- [x] Verified load_dotenv fix resolves 0-session issue when caller sets env vars
- [x] Review prompt versioning across cycles (git diff agent/prompts.py)


caohy1988 (commented):
Solid end-to-end demo that tells a compelling story. The SDK-level additions in `scripts/quality_report.py` are small, additive, and sensible. My concerns are mostly in the `--output-json` surface plus a few safety flags in the improver and shell orchestration.

This is a code-read review against `origin/main..pr-61`; I did not run the full suite locally.

---

## Medium — `scripts/quality_report.py`

### 1. `--output-json -` (stdout) corrupts piped output

`scripts/quality_report.py:541-547`

```python
if args.output_json == "-":
    json.dump(output, sys.stdout, indent=2, default=str)
    print(f"\n  JSON report: (stdout)")
```

`json.dump` writes to stdout, then a trailing `print` goes to stdout too. Any `… --output-json - | jq` consumer gets invalid JSON (`  JSON report: (stdout)` trailing the document). Fix: write that status line to `sys.stderr`, or skip it entirely when the target is stdout.

### 2. `_build_json_output` ignores the `--samples` cap

`scripts/quality_report.py:1007-1056`

The text printer respects `samples=args.samples` to cap per-category output; the JSON writer dumps every session. That's fine for the improver (which wants all of them), but callers see `--samples` and `--output-json` as independent knobs and reasonably assume the JSON respects the cap. Worth a one-line note on the `--output-json` help string (`… writes all sessions regardless of --samples`).

---

## Medium — `improver/improve_agent.py`

### 3. Only quality check on the improved prompt is `len(...) < 50`

`examples/agent_improvement_cycle/improver/improve_agent.py:203-205`

The `compile()` check at line 240 catches Python syntax errors — good belt-and-suspenders. But it won't catch a case where Gemini hallucinates away most of the original prompt and returns 60 chars of unrelated text. For a demo that showcases a virtuous cycle, a soft sanity check (e.g. verify at least one of `pto`, `sick_leave`, `remote_work` survives, or reject prompts strictly shorter than the original) would prevent embarrassment. Not blocking for a demo; worth a TODO comment so readers who copy this pattern understand the limitation.

### 4. `add_eval_cases` doesn't validate the LLM-returned case schema

`examples/agent_improvement_cycle/improver/improve_agent.py:253-270`

```python
existing_ids = {c[\"id\"] for c in data[\"eval_cases\"]}
...
for case in new_cases:
    if case[\"id\"] not in existing_ids:
        data[\"eval_cases\"].append(case)
```

If Gemini returns a case missing `question` or `category`, the improver writes it to disk and `run_eval.py` crashes on the next cycle with `KeyError: 'question'`. Add a schema check (required keys: `id`, `question`, `category`, `expected_tool`) before appending, and skip + log on failure so one bad case doesn't poison the suite.

---

## Low / nitpick

### 5. `run_cycle.sh` has a fixed `sleep 5` for BigQuery propagation

`examples/agent_improvement_cycle/run_cycle.sh:181`

Five seconds is usually fine on the happy path but flaky on slow networks or cold datasets. A retry loop on the quality-report query (up to ~60s) would be more resilient for Cloud Shell users.

### 6. In-place mutation of checked-in files between runs

The improver writes to `agent/prompts.py` and `eval/eval_cases.json` in place — that's the point. But the README doesn't tell users what to do between fresh demo runs. A one-liner like

> Between fresh runs, reset the demo with `git checkout -- examples/agent_improvement_cycle/agent/prompts.py examples/agent_improvement_cycle/eval/eval_cases.json`

would save pain.

### 7. Reproducibility claim \"30% → 90%+ across 3 cycles\"

`examples/agent_improvement_cycle/README.md:34`

Depends on non-deterministic Gemini output. Fine as a demo claim; consider softening to \"~30% → ~90% on typical runs\" or linking to a stored sample report.

### 8. `--app-name` filter relies on `attributes.root_agent_name` being populated

`scripts/quality_report.py:389, 446`

`TraceFilter.root_agent_name` matches `JSON_VALUE(attributes, '$.root_agent_name')` (verified at `src/bigquery_agent_analytics/trace.py:576`). Sessions logged by `BigQueryAgentAnalyticsPlugin` populate this; sessions from other sources may not. Worth a one-line help note so users who get empty results know where to look.

### 9. `run_eval.py:49-53` requests a specific `session_id` but then uses `session.id`

The ADK runner may or may not honor the caller-supplied ID. Unlikely to matter for the demo (filtering is by `app_name`, not session prefix), so this is a latent inefficiency at worst. Consider dropping the explicit UUID generation since it's unused downstream.

---

## Kudos

- `setup.sh` is properly idempotent (`bq show` before `bq mk`, `.env` overwrite guard, APIs enabled safely).
- The `improve_agent.py` retry loop on `compile()` failure is a good guardrail against malformed LLM output.
- `run_cycle.sh` explains every step in long-form comments — great teaching tool.
- `GOOGLE_GENAI_USE_VERTEXAI=True` + explicit `GOOGLE_CLOUD_PROJECT` / `GOOGLE_CLOUD_LOCATION` in `agent.py` is the right Cloud Shell portability choice.

---

**Recommendation:** Address (1), (2), and (4) before merge — those have clear user-visible impact (`jq` pipelines break, `--samples` surprise, next-cycle crashes). (3) is worth at least a TODO comment. Everything else is soft / optional.

caohy1988:
Additional code-read findings that were not covered in the earlier commented review:

1. Medium: `examples/agent_improvement_cycle/run_cycle.sh:182-190`

Step 2 swallows `quality_report.py` failures with `|| true`, then only checks whether `quality_report_cycle_${cycle}.json` exists. Because that filename is stable across reruns, an old report can satisfy the check and Step 3 can proceed using stale data after a fresh evaluation failure. I think this should either fail hard here, or delete/check freshness of the target JSON before invoking `quality_report.py`.

2. Medium: `examples/agent_improvement_cycle/run_cycle.sh:183-187`, `examples/agent_improvement_cycle/improver/improve_agent.py:102`

The demo advertises `--cycles 3`, but Step 2 hardcodes `--limit 15` while the improver is instructed to add `2-4` new eval cases every cycle. The suite starts at 10 cases, so by cycle 3 the quality report can stop evaluating the full current suite and the headline improvement number becomes a suffix-of-run metric instead of a whole-suite metric.

3. Medium: `examples/agent_improvement_cycle/improver/improve_agent.py:287-295`

The improver still rewrites `prompts.py` / `eval_cases.json` when the JSON report contains zero sessions. If BigQuery propagation is slower than the fixed sleep or the `--app-name` filter misses, `format_problem_sessions()` becomes "No problem sessions found." and the script still asks Gemini to mutate the agent on effectively no signal. A guard on `summary.total_sessions == 0` (or a minimum-session threshold) would make the loop much safer.

4. Medium: `examples/agent_improvement_cycle/eval/run_eval.py:97-114`

Per-case exceptions are converted into local `ERROR: ...` records and the loop continues, but those failed cases never make it into BigQuery, so Step 2 scores only successful traces. That can inflate `meaningful_rate` exactly when eval cases are crashing. I’d either fail the cycle on any case error or explicitly fold the error count into the reported score.

5. Low: `examples/agent_improvement_cycle/setup.sh:59-60,93-95`, `examples/agent_improvement_cycle/README.md:190-209`

The quickstart docs tell users to export `PROJECT_ID` and optionally `DATASET_LOCATION`, but `setup.sh` actually reads the project from `gcloud config` and the location from `BQ_LOCATION`. Following the README literally can still pick the wrong project or silently fall back to `us-central1`. I’d either honor the documented env vars or update the docs to match the script.

This was a code-read against refreshed `origin/main..pr-61`; I did not rerun the full demo.

caohy1988 (commented):
Re-reviewed at `e0d09ec`. Every item from my earlier pass was addressed, most with stronger designs than I proposed. Specifically the replacement of a heuristic \"does the prompt look OK\" check with an actual behavioral regression gate is a material upgrade — that's now a correctness guarantee rather than a smell test. Good work.

## Verified fixes

| Original finding | Resolution |
| --- | --- |
| `--output-json -` corrupts stdout | ✅ status line redirected to `sys.stderr` |
| `--samples` vs `--output-json` docs | ✅ help text updated |
| `--app-name` / `root_agent_name` docs | ✅ help text clarified |
| Weak prompt quality check (`len < 50` only) | ✅ **replaced** with golden eval gate |
| `add_eval_cases` no schema validation | ✅ `_REQUIRED_CASE_KEYS`, skip-and-log on malformed |
| Fixed `sleep 5` BQ propagation | ✅ 6-attempt retry with 10s backoff |
| In-place file mutation without reset docs | ✅ README now has `git checkout --` instructions |
| Reproducibility claim | ✅ softened with non-determinism caveat |
| Unused UUID in `run_eval.py` | ✅ removed |

Dedup-by-question-text in `add_eval_cases` is a nice touch beyond just dedup-by-ID. Type annotations were added across the board — bonus.

---

## New findings in the added code (`e0d09ec`)

### Medium — Golden eval cost grows unbounded

`improver/improve_agent.py:221-282`

Each improvement attempt runs the full golden set: N agent calls + N judge calls = 2N Gemini requests per attempt. With up to 3 retries that's **6N Gemini requests per cycle**, purely for validation. Since `extract_failed_cases` grows the golden set every cycle, this compounds — cycle 1 might validate against 10 cases, cycle 5 could be 25+. The README doesn't flag this cost/runtime growth, and there's no cap.

Consider either:
- Capping golden-set growth (e.g. keep the N hardest cases), or
- Sampling a subset of the golden set for the gate (with full run on the final accepted candidate), or
- At minimum, calling it out in the README's \"Cost notes\" section.

### Medium — Golden gate has no graceful degradation

`improver/improve_agent.py:461-466`

```python
if not passed_all:
    print(\"  Golden eval FAILED: candidate breaks existing cases.\")
    if attempt == 2:
        raise ValueError(\"Golden eval failed after 3 attempts\")
    ...
```

If three consecutive Gemini drafts all regress the golden set, the cycle hard-fails. Plausible scenario: the golden set has absorbed hard extracted cases that no reasonable prompt can fully solve. The demo halts mid-cycle with a traceback, which is a rough UX for \"quality improved partway but we can't lock it in.\"

Consider falling back to the highest-passing candidate after N retries rather than raising, or at least logging clearly which golden cases keep regressing so the user knows where the cycle got stuck.

### Low — Lenient judge prompt doesn't check tool use

`improver/improve_agent.py:99-114`

```
A response PASSES if it provides a specific, substantive answer to the question.
A response FAILS if it says \"I don't know\", defers to HR, or gives vague/generic information without specifics.
```

An agent that hallucinates confident-sounding specifics passes this gate. Since each eval case has an `expected_tool` field, the judge could cross-check whether the tool was actually invoked. Without that, the gate catches unhelpful responses but not ungrounded ones — and ungrounded responses are exactly what `task_grounding` in the quality report is designed to catch.

Passing the expected tool into the judge prompt and failing responses that claim specifics without tool evidence would close this gap.

### Low — `generate_traffic.py` doesn't enforce its own constraints

`eval/generate_traffic.py:89-114`

The prompt asks Gemini to \"avoid duplicating\" existing questions and to generate `{count}` diverse questions. Neither is enforced in code:
- No post-check that returned cases are disjoint from `existing_questions` (trusting the model 100%).
- No check that `len(cases) >= count`; if Gemini returns fewer, the cycle silently uses whatever came back.

For a demo this is acceptable, but a `if len(cases) < count / 2: retry` fallback and a post-dedup filter would harden the traffic generator against bad runs.

### Low — Golden-eval agent drops the production agent's retry options

`improver/improve_agent.py:246-253`

```python
test_agent = Agent(
    name=\"golden_eval_agent\",
    model=model_id,  # plain string
    ...
)
```

The production agent in `agent/agent.py` uses `Gemini(model=MODEL_ID, retry_options=types.HttpRetryOptions(attempts=3))`. The throwaway test agent passes the model as a string, losing retry-on-transient-error. On a flaky network the gate can spuriously fail. Mirror the production agent's model configuration here.

### Low — `extract_failed_cases` throws away category

`improver/improve_agent.py:286-324`

Every extracted case gets `category=\"unknown\"` and `expected_tool=\"lookup_company_policy\"`. But the synthetic traffic JSON that produced the failed session has the real `category` (pto, benefits, expenses, etc.). Joining back on `session.question` against the synthetic traffic file would preserve that metadata. Makes the golden set more useful for per-topic analysis over time.

---

## Kudos

- Golden eval gate as a replacement for LLM-based \"validation\" is a substantively better design. Switching from \"does this look right?\" to \"does this actually work?\" is the right call.
- The eval case schema validation with skip-and-log (not skip-and-silently-drop) is clean.
- Six-attempt retry loop with 10s backoff for BigQuery propagation is well-calibrated.
- Dedup by both ID and question text.
- Type annotations throughout.
- README's new Guardrails section documents the defense-in-depth clearly.

---

## Recommendation

None of the new findings are blocking. The two mediums are about cost/UX of a longer-running demo, not correctness. The three lows are quality improvements for a future iteration. I'd be comfortable merging as-is; the Medium-1 cost note is worth a docs line before ship.

evekhm:
## Latest changes (830cbff)

**Fixes:**
- **dotenv override bug**: `quality_report.py` used `load_dotenv(override=True)` which overwrote the demo's `DATASET_ID`/`TABLE_ID` env vars with values from the repo root `.env`, causing the quality report to query the wrong BQ table (0 sessions found). Changed to `override=False` so caller env vars take precedence.
- **BQ retry validation**: Step 3 retry loop now validates `total_sessions > 0` before accepting the quality report. Previously it broke on exit 0 even with an empty report.
- **BQ staleness guard**: Step 5 retry loop compares session IDs against Step 3 report to detect stale data from streaming buffer propagation delay.

**Improvements:**
- **Parallel eval**: `run_all_cases` and `run_golden_eval` now use `asyncio.gather` for ~5-8x speedup on eval runs.
- **Remove duplicate regression check**: Step 4 already validates the candidate against all golden cases — removed redundant re-run in Step 5.
- **Suppress warnings**: Added `PYTHONWARNINGS=ignore` to suppress authlib deprecation warnings.
- **Reset via git checkout**: `reset.sh` uses `git checkout` instead of copying baseline files. Removed redundant `prompts_v1.py` and `eval_cases_v1.json`.

**Demo results (single cycle):**
- Before (V1): 40% meaningful
- After (V2): 90% meaningful

evekhm:
## SDK fix: `quality_report.py` dotenv override

**File:** `scripts/quality_report.py`  
**Change:** `load_dotenv(candidate, override=True)` → `load_dotenv(candidate, override=False)`

### Problem

`_load_dotenv()` in `quality_report.py` loads the repo root `.env` with `override=True`. This overwrites any environment variables already set by the caller (e.g., `DATASET_ID`, `TABLE_ID`, `PROJECT_ID`).

When another script (like the demo's `run_cycle.sh`) exports its own config and then calls `quality_report.py`, the repo root `.env` silently replaces those values — causing the quality report to query the wrong BigQuery table and return 0 sessions.

### Fix

Change to `override=False`, which is the standard `python-dotenv` convention: environment variables set by the caller take precedence over `.env` file values. The `.env` file acts as a fallback for values not already in the environment.

This is a one-line fix but affects any workflow that sets env vars before calling `quality_report.py`.

caohy1988:
## Latest-version re-review (e2a061a6)

Prior findings addressed cleanly: stale report removal, 0-session guard, golden-set cost note, graceful skip-on-failure, `HttpRetryOptions` for 429s. The flow now looks coherent end-to-end. Below are issues I didn't flag before that are worth considering before merge.

---

### High — Pre-flight auto-fix silently rewrites the committed baseline

`run_cycle.sh` pre-flight block (lines ~119-152): if the committed V1 prompt fails the committed golden set, the script calls `improve_agent.py --from-eval-results` to auto-overwrite `prompts.py` with a new version before any cycle runs. That means the committed baseline can be silently edited in a scenario that is itself a regression signal (the shipping baseline does not pass its own tests).

This is useful for robustness but hides the signal. For a demo that will be shown to others, I'd prefer:

- Print a clear `WARNING: committed V1 fails golden eval (N/M). Auto-recovery enabled.` before overwriting, or
- Require `--auto-recover` to opt in, with a non-zero exit otherwise.

Either makes the regression visible and the behavior predictable in screencasts.

### High — Golden judge rewards form, not correctness

`run_eval.py:161-169` and `improve_agent.py:103-108` use this judge prompt:

> A response PASSES if it provides a specific, substantive answer to the question.  
> A response FAILS if it says "I don't know", defers to HR, or gives vague/generic information without specifics.

This judges the *shape* of the answer, not whether it matches tool output. A prompt that confidently confabulates "meal limit is $500, receipts required over $100" would pass — the tools return different numbers, but the judge has no ground truth. With `temperature=0.0` on the agent this may not trigger often in practice, but the demo's headline claim is "quality 30% → 90%". If someone stress-tests by deliberately perturbing the tool data vs. the prompt's internal "knowledge", the gate would not catch factually wrong but confidently worded answers.

Cheap fix: include the expected tool and an expected-fact sketch in each golden case and have the judge compare. The existing `expected_tool` field is already on each case and is currently unused by the judge.

### Medium — `extract_failed_cases` hardcodes `expected_tool="lookup_company_policy"`

`improve_agent.py:344`: every extracted case gets `"expected_tool": "lookup_company_policy"`, regardless of whether the failing question was about dates (should use `get_current_date`), out-of-scope topics (should be rejected), or sessions that had tool errors. This (a) makes the golden set misleading as documentation of the agent's expected behavior, and (b) if a future improvement *does* wire `expected_tool` into the judge (per the finding above), the assertion will be wrong.

Minimal fix: classify from the question (`"today" / "date" / "holiday"` → `get_current_date`), default to `unknown` rather than `lookup_company_policy`.

### Medium — Step 5 freshness guard accepts mixed old+new sessions

`run_cycle.sh:380-395`:

```python
print('yes' if new and new != old else 'no')
```

If Step 5 returns a set that partially overlaps Step 2 (e.g. 4 fresh + 6 stale due to a partial propagation window and `--limit $TRAFFIC_COUNT=10`), `new != old` is `True` and the guard accepts. The "after" score is then computed over a mixed population, skewing the before/after delta.

Correct check:

```python
print('yes' if new and new.isdisjoint(old) else 'no')
```

### Medium — `write_improved_prompt` text-substitutes on an LLM-controlled blob

`improve_agent.py:369-395`: the whole prompts.py is rewritten by:

1. `content.replace(triple_q, ...)` — escapes triple quotes in the improved prompt.
2. `re.sub(r"CURRENT_PROMPT\s*=\s*PROMPT_V\d+", ...)` — rewrites the current-version pointer.
3. `content.replace(current_prompt_line, new_block + "\n" + current_prompt_line)` — inserts the new block.

Failure modes:

- If the improver returns a prompt containing the literal substring `CURRENT_PROMPT = PROMPT_V1` (say, because the improved prompt quotes an example), step (2) will also rewrite that occurrence inside the new prompt block.
- If the improver returns `improved_prompt` with a bare `"""` that doesn't match the simple escape (e.g. in an f-string template example), the final `compile()` check catches it, but you lose the attempt silently without surfacing what the model emitted.

Both are edge cases, but the file you're mutating is load-bearing. Consider:

- Parse with `ast` and replace the `PROMPT_V{n+1}` assignment node, or
- Maintain prompts in a separate YAML/JSON file and read from there; `prompts.py` becomes a thin loader. That also makes version diffs readable without re-running the cycle.

### Low — Parallel Gemini calls without a concurrency cap

`run_eval.py` and `improve_agent.py` both use `asyncio.gather(*[_ for c in cases])` with no semaphore. For the default `TRAFFIC_COUNT=10` plus a judge call per case, that's 20 concurrent Gemini calls at Step 5 golden-eval time, plus another N concurrent during the improver regression gate (N = growing golden set size). Commit `6591599` added a `Semaphore(3)` throttle and then reverted it at `f91101f` in favor of `HttpRetryOptions(attempts=3)`. Retries will mask transient 429s but won't fix a sustained overload pattern: 3 attempts × `initial_delay=10s` is 30s of backoff against a quota that resets per-minute. On a first-try Vertex project with tight quotas this will flake.

Minimal: re-add a semaphore (say, `Semaphore(5)`). Keep `HttpRetryOptions` as the safety net.

### Low — `improve_agent.py:run_golden_eval` is invoked via `asyncio.run(...)` inside a `for` loop

`improve_agent.py:555`: each retry attempt calls `asyncio.run(run_golden_eval(...))`, which tears down the event loop between attempts. That in turn forces re-creation of the `InMemoryRunner` and re-import-from-sys.path each attempt (via `_create_eval_agent`). Not a correctness issue but doubles the setup cost per retry. One outer `async def main(): for attempt in range(3): await run_golden_eval(...)` would keep the runner/session service warm.

---

Nothing above is blocking — the demo works. The golden-judge-rewards-form and auto-recovery-hides-regression items are the only ones I'd consider pre-merge polish for a public demo; the rest are follow-ups.

caohy1988:
## Pre-merge re-review (latest, `19f6e3d1`)

Substantial rework since the prior round. All seven CI jobs pass. Mergeable but **`BEHIND` origin/main** — needs a rebase before merge (main has #62, #68, #70 since the branch diverged).

### What got significantly better

- **Proper package split.** `improver/improve_agent.py` (single 625-line file) → `agent_improvement/` package with `config.py`, `config_loader.py`, `eval_runner.py`, `improver_agent.py`, `prompt_adapter.py`, `prompts.py`, `tool_introspection.py`. Much cleaner separation of concerns.
- **LoopAgent-based improver.** Instead of a procedural retry loop, the improver is now an ADK `LoopAgent` wrapping an inner `Agent` with six tools (`read_quality_report`, `read_current_prompt`, `generate_candidate`, `test_candidate`, `write_prompt`, `exit_loop`). Idiomatic for ADK and the loop exit semantics are explicit.
- **Teacher-agent ground truth** (`improver_agent.py:_generate_ground_truth`). Directly addresses the prior **"golden judge rewards form, not correctness"** High finding — the teacher runs with the same tools and a stronger prompt, producing ground-truth answers grounded in real tool output. Fed to the Vertex Optimizer as `target_response` examples.
- **Vertex AI Prompt Optimizer integration** (`_generate_via_vertex_optimizer`) with `OPTIMIZATION_TARGET_FEW_SHOT_TARGET_RESPONSE` mode and `HttpRetryOptions(attempts=6, codes=[429,503])`.
- **Vertex AI Prompt Registry storage** (`VertexPromptAdapter`) with local-file mirror for git visibility.
- **Config-driven** via `config.json` (app_name, agent_module, prompts_path, eval_cases_path, traffic_generator, model_id, prompt_storage, vertex_prompt_id, use_vertex_optimizer) — demo now works for any ADK agent, not just the HR-policy one.
- **Tool introspection** (`tool_introspection.py`) builds the "Available Tools" block from live function signatures + docstrings, so the improver prompt stays consistent with the agent's real tool set.
- **`jq` replaces inline Python JSON parsing** in `run_cycle.sh`.

---

### Still open from prior review

#### High — Pre-flight auto-fix silently rewrites the committed baseline

`run_cycle.sh:227`:

```bash
echo "  PRE-FLIGHT FAILED: Auto-improving prompt to pass golden eval set..."
```

If the committed V1 fails its own golden eval, the script auto-improves V1 in place. The baseline-regression signal gets hidden; a screencast-driving operator never sees why V1 failed. Recommendation unchanged: require `--auto-recover` to opt in, or surface a loud `WARNING` before the auto-fix. For a demo that ships as canonical example code, committing with this behavior is fine **only** if the first-run behavior is deterministic (i.e. V1 always passes its own golden eval at `main`); if not, reviewers will see a committed baseline that fails its own tests.

#### Medium — Step 5 freshness guard accepts mixed old+new

`run_cycle.sh:455`:

```bash
IS_FRESH=$( [[ -n "$NEW_IDS" && "$NEW_IDS" != "$OLD_IDS" ]] && echo "yes" || echo "no" )
```

Still inequality, not disjoint. A partial-propagation window where Step 5 returns 6 fresh + 4 stale (under `--limit 10`) satisfies `!=` and gets accepted. The "after" score is then computed over a mixed population. Fix unchanged — use set-intersection-empty:

```bash
IS_FRESH=$(python3 -c "
import json, sys
old = set(json.load(open('$REPORT_JSON')).get('sessions', []) and [s['session_id'] for s in json.load(open('$REPORT_JSON')).get('sessions', [])] or [])
new = set([s['session_id'] for s in json.load(open('$FRESH_REPORT')).get('sessions', [])])
print('yes' if new and new.isdisjoint(old) else 'no')
")
```

Or re-run Step 2's BQ query filter with `timestamp > step5_start_time` to scope bulk-to-fresh.

---

### New in this round — should consider before merge

#### Medium — Domain leak in `_classify_question` despite the "remove domain leaks" commit

`improver_agent.py:~530`. Commit `eb2eeaa1` is titled *"remove domain leaks in reusable module"*, but the keyword table inside `_classify_question` is still HR-policy-specific:

```python
_TOPIC_KEYWORDS = {
    "pto": ["pto", "paid time off", "vacation", ...],
    "sick_leave": ["sick", "illness", "doctor's note", ...],
    "remote_work": ["remote", "work from home", "wfh", ...],
    "expenses": ["expense", "receipt", "reimburs", ...],
    "benefits": ["benefit", "401k", "401(k)", ...],
    "holidays": ["holiday", "christmas", "thanksgiving", ...],
    "date": ["next friday", "next monday", ...],
}
```

This function lives in `agent_improvement/` — the reusable module. A second agent using this package gets wrong classifications on its domain. Two options:

- Move this table to the demo agent's own module and pass it in via `ImprovementConfig.classify_fn` or similar.
- Derive categories and keywords from the tool docstrings already exposed by `tool_introspection.format_tool_signatures` (the `tools` parameter is threaded in but unused — see next finding).

Leaving this as-is defeats the "reusable module" premise of the rewrite.

#### Low — `_classify_question(question, tools)` has unused `tools` parameter

Same function. The signature accepts `tools: list` but the body never references it — the keyword table is hardcoded. Either wire the tools through (the natural path) or drop the param. Mixed signal for the next reader.

#### Low — Module-level `_state` dict for LoopAgent tool functions

`improver_agent.py:53`:

```python
_state: dict = {}
```

The LoopAgent's tool functions (`read_quality_report`, etc.) read context from this module-global dict, which `run_improvement()` mutates on entry. Fine for the current "one cycle per process, sequential" usage; races if anyone ever runs two cycles concurrently in the same process. Add a one-line comment pinning the single-cycle-per-process assumption so a future caller doesn't accidentally break it via `asyncio.gather`.

#### Low — `VertexPromptAdapter.write_prompt` double-write has no transactional guarantee

`prompt_adapter.py:~290`. Writes to Vertex AI Prompt Registry, then mirrors to `prompts.py` via `_local_backup`. If Vertex succeeds and the local write fails (disk full, permissions), the two drift — next `read_prompt()` reads Vertex (advanced), but `git diff` shows older state. Demo-acceptable; flag the failure mode in the docstring so operators don't trust `git diff` as the authoritative history of prompt evolution when this adapter is active.

#### Low — `[improvement]` extra pulls pandas + aiplatform into `[all]`

`pyproject.toml`:

```toml
all = [
  "bigquery-agent-analytics[llm,bigframes,owl,cli,dev,improvement]",
]
```

Adding `improvement` to `all` drags `pandas>=2.0` and `google-cloud-aiplatform>=1.148.0` into any `pip install .[all]`. The improvement cycle is demo-only; users installing `[all]` for core SDK features get a heavier dependency closure. Consider keeping `improvement` outside `all` since it's demo-scoped, not SDK-core.

#### Informational — Vertex Optimizer tool-use directive re-append

`_generate_via_vertex_optimizer:210+` — the optimizer strips tool instructions from the rewritten prompt; the code re-appends the tool-use directive if `lookup_company_policy` isn't in the returned text. Comment explains. This is a known-ugly seam around the optimizer's current behavior — justified, but the substring check `"lookup_company_policy" not in improved` is itself a domain leak. If this module is reusable, the check should use a tool name from the config, not a hardcoded HR tool name.

---

### Verdict

The rework is a significant improvement on the prior version — proper package split, LoopAgent idiom, teacher-agent ground truth solving the judge-form problem, Vertex Optimizer integration, config-driven reusability. CI is green. Two prior findings (pre-flight auto-fix, Step 5 freshness) are unaddressed and worth a decision before merge. The domain leaks in `_classify_question` and the Vertex re-append check are the newest signals — both undermine the "reusable module" claim of the rewrite and are <30 lines to fix.

After those items land (or are explicitly deferred with tracking issues) and the branch rebases onto current main, this is mergeable.

caohy1988:
Latest review against current PR head `64dfbb0`, focusing only on diff-local issues in this PR, not branch freshness.

1. High: the golden-eval gate is still not a reliable correctness check, and the latest `expected_tool` tweak makes it internally inconsistent. `run_single_case()` only captures plain text from agent events, not tool-call events, yet `judge_case()` now tells the judge to fail answers that lack evidence the expected tool was called. That means correct answers from an agent that silently used the tool can still fail, while the gate still never compares against a ground-truth answer. Refs: `examples/agent_improvement_cycle/agent_improvement/eval_runner.py:72`, `:106`, `examples/agent_improvement_cycle/agent_improvement/prompts.py:17`.

2. High: the pre-flight auto-fix still silently rewrites the committed baseline before the cycle starts. If golden eval fails, `run_cycle.sh` immediately runs `run_improvement.py --from-eval-results` and then continues without preserving or surfacing the original failing version, so the later baseline score is already post-fix. Refs: `examples/agent_improvement_cycle/run_cycle.sh:218`, `:224`, `:245`.

3. Medium: the new traffic dedup logic can now return fewer cases than requested, but `run_cycle.sh` still asks `quality_report.py` for `--limit "$TRAFFIC_COUNT"`. If Gemini returns 6 deduped questions for a requested 10, Step 3/5 can pull 4 stale sessions from earlier runs to fill the limit, polluting both baseline and after scores. Refs: `examples/agent_improvement_cycle/eval/generate_traffic.py:131`, `:142`, `examples/agent_improvement_cycle/run_cycle.sh:260`, `:318`, `:443`.

4. Medium: Step 5s freshness guard still accepts mixed old/new populations. It only checks `NEW_IDS != OLD_IDS`, so any overlap with a few newly propagated sessions is enough to pass. That still allows after metrics to include stale Step 3 traces. Refs: `examples/agent_improvement_cycle/run_cycle.sh:449`.

5. Medium: the reusable improver still contains hardcoded HR-domain assumptions. `_classify_question()` ignores its `tools` parameter and hardcodes PTO/sick-leave/benefits/holiday keywords, and the optimizer re-append path still special-cases the literal tool name `lookup_company_policy`. That still leaks the demo domain into the supposedly generic module. Refs: `examples/agent_improvement_cycle/agent_improvement/improver_agent.py:373`, `:492`.

Residual low-severity items still look open too: the Vertex/local mirror double-write still lacks a drift warning in `VertexPromptAdapter`, the module-level `_state` still does not document its single-process assumption, and the Vertex location mismatch remains between `setup_vertex.py`, `run_cycle.sh`, and `show_prompt.sh`.

caohy1988:
## Fresh-eye re-review against current head (`35e01669`)

Rebased onto current `main`, `mergeStateStatus: CLEAN`, all 7 CI jobs pass. Major improvements on every Medium I raised last round. One **new High** surfaced while reading the files cold.

### Resolved since prior review

- **Judge uses captured tool-call evidence, not text inference.** `run_single_case()` now collects `tools_called` from `function_call` events, and `judge_case()` passes an objective `tool_check` block plus a conditional `tool_fail_rule` to the judge. Refs: `agent_improvement/eval_runner.py:73-117, 118-163`. Cleanly fixes the prior-round High.
- **`_classify_question` is now fully dynamic.** Extracts keywords from tool `__name__` and `__doc__`, normalizes via `_word_forms`, filters `_CLASSIFIER_STOP`, scores by overlap. No HR-domain keyword table. Refs: `improver_agent.py:603-683`. Fixes the prior Medium.
- **Vertex re-append check is tool-name agnostic.** `tool_names = [getattr(t, "__name__", "") for t in config.agent_tools]; has_tool_ref = any(name and name in improved for name in tool_names)`. Replaces the hardcoded `"lookup_company_policy" not in improved`. Fixes the prior Informational.
- **Step 5 freshness guard uses set-overlap-is-empty.** `IS_FRESH=$( [[ "$NEW_COUNT" -gt 0 && "$OVERLAP_COUNT" == "0" ]] && ... )`. Fixes the prior Medium.
- **Traffic / limit mismatch fixed.** `ACTUAL_TRAFFIC_COUNT=$(jq '.eval_cases | length' "$TRAFFIC_JSON")` + `--limit "$ACTUAL_TRAFFIC_COUNT"`; same for `ACTUAL_FRESH_COUNT` in Step 5. Fixes the prior Medium.
- **Vertex mirror drift warning.** `VertexPromptAdapter.read_prompt` logs a warning when `local_version != vertex_version`. Fixes the prior Low.
- **`_state` single-process assumption documented.** Comment above the module-level dict explicitly calls out the single-cycle-per-process constraint. Fixes the prior Low.
- **Vertex location consistent via config.** Every call site reads `config.get("vertex_location", "us-central1")`; only remaining literal is in a `prompt_adapter.py` docstring example, which is fine. Fixes the prior Low.

---

### Still open from prior reviews

#### High (unchanged) — pre-flight auto-improve still advances the committed baseline before the cycle measures it

`run_cycle.sh:218-262`: on `PREFLIGHT_EXIT != 0`, the script calls `run_improvement.py --from-eval-results ...` and continues. This round improved the visibility — failing cases are now listed with their reason via a small Python snippet — but the baseline V1 is still silently rewritten. See the new High below, which makes this effectively mandatory for every first run of the demo.

No opt-out flag. No `--allow-failing-baseline`. The advertised "30%" baseline in the DEMO_SCRIPT / README will be measured on the post-auto-fix version, not the committed V1.

---

### New in this round

#### High — committed golden set + new tool-call evidence judge guarantees pre-flight always fails V1

This is a regression the tool-call fix unintentionally introduced.

`eval/eval_cases.json` commits three cases, all tagged `expected_tool: "lookup_company_policy"`:

```json
{ "id": "pto_balance", "question": "How many PTO days...", "expected_tool": "lookup_company_policy", "notes": "V1 has this info in the prompt, should pass" },
{ "id": "sick_leave_days", ..., "expected_tool": "lookup_company_policy", "notes": "V1 has this info in the prompt, should pass" },
{ "id": "remote_work_days", ..., "expected_tool": "lookup_company_policy", "notes": "V1 has this info in the prompt, should pass" }
```

`DEMO_SCRIPT.md:122` asserts *"all 3 cases should pass with V1."*

But `prompts.py:27` (V1) says:

> You have the following knowledge about company policies: ... Answer questions using only the information above. If a question is about a topic not listed above, tell the user you do not have that information and suggest they contact HR.

V1 is designed to answer from inline knowledge, not call tools. With the new tool-call evidence judge (which is the correct fix from last round):

```python
# eval_runner.py:128-143
if expected_tool in tools_called:
    tool_check = f"Expected tool '{expected_tool}' was called. ..."
    tool_fail_rule = ""
else:
    tool_check = f"Expected tool '{expected_tool}' was NOT called. ..."
    tool_fail_rule = f"\nA response FAILS if the expected tool '{expected_tool}' was not called, as the answer may be hallucinated without tool grounding."
```

So for every first run of `./run_cycle.sh`:

- V1 answers "20 days" for PTO from its prompt → `tools_called = []`
- Judge sees: tool was NOT called → `tool_fail_rule` fires → response FAILS
- Pre-flight fails on all three committed cases
- Auto-improve runs → V1 → V2 before the cycle's baseline measurement
- Cycle 1's advertised "baseline" is actually V2, not V1
- The committed golden set's own `notes: "V1 ... should pass"` is now false

This contradicts `DEMO_SCRIPT.md:122` and silently undermines the 30%→90% demo arc. Three minimally invasive options:

1. **Preferred — drop `expected_tool` for these three committed cases** (set to `"unknown"`). They're the V1-is-allowed-to-answer-from-memory baseline; they shouldn't require tool-call evidence. The rest of the golden set will grow organically via `extract_failed_cases` with `expected_tool` inferred by the classifier.
2. Change V1 to tell the agent to call `lookup_company_policy` even for the three baseline topics. Defeats V1's intentional-flaw design.
3. Add a per-case `require_tool_call: bool` field to eval cases, defaulting true for classifier-inferred cases and false for the three committed baselines. More surface but most explicit.

(1) is the 4-line fix.

#### Low — `_classify_question` tie-breaking is tool-order-dependent

`improver_agent.py:679-683`:

```python
if score > best_score:
    best_score = score
    best_tool = name
```

Strict `>` means if two tools tie on keyword overlap, whichever appears first in `config.agent_tools` wins. For the two-tool demo (`lookup_company_policy`, `get_current_date`) the practical risk is low, but for agents with many similar tools the classifier silently becomes order-dependent. Fix: on tie, prefer lexicographic first (deterministic) or return `"unknown"` (abstain) — either is better than tool-list order.

#### Low — `_word_forms` misses `-ies` plural

`improver_agent.py:603-616`:

```python
def _word_forms(word: str) -> list[str]:
    forms = [word]
    if len(word) > 4:
        if word.endswith("ing"): forms.append(word[:-3])
        elif word.endswith("ly"): forms.append(word[:-2])
        elif word.endswith("ed"): forms.append(word[:-2])
    if len(word) > 3 and word.endswith("s"):
        forms.append(word[:-1])
    return forms
```

Handles plural `s` → stem, but misses `-ies` (e.g., `policies` → `policy`, `companies` → `company`). For a classifier that scores against tool-name words like `lookup_company_policy` and questions like "What is company policy on...?", this matters. Add:

```python
if len(word) > 4 and word.endswith("ies"):
    forms.append(word[:-3] + "y")
```

#### Low — custom `judge_prompt` without `{tool_check}` / `{tool_fail_rule}` slots will `KeyError`

`eval_runner.py:149-154`:

```python
judge_prompt = self._judge_prompt.format(
    question=case["question"],
    response=result["response"][:500],
    tool_check=tool_check,
    tool_fail_rule=tool_fail_rule,
)
```

`ImprovementConfig.judge_prompt` is documented as *"Must contain `{question}` and `{response}` placeholders"* (`config.py:~56`). But if a user supplies a custom prompt with only those two slots, `format(...tool_check=..., tool_fail_rule=...)` passes extra kwargs — which is fine for `str.format` (silently ignored). Conversely, if a user's custom prompt has `{tool_check}` / `{tool_fail_rule}` placeholders, this works. So this is only a problem if the format string has a literal `{` somewhere else — e.g. if a user writes `JSON: { "pass": true }` in the prompt body, `str.format` will raise `KeyError` on the literal `{`. Recommend either:

- Document all four placeholders in `ImprovementConfig.judge_prompt`'s docstring.
- Or use `str.Template` / explicit substitution to tolerate literal braces.

Minor — not a blocker — but someone swapping the judge prompt will hit this, and the resulting traceback doesn't point at the root cause.

---

### Verdict

The rework cleaned up **every Medium from last round** and the architectural split is now solid. CI green. Branch clean. The **one High (committed-golden-set vs tool-call-evidence judge)** is a correctness bug the new judge fix accidentally introduced against the existing committed data; a 4-line edit to `eval/eval_cases.json` resolves it. The pre-flight auto-improve remains a philosophical choice — the demo narrative is clearer with a warning-only failing pre-flight, but this round's surfaced failing-cases output at least makes the behavior visible.

After the High is addressed (and ideally the three Lows), this is a clean merge.

caohy1988:
Additional fresh-eye pass on current head `35e01669`, after reading the latest review thread. I agree with the latest comment's High around the committed golden set vs. tool-call-evidence judge. A couple of extra details still look worth fixing before merge:

### Medium — Step 3/5 still do not prove the scored BQ sessions are the sessions from this cycle

`eval/run_eval.py` writes the actual run results, including `session_id`, to `reports/latest_eval_results.json` (`eval/run_eval.py:226-230`). But `run_cycle.sh` ignores those IDs and asks `quality_report.py` for the latest N sessions by app/time window:

```bash
quality_report.py --app-name "$APP_NAME" --limit "$ACTUAL_TRAFFIC_COUNT" --time-period 24h
```

Refs: `run_cycle.sh:356-360`, `run_cycle.sh:482-486`. The underlying categorical query orders by `MAX(timestamp) DESC` and applies `LIMIT @trace_limit` (`categorical_evaluator.py:230-253`), so any previous or concurrent sessions for the same `root_agent_name` in the last 24h can enter the scored population. The new Step 5 overlap guard only proves the after report does not reuse Step 3 IDs; it still does not prove the after report equals the fresh traffic IDs.

This can skew both the baseline and after scores, especially after reruns or partial BigQuery propagation. Minimal fix: after each `run_eval.py` call, copy `reports/latest_eval_results.json` to a cycle-specific expected-results file and require:

- no traffic-mode case returned an empty `session_id`;
- `set(report.sessions[*].session_id) == set(expected_results[*].session_id)`.

Longer-term, `quality_report.py` could accept `--session-ids` and evaluate exactly the sessions produced by the current run.

### Medium — traffic-mode eval still masks per-case execution failures

`eval/run_eval.py:111-139` catches per-case timeout/exception and returns an `ERROR: ...` result with `session_id: ""`, but non-golden mode still exits successfully because `main()` sets `failed = 0` for traffic mode (`eval/run_eval.py:219-224`). Then `run_cycle.sh` continues to score whatever BigQuery returns, which may omit the failed cases entirely.

This was one of the earliest findings and is still present in the package-split version. I would fail traffic mode if any case returns an empty `session_id` or `response` starting with `ERROR:`; otherwise the demo can report high quality while some generated cases never reached BigQuery.

### Low — README link points to a nonexistent relative path

`examples/agent_improvement_cycle/README.md:25` links to `[quality_report.py](scripts/quality_report.py)`, which resolves under `examples/agent_improvement_cycle/scripts/`. It should point to `../../scripts/quality_report.py`.

### Low — whitespace check failure

`git diff --check 939be6511ce12f4baf47f7c8bb56bceff4823785...HEAD` reports:

```text
examples/agent_improvement_cycle/run_cycle.sh:580: new blank line at EOF.
```

I also ran:

```text
python3 -m compileall -q examples/agent_improvement_cycle scripts/quality_report.py
python3 scripts/quality_report.py --help
uv run pytest tests/test_quality_report_helpers.py
```

All three passed locally.

caohy1988:
## Final fresh-eye pass against current head (`f0215f56`)

CI green across all 7 jobs. Eight commits since my last review — including `126ff0b1` which addressed every prior High/Medium I flagged, plus `bce9c002`, `aa3f2b05`, and three polish commits.

### All three Highs from prior reviews are now resolved

| Prior finding | Fix commit | Status |
|---|---|---|
| Judge rewards form, not correctness (missing tool-call evidence) | earlier round | ✅ Resolved via `tools_called` capture in `eval_runner.py` |
| Committed golden set + tool-call evidence → V1 always fails pre-flight | `126ff0b1` | ✅ `expected_tool: "unknown"` on the three V1-baseline cases; notes rewritten |
| Pre-flight auto-fix silently advanced the baseline | `126ff0b1` | ✅ Now shows a prominent WARNING banner with `./reset.sh` instructions and lists failing cases with reasons |

### Prior Lows resolved

- `_word_forms` `-ies` plural — `126ff0b1`. Correctly guarded with `and not word.endswith("ies")` on the `-s` branch so `policies` doesn't collide with two stems.
- `judge_prompt` placeholders now documented in `config.py` — `126ff0b1`.
- README link to `quality_report.py` path fixed — `126ff0b1`.
- Mirror drift warning on `VertexPromptAdapter` — earlier round.
- `_state` single-process assumption documented — earlier round.
- Vertex location consistently from config — earlier round.

### New in this round — worth flagging

#### Low — soft session-verification

`run_cycle.sh:386-402, 546-562`:

```bash
if [[ "$MISSING" -gt 0 ]]; then
  echo "  Note: $MISSING expected session(s) not yet in quality report (BQ propagation delay)."
fi
```

Prints a note, continues with partial coverage. Good for demo robustness (BQ propagation occasionally lags past the 30s×6 retry budget), bad as a strict correctness gate. At `TRAFFIC_COUNT=10` the worst realistic case is 2 missing out of 10, shifting the quality score by ≤20 points. Worth being honest in the docs that "after" scores can be computed over partial population; or make it strict when `$MISSING > ceil(TRAFFIC_COUNT * 0.2)` so small gaps log and large gaps fail.

#### Low — `category` vs `expected_tool` field inconsistency in the golden set

Committed baseline cases (`eval/eval_cases.json`) have semantic categories:

```json
{"id": "pto_balance", "category": "pto", "expected_tool": "unknown"}
```

But `extract_failed_cases → _classify_question` (for cases extracted during cycle runs) returns `(best_tool, best_tool)` — so both fields become the tool name:

```json
{"id": "extracted_...", "category": "lookup_company_policy", "expected_tool": "lookup_company_policy"}
```

Paired with the new golden-set display at `run_cycle.sh:224-230`:

```
    [pto_balance] How many PTO days...? (pto)
    [extracted_how_much_is_meal_limit] How much is the meal limit? (lookup_company_policy)
```

Mixed vocabulary for committed vs extracted rows. Not a correctness bug; just reads oddly to the demo viewer. Two ways to unify:
- Drop `category` from extracted cases (or set to `"extracted"`); `expected_tool` carries the tool identity either way.
- Or keep the tool-name-as-category and retroactively rename the committed cases' categories to their tool names (`"lookup_company_policy"`).

#### Low — `_classify_question` still tool-order-dependent on ties

`improver_agent.py:~680`:

```python
if score > best_score:
    best_score = score
    best_tool = name
```

Strict `>` not `>=`. Still returns whichever tool appears first in `config.agent_tools` when two tools score equal. Same finding as my prior fresh-eye pass, unchanged in `126ff0b1`. For the two-tool demo it never bites; for agents with many similar tools it silently depends on tool-list order. Deterministic tiebreak (lexicographic, or abstain on tie) would be <5 lines.

### Pre-existing concern, unaddressed — brittle Vertex private imports

`setup_vertex.py` and `improver_agent.py` import from `vertexai._genai.types.common`:

```python
from vertexai._genai.types.common import CreatePromptConfig
from vertexai._genai.types.common import DeletePromptConfig
from vertexai._genai.types.common import OptimizeConfig
from vertexai._genai.types.common import OptimizeTarget
from vertexai._genai.types.common import Prompt
from vertexai._genai.types.common import PromptData
```

The `_genai` underscore prefix marks this as a private module that Vertex AI can restructure without a semver bump. `bce9c002`'s move from dict-based `config={"timeout": 300}` to typed `DeletePromptConfig(timeout=300)` is nicer ergonomically but tightens the coupling to the private API. The `[improvement]` extra pins `google-cloud-aiplatform>=1.148.0` with no upper bound, so a minor-version bump that moves or renames these types will break the demo silently.

Mitigations (pick one):
- Pin an upper bound in `pyproject.toml`: `google-cloud-aiplatform>=1.148.0,<2.0.0`.
- Or wrap the imports in a `try/except ImportError` with a clear error message naming the version compatibility requirement.
- Or accept the risk — the demo gets updated on each Vertex AI release.

For a feature-complete demo this is acceptable; worth a comment in `setup_vertex.py`'s header so the next maintainer knows why imports resolve through `_genai`.

### What I verified end-to-end

- `run_cycle.sh` full flow: banner → show prompt → show golden set → pre-flight → cycles. The new "display golden set" step (`aa3f2b05`) gives the demo viewer immediate visibility into what's being tested.
- Pre-flight warning banner + failing-case printout gives operators enough context to decide whether to `Ctrl+C` and run `./reset.sh` vs. let the auto-fix proceed.
- Traffic-mode eval now fails on `session_id == ""` or `response.startswith("ERROR:")` (`run_eval.py:224-231`), so BQ-invisible sessions can't pollute downstream scoring.
- Session-ID cross-check at Step 3/5 confirms the quality report covers the traffic we actually ran.
- Step 5 freshness guard uses `OVERLAP_COUNT == 0` (set-disjoint, not inequality) — prior Medium resolved.
- Teacher-agent ground truth path in `_generate_via_vertex_optimizer` is unchanged and still grounds improvements in real tool output.

### Verdict

**Mergeable.** All three Highs from prior reviews and every Medium are resolved. The three new items flagged above are Low-severity cosmetics (session-verification soft check, category vocabulary mix, tool-order tie-breaking) and the brittle-Vertex-private-imports concern pre-exists this PR's scope. If Eva wants, the three Lows are ~30 lines of follow-up; if not, this ships as a demo with a clean end-to-end story.

Nothing blocking left in the diff.

caohy1988:
Blocking issue still present on latest PR head `f0215f56`:

High: `examples/agent_improvement_cycle/run_cycle.sh` does not parse under Bash, so the main demo entrypoint cannot start. Repro from the fetched PR head:

```text
git show pr61-latest-check:examples/agent_improvement_cycle/run_cycle.sh > /tmp/pr61_run_cycle_latest.sh
bash -n /tmp/pr61_run_cycle_latest.sh
/tmp/pr61_run_cycle_latest.sh: line 52: syntax error near unexpected token `;;'
/tmp/pr61_run_cycle_latest.sh: line 52: `    *AuthlibDeprecationWarning*|*"authlib.jose module is deprecated"*|*"from authlib.jose import"*|*"compatible before version"*) ;;'
```

The failing block is the stderr-filter `case` statement inside process substitution at `run_cycle.sh:50-55`. Suggested fix: replace the `case` pattern with an `if [[ "$line" == *...* || ... ]]` filter, or remove the stderr filter entirely.

I would not merge until `bash -n examples/agent_improvement_cycle/run_cycle.sh` and `./examples/agent_improvement_cycle/run_cycle.sh --help` both succeed.

caohy1988:
## Follow-up: High — `run_cycle.sh` fails `bash -n` on macOS default bash

Promoting this to the top of the open-findings list because it blocks the main demo entrypoint on a significant fraction of developer machines and it was **not** surfaced by any prior review including my own.

### Reproduction

Against current PR head (`f0215f56`), on macOS default `/bin/bash`:

```
$ bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin25)

$ bash -n examples/agent_improvement_cycle/run_cycle.sh
run_cycle.sh: line 52: syntax error near unexpected token `;;'
run_cycle.sh: line 52: `    *AuthlibDeprecationWarning*|*"authlib.jose module is deprecated"*|*"from authlib.jose import"*|*"compatible before version"*) ;;'
```

Minimal reproducer isolates the issue — the `case` statement alone parses fine, but **only breaks when wrapped in the `>(...)` process substitution**:

```bash
# parses OK on bash 3.2
case "$line" in
  *AuthlibDeprecationWarning*|*"authlib.jose module is deprecated"*) ;;
  *) ... ;;
esac

# same block inside process substitution — fails on bash 3.2, parses on bash 4+
exec 2> >(while IFS= read -r line; do
  case "$line" in
    *AuthlibDeprecationWarning*|*"authlib.jose module is deprecated"*) ;;
    ...
```

Root cause: a known bash 3.2 parser limitation with mixed-pattern alternation (unquoted glob + quoted string) inside process substitutions. Fixed in bash 4.0+ (2009).

### Why CI didn't catch it

GitHub Actions jobs run Ubuntu (bash 5.x) and exercise the Python test suite, not `run_cycle.sh`. The shell script has never been shell-parsed in CI.

### Where it bites

| Environment | Bash | Status |
|---|---|---|
| macOS default `/bin/bash` | 3.2.57 | **Fails** |
| Google Cloud Shell | 5.x | OK |
| Ubuntu / Debian / most Linux | 5.x | OK |
| macOS + Homebrew bash | 5.x | OK (but not `PATH`-default) |

The shebang `#!/usr/bin/env bash` resolves to whatever's first in `PATH`, which on stock macOS is `/bin/bash` (3.2.57 — Apple has not shipped a newer bash since GPLv3). Any Mac developer cloning the repo and running `./run_cycle.sh` hits this before Step 1. Cloud Shell users (the target environment per `DEMO_SCRIPT.md`) don't.

### Severity

Demo targets Cloud Shell per the script, but the `run_cycle.sh` entrypoint is also the natural thing a developer runs locally after cloning. On that path macOS is a primary platform, and the demo breaks with a cryptic parser error before any setup check runs. Calling this **High** — not because the CI environment is broken, but because the out-of-box path on a common dev platform is.

### Suggested fix

The `case` block is only doing stderr filtering; two minimal-diff options:

**Option A — `if/elif` chain (preserves structure):**

```bash
exec 2> >(while IFS= read -r line; do
  if [[ "$line" == *AuthlibDeprecationWarning* ]] || \
     [[ "$line" == *"authlib.jose module is deprecated"* ]] || \
     [[ "$line" == *"from authlib.jose import"* ]] || \
     [[ "$line" == *"compatible before version"* ]]; then
    continue
  fi
  printf '%s\n' "$line" >&2
done)
```

**Option B — `grep -v` (shortest):**

```bash
exec 2> >(grep -v -E 'AuthlibDeprecationWarning|authlib\.jose module is deprecated|from authlib\.jose import|compatible before version' >&2)
```

Both parse on bash 3.2+ and preserve the filtering intent. I'd lean toward Option A since the existing code already speaks in line-processing idioms.

### Belt-and-braces recommendation

Add a CI job that runs `bash -n examples/agent_improvement_cycle/*.sh` so the next syntax regression is caught before merge. One line in the workflow, costs nothing.

### Verdict

Worth an add-on commit before merge, since the fix is <10 lines and turns a "works for most people" demo into a "works for everyone." Not a blocker against the substantive content of the PR — which as I noted in the prior review is materially complete — but a blocker against the "working demo" framing in the README.

caohy1988:
## Final pass against current head (`050036c4`)

Single commit since last round (`050036c4` — "Remove bash 3.2-incompatible stderr filter, fix classifier tie-breaking") — addresses both my latest findings in 10 lines.

### Both prior findings resolved

**High — bash 3.2 parse failure.** Fixed by deleting the entire stderr-filter block (lines 47-55 in the prior version). Verified locally:

```
$ bash --version
GNU bash, version 3.2.57(1)-release
$ bash -n examples/agent_improvement_cycle/run_cycle.sh
$ echo $?
0
```

Trade-off: authlib warnings previously suppressed by that block will now surface when a newer authlib lives in `~/.local/lib` and bypasses the Python-level filters (`PYTHONWARNINGS="ignore"`, `-W ignore`). Acceptable — the common path is already covered by the Python-level filters. The specific "newer authlib in `~/.local/lib`" edge case is narrow enough that portability wins.

If cosmetic cleanup matters, my earlier suggestion (Option A if/elif or Option B `grep -v >&2`) would restore filtering without the 3.2 parser bug. Not requesting — the "just delete it" call is defensible for a demo.

**Low — classifier tie-breaking.** Lexicographic tiebreak added:

```python
if score > best_score or (score == best_score and name < (best_tool or "")):
```

Traced through the edge cases:
- First tool with score > 0 wins by `>`.
- Tied tool replaces only if its name sorts earlier — deterministic regardless of `config.agent_tools` order.
- `or ""` defensive guard for `best_tool is None`; harmless.
- Zero-score tools never replace a non-zero-score winner (0 == 0 tiebreak would need `best_tool is None` which means no prior tool scored non-zero either → doesn't apply).

Correct. Turns the prior order-dependent behavior into a stable one.

### Outstanding items (non-blocking)

- `mergeStateStatus: BEHIND` — needs a rebase onto current `main` before merge. Routine.
- Soft session-verification (prior Low, unchanged) — `MISSING > 0` prints a note, doesn't fail. Acceptable at demo scale.
- `category` vs `expected_tool` vocabulary mix in the golden set between committed and extracted cases (prior Low, unchanged). Cosmetic.
- `vertexai._genai.types.common` private-module imports in `setup_vertex.py` and `improver_agent.py` with no upper bound on `google-cloud-aiplatform` (pre-existing concern). A Vertex AI minor-version bump could break silently.

### Belt-and-braces recommendation — still open

CI job running `bash -n examples/agent_improvement_cycle/*.sh` on every PR would have caught the bash 3.2 issue before merge (CI bash 5.x still catches bash-5.x-invalid syntax; wouldn't have caught the 3.2-specific issue, but would catch regressions against its own version). Cheap insurance; one line in the workflow. Not a blocker.

### Verdict

**Mergeable after rebase.** All Highs and Mediums across the full review history are resolved. The three remaining Lows are cosmetic / edge-case and can be follow-ups or deferred. CI all green (Build package pending, other 8 pass). Demo now parses and runs on macOS default bash 3.2.57.

The PR is in good shape to land.

Issue #57, opened 2026-04-18
Feat: SKOS import support alongside OWL (design proposal — feedback wanted)


libei:
> **Status**: Design proposal. Not yet implemented. Comments welcome — especially on the "Open questions" section at the bottom.

## Goal

Many real-world ontologies mix OWL (formal classes and properties) with SKOS (human-readable labels, taxonomies, cross-vocabulary mappings). The current `gm import-owl` importer recognizes SKOS only partially: `skos:altLabel` and `skos:prefLabel` populate `synonyms`, and everything else (`skos:definition`, `skos:broader`, `skos:scopeNote`, `skos:notation`, `skos:exactMatch`, standalone `skos:Concept` resources, etc.) is silently dropped.

This issue proposes first-class SKOS support in the same `gm import-owl` command, so users importing FIBO, SNOMED excerpts, library vocabularies, and other hybrid OWL+SKOS sources don't lose descriptive metadata or taxonomy structure.

## Audit of current behavior

On `main` (`src/bigquery_ontology/owl_importer.py`):

- `skos:prefLabel`, `skos:altLabel` → `synonyms` (lines 213–218)
- All other SKOS predicates → **silently dropped** (no code path reads them)
- Standalone `skos:Concept` (no `owl:Class` typing) → **not imported** (entity iteration is `g.subjects(RDF.type, OWL.Class)` only, line 350)
- Design doc `docs/ontology/owl-import.md:168-170` *claims* literal annotations are preserved as `annotations: {iri: value}`, but the implementation only covers a fixed OWL allowlist — SKOS literals fall through unhandled
- No tests or fixtures cover SKOS input

## Guiding principle: informational by default

SKOS and OWL make different kinds of claims, and the importer should preserve the distinction:

- **OWL** makes *formal* claims. `rdfs:subClassOf` means every instance of the child is an instance of the parent. Drives inheritance, keys, substitutability.
- **SKOS** makes *informational* claims. `skos:broader` means "this is a narrower topic" — explicitly *not* subsumption per [the W3C SKOS Primer](https://www.w3.org/TR/skos-primer/). Drives documentation, browsing, search.

Consequences for the importer:

1. OWL constructs flow into structural fields (`extends`, `from`/`to`, `keys`, `properties`).
[...]

PR #62, opened 2026-04-21
feat(owl-import): SKOS support alongside OWL

libei:
## Summary

Implements the ontology-layer portion of the SKOS import design in #57. Hybrid OWL+SKOS sources (FIBO, SNOMED excerpts, library vocabularies) no longer silently drop descriptive metadata and taxonomy structure.

- `Entity` and `Relationship` gain `abstract: bool = False`. Abstract elements are declared but not bound to a BigQuery table; keys are optional on abstract entities; bindings that target abstract elements are rejected.
- `skos:Concept` (without `owl:Class`) → abstract entity with `skos_` name prefix. Mixed `owl:Class` + `skos:Concept` → concrete entity unprefixed (OWL wins structure).
- `skos:broader`/`narrower` → abstract `skos_broader` (narrower normalized). `skos:related` → abstract `skos_related`. `skos:exactMatch` etc. → abstract relationship if target imported, annotation with full IRI if external.
- `skos:definition`, `notation`, `scopeNote`, etc. → annotations with `skos:` prefix.
- New `--language` CLI flag (default `en`) selects labels by BCP-47 tag; non-selected languages become language-suffixed annotations.
- Generic literal-annotation pass closes the silent-drop gap for Dublin Core and other non-OWL/SKOS literal predicates (previously docs claimed this but code only handled a fixed OWL allowlist).

## Scope

This PR covers the **ontology package only** (`src/bigquery_ontology/`). The SDK runtime choke-point filter (`bigquery_agent_analytics.runtime_spec`) called out in the issue's reviewer comment is **not** included — it belongs in a follow-up PR.

## Validation rules added

- Concrete relationships must have concrete endpoints (cannot bind a rel to an abstract entity).
- Abstract relationships use relaxed uniqueness: `(name, from, to)` instead of `(name,)`.
- Abstract relationships forbid `extends` (the mixed concrete/abstract parent-map has ambiguous uniqueness; informational-only elements don't need inheritance).
- Abstract entities/relationships with declared keys still follow shape rules.
- Endpoint existence still enforced on abstract relationships.

## Test plan

- [x] `pytest tests/bigquery_ontology/` — 235 passed (199 existing + 36 new)
- [x] `pytest tests/` — 1952 passed, 4 skipped (full repo, no regressions)
- [x] Pure SKOS taxonomy fixture imports cleanly: `tests/fixtures/skos_taxonomy.ttl`
- [x] Mixed OWL+SKOS fixture imports cleanly: `tests/fixtures/mixed_owl_skos.ttl`
- [x] Regression: OWL ObjectProperty with range on pure SKOS concept resolves to `skos_<name>` endpoint via shared IRI→name map
- [x] Multilingual label selection with `--language fr`
- [x] Emitted YAML round-trips through `load_ontology_from_string`

## Follow-ups (not in this PR)

- SDK runtime adapter filter (`runtime_spec.graph_spec_from_ontology_binding`) to exclude abstract elements before building name-indexed maps (issue's reviewer comment, Step 2 in their suggested sequence).
- `skos:ConceptScheme` as ontology-level annotation (noted in issue mapping table, deferred here).
- Opt-in promotion of `skos:broader` → `extends` (issue's open question 9; conservative default is no promotion).

### DIFF
--- a/docs/ontology/binding.md
+++ b/docs/ontology/binding.md
@@ -221,9 +221,10 @@ correctly).
 2. The ontology named by `ontology:` exists and loads without errors.
 3. `target.backend` is supported; backend-specific required fields are
    present.
-4. Every `EntityBinding.name` names an entity declared in the ontology.
+4. Every `EntityBinding.name` names an entity declared in the ontology,
+   and the entity must not be abstract (see `ontology.md` §3a).
 5. Every `RelationshipBinding.name` names a relationship declared in the
-   ontology.
+   ontology, and the relationship must not be abstract.
 6. No duplicate entity or relationship names within the binding.
 7. For every bound entity, every non-derived property (including
    inherited) has exactly one `PropertyBinding` with a non-empty `column`.
--- a/docs/ontology/cli.md
+++ b/docs/ontology/cli.md
@@ -166,10 +166,15 @@ Read OWL sources and emit a `*.ontology.yaml` (see `owl-import.md`).
 
 ```
 gm import-owl <source>... --include-namespace <iri>... [-o <out>]
+              [--format ttl|rdfxml] [--language <tag>] [--json]
 ```
 
 - One or more OWL source files (Turtle, RDF/XML).
 - At least one `--include-namespace` required; multiple allowed.
+- Recognizes both OWL and SKOS constructs; SKOS concepts become
+  abstract entities, SKOS graph predicates become abstract
+  relationships, and SKOS literals become annotations. See
+  `owl-import.md` §19 for the full mapping.
 - Output uses `FILL_IN` for ambiguities and annotations for dropped OWL
   features (see `owl-import.md` §11, §13). `FILL_IN` causes
   `gm validate` to fail until resolved.
@@ -181,9 +186,14 @@ gm import-owl <source>... --include-namespace <iri>... [-o <out>]
 | `--include-namespace <iri>` | Required, repeatable. |
 | `-o <file>`, `--out <file>` | Write YAML to file instead of stdout. |
 | `--format ttl\|rdfxml` | Override parser selection from file extension. |
+| `--language <tag>` | BCP-47 language tag for label selection (default `en`). Other-language labels become language-suffixed annotations. |
 | `--json` | Structured drop-and-placeholder summary. |
 
-Drop summary is printed to stderr regardless of `--json`.
+Drop summary is printed to stderr regardless of `--json`. It includes
+SKOS counts (concepts imported as abstract, relationships imported as
+abstract, annotations preserved, labels discarded by language
+selection, external match targets, generic literal annotations) and a
+hint when every imported entity is abstract.
 
 ## 7. Open questions
 
--- a/docs/ontology/compilation.md
+++ b/docs/ontology/compilation.md
@@ -276,6 +276,11 @@ On top of ontology-level (`ontology.md` §10) and binding-level
 3. No cycles among derived properties.
 4. Every logical property type is supported by the target backend
    (`ontology.md` §7).
+5. **No abstract elements in bindings.** The binding loader already
+   rejects bindings that target abstract entities or relationships
+   (`ontology.md` §3a). The compiler guards against this as
+   defense-in-depth so a hand-constructed `Binding` object that skips
+   the loader still raises rather than emitting unresolvable DDL.
 
 Warnings: bound entity referenced by no relationship.
 
--- a/docs/ontology/ontology.md
+++ b/docs/ontology/ontology.md
@@ -173,27 +173,66 @@ additional: [<string>, ...]             # relationships only, mode 2: extends (f
 | Field | Type | Required | Notes |
 |---|---|---|---|
 | `name` | string | yes | Unique within the ontology. |
+| `abstract` | boolean | no | Defaults to `false`. See §3a. |
 | `extends` | string | no | Name of the parent entity. Single-parent. |
-| `keys` | Keys | yes | See §6. |
+| `keys` | Keys | yes on concrete; optional on abstract | See §6. |
 | `properties` | list\<Property\> | no | See §5. |
 | `description` | string | no | Free-form. |
 | `synonyms` | list\<string\> | no | Alternate names for search / LLM grounding. |
 | `annotations` | map\<string, string \| list\<string\>\> | no | Free-form metadata. Values may be a string or a list of strings. |
 
+## 3a. Abstract elements
+
+An element with `abstract: true` is declared in the ontology but is
+**not bound to a physical table**. Abstract elements are informational:
+documentation, taxonomy structure, or cross-vocabulary mappings that
+don't back any row data. The field exists primarily to let the OWL
+importer represent standalone SKOS concepts (see `owl-import.md` §19),
+but it is part of the core ontology model and may be used by
+hand-authored ontologies as well.
+
+Rules:
+
+- **Abstract entities** do not require `keys.primary`. Declared keys
+  still follow the shape rules in §6.
+- **Abstract relationships** must not use `extends`; see §4. Declared
+  keys still follow the shape rules in §6.
+- A **concrete relationship** cannot have an **abstract endpoint**:
+  a relationship that will be materialized needs bindable endpoints
+  on both sides.
+- An **abstract relationship** may point at concrete entities,
+  abstract entities, or any combination.
+- A concrete relationship and an abstract relationship cannot share a
+  name.
+- Bindings that target an abstract entity or abstract relationship are
+  rejected by the binding loader. `gm scaffold` skips abstract
+  elements entirely when emitting DDL and binding stubs.
+
 ## 4. Relationship
 
 | Field | Type | Required | Notes |
 |---|---|---|---|
-| `name` | string | yes | Unique within the ontology. |
-| `extends` | string | no | Name of the parent relationship. Single-parent. |
+| `name` | string | yes | See §4a for uniqueness rules. |
+| `abstract` | boolean | no | Defaults to `false`. See §3a. |
+| `extends` | string | no | Name of the parent relationship. Single-parent. Forbidden on abstract relationships. |
 | `keys` | Keys | no | See §6. |
-| `from` | string | yes | Source entity name. |
-| `to` | string | yes | Target entity name. |
+| `from` | string | yes | Source entity name. Concrete relationships require a concrete source. |
+| `to` | string | yes | Target entity name. Concrete relationships require a concrete target. |
 | `cardinality` | enum | no | `one_to_one` \| `one_to_many` \| `many_to_one` \| `many_to_many`. |
 | `properties` | list\<Property\> | no | See §5. |
 | `description` | string | no | |
 | `synonyms` | list\<string\> | no | |
 | `annotations` | map\<string, string \| list\<string\>\> | no | |
+
+### 4a. Relationship name uniqueness
+
+- **Concrete relationships** require unique `(name,)` within the
+  ontology (same rule as entities).
+- **Abstract relationships** require unique `(name, from, to)`. Two
+  abstract relationships with the same name but different endpoints
+  are legal — this is what lets an imported taxonomy emit multiple
+  `skos_broader` edges without synthetic naming.
+- Concrete and abstract relationships cannot share a name.
 
 ## 5. Property
 
@@ -339,10 +378,14 @@ child `alumni: Person → School` is valid iff `Person extends Party` and
 
 Enforced by the ontology loader before any downstream stage runs.
 
-1. Every entity and relationship name is unique within the ontology.
+1. Entity names are unique within the ontology. Concrete relationship
+   names are unique. Abstract relationships are unique by
+   `(name, from, to)`. A concrete and abstract relationship cannot
+   share a name. Entities and relationships cannot share a name.
 2. Every property name is unique within its entity or relationship.
 3. `extends` must resolve to a declared entity (entity-to-entity) or
    relationship (relationship-to-relationship) of the correct kind.
+   Abstract relationships must not use `extends`.
 4. No cycles in `extends` chains.
 5. Redeclaring an inherited property (by name) is an error.
 6. Redeclaring inherited keys is an error.
@@ -354,9 +397,13 @@ Enforced by the ontology loader before any downstream stage runs.
    On entities, `additional` is not allowed.
 10. Relationships with `extends`: `from` / `to` satisfy covariant narrowing
     against the parent's endpoints.
-11. Every entity must declare `keys.primary`.
+11. Every **concrete** entity must declare `keys.primary`. Abstract
+    entities may omit it.
 12. Property `type` is one of the valid types (§7).
 13. Unknown YAML keys anywhere in the spec are an error (`extra="forbid"`).
+14. Relationship endpoints must reference declared entities (concrete
+    or abstract). A concrete relationship cannot point at an abstract
+    entity.
 
 ## 11. Out of scope / future docs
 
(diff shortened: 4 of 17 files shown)


caohy1988 (commented):
Code-read review against refreshed `origin/main..origin/feat/skos-import-support`.

## High

1. `src/bigquery_ontology/owl_importer.py:588`, `src/bigquery_ontology/ontology_loader.py:424`

The new IRI→name resolution lets an OWL `ObjectProperty` point at a pure SKOS concept, but the emitted relationship stays concrete while the endpoint entity is abstract. That produces YAML the loader rejects.

I verified this in the PR worktree with a minimal fixture:

- importer emitted `belongsTo` with `from: Account`, `to: skos_Category`
- then `load_ontology_from_string(...)` failed with:
  `Relationship 'belongsTo' is concrete but its 'to' endpoint 'skos_Category' is abstract.`

The PR summary calls this case out as a supported regression fix, but in its current form it generates an ontology that cannot round-trip.

2. `src/bigquery_ontology/owl_importer.py:277`, `src/bigquery_ontology/owl_importer.py:1288`

OWL inheritance still resolves parents by raw local name before SKOS-only concepts are renamed to `skos_<name>`. So `owl:Class Account rdfs:subClassOf :Category` where `:Category` is a pure `skos:Concept` emits `extends: Category`, while the imported entity is `skos_Category`.

I verified that this YAML immediately fails validation with:
`Entity 'Account' extends 'Category', which is not a declared entity.`

This is the same cross-kind naming problem as above, just on the `extends` path instead of relationship endpoints.

## Medium

3. `src/bigquery_ontology/owl_importer.py:739`

`skos:inScheme` / `skos:topConceptOf` annotations currently store `_local_name(obj)` instead of preserving enough identity for later runtime use. Two distinct scheme IRIs like `http://ns1.example/Scheme` and `http://ns2.example/Scheme` both import as:

```yaml
annotations:
  skos:inScheme: [Scheme, Scheme]
```

That is already lossy in this PR, and it directly undercuts issue #58's planned `in_scheme()` / concept-index work, because the runtime layer can no longer tell which scheme membership was asserted.

4. `src/bigquery_ontology/owl_importer.py:195`, `src/bigquery_ontology/owl_importer.py:243`

Non-selected-language labels are accumulated in a `dict[str, str]`, so multiple values for the same `(predicate, language)` overwrite each other. I verified a pure SKOS concept with two French `skos:altLabel`s and one French `skos:hiddenLabel`; only the last alt label survived in the emitted annotations.

Since issue #57's goal is to stop silently dropping SKOS metadata, this is still silent data loss.

I did not run the full repo test suite. The findings above were verified with targeted importer/loader probes in an isolated worktree on the PR head.

caohy1988 (commented):
Re-reviewed at `a5eaba1` against `origin/main`, `#57`, and `#58`. Five new findings not captured in my earlier pass. My previous 4 highs/mediums (concrete-relationship-to-abstract-endpoint, pre-rename `extends` path, `skos:inScheme` local-name collapse, non-selected-language label dict overwrite) remain open.

## Medium

### 1. Generic literal-annotation pass strips the namespace prefix

`src/bigquery_ontology/owl_importer.py:948`

```python
ann_key = _local_name(pred)
collected.setdefault(ann_key, []).append(str(obj))
```

The key for any non-OWL/RDF/RDFS/SKOS literal predicate is its bare local name. That has two problems:

- **Cross-namespace collision.** A resource carrying both `dc:title "Foo"` and `dcterms:title "Bar"` merges under a single `title` annotation as `["Bar", "Foo"]`, with source-vocabulary provenance lost. Round-trip back to RDF cannot reconstruct which was `dc:` and which was `dcterms:`.
- **Inconsistent convention across passes.** SKOS literals deliberately retain the `skos:` prefix (`skos:definition`, `skos:notation`, etc. — `owl_importer.py:671-680`). OWL literals retain their `owl:` prefix (line 357 region). The generic pass is the odd one out, stripping prefixes entirely. Issue #58's planned `rt.annotation(name, key)` accessor expects callers to look up `\"dc:title\"` — under the current code they'd have to look up `\"title\"` instead, which collides with any other vocabulary's local `title`.

Consider emitting `<prefix>:<local>` as the annotation key (derived from rdflib's `g.namespace_manager.normalizeUri(pred)` or a prefix map). If a prefix can't be resolved, fall back to the full IRI rather than the bare local name.

### 2. Binding rejection of abstract elements gives no user guidance

`src/bigquery_ontology/binding_loader.py:230, 240`

```
Entity binding 'Foo' targets an abstract entity, which cannot be bound to a
physical table.
```

For a user who encounters \"abstract\" for the first time — likely, given this PR introduces the concept — the message doesn't explain *why* this is abstract (imported as SKOS-only?), *how* they might bind it (promote the source to `owl:Class`? add a concrete relationship?), or where to read more. A one-line pointer to the relevant doc section would materially soften the rough edge:

```
Entity binding 'Foo' targets an abstract entity (declared for documentation
but not backed by a table). Abstract entities arise from SKOS-only concepts
or from ontologies that declare structure without physical realization. See
docs/ontology/binding.md#abstract-elements. Remove this binding, or promote
'Foo' in the ontology if you need to back it with a table.
```

## Low

### 3. Abstract-element defense-in-depth in the DDL compiler is unreachable

`src/bigquery_ontology/graph_ddl_compiler.py:108-119`

```python
# Defense-in-depth: abstract elements must never reach the compiler
# (the binding loader rejects them).
for eb in binding.entities:
  if entity_map[eb.name].abstract:
    raise ValueError(f\"Internal error: abstract entity {eb.name!r} in binding.\")
```

`binding_loader._validate_binding` already rejects abstract targets before any compile path touches them, so this branch is unreachable in normal use and untested. Either:

- Document it as an intentional defensive backstop (and add a unit test that constructs a malformed state bypassing the loader, to lock it down), or
- Delete it and rely on the loader as the single gate.

As-is, the \"Internal error\" framing suggests a genuine invariant that no caller can actually trigger from outside — confusing to future maintainers.

### 4. No test asserts `abstract: false` default round-trips cleanly

`tests/bigquery_ontology/test_owl_importer.py`

The `abstract: bool = False` default on `Entity`/`Relationship` is a public-API change. The test file has 36 new abstract-related tests, but none verifies:

- A concrete entity emitted to YAML has no `abstract: false` line (verifies emitter suppresses the default to keep YAML terse).
- YAML that omits `abstract` parses back to `abstract = False` (verifies loader default propagation).

Without these, a future emitter change that accidentally writes `abstract: false` on every concrete entity would pass CI while bloating every real-world ontology file. Suggest two small round-trip tests covering both directions.

### 5. No test exercises `skos:inScheme` or `skos:topConceptOf` annotation extraction

`tests/bigquery_ontology/test_owl_importer.py`

Matching the earlier finding on `_extract_skos_annotations` storing `_local_name(obj)` — the reason that bug slipped through is there's no test that imports a fixture with `skos:inScheme` and asserts the resulting annotation shape. Adding even one test (and one where two schemes share a local name) would turn that silent data loss into a visible CI failure, and would serve as the check-in contract for #58's `rt.in_scheme()` once that surface lands.

---

## Cross-reference notes against #57 and #58

- **#57 mapping table `skos:ConceptScheme` → ontology-level annotation** — acknowledged as deferred in the PR body; no code path today. OK.
- **#57 open question 9 (opt-in `skos:broader` → `extends` promotion)** — correctly left off; the conservative default holds.
- **#58 identity rules: \"relationships are traversal-first, not name-addressed\"** — `ontology_loader.py:80-93` enforces exactly `(name, from, to)` for abstract relationships while keeping concrete relationships globally unique by name, matching #58's requirement. ✅
- **#58 `rt.in_scheme()`** — blocked by the `_local_name(obj)` scheme collapse (my earlier finding #3). Fixing scheme-annotation IRI preservation is a prerequisite for that runtime accessor to be implementable.
- **#58 `rt.annotation(name, key)`** — blocked by (1) above for non-OWL/non-SKOS literal predicates. Callers can't look up by the conventional `dc:title` if the stored key is `title`.

---

No blocking items in this pass; (1) and (2) would measurably improve the surface before shipping, (3)-(5) are cleanup / coverage.

caohy1988 (commented):
Re-reviewed at `e34fb72`. All nine prior findings are correctly addressed; verified with targeted diff reads:

- IRI→name precompute (`_precompute_iri_to_name`, `owl_importer.py:439-460`) resolves both relationship endpoints and `extends` in one pass ✅
- OWL ObjectProperty whose endpoint is a pure SKOS concept now gets auto-promoted to `abstract: true` (`owl_importer.py:1445-1451`), so the emitted YAML round-trips ✅
- `skos:inScheme` / `skos:topConceptOf` preserve full IRI (`owl_importer.py:822-833`) ✅
- Non-selected-language labels accumulate as `dict[str, list[str]]` (`owl_importer.py:204, 222, 244`) ✅
- Generic annotations keyed via `_qname_or_iri` (`owl_importer.py:110-122`) with full-IRI fallback ✅
- Binding-rejection messages expanded (`binding_loader.py:231-252`) ✅
- Unreachable defense-in-depth check removed from `graph_ddl_compiler.py` ✅
- `test_concrete_entity_no_abstract_key_emitted` + `test_omitted_abstract_key_loads_as_false` lock the default round-trip ✅
- `test_in_scheme_preserves_full_iri` + `test_top_concept_of_preserves_full_iri` lock IRI identity ✅

Four new findings in this pass, all medium-or-low — not blockers, but worth tightening before merge.

---

## Medium

### 1. OWL → abstract promotion is silent — no log line, no `_DropSummary` counter

`src/bigquery_ontology/owl_importer.py:1445-1451`

```python
for rel in owl_relationships.values():
  if rel.abstract:
    continue
  from_abstract = rel.from_entity in entities and entities[rel.from_entity].abstract
  to_abstract = rel.to_entity in entities and entities[rel.to_entity].abstract
  if from_abstract or to_abstract:
    rel.abstract = True
```

Correct behavior, necessary for round-trip. But there's no diagnostic signal anywhere:

- `_DropSummary` (line 180) has counters for SKOS concepts imported as abstract, SKOS relationships imported as abstract, and SKOS external matches — but no counter for "OWL relationships promoted to abstract because an endpoint was abstract."
- No `logger.info` / `logger.warning` at the promotion site.
- The drop summary rendered by `_format_drop_summary` therefore hides this transformation entirely.

The downstream failure mode is the sharp one: a user authored `:belongsTo a owl:ObjectProperty` and wrote a binding for `belongsTo`. At bind time they hit the excellent new error message \"...targets an abstract relationship, which cannot be bound...\" — and now they have to figure out, unaided, that the cause is `rdfs:range :Category` where `:Category` happens to be a pure `skos:Concept`. They never wrote `abstract: true` anywhere.

Suggest:
1. Add a field `owl_relationships_promoted_to_abstract: int` on `_DropSummary`.
2. Increment it inside the loop.
3. Surface it in `_format_drop_summary` alongside the existing counters, with a one-line explanation of the cause (\"OWL ObjectProperty with a SKOS-only endpoint — see docs/ontology/owl-import.md#abstract-promotion\").

### 2. Docs don't describe the implicit promotion rule

`docs/ontology/owl-import.md` and `docs/ontology/ontology.md`

The docs now cover abstract entities, abstract relationships, SKOS naming conventions, `gm validate` / `gm scaffold` / `gm bind` behavior around abstract elements. What's missing is the single rule that creates surprise: **\"An OWL ObjectProperty whose `rdfs:domain` or `rdfs:range` resolves to an abstract (SKOS-only) entity is imported as an abstract relationship, keeping its unprefixed OWL name.\"**

A two-sentence addition to §19 of `owl-import.md` would close the loop for readers who see `abstract: true` in their emitted YAML on a relationship they didn't mark abstract in RDF.

---

## Low

### 3. Generic-annotation prefix depends on namespace-manager lookup order

`src/bigquery_ontology/owl_importer.py:110-122`

`_qname_or_iri` uses `g.namespace_manager.compute_qname(iri, generate=False)`. When multiple source files each `@prefix` the same namespace under a different alias (rare but legal), rdflib's namespace manager resolves to whichever alias was registered first, which depends on file-parse order. The resulting annotation key — `dc:title` vs `dcterms:title` vs raw IRI fallback — can therefore vary across runs or across invocations with different input file order.

Realistically rare, but the fix is small: pin a deterministic prefix registry (pass `namespace_manager` an explicit alias map, or sort `g.namespaces()` before binding) and document the chosen prefix for each common vocabulary. Alternatively, accept the non-determinism and document it.

### 4. No end-to-end test for \"author declared concrete OWL rel → imports as abstract → binding rejected\"

`tests/bigquery_ontology/test_owl_importer.py`

`test_binding_abstract_relationship_fails` (line 1111) covers explicitly-declared abstract relationships. `test_owl_objectproperty_range_on_skos_concept` (line 1202) covers the import side of the promotion. But no single test follows the full user flow:

1. Author RDF that declares an OWL ObjectProperty with a SKOS range.
2. Import → YAML has `abstract: true` on the rel (implicit promotion).
3. Author a binding targeting that rel.
4. Loader rejects with the new guided error message.

This is the worst-first-time-user journey for this feature, and covering it would catch any future regression where either the promotion or the error message drift. A single test plus a snapshot of the rejection message would lock this in.

---

## Cross-reference against #57 and #58

- **#57 mapping table** — all proposed mappings now implemented except the acknowledged `skos:ConceptScheme` ontology-level annotation. ✅
- **#57 silent-drop guarantee** — every transformation from RDF to YAML is now either directly representable or counted in `_DropSummary`, except the OWL→abstract promotion (finding 1 above).
- **#58 `rt.in_scheme()` readiness** — scheme annotations now preserve full IRI, so the planned accessor can return either concept-IRIs-by-scheme-IRI or concept-IRIs-by-scheme-local-name at the caller's choice. ✅
- **#58 `rt.annotation(name, key)` readiness** — annotation keys now consistently use `prefix:local` form (with full-IRI fallback), so callers asking for `dc:title` will find it. ✅
- **#58 relaxed `(name, from, to)` uniqueness for abstract rels** — `ontology_loader.py:80-93` uses that exact tuple, and `test_abstract_rel_duplicate_names_different_endpoints` exercises it. ✅

---

None of the four findings above block merge. (1) and (2) are the recommended fixes because they together close the \"I didn't write this, why is my rel abstract?\" debugging gap that will otherwise bite every FIBO / SNOMED user on day one.

caohy1988:
Follow-up review against latest `origin/main` (`dcce721`) and current PR head (`e34fb72`). I re-ran the focused importer suite (`uv run pytest tests/bigquery_ontology/test_owl_importer.py`, 83 passed) and found 3 remaining items:

1. Medium: `skos:related` is still implemented/documented inconsistently. `owl_importer.py` comments and `docs/ontology/owl-import.md` say it is emitted once per subject-object pair, but the dedupe key is ordered `(name, from, to)`, so mutual assertions still produce two abstract edges. I verified with a minimal fixture that `:c1 skos:related :c2` plus `:c2 skos:related :c1` emits both `skos_c1 -> skos_c2` and `skos_c2 -> skos_c1`. For a symmetric SKOS predicate, that duplicates the informational edge and contradicts the stated normalization rule.

2. Low: the docs are still behind the implementation on annotation identity. The code now preserves full IRIs for `skos:inScheme` / `skos:topConceptOf`, and generic literal annotations use `prefix:local` or full IRI via `_qname_or_iri`, but the docs still say these are stored by local name. The stale spots I saw are `docs/ontology/owl-import.md` §9 / §19.3 / §19.7 and `src/bigquery_ontology/docs/user_manual.md` in the OWL-import section.

libei:
Thanks — both addressed in fb77ce1:

1. **Medium — `skos:related` symmetric dedupe.** `_extract_skos_relationships` now sorts `(from_name, to_name)` before keying the `seen` set for `skos:related`, so mutual assertions collapse to one edge. Regression test `test_related_is_symmetric_single_edge` covers the `:c1 related :c2` + `:c2 related :c1` case.

2. **Low — stale docs on annotation identity.** Updated `docs/ontology/owl-import.md` §9, §19.3 (Reference predicates table now says "IRI target stored verbatim"), and §19.7, plus the matching spots in `src/bigquery_ontology/docs/user_manual.md` (§OWL-import mapping table, SKOS summary table, Generic literal annotations) to reflect that `skos:inScheme` / `skos:topConceptOf` preserve full IRIs and generic literal annotations key on `prefix:local` (or full IRI fallback).

`uv run pytest tests/bigquery_ontology/test_owl_importer.py` → 84 passed.

PR #68, opened 2026-04-23
feat(runtime): filter abstract upstream elements in both SDK adapters

caohy1988:
## Summary

Follow-up to #62 (SKOS import support). PR #62 added `abstract: bool = False` on upstream `Entity`/`Relationship` and relaxed relationship uniqueness to `(name, from, to)` for abstract relationships so SKOS imports can emit multiple `skos_broader` edges with different endpoints. Both SDK adapters still build `{name: obj}` maps over the full upstream ontology, which last-write-wins-collapses duplicate-name abstract relationships.

This PR adds the reviewer-requested "choke-point filter" at both SDK adapter boundaries so abstract upstream elements never enter SDK runtime artifacts.

## Scope

**Both SDK adapters** are patched symmetrically:

- `src/bigquery_agent_analytics/runtime_spec.py` — `graph_spec_from_ontology_binding()` (→ `GraphSpec`)
- `src/bigquery_agent_analytics/resolved_spec.py` — `resolve()` (→ `ResolvedGraph`)

Each adapter now:

1. Filters `abstract=True` entities and relationships out of the upstream ontology **before** building the name-indexed lookup maps.
2. Raises a clear error if every entity in the ontology is abstract — distinct from the existing empty-ontology case.
3. Documents the concrete-only contract in its docstring.

## Why both adapters

`resolved_spec.py:226` has the same `{e.name: e for e in ontology.entities}` / `{r.name: r for r in ontology.relationships}` pattern as `runtime_spec.py:229`. Filtering only in one would leave the other exposed.

## Why not a `GraphSpec` validator

SDK `EntitySpec` and `RelationshipSpec` (`ontology_models.py:112, :134`) intentionally do not carry an `abstract` field — the concept doesn't exist in SDK runtime models. The invariant lives at the adapter boundary: *"GraphSpec and ResolvedGraph carry only concrete, bindable elements by construction."* Stated in both adapter docstrings.

## Safety framing: defense-in-depth, not live regression

The validated `load_binding(...)` path already rejects bindings that target abstract elements (`src/bigquery_ontology/binding_loader.py:229, :245`) with clear error messages. Normal users going through `load_ontology` + `load_binding` cannot hit the collapse. This PR protects callers that construct `Ontology` and `Binding` objects programmatically in Python and bypass upstream validation.

## Test plan

- [x] New test file `tests/test_abstract_adapter_filter.py`, 12 tests, symmetric across both adapters:
  - abstract entities/relationships filtered out of output
  - all-abstract ontology raises clear "no concrete entities" error
  - duplicate-endpoint abstract relationships (programmatic construction) do not collapse
  - programmatic `Binding` targeting an abstract entity → `not defined` error
  - programmatic `Binding` targeting an abstract relationship → `not defined` error (was silently succeeding with collapsed map entry before this fix)
- [x] `python -m pytest tests/` → **1991 passed, 4 skipped, 0 failures** (baseline was 1991 on latest `origin/main`).

## Follow-ups (not in this PR)

- Issue #58 (Runtime entity resolution primitives) is the larger next-wave work. This PR does not implement any of it; it only ensures the PR #62 foundation is safe for the SDK runtime before that design lands.

### DIFF
--- a/src/bigquery_agent_analytics/resolved_spec.py
+++ b/src/bigquery_agent_analytics/resolved_spec.py
@@ -251,12 +251,8 @@ def resolve(
   # This also prevents the ``(name, from, to)``-unique abstract
   # relationships relaxed in PR #62 from collapsing under
   # ``{r.name: r}`` last-write-wins.
-  concrete_entity_names = {
-      e.name for e in ontology.entities if not e.abstract
-  }
-  concrete_relationships = [
-      r for r in ontology.relationships if not r.abstract
-  ]
+  concrete_entity_names = {e.name for e in ontology.entities if not e.abstract}
+  concrete_relationships = [r for r in ontology.relationships if not r.abstract]
 
   if not concrete_entity_names:
     abstract_count = len(ontology.entities)
--- a/src/bigquery_agent_analytics/resolved_spec.py
+++ b/src/bigquery_agent_analytics/resolved_spec.py
@@ -235,14 +235,30 @@ def resolve(
   project = binding.target.project
   dataset = binding.target.dataset
 
-  # Concrete-only filter: abstract upstream elements never enter the
-  # name-indexed maps. See the docstring above for the rationale.
-  concrete_entities = [e for e in ontology.entities if not e.abstract]
+  # Concrete-only contract: abstract upstream elements are not
+  # bindable, but for entities we must keep abstract parents reachable
+  # in the name-index map because a concrete child may ``extends`` an
+  # abstract parent and inherit its keys/properties (upstream
+  # validation allows this shape; ``_effective_keys`` /
+  # ``_effective_properties`` walks the ancestor chain via
+  # ``item_map[cur]``).
+  #
+  # Entities: full map for inheritance traversal, concrete-only set
+  # used to enforce bindability.
+  # Relationships: abstract relationships do not participate in
+  # inheritance (upstream's ``_check_extends_targets`` runs only on
+  # concrete relationships), so filter them before building the map.
+  # This also prevents the ``(name, from, to)``-unique abstract
+  # relationships relaxed in PR #62 from collapsing under
+  # ``{r.name: r}`` last-write-wins.
+  concrete_entity_names = {
+      e.name for e in ontology.entities if not e.abstract
+  }
   concrete_relationships = [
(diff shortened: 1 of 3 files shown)


caohy1988:
Good catch, confirmed and fixed in `62e0922`.

Reproduced the `KeyError 'AbstractParent'` exactly as described: a programmatic ontology with `Child extends AbstractParent` where `AbstractParent` is `abstract: true` and owns the primary key crashed in `_ancestors` during `_effective_keys` traversal.

Applied fix option A from the review: full map for inheritance, concrete-only set for bindability.

**`resolved_spec.py`:**

- `ont_entity_map` keeps all ontology entities (abstract parents reachable for `_ancestors` walks).
- New `concrete_entity_names` set is the bindability gate — `binding.entities` loop checks membership before lookup.
- Relationships keep the concrete-only filter: upstream's `_check_extends_targets` runs only on concrete relationships (`ontology_loader.py:115-117`), so abstract relationships never participate in inheritance, and the filter also blocks the PR #62 `(name, from, to)`-unique abstract relaxation from collapsing under `{r.name: r}`.
- Error message on abstract binding now reads: *"Binding references entity X which is not defined in ontology ... (or is declared abstract and therefore not bindable)."*

**`runtime_spec.py`: unchanged.** It doesn't walk ancestors today (`_build_key_spec` uses `entity.keys` directly), so the simpler concrete-only entity filter is sufficient. A concrete child with inherited keys has always failed `runtime_spec` with "no primary key" — pre-existing, not introduced by #68. Flagging it as a follow-up rather than widening scope here.

**New test:**

`test_concrete_child_extends_abstract_parent_inherits_keys` — constructs the exact shape from the review (`Child` extends abstract `AbstractParent` with `keys=Keys(primary=["pid"])`, binding maps `pid -> child_id`), asserts:

- Output contains only `Child` (abstract parent correctly excluded from output).
- `Child.key_columns == ("child_id",)` (inheritance walked through abstract parent).
- Child inherits the parent's `label` property, remapped to `child_label`.

I reproduced the crash on pre-fix code before adding the fix — classic RED/GREEN.

**Re-ran:**

- `tests/test_abstract_adapter_filter.py` → 13 passed (12 from before + the new inheritance test).
- `tests/` → 1992 passed, 4 skipped, 0 failures.

PR #71, opened 2026-04-23
feat(ontology): A1 — internal _fingerprint module for concept-index provenance

caohy1988:
## Summary

First code slice of the concept index + runtime entity resolution work (issue #58). Per the [in-repo implementation plan](docs/implementation_plan_concept_index_runtime.md) — item **A1** in the work breakdown.

Adds `src/bigquery_ontology/_fingerprint.py` — the **single source of truth** for the canonical model fingerprint and the 12-hex-char `compile_id` pair-consistency tag. Both the ontology compiler (writer, landing in A3) and the SDK runtime (verifier, landing in B1) import from here via absolute import so the contract is guaranteed to agree bit-for-bit between writer and reader.

## API

Two functions, both internal (underscore-prefix module, not re-exported):

```python
from bigquery_ontology._fingerprint import fingerprint_model, compile_id

ont_fp = fingerprint_model(ontology)
# "sha256:6f9a…"  (64 hex chars)

bnd_fp = fingerprint_model(binding)
# "sha256:3c2e…"

cid = compile_id(ont_fp, bnd_fp, "bigquery_ontology 0.2.1")
# "a1b2c3d4e5f6"  (12 hex chars, 48 bits)
```

## Contract pinning (W1 watchpoint)

The serialization contract is pinned in the module docstring:

- `model_dump(mode="json", by_alias=False, exclude_none=False)` — normalizes enums, datetimes, and Pydantic types to JSON-safe primitives so the hash is stable across Python versions; keeps optional-but-declared fields in the output so adding a default later doesn't silently collide with the pre-default fingerprint.
- `json.dumps(..., sort_keys=True, separators=(",", ":"), ensure_ascii=False)` — sort keys at every nesting level, no whitespace, UTF-8.
- SHA-256 over the encoded bytes; output `"sha256:" + hexdigest`.

The docstring explicitly calls out the three downgrade paths that must **never** be taken: `yaml.dump(model)`, `str(model)`, or `model.model_dump()` without `mode="json"`. All three break cross-version or cross-type stability.

## Why underscore

- Non-public — module path pins the semver-unstable status for the functions.
- Renaming `fingerprint_model` / `compile_id` or changing their signatures does not trigger a package minor bump.
- Matches the existing pattern for other internal modules.

Not re-exported from `bigquery_ontology/__init__.py`; a test case pins that.

## Tests

`tests/bigquery_ontology/test_fingerprint.py` — 15 tests, all new.

**Determinism (what must stay stable):**
- Same model → same fingerprint across repeated calls.
- Non-semantic YAML edits (whitespace, comments, indentation, blank lines) → identical fingerprint.

**Semantic sensitivity (what must shift):**
- Entity rename.
- Property type change.
- Binding target dataset change.
- Key-column order change (lists are declaration-ordered, not sorted — per the plan).

**`compile_id`:**
- 12 hex chars, lowercase.
- Deterministic given same inputs.
- Responds independently to each of the three inputs (`ontology_fingerprint`, `binding_fingerprint`, `compiler_version`).

**Private surface:**
- Not exposed at `bigquery_ontology.*` top level.
- Still importable via the pinned absolute path.

## Test plan

- [x] `python -m pytest tests/bigquery_ontology/test_fingerprint.py -v` → 15 passed.
- [x] `python -m pytest tests/` → 2007 passed, 4 skipped, 0 failures.
- [x] `bash autoformat.sh` → no diff remaining.

## Scope

Phase 1 of the implementation plan. This is A1 only; the subsequent PRs stack cleanly on top:

- **A2** — `concept_index.py` row builder (depends on A1 and #57's `abstract` field; both now in place).
- **A3-A5** — `compile_concept_index` + inline-UNNEST SQL emission.
- **A7** — CLI flags `--emit-concept-index` + `--concept-index-table`.
- **A8** — user-facing `docs/ontology/concept-index.md`.

None of those are in this PR.

### DIFF
--- a/src/bigquery_ontology/_fingerprint.py
+++ b/src/bigquery_ontology/_fingerprint.py
@@ -1,0 +1,109 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Canonical model fingerprint + compile_id for concept-index provenance.
+
+Internal module. Both the ontology compiler (``graph_ddl_compiler`` when
+it gains ``compile_concept_index``) and the SDK runtime (``OntologyRuntime``
+in ``bigquery_agent_analytics``) import from here via absolute import —
+the contract has to agree bit-for-bit between the side that writes meta
+rows and the side that verifies them, so there is exactly one definition.
+
+The underscore prefix marks this as non-public. Not re-exported from
+``bigquery_ontology/__init__.py``. See
+``docs/implementation_plan_concept_index_runtime.md`` watchpoint W1.
+
+Serialization contract (pinned):
+
+- Input: ``BaseModel.model_dump(mode="json", by_alias=False, exclude_none=False)``.
+  ``mode="json"`` normalizes enums, datetimes, and Pydantic types to
+  JSON-safe primitives so the hash is stable across Python versions.
+  ``exclude_none=False`` keeps optional-but-declared fields in the
+  output so adding a default value later doesn't silently collide
+  with the pre-default fingerprint.
+- Encoding: ``json.dumps(..., sort_keys=True, separators=(",", ":"),
+  ensure_ascii=False)``. Sorted keys at every nesting level, no
+  whitespace, UTF-8.
+- Hash: SHA-256 over the encoded bytes; output ``"sha256:" + hexdigest``.
+
+Never fingerprint over ``yaml.dump(model)``, ``str(model)``, or
+``model.model_dump()`` without ``mode="json"``. Each of those breaks
+cross-version or cross-type stability.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+
+from pydantic import BaseModel
+
+_FINGERPRINT_PREFIX = "sha256:"
+_COMPILE_ID_LEN = 12  # 48 bits of entropy, per issue #58.
+
+
+def fingerprint_model(model: BaseModel) -> str:
+  """Return the canonical SHA-256 fingerprint of a validated Pydantic model.
+
+  Args:
+      model: Any validated Pydantic model (Ontology, Binding, or similar).
+
+  Returns:
+      ``"sha256:" + 64 hex chars`` — a stable content hash over the
+      model's semantic fields, invariant to source-file whitespace,
+      key ordering, comments, and Pydantic-default materialization.
+  """
+  dumped = model.model_dump(
+      mode="json",
+      by_alias=False,
+      exclude_none=False,
+  )
+  canonical = json.dumps(
+      dumped,
+      sort_keys=True,
+      separators=(",", ":"),
+      ensure_ascii=False,
+  ).encode("utf-8")
+  return _FINGERPRINT_PREFIX + hashlib.sha256(canonical).hexdigest()
+
+
+def compile_id(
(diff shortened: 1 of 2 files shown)


caohy1988:
## Updated to Option 2 contract (additive, non-breaking) — commit \`b61d88a\`

Brings this PR into alignment with the Option 2 decision from [issue #58#issuecomment-4311684137](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/58#issuecomment-4311684137) and the docs PR #80.

### What changed

- **New export: \`compile_fingerprint(ont_fp, bnd_fp, compiler_version) -> str\`**. Full 64-hex SHA-256 over the concatenated triple. This is the canonical integrity key used by strict verification for main↔meta pair consistency and runtime freshness checks.
- **\`compile_id\` now derives from \`compile_fingerprint\`**: its body is just \`return compile_fingerprint(...)[:_COMPILE_ID_LEN]\`. Return value is byte-identical to the prior release — non-breaking for existing callers.
- **Module docstring** rewritten to describe all three exports, the structural invariant, and the Option 2 linkage.

### Tests (23 total, +8)

- **\`TestCompileFingerprint\` (5 new)**: length (64 hex), determinism, responds-to-each-input-independently.
- **\`TestCompileIdFingerprintInvariant\` (3 new)**: asserts \`compile_id(*args) == compile_fingerprint(*args)[:12]\` across varied inputs including degenerate and long cases. Includes a regression guard against re-implementing \`compile_id\` with a different hash payload.
- **\`TestPrivateSurface\`**: extended to cover \`compile_fingerprint\` — must not be re-exported at \`bigquery_ontology.*\`, must be importable via the pinned absolute path.
- **\`TestCompileId\`** (existing 5): all still pass. Return-value contract is unchanged.

### Why structural derivation matters

\`compile_id\` never computes its own hash. The invariant \`compile_id == compile_fingerprint[:12]\` is enforced at the function boundary, not by convention. A future refactor would have to delete the derivation to break it — and that's visible in review.

This prevents the two provenance columns in the concept index (\`compile_id\` for display, \`compile_fingerprint\` for integrity) from ever drifting out of sync.

### Full suite

- \`pytest tests/bigquery_ontology/test_fingerprint.py\` → 23 passed.
- \`pytest tests/\` → 2015 passed, 4 skipped.
- \`bash autoformat.sh\` → no diff remaining.

### Downstream sequencing (unchanged, now unblocked)

1. ✅ **This PR (A1)** — fingerprint primitives settled.
2. Docs PR #80 — concept-index schema and W2 rewritten under Option 2.
3. A2 row builder — consumes both provenance columns.
4. A3 emission SQL — writes \`compile_id\` and \`compile_fingerprint\` on main and meta.
5. B1/C4 verifier — queries \`compile_fingerprint\` only.

caohy1988:
Addressed findings 2 and 3 in \`69765f6\`.

- **Finding 2 (exact payload contract)** — docstring for ``compile_fingerprint`` now pins the payload explicitly as NUL-delimited UTF-8 of the three inputs, with a rationale (NUL cannot appear in any input — fingerprints are hex, version strings are ASCII — so the delimited encoding is unambiguous). Explicit instruction to import :func:\`compile_fingerprint\` rather than reimplement. Matching change in docs PR #80 (commit ``d78b46b``) removes the ambiguous ``||`` notation.
- **Finding 3 (golden vector)** — new \`test_golden_vector\` pins the byte-exact digest for ``("sha256:aa", "sha256:bb", "gm-1.0.0")``:

  ```python
  full  == "dd301e2874a3bc55dd3ab6ea428ad1046f6b4834b4dccd7a87cf82a1991e0184"
  short == "dd301e2874a3"
  ```

  A refactor that silently changes the separator, order, or encoding would still pass length/determinism/input-sensitivity tests and the ``compile_id == compile_fingerprint[:12]`` invariant (since ``compile_id`` is a structural truncation), but the golden vector flips and fails the test. That closes the last hole where the cross-package bit-for-bit contract could drift under refactoring pressure.

Test suite: 24 passed (was 23).

caohy1988:
Closing in favor of https://github.com/haiyuan-eng-google/BigQuery-Agent-Analytics-SDK-fork/pull/40 — same head commit (`69765f6`), same diff, just retargeted per maintainer request. The work continues there.

PR #90, opened 2026-04-27
Add operational metrics, LLM classifier, and demo polish

evekhm:
## Summary

- **Operational metrics gate** (`eval/operational_metrics.py`): Adds deterministic checks using the SDK's `CodeEvaluator` — latency, token efficiency, turn count, and tool error rate. Runs as a baseline in Step 3 and as a before/after comparison in Step 5, verifying the improved prompt doesn't trade quality for cost.
- **LLM-based eval classifier**: Replaces the brittle keyword-matching heuristic in `_classify_question` with a Gemini call. Fixes false `expected_tool` assignments (e.g., sick leave questions incorrectly mapped to `get_current_date`) and returns proper topic categories (`benefits`, `expenses`) instead of tool names.
- **Weighted traffic generation**: Biases synthetic traffic toward expenses, benefits, and holidays (at least 6/10 questions) to reliably expose V1's blind spots and produce a clear ~40-50% baseline.
- **Scale-safe report filtering**: Strips meaningful sessions from the quality report before entering the improvement pipeline, so only failures are loaded into memory and LLM context.
- **Demo narration** (`DEMO_NARRATION.md`): Replaces `DEMO_SCRIPT.md` with a speaker-ready narration transcript covering all steps, done summary, and reset instructions.
- **README overhaul**: Adds `CodeEvaluator` as a key component, blog post links (Further Reading), V1 flaws table, observed model behavior note, and architecture updates.

## Test plan

- [ ] Run `./reset.sh && ./run_cycle.sh` — verify baseline metrics appear after Step 3 and comparison table after Step 5
- [ ] Confirm all 4 metrics show PASS when averages are within budget
- [ ] Verify extracted eval cases have correct topic categories (e.g., `benefits` not `lookup_company_policy`)
- [ ] Check V1 baseline score lands ~40-50% meaningful (not 80%) with weighted traffic


caohy1988:
## Review against current head (`4de4df65`)

CI: 7/9 pass. **Format check fails** — that's the immediate merge blocker. Tests and `scan-pr` both green.

The PR is a substantive improvement to the demo: the operational-metrics gate is the clearest signal yet that the cycle didn't trade quality for cost, the LLM classifier replaces the brittle keyword-scoring path with something actually correct, and the weighted-traffic + scale-safe-filter changes make the demo's narrative more honest. Most of my findings below are diff-local; one is a regression of a prior High I flagged on the previous round.

---

### High — `expected_tool` revert reintroduces the pre-flight-always-fails bug

Commit `2b8d983a` ("Set expected_tool to lookup_company_policy in golden eval cases") undoes the fix Eva landed in `126ff0b1` on the prior round. The three committed V1-baseline cases in `eval/eval_cases.json` go from:

```json
{"expected_tool": "unknown", "notes": "V1 answers from inline knowledge (no tool call)."}
```

back to:

```json
{"expected_tool": "lookup_company_policy"}
```

V1's prompt still tells the agent *"Answer questions using only the information above ... If a question is about a topic not listed above, tell the user you do not have that information"* (`agent/prompts.py`). V1 doesn't call the tool for PTO / sick / remote-work — those are inline. The new tool-call evidence judge (`eval_runner.py:judge_case`) therefore sees `expected_tool='lookup_company_policy'` + `tools_called=[]` and FAILS the case for "tool was not called."

End-to-end consequence on a fresh `./run_cycle.sh`:
1. Pre-flight fails on all three committed cases.
2. Auto-improve fires (the WARNING banner now visible from the prior round, but the behavior remains).
3. Cycle 1's "baseline" measurement is the post-auto-fix V1.5, not V1.
4. The advertised ~40-50% baseline arc in the README is measured from V1.5, not V1.

This is exactly the High I documented at https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/61#issuecomment-4311038530 and that `126ff0b1` resolved. Two clean fixes:

- **Revert this revert** (preferred — three lines): set `expected_tool` back to `"unknown"` for the three committed cases. The new LLM classifier (this PR) handles real `expected_tool` values for *extracted* cases at runtime, so committed-baseline cases don't need a hardcoded value.
- **OR change V1 to actually call the tool** for the three baseline questions. Defeats the intentional-flaw design but unambiguously satisfies the judge.

Without either, the demo's "30% → 90%" arc starts measuring from the wrong baseline on every run.

---

### Medium — Format check failing on `operational_metrics.py`

`eval/operational_metrics.py` uses 4-space indentation throughout; the rest of the project (and CI's `pyink` config) is 2-space. That's almost certainly the format-check failure. Fix:

```bash
bash autoformat.sh
git add examples/agent_improvement_cycle/eval/operational_metrics.py
git commit --amend --no-edit && git push -f
```

Or as a separate commit, whatever the maintainer prefers.

---

### Medium — `_classify_question` LLM path swallows all exceptions and has no retry

`improver_agent.py` (LLM classifier replacing the keyword path):

```python
try:
    client = genai.Client()
    response = client.models.generate_content(
        model=_CLASSIFY_MODEL, contents=prompt,
        config=GenerateContentConfig(temperature=0.0, response_mime_type="application/json"),
    )
    result = json.loads(response.text)
    ...
except Exception:
    logging.warning("LLM classification failed for: %s", question[:80])
return "unknown", "unknown"
```

Two issues:
1. **`except Exception:` is too broad.** Catches everything that isn't `BaseException`, including transient `genai` API errors but also typing errors, `KeyboardInterrupt` … wait `KeyboardInterrupt` is a `BaseException` subclass, so OK there. But still: `AttributeError` from the `genai` SDK (e.g., on a structural change) would be swallowed silently. Recommend `except (genai.errors.APIError, json.JSONDecodeError, ValueError, KeyError)` — explicit list of the failures we know we want to recover from.
2. **No retry for transient 429s.** The rest of this codebase uses `HttpRetryOptions(attempts=3, http_status_codes=[429])` for Gemini calls (see `eval_runner.py`, `improver_agent.py:_generate_via_vertex_optimizer`). The classifier path doesn't use it. A flaky network or quota hit silently downgrades the extracted case to `("unknown", "unknown")` and the golden set acquires bad metadata. At cycle scale (4-6 failed cases per cycle) the silent-downgrade probability is real.

Fix: thread `HttpOptions(retry_options=HttpRetryOptions(attempts=3, ...))` into the `genai.Client()` constructor here, and tighten the `except` clause.

---

### Low — Dead-comment code in `run_cycle.sh`

```bash
-  echo "  While we wait, here are the questions that were sent to the agent:"
-  jq -r '.eval_cases[] | "    [\(.id)] \(.question)"' "$TRAFFIC_JSON" 2>/dev/null || true
-  echo ""
+#  echo "  While we wait, here are the questions that were sent to the agent:"
+#  jq -r '.eval_cases[] | "    [\(.id)] \(.question)"' "$TRAFFIC_JSON" 2>/dev/null || true
+#  echo ""
```

Either delete the lines or restore them. Don't ship commented-out code; the next maintainer can't tell whether it's intentional.

---

### Low — Comparison-mode reports PASS when there's no data

`operational_metrics.py:print_comparison`:

```python
if isinstance(a_val, (int, float)):
    status = "PASS" if a_val <= threshold else "WARN"
else:
    status = "PASS"
```

If `a_val == "n/a"` (no observations from BQ), status silently becomes `"PASS"`. A genuinely empty fresh-traffic run would render an all-pass comparison table that's misleading. Either status should be `"NO DATA"` / `"WARN"` for non-numeric values, or non-numeric values should set `all_pass=False`.

---

### Low — README + DEMO_NARRATION rewrite not reviewed in detail

This PR replaces `DEMO_SCRIPT.md` (255 lines deleted) with `DEMO_NARRATION.md` (218 new), and overhauls `README.md` (+168/-73). Substantial user-facing surface change. Worth a separate doc-only review pass before merge — I haven't inspected the prose for accuracy against the new operational-metrics-gate output or the LLM-classifier behavior.

---

### Worth highlighting (positive)

- **Operational-metrics gate is the right shape.** `CodeEvaluator.{latency, token_efficiency, turn_count, error_rate}` running against the `EXPECTED_IDS` set, both as Step 3 baseline and Step 5 before/after — this is exactly the "did the cycle trade quality for cost?" signal the prior demo lacked.
- **LLM classifier > keyword scoring.** The keyword path I helped land in the prior round was a pragmatic stopgap; an LLM call with the explicit topic enumeration is genuinely correct. Sick-leave-mapped-to-`get_current_date` was a real false positive.
- **Scale-safe report filtering** (improver only sees unhelpful/partial sessions) is the correct fix for "improver context grows linearly with cycle size." Single block, ~10 lines.
- **`agent.py` Vertex error message** with prompt ID + project + location + reset.sh hint is genuinely operator-friendly.

---

### Verdict

**Mergeable after the High and the format check.** The format-check fix is `bash autoformat.sh`. The High is a 3-line revert of `2b8d983a`. The two Mediums-on-the-classifier (broad except + no retry) are 5-10 lines each but soft — could land as a follow-up if the maintainer wants to ship the demo polish first.

The README/DEMO_NARRATION assessment is its own review effort; happy to do that pass separately if useful.

caohy1988:
Follow-up pass on top of the earlier review: I agree with the High finding and the merge blocker. I found a few additional comments worth addressing.

**Medium**

- `run_cycle.sh` currently absorbs operational-metrics failures with `|| true` in both baseline and comparison calls:
  - `examples/agent_improvement_cycle/run_cycle.sh:415-418`
  - `examples/agent_improvement_cycle/run_cycle.sh:598-603`

  That makes the new operational metrics advisory in practice, not a gate. `operational_metrics.py` returns `sys.exit(0 if all_pass else 1)` in comparison mode, but `run_cycle.sh` discards that status, so a latency/token/turn/error budget regression can print WARN and still let the cycle complete successfully. If this is intended to gate, remove `|| true` at least from the comparison-mode call. If it is intended to be advisory-only, the README/script wording should not call it a deterministic metrics gate.

**Low**

- The format failure is a little broader than the previous comment describes. `uv run pyink --check ...` reports both `examples/agent_improvement_cycle/eval/operational_metrics.py` and `examples/agent_improvement_cycle/agent_improvement/improver_agent.py` would be reformatted, and `git diff --check origin/main...HEAD` also reports trailing whitespace throughout `examples/agent_improvement_cycle/DEMO_NARRATION.md`.

- `examples/agent_improvement_cycle/README.md:107` has a broken local link: `../../bigquery_agent_analytics/evaluators.py` does not exist from that location. The target appears to be `../../src/bigquery_agent_analytics/evaluators.py`.

- `examples/agent_improvement_cycle/DEMO_NARRATION.md:12-13` has a malformed list item: `expenses, benefits, and holidays` is rendered as its own bullet instead of continuing the `lookup_company_policy(topic)` description.

The earlier High remains the main behavioral issue: the three committed V1 baseline cases should keep `expected_tool: "unknown"` unless the V1 prompt is changed to actually require tool use.

evekhm:
**This comment is based on an incorrect premise.** The claim that "V1 doesn't call the tool for PTO / sick / remote-work — those are inline" is not what actually happens when you run the agent.

We documented this observed behavior explicitly in the README and DEMO_NARRATION because it's a key insight of the demo:

> **Observed model behavior with V1.** The model does not fail uniformly across all topics. For topics **mentioned** in V1's inline knowledge (PTO, sick leave, remote work), the model often calls `lookup_company_policy` anyway — even though the prompt says "answer from the knowledge above." The inline mention acts as a signal that the topic is valid, which encourages the model to explore available tools for more detail.

This is not speculation — this is what happens on every run. The pre-flight passes consistently with `expected_tool: "lookup_company_policy"` for these three cases because the agent **does** call the tool. You can verify this yourself: run `./reset.sh && ./run_cycle.sh` and watch the pre-flight step. All three cases pass, and the tool call traces in BigQuery confirm `lookup_company_policy` was called.

The V1 prompt's flaw is **selective**, not total. It blocks tool usage only for topics **not mentioned** in the inline knowledge (expenses, benefits, holidays). For topics that are mentioned, the model recognizes the topic name, infers the tool is relevant, and calls it. This is precisely what makes the demo interesting — it's not a binary "tools work / tools don't work" story, it's a nuanced behavior where the prompt's inline knowledge acts as an implicit tool-routing signal.

Setting `expected_tool: "unknown"` for the baseline cases would be the incorrect fix — it would mean the eval judge doesn't verify that the agent actually used its tools, which is the behavior we want to validate and protect as a regression gate.

Issue #107, opened 2026-05-02
Doc: V5 migration notebook storyboard for the four-guarantee decision-lineage narrative


caohy1988:
# Doc: V5 migration notebook storyboard for the four-guarantee decision-lineage narrative

## Goal

Land a per-cell storyboard for `examples/migration_v5_demo_notebook.ipynb` that ties each narrative beat in the post-shipping decision-lineage demo to a concrete notebook cell. The storyboard ships **alongside** the last issue in the [working plan](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/96#issuecomment-4363301699), so the storytelling artifact is ready when the technical work is.

This issue does **not** rewrite the notebook. It produces a cell-by-cell plan (markdown + code-cell sketches) that a notebook author can apply mechanically.

## What the user actually has to provide (minimum hand-authored input)

Before describing the storyboard, lock down what the demo asks the user to **author**. The demo will create or consume several other artifacts (auto-scaffolded `binding.yaml`, generated `table_ddl.sql`, plugin-emitted `agent_events`) — those don't count toward the user's authored-input floor because they're produced by tooling. The introduction needs to say "you write one file; the SDK and plugin produce the rest."

### Mandatory today

1. **An ontology file.** Either `ontology.yaml` (hand-authored), or a `.ttl` (OWL/SKOS) file run through `gm import-owl` to produce `ontology.yaml`. Mandatory because it defines what entities/relationships the SDK extracts from `agent_events`. There is no path that skips it.
2. **A binding file.** Either hand-authored `binding.yaml` (mandatory if pointing at user-pre-defined tables — required for the #104 / "user owns the graph" beat) **or** auto-generated by `gm scaffold --ontology X.yaml --dataset Y --project Z --out outdir/` (works today, see [`bigquery_ontology/cli.py:453`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/src/bigquery_ontology/cli.py#L453)) for the SDK-creates-tables case. The scaffolder writes `binding.yaml` + `table_ddl.sql` and the binding is immediately valid as input to `gm compile`.
3. **Populated `agent_events`.** Generated by the BQ AA plugin running an agent. Demo includes a sandbox-runner cell that produces this.

[...]

Issue #105, opened 2026-05-02
Feat: validate ontology binding against existing BigQuery table schemas (pre-flight)


caohy1988:
# Feat: validate ontology binding against existing BigQuery table schemas (pre-flight)

## Goal

Add a pre-flight validator that checks whether the BigQuery tables a binding YAML points at physically exist with the columns and types the binding requires, **before** the SDK starts extraction or materialization.

This is a different validator from #76. #76 validates `ExtractedGraph` against `ResolvedGraph` (extracted output vs. logical spec). This issue validates `Binding`+target tables against actual BigQuery schemas (logical spec vs. physical reality).

## Motivation

Today the SDK supports populating user-pre-defined BigQuery property graphs from BQ AA traces (see CLI `ontology-build` and the Python composition path). The binding YAML expresses node/edge table topology cleanly via `EntityBinding.source` + `properties[].column` and `RelationshipBinding.source` + `from_columns`/`to_columns`.

But the loader does not check whether those tables and columns actually exist with compatible types. From [`bigquery_ontology/binding_models.py:71–84`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/src/bigquery_ontology/binding_models.py#L71):

> "Type compatibility is not checked here — the physical column type must already match the ontology property type, upstream."

Failure modes today, all surfacing late and with poor error messages:

- **Missing table** — surfaces during `OntologyMaterializer._batch_load_table` when `bq_client.get_table(table_ref)` raises `NotFound`. Materialization for that table is then logged as `delete_failed` or `insert_failed`. Other tables may proceed; the run reports partial success.
- **Missing column on a node table** — `INSERT INTO <table> (col1, col2, ...) SELECT ... FROM staging` fails with a BigQuery schema error after staging data has been loaded. Cost wasted; debugging is on the user.
- **Type mismatch** — surfaces at INSERT-from-staging time as a cast failure. The error names the BQ column, not the ontology property, so users have to map back to the binding by hand.
- **Wrong endpoint key column** on an edge table — INSERT may succeed (if the column exists with a compatible type) but produces semantic garbage: edges referencing rows that don't exist on the endpoint table. No error raised; data quality drops silently.

[...]

Issue #104, opened 2026-05-02
Feat: `bq-agent-sdk ontology-build --skip-property-graph` to populate base tables only


caohy1988:
# Feat: `bq-agent-sdk ontology-build --skip-property-graph` to populate base tables only

## Goal

Add an opt-in flag on the `ontology-build` CLI that runs phases 1–4 (load spec → extract → create tables → materialize) and stops before phase 5 (`CREATE OR REPLACE PROPERTY GRAPH`). Lets users with a pre-defined BigQuery property graph populate its base tables from BQ AA traces without overwriting their graph DDL on every run.

## Motivation

`build_ontology_graph(...)` ([`ontology_orchestrator.py:291`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/src/bigquery_agent_analytics/ontology_orchestrator.py#L291)) runs five phases in sequence:

1. Load spec.
2. Extract `ExtractedGraph` via `AI.GENERATE` on `agent_events`.
3. `OntologyMaterializer.create_tables()` — `CREATE TABLE IF NOT EXISTS` per entity/relationship table. Idempotent against pre-existing tables.
4. `OntologyMaterializer.materialize(...)` — staging-table → DELETE-by-session → INSERT FROM staging. Non-destructive, schema-aware via `bq_client.get_table(...).schema`.
5. `OntologyPropertyGraphCompiler.compile_property_graph_ddl(...)` then execute → **`CREATE OR REPLACE PROPERTY GRAPH`** ([`ontology_property_graph.py:307`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/src/bigquery_agent_analytics/ontology_property_graph.py#L307)).

Phase 5 is destructive against a user-defined property graph. A user with their own `CREATE PROPERTY GRAPH` DDL — managed by Terraform, dbt, or hand-authored to express graph-object features the SDK doesn't generate yet — gets it overwritten on every `ontology-build` run.

The clean workaround today is dropping into Python and calling `OntologyGraphManager.extract_graph(...)` + `OntologyMaterializer.materialize(...)` directly, bypassing the orchestrator. That works but loses the CLI surface (`--ontology`, `--binding`, `--session-ids`, format output, error handling).

## Proposed change

[...]

Issue #96, opened 2026-04-29
BigQuery Agent Analytics Roadmap


caohy1988:
# BigQuery Agent Analytics — Roadmap

A tiered roadmap built from a survey of the 19 open issues on this repo, the
22 sections of the current `SDK.md` user manual, the 38 Python modules in
`src/bigquery_agent_analytics/`, and the ADK plugin
(`google/adk/plugins/bigquery_agent_analytics_plugin.py`, ~3,500 LOC, 14
lifecycle callback hooks, BigQuery Storage Write API path with GCS offload).

Filed for discussion. Estimates are calibrated to a single experienced
SDK / plugin engineer; multiply for parallel streams. Impact and effort
are best-effort; the implementer of each item will refine them.

> **Ground rules for this doc.** Impact = `H/M/L` on a "downstream user
> adoption + DevX uplift" axis, not revenue. Effort = engineer-weeks
> (full-time equivalent), confidence band in parens. Items marked
> *strategic decision pending* need a maintainer call before
> sequencing.

---

## Director summary

For an exec audience, the actionable shape is:

- **Next 30 days (1 engineer):** P0 (~2 eng-weeks) plus Quality Scorecard
  Phase 1 (~2 eng-weeks). That's the realistic month. Don't promise
  more on a single track.
- **Next quarter (1 engineer):** P0 + most/all of P1 (~15 eng-weeks
  total). **No P2 unless P1 scope is cut.** If a strategic bet
  matters more than completing P1, the realistic trade is to drop
  one P1 item (typically inheritance compilation) and pull
  ReasoningBank up — but pick one, not both.
- **Next quarter (2 engineers in parallel):** ship all P0/P1 work and
  fund **ReasoningBank** as the first strategic P2 bet — it converts
  the existing `agent_improvement_cycle` demo into a real product
  surface and has lower execution risk than ontology consolidation.
  Defer `bigquery_ontology` migration (#38) until its ADR is approved;
  treat it as the *next* quarter's strategic bet, not this one.
- **Always P3:** auto-skills, advanced resolvers, Spanner backends,
  V5 context graph. Research arm or partner-team work; don't block
  production roadmap on them.

**Reality check on calendar math**: P0 ≈ 2 eng-weeks, P1 ≈ 13 eng-weeks,
**P0+P1 ≈ 15 eng-weeks ≈ 1 quarter for 1 engineer**, not 1 month.
[...]

PR #108, opened 2026-05-03
feat(ontology): add --skip-property-graph for user-owned graph DDL (#104)

caohy1988:
Implements [#104](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/104) — `--skip-property-graph` for users with their own `CREATE PROPERTY GRAPH` DDL (Terraform / dbt / hand-authored). Lets the SDK populate base tables from BQ AA traces without overwriting the user's graph object on every run.

## Changes

**`ontology_orchestrator.py`**
- `build_ontology_graph(...)` gains `skip_property_graph: bool = False`.
- When `True`, phase 5 short-circuits: no `OntologyPropertyGraphCompiler` is constructed, no `CREATE OR REPLACE PROPERTY GRAPH` runs.
- Result dict gains `property_graph_status` (`"created"` / `"failed"` / `"skipped:user_requested"`) and `skipped_reason` (`"user_requested"` only when phase 5 was skipped).

**`cli.py` (`ontology-build`)**
- New `--skip-property-graph` flag, threaded through to `build_ontology_graph`.
- Curated CLI output dict gains `property_graph_status` so JSON consumers can distinguish skipped from failed without parsing stderr.
- Exit handling: `skipped_reason == "user_requested"` exits 0 silently; existing exit-1-with-error behavior preserved for actual graph-creation failures (`property_graph_created=False` for any other reason).

**Docs (`docs/ontology/ontology-build.md`, new)**
- Documents the SDK orchestrator end-to-end with a dedicated `--skip-property-graph` section.
- Status-field reference table mapping `property_graph_status` → `property_graph_created` + exit code, so JSON consumers have a stable contract.
- "When to use this" guidance covering Terraform/dbt-managed DDL, custom graph options, and decoupled refresh cadence.
- Python API example with the expected result-dict shape.
- Linked from `docs/README.md` under Ontology Reference.

## Tests

**Unit** (`tests/test_ontology_orchestrator.py`):
- `test_skip_property_graph_does_not_construct_compiler` — asserts `OntologyPropertyGraphCompiler` is never called when the flag is set.
- `test_property_graph_status_created_on_success` — default flow with successful creation reports `"created"`, no `skipped_reason`.
- `test_property_graph_status_failed_on_compiler_false` — default flow with `create_property_graph()=False` reports `"failed"`, distinct from `"skipped:user_requested"`.

**CLI** (`tests/test_cli.py::TestOntologyBuild`):
- `test_skip_property_graph_exits_zero_with_status` — exit 0, status `"skipped:user_requested"`, no "Property Graph creation failed" message, flag threaded through.
- `test_default_invocation_omits_skip_flag` — default invocation passes `skip_property_graph=False`.
- `test_skip_property_graph_status_visible_in_text_format` — pins the contract that `property_graph_status` appears in `--format=text` output, not just JSON.
- `test_property_graph_failure_status_failed` — exit 1 with status `"failed"` and the existing error message preserved.

**Live integration** (`tests/test_integration_ontology_binding.py::TestSkipPropertyGraph`, gated on `RUN_LIVE_BIGQUERY_TESTS=1`):
- `test_skip_property_graph_issues_no_create_graph_job` — full end-to-end test:
  1. Create authored `CREATE OR REPLACE PROPERTY GRAPH` directly via SQL (simulating Terraform/dbt-managed DDL).
  2. Capture `CURRENT_TIMESTAMP()` **after** the authored DDL completes (closes the false-positive trap from #107 cell 1.3 where the JOBS_BY_PROJECT filter would otherwise catch the test's own setup job).
  3. Run `build_ontology_graph(..., dataset_id=_DATASET, table_id=_TABLE, skip_property_graph=True)` — extraction reads from real `agent_events`, materializer writes to `scratch_dataset` per spec, no DDL job for phase 5.
  4. Assert `property_graph_status == "skipped:user_requested"` AND `sum(rows_materialized.values()) > 0` (catches the silent-empty-graph trap).
  5. Query `INFORMATION_SCHEMA.JOBS_BY_PROJECT WHERE creation_time > @before AND UPPER(query) LIKE '%CREATE OR REPLACE PROPERTY GRAPH%' AND query LIKE '%{_PROJECT}.{scratch_dataset}.{spec.name}%'` — assert zero rows. The graph-ref filter prevents false-fail from unrelated DDL jobs running concurrently in the same project.
  6. Re-run the showcase GQL query — confirms the user's graph object survived intact.

This is exactly the contract #107 cells 1.3–1.5 will demonstrate in the V5 migration notebook when the storyboard moves to execution.

7/7 ontology-build CLI tests + 4/4 orchestrator tests pass; all 135 mocked tests in `test_ontology_orchestrator.py` and `test_cli.py` pass.

## Next

After this lands, [#105 (binding pre-flight)](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/105) can start independently per the [working plan](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/96#issuecomment-4363301699). #76 follows; #75 C1 is gated on #76.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -36,6 +36,7 @@ architecture, rationale, and implementation plans behind key SDK features.
 | [ontology/compilation.md](ontology/compilation.md) | Compilation — resolving ontology + binding into backend DDL |
 | [ontology/cli.md](ontology/cli.md) | CLI design for the `gm` tool (validate, compile, import-owl) |
 | [ontology/owl-import.md](ontology/owl-import.md) | OWL import — converting OWL ontologies to YAML format |
+| [ontology/ontology-build.md](ontology/ontology-build.md) | `bq-agent-sdk ontology-build` orchestrator + `--skip-property-graph` reference |
 
 ## Deployment Surfaces
 
--- a/docs/ontology/ontology-build.md
+++ b/docs/ontology/ontology-build.md
@@ -1,0 +1,84 @@
+# `bq-agent-sdk ontology-build` — End-to-End Orchestrator
+
+`bq-agent-sdk ontology-build` runs the SDK's full ontology pipeline end-to-end against a populated `agent_events` table:
+
+1. Load the spec (`--ontology X.yaml --binding Y.yaml`).
+2. Extract an `ExtractedGraph` from agent telemetry via `AI.GENERATE`.
+3. Create physical entity/relationship tables (`CREATE TABLE IF NOT EXISTS`).
+4. Materialize extracted nodes/edges into those tables.
+5. Run `CREATE OR REPLACE PROPERTY GRAPH` to wire the BigQuery property graph object.
+
+The Python entry point is `bigquery_agent_analytics.ontology_orchestrator.build_ontology_graph(...)`. The CLI is a thin wrapper.
+
+## Skipping property-graph DDL
+
+Use `--skip-property-graph` when **the caller owns their own `CREATE PROPERTY GRAPH` DDL** — e.g., the property graph is provisioned via Terraform, dbt, or hand-authored SQL — and only wants the SDK to populate base tables.
+
+```
+bq-agent-sdk ontology-build \
+  --project-id my-project \
+  --dataset-id my-dataset \
+  --ontology my.ontology.yaml \
+  --binding my-bq-prod.binding.yaml \
+  --session-ids sess-1,sess-2 \
+  --skip-property-graph
+```
+
+Behavior with the flag set:
+
+- Phase 5 short-circuits. No `OntologyPropertyGraphCompiler` is constructed, no `CREATE OR REPLACE PROPERTY GRAPH` job runs. The user's existing graph object is unchanged.
+- Phases 1–4 run normally. Tables are created (`CREATE TABLE IF NOT EXISTS` is a no-op against pre-existing tables) and rows are materialized.
+- The CLI exits 0.
+- The output dict reports:
+
+  ```json
+  {
+    "property_graph_created": false,
+    "property_graph_status": "skipped:user_requested",
+    ...
+  }
+  ```
+
+  JSON consumers should read `property_graph_status` (not just `property_graph_created`) to distinguish a deliberate skip from a creation failure.
+
+## Status field reference
+
+The CLI's `property_graph_status` field has three values:
+
+| `property_graph_status` | `property_graph_created` | Exit code | Meaning |
+|---|---|---|---|
+| `"created"` | `true` | 0 | Phase 5 ran and BigQuery confirmed the graph object. |
+| `"failed"` | `false` | 1 | Phase 5 ran but the graph object was not created. The CLI prints "Property Graph creation failed" to stderr. Tables and rows were still materialized. |
+| `"skipped:user_requested"` | `false` | 0 | `--skip-property-graph` was set. Phase 5 did not run. No error message. |
+
+Without `--skip-property-graph`, the existing exit-1 behavior on graph-create failure is preserved exactly.
+
+## When to use this
+
+- **You already manage `CREATE PROPERTY GRAPH` in Terraform / dbt / a SQL file.** The SDK's `CREATE OR REPLACE PROPERTY GRAPH` would clobber your DDL on every run.
+- **Your property graph definition uses options the SDK doesn't generate.** You hand-authored the graph DDL to express features (custom labels, additional indexes, dialect-specific options) the SDK's compiler doesn't emit.
+- **You want to populate your tables on a different cadence than you redefine the graph.** The graph definition rarely changes; the data is refreshed continuously.
+
+For all other cases, leave the flag off and let the SDK manage the property graph end-to-end.
+
+## Python API
+
+The flag is also available on `build_ontology_graph(...)`:
+
+```python
+from bigquery_agent_analytics.ontology_orchestrator import build_ontology_graph
+
+result = build_ontology_graph(
+    spec=resolved_spec,
+    session_ids=["sess-1"],
+    project_id="my-project",
+    dataset_id="my-dataset",
+    skip_property_graph=True,  # phase 5 skipped
+)
+
+assert result["property_graph_status"] == "skipped:user_requested"
+assert result["skipped_reason"] == "user_requested"
+assert result["property_graph_created"] is False
+```
+
+`skipped_reason` is only present when the phase was skipped; it is omitted when phase 5 ran (whether or not it succeeded).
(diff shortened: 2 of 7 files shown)


caohy1988:
## Follow-up: closed the two #104 AC gaps from review

**Docs added** — `docs/ontology/ontology-build.md` documents the SDK orchestrator end-to-end with a dedicated `--skip-property-graph` section. Includes:
- Per-flag behavior table mapping `property_graph_status` (`created` / `failed` / `skipped:user_requested`) to `property_graph_created` and CLI exit code, so JSON consumers have a stable reference for status disambiguation.
- "When to use this" section covering Terraform/dbt-managed DDL, custom graph options, and decoupled refresh cadence.
- Python API example showing `skip_property_graph=True` with the expected result-dict keys.

**Live integration test added** — new `TestSkipPropertyGraph::test_skip_property_graph_issues_no_create_graph_job` in `tests/test_integration_ontology_binding.py`. Gated on `RUN_LIVE_BIGQUERY_TESTS=1` matching the existing live-test pattern. Sequence:

1. Create authored `CREATE OR REPLACE PROPERTY GRAPH` directly (simulating user-managed DDL).
2. Capture `CURRENT_TIMESTAMP()` **after** the authored DDL completes (avoids the false-positive trap from #107 cell 1.3 where the JOBS_BY_PROJECT filter would catch the user's own setup job).
3. Run `build_ontology_graph(..., skip_property_graph=True)` — assert `property_graph_status == "skipped:user_requested"`.
4. Query `INFORMATION_SCHEMA.JOBS_BY_PROJECT WHERE creation_time > @before AND UPPER(query) LIKE '%CREATE OR REPLACE PROPERTY GRAPH%'` — assert zero rows.
5. Re-run the showcase GQL query against the user's graph — confirms the graph object survived the SDK run intact.

This is the exact contract #107 cells 1.3–1.5 will demonstrate in the V5 migration notebook once the storyboard moves to execution.

The unit-test coverage from the original commit is unchanged and still passes (135/135).

caohy1988:
## Review fixes folded in (commit 8d45671)

**(1) Live test now exercises real extraction/materialization.** Was passing `dataset_id=scratch_dataset` so extraction queried `scratch_dataset.agent_events` (which doesn't exist). The AI extraction path silently returns an empty graph when the source table is missing (`ontology_graph.py:683`), so the test could pass while phases 1-4 did nothing. Fixed: now passes `dataset_id=_DATASET, table_id=_TABLE` so extraction reads the real `agent_events` table; materializer still writes to `scratch_dataset` because spec entity sources arrive 3-part-qualified to `binding.target.dataset` via `_qualify_source` (`resolved_spec.py:141`), so `dataset_id` doesn't influence output table location. Added `assert sum(rows_materialized.values()) > 0` to catch the silent-empty trap.

**(2) JOBS_BY_PROJECT assertion narrowed.** Previously filtered every project-level job by `LIKE '%CREATE OR REPLACE PROPERTY GRAPH%'`, which could false-fail on an unrelated graph-DDL job from another developer or test in the same project. Now also requires the query to contain the test's fully-qualified graph reference (`{_PROJECT}.{scratch_dataset}.{spec.name}`), scoping the assertion to the test's own scratch dataset.

**(3) `docs/README.md`** gains a row for `ontology/ontology-build.md` under Ontology Reference, so the new doc is discoverable.

**(4) New CLI test `test_skip_property_graph_status_visible_in_text_format`** asserts `property_graph_status` appears in `--format=text` output. Pins the contract that the status field is not JSON-only.

**(5) PR body updated** to reflect the docs + live test from commit 054e141 and the hardening from 8d45671. Tests section now lists unit / CLI / live coverage explicitly.

**(6) CLA check** — that's an admin/account-side gate that needs your action; I can't resolve it via PR commits.

Verified locally:
- `pytest tests/test_ontology_orchestrator.py tests/test_cli.py` → 135 passed.
- `pytest tests/test_cli.py::TestOntologyBuild` → 7/7 pass (was 6 before adding the text-format check).
- Live test (`TestSkipPropertyGraph::test_skip_property_graph_issues_no_create_graph_job`) collects clean; runs only with `RUN_LIVE_BIGQUERY_TESTS=1`.

caohy1988:
## Review fixes folded in (commit e98f1c2)

**(1) DDL-detection blind spot fixed.** The previous filter required a regressed `CREATE OR REPLACE PROPERTY GRAPH` to target `_PROJECT.<scratch_dataset>.<spec.name>`. But the orchestrator's compiler builds the DDL with `dataset_id=_DATASET` in this test (since the test passes `_DATASET` for extraction), so a regression would fire DDL targeting `_PROJECT._DATASET.<spec.name>` — and the old filter would miss it.

Replaced with two narrower constraints that catch a regression in **either** dataset:

- **Graph name (`spec.name`)** — appears in the DDL string regardless of which dataset the compiler targets.
- **`sdk_feature='ontology-gql'` label** — only SDK-issued property-graph jobs carry this label per `ontology_property_graph.py:465`. The test's setup `CREATE PROPERTY GRAPH` (issued via plain SQL, not via the SDK) does not get this label, so it doesn't trip the assertion. Cleaner than dataset-scoped filtering.

This catches the regression case the test was actually meant to detect.

**(2) `graph_ref` source/target limitation documented.** Added a "Known limitation" section to `docs/ontology/ontology-build.md` noting that `result["graph_ref"]` reports the extraction dataset, not the binding's target dataset, in split source/target setups. The materialized base tables themselves still go to the binding's target dataset per the resolved spec; only the reported string is affected. Worth a follow-up issue to either rename the field or compute it from `binding.target` — flagged here as out-of-scope for #108.

**(3) Docs wording softened.** `"additional indexes, dialect-specific options"` was overreaching for BigQuery property graphs; tightened to `"custom labels or other DDL details the SDK's compiler doesn't generate."`

**(4) CLI live smoke test (finding #3 — non-blocker).** Not added in this commit. The unit tests cover CLI wiring (`property_graph_status` in JSON + text output, exit codes for skip / fail / success, flag threading); the live test exercises the orchestrator path end-to-end. A literal CLI live smoke test (running `subprocess.run(['bq-agent-sdk', 'ontology-build', '--skip-property-graph', ...])`) is a reasonable follow-up but doesn't add coverage beyond what's already there.

Verified locally:
- `pytest tests/test_ontology_orchestrator.py tests/test_cli.py` → 136 passed.
- Live test (`TestSkipPropertyGraph::test_skip_property_graph_issues_no_create_graph_job`) collects clean; runs only with `RUN_LIVE_BIGQUERY_TESTS=1`.

PR #112, opened 2026-05-04
feat(cli): add bq-agent-sdk binding-validate + ontology-build --validate-binding flags (#105 PR 2b)

caohy1988:
Implements [#105 PR 2b](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/105) per the [working plan on #96](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/96#issuecomment-4363301699): the user-facing CLI surface and docs. PR 2a (#109, merged) shipped the validator core + unit tests + a live integration test.

### Standalone CLI: `bq-agent-sdk binding-validate`

```bash
bq-agent-sdk binding-validate \
  --project-id my-project \
  --ontology my.ontology.yaml \
  --binding my-bq-prod.binding.yaml \
  --location US \
  [--strict] \
  [--format json|text|table]
```

- Loads ontology + binding, builds a `bigquery.Client`, calls the validator, prints a structured JSON report on stdout.
- Failure-code values are emitted as **lowercase strings** (e.g. `"missing_table"`); the docs reference table shows both the Python `FailureCode` enum name and the JSON value.
- Warnings always print to stderr (one line each) so CI logs surface advisory drift even when stdout is consumed by a JSON-aware tool.
- Exit codes: `0` ok, `1` failure(s), `2` unexpected error (load failure, missing flags).

### `ontology-build` integration flags

```bash
bq-agent-sdk ontology-build ... --validate-binding         # default mode
bq-agent-sdk ontology-build ... --validate-binding-strict  # strict mode
```

- When set, run the validator **before phase 2 (extraction)**. On any failure, the build short-circuits **before any `AI.GENERATE` call fires** — authoring drift never costs tokens.
- Default-mode warnings print to stderr but don't block.
- Mutually exclusive; both incompatible with deprecated `--spec-path` (validator needs the unresolved `Ontology + Binding` pair).
- Helper `_run_binding_preflight()` centralizes the load + validate + print + exit logic.

### Plus: `--location` on `ontology-build`

`build_ontology_graph()` has supported `location` since PR #108. The CLI was building without forwarding it. This PR threads it through.

## Tests

**14 new CLI tests, all pass** (using patched validators — no live BQ; live coverage is in PR 2a's `TestBindingValidationLive`):

[...]

Issue #75, opened 2026-04-23
Epic: Compile-time code generation for structured trace extractors (scoped rework)


caohy1988:
# Epic: Compile-time code generation for structured trace extractors

> **Scope revision (based on review feedback, 2026-04-23)**: the original proposal here treated this as a single implementation. That was too broad. Review found (a) the per-field fallback depended on ontology-level validation that doesn't exist today, (b) the runtime execution target was underspecified — current extraction runs *inside BigQuery* as server-side `AI.GENERATE`, not client-side Python, and emitting Python bundles is an architectural choice, not an implementation detail, (c) the proposal overgeneralized the paper's wins on structured workflow tasks to free-text extraction from arbitrary `LLM_RESPONSE` prose, and (d) the "every trace → LLM call" framing was imprecise — `ontology_graph.py` already aggregates events per session before the LLM call, only `context_graph.py`'s BizNode path is row-level. Reworking as an epic with scoped phases, prerequisites, and an explicit Phase 1 limited to compiling deterministic extractors from **known structured event schemas** (where the paper's evidence maps cleanly). Runtime `AI.GENERATE` stays as the semantic fallback until Phase 1 has measured precision/recall on real traces.

## Motivation

[arXiv:2604.05150 — *Compiled AI: Deterministic Code Generation for LLM-Based Workflow Automation*](https://arxiv.org/abs/2604.05150) (Trooskens et al., 2026, submitted April 6, 2026) treats the LLM as a **compile-time** code generator whose output is constrained to fill validated templates. Reported results on comparable tasks: 96% completion with zero execution tokens on function-calling (BFCL, n=400), 57× token reduction at 1,000 transactions, 80.4% accuracy on structured document extraction (DocILE, n=5,680).

The paper's evidence is strongest on **structured-to-structured** transformations — the target schema is known, the input has exploitable shape, and the task repeats at scale. That's exactly the profile of the SDK's structured event extractors (`structured_extraction.py`). It is **not** the profile of free-text semantic extraction from `LLM_RESPONSE` prose, which this issue no longer tries to compile in Phase 1.

## Current architecture (accurate characterization)

Three distinct extraction paths, with different costs:

1. **`context_graph.py:299` — row-level BQ-native `AI.GENERATE`** inside SQL/MERGE, invoked per row to pull business entities from `LLM_RESPONSE` text. This is the expensive, open-ended, free-text path.
[...]

PR #113, opened 2026-05-05
feat(ontology): add validate_extracted_graph (#76)

caohy1988:
Implements [#76](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/76) per the [working plan on #96](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/96#issuecomment-4363301699). Hard prerequisite for #75 (compile-time extractor code generation). Independent of #105 (which has fully merged).

## What this is

Ontology-aware validator that checks an `ExtractedGraph` (output of the extraction pipeline — LLM `AI.GENERATE`, hand-written `structured_extraction`, or future compiled extractors) conforms to the ontology declared in a `ResolvedGraph`. Catches ontology drift, malformed extractor output, and orphan-edge bugs that would otherwise silently corrupt the materialized property graph.

Sibling to #105's binding validator:

| | This PR (#76) | #105 (merged) |
|---|---|---|
| Inputs | `ResolvedGraph` + `ExtractedGraph` | `Ontology` + `Binding` + live `bq_client` |
| Phase | post-extraction, pre-materialization | pre-extraction |
| Surfaces | extractor output ↔ ontology spec drift | binding ↔ BigQuery schema drift |

## Public API

\`\`\`python
from bigquery_agent_analytics import (
    validate_extracted_graph,
    validate_extracted_graph_from_ontology,
    ValidationReport,
    ValidationFailure,
    FallbackScope,
)
\`\`\`

`ValidationReport.ok` is True iff `failures` is empty. `report.by_scope(FallbackScope.NODE)` filters by scope.

\`\`\`python
# Lineage-edge batches whose endpoint nodes were materialized in
# earlier passes opt out of the in-graph node lookup. The
# endpoint-key parse and the entity-segment check still run, so
# silent FK-corruption / wrong-entity-routing stays caught.
report = validate_extracted_graph(
    spec, lineage_only_graph, allow_external_endpoints=True
)
\`\`\`

## Failure codes (12 total, all NODE/FIELD/EDGE)

| Scope | Codes |
|---|---|
| `NODE` | `unknown_entity`, `missing_node_id`, `duplicate_node_id`, `missing_key`, `key_mismatch` |
| `FIELD` | `unknown_property`, `type_mismatch`, `unsupported_type` |
| `EDGE` | `unknown_relationship`, `unresolved_endpoint`, `wrong_endpoint_entity`, `missing_endpoint_key` |

`FallbackScope.EVENT` exists in the enum so #75 C2's compiled-extractor runtime can classify event-level fallbacks. **#76 does not emit `EVENT` failures** — that requires per-`event_type` expectations from #75's compile-time inputs, not in `ResolvedGraph`. Aligned with the working plan's EVENT-scope deferral.

## Validator/materializer contract alignment

A subtle class of bug the validator must defend against: an extraction can validate clean and then silently corrupt at INSERT time if the validator and materializer disagree on what an extractor is allowed to emit. Six contract points are now jointly owned by both modules via a shared private `_ontology_routing` helper:

- **Property-name resolution** — both accept logical names AND physical columns; logical name wins on collision.
- **Endpoint-key parsing** — both parse `{session}:{entity}:k=v` segments out of `from_node_id` / `to_node_id` to derive FK column values; short-form ids like `d1` parse to `{}` and trigger `missing_endpoint_key`. The parser splits on `:` *at most twice*, so values containing `:` round-trip cleanly (`sess:Decision:decision_id=a:b` → `{"decision_id": "a:b"}`).
- **Materializer routing** — `_route_node` / `_route_edge` use the same two-pass lookup, so renamed-binding properties emitted under a logical name route to the right physical column.
- **BigQuery date/timestamp shape** — `parse_iso_date` / `parse_iso_datetime` gate on a BigQuery JSON-input regex (dashed `YYYY-MM-DD` only; up to 6 fractional-second digits) before reaching `fromisoformat`. Compact (`20260505`), week-date (`2026-W19-2`), and nanosecond (`...123456789Z`) forms all reject up front.
- **Permissive-mode entity check** — `allow_external_endpoints=True` skips the in-graph node lookup but still parses the entity segment from the node-id and emits `wrong_endpoint_entity` on mismatch (including empty entity segments like `sess1::outcome_id=o1`).
- **Node primary-key vs node-id agreement** — `key_mismatch` fires when (a) the materializer-routed key value disagrees with `parse_key_segment(node_id)`, OR (b) two extracted properties route to the same key column with conflicting values, OR (c) the `node_id` entity segment disagrees with `ExtractedNode.entity_name` (including empty segments). Closes the silent-corruption gap where the materializer writes node rows from properties (last-wins on duplicate routing) but edge FKs from the parsed node-id; it also catches unescaped `,` in primary-key values that `_build_key_string` doesn't escape, and keeps node-id semantics consistent between in-graph edges (which resolve through `entity_name`) and lineage-only batches (which resolve through the parsed segment). `=` and `:` both round-trip cleanly.

## Tests (52 unit tests in `test_graph_validation.py`, all pass)

- `TestCleanBaseline` — well-formed graph with key-segment node_ids validates clean end-to-end.
- `TestNodeScopeCodes` (16 tests) — each NODE code, including duplicate-detection on unknown-entity nodes, missing_key on empty-string value, a direct `parse_key_segment` regression for `:`, and nine `key_mismatch` cases (direct mismatch, unescaped comma, short-form-id suppression, duplicate-name conflict on a renamed key, materializer-routed value comparison, `=`-in-value passes cleanly, `:`-in-value passes cleanly, node-id entity segment vs entity_name disagreement, empty entity segment).
- `TestFieldScopeCodes` (12 tests) — each FIELD code plus type-acceptor edge cases (`bool` rejected for `int64`/`double`, naive vs tz-aware datetime, ISO-8601 acceptance, list/dict on scalar, compact/week-date rejection, nanosecond-precision rejection, microsecond boundary).
- `TestEdgeScopeCodes` (5 tests) — each EDGE code; short-form ids correctly fail `missing_endpoint_key`.
- `TestRenamedColumnRoundTrip` (3 tests) — logical-name extraction validates AND materializes; physical-column form works; logical-name-wins-on-collision regression.
- `TestExternalEndpoints` (5 tests) — strict-mode default; permissive-mode passes; permissive still validates endpoint-key format; permissive catches wrong entity segment; permissive rejects empty entity segment.
- `TestOntologyAdapter` — `validate_extracted_graph_from_ontology` smoke test.
- `TestReportShape` — `by_scope` filter, `ok` property.
- `TestBkaDecisionEventRegression` — `extract_bka_decision_event`'s output validates clean.

**Verified locally**: 308/308 tests pass across `test_graph_validation.py`, `test_ontology_materializer.py`, `test_ontology_orchestrator.py`, `test_resolved_spec.py`, `test_binding_validation.py`, `test_cli.py`. Full repo suite: 2220 passed, 9 skipped. Autoformat clean.

## Modules touched

- **New**: `src/bigquery_agent_analytics/graph_validation.py` — the validator.
- **New**: `src/bigquery_agent_analytics/_ontology_routing.py` — shared private helper used by both validator and materializer (single source of truth for property-name resolution, node-id key parsing, BigQuery date/timestamp shape narrowing, and JSON-compat normalization).
- **Modified**: `src/bigquery_agent_analytics/ontology_materializer.py` — `_route_node` / `_route_edge` now use the shared lookup and `normalize_property_value`; backward-compat alias `_parse_key_segment = parse_key_segment` preserved.
- **New**: `tests/test_graph_validation.py`.
- **New**: `docs/ontology/validation.md`.
- **Modified**: `docs/README.md`, `CHANGELOG.md`.

Public API exported from `bigquery_agent_analytics/__init__.py`.

## Working plan position

\`\`\`
🔄 #76  — extracted-graph validator (this PR)
   #75 C1 — compile harness (this PR unblocks it)
\`\`\`


caohy1988:
## Review fixes folded in (rebased + force-pushed; head `deecdfe`)

PR rebased onto fresh main (which now includes #112 and the full #105 split). CHANGELOG and `docs/README.md` conflicts resolved to keep both #105 and #76 entries.

**Four substantive fixes:**

**(1, P1) `missing_endpoint_key` now parses node-id segments to match the materializer.** The validator imports `_parse_key_segment` from `ontology_materializer` and runs the same FK-derivation parse the materializer uses (`{session}:{entity}:k=v[,...]` format). Earlier behavior checked endpoint properties on the referenced node, which let a node_id like `d1` validate clean while the materializer silently produced empty FK columns at INSERT. New tests:
- `TestEdgeScopeCodes::test_missing_endpoint_key_short_node_id` — asserts short-form ids fail with the right code.
- `test_missing_endpoint_key_segment_missing_column` — covers the parses-but-missing-column case.
- Clean baseline updated to use key-segment node_ids.

**(2, P1) Materializer routing accepts both `logical_name` and `column`.** `_route_node` and `_route_edge` now build a `{logical_name: physical_column, column: physical_column}` lookup mirroring the validator's `_build_property_lookup`. Earlier behavior only matched `.column`, so a binding-renamed property emitted under its logical name would validate clean and silently drop at INSERT. New `TestRenamedColumnRoundTrip` class:
- `test_logical_name_validates_and_materializes` — asserts the row dict gets the renamed physical column populated when the extractor emits the logical name.
- `test_physical_column_name_also_works` — backward-compat coverage.

**(3, M) `duplicate_node_id` runs at the graph level (Pass 0).** Detection moved out of `_validate_node`. Two unknown-entity nodes with the same node_id now correctly trigger `duplicate_node_id` alongside the `unknown_entity` failures. New `test_duplicate_node_id_detected_on_unknown_entity_nodes`.

**(4, P1) Rebase**: dropped the stale-against-main state. CHANGELOG `[Unreleased]` now has both #105 PR 2b's entries (binding-validate CLI, `--validate-binding[-strict]`, `--location`, Python API) and the new #76 entries. `docs/README.md` lists both `binding-validation.md` and `validation.md`.

Verified locally: **282 passed** across `test_graph_validation.py`, `test_ontology_materializer.py`, `test_ontology_orchestrator.py`, `test_resolved_spec.py`, `test_binding_validation.py`, `test_cli.py`. Autoformat clean.

caohy1988:
## Review fixes folded in (commit 78ca400)

**(1, P1) Validator-clean values now produce JSON-serializable rows.** Two halves of the same gap:
- **Validator tightening**: `_accepts_date` / `_accepts_timestamp` use `datetime.date.fromisoformat` / `datetime.datetime.fromisoformat` (with trailing-Z translation) instead of regex-only matching. Strings like `"9999-99-99"` that passed the regex but aren't valid dates now fail with `type_mismatch`.
- **Materializer normalization**: `_route_node` / `_route_edge` call new `_ontology_routing.normalize_property_value(value, sdk_type)` before writing. `bytes`/`bytearray` → base64 string, `datetime`/`date` → ISO string. These values were correctly accepted by the validator but would crash `insert_rows_json` / `load_table_from_json` at INSERT time.

New tests assert end-to-end JSON-serializability of validator-clean rows (`test_bytes_value_normalizes_to_base64_string`, `test_date_value_normalizes_to_iso_string`) and the strict-parsing regression (`test_invalid_iso_date_string_now_rejected`).

**(2, M) SDK-type aliases now type-checked.** `_DDL_TYPE_MAP` supports `float64` (alias for `double`) and `bool` (alias for `boolean`) alongside the canonical names. Earlier `_TYPE_ACCEPTORS` only registered canonical names, so a GraphSpec property typed `float64` or `bool` fell through the unknown-sdk-type fallback (which returns True for forward-compat) and accepted any value. Fix: alias entries `float64 → _accepts_double` and `bool → _accepts_boolean`. New `test_sdk_type_alias_float64_is_type_checked` and `test_sdk_type_alias_bool_is_type_checked` cover both.

**(3, M) `allow_external_endpoints` opt-in for lineage-edge batches.** Pipelines that materialize lineage edges in a graph with `nodes=[]` (after endpoint nodes were materialized in earlier passes) were false-failing on `unresolved_endpoint`. Fix: `validate_extracted_graph` and `validate_extracted_graph_from_ontology` gain a kwarg `allow_external_endpoints: bool = False`. When True, in-graph node lookup is skipped but the endpoint-key parse still runs — short-form ids still fail `missing_endpoint_key`. Three new tests cover default-strict-fails, permissive-passes, and permissive-still-catches-bad-format. `validation.md` documents the kwarg and the lineage-edge use case.

**(4, hygiene) Test count refreshed**: 26 → **35 unit tests / 291/291 across touched modules**. PR body updated.

Verified locally: 291 passed across `test_graph_validation.py`, `test_ontology_materializer.py`, `test_ontology_orchestrator.py`, `test_resolved_spec.py`, `test_binding_validation.py`, `test_cli.py`. Autoformat clean.

caohy1988:
Folded both findings from the latest review round in `fff12aa`:

**1. BigQuery date/timestamp shape narrowing** (`_ontology_routing.py`)

`date.fromisoformat` / `datetime.fromisoformat` accept compact and week-date forms (`20260505`, `2026-W19-2`, `20260505T120000`) that BigQuery JSON rejects. `parse_iso_date` and `parse_iso_datetime` now gate on a BigQuery-shape regex first and only reach `fromisoformat` for the semantic check (valid month/day/time-of-day). The validator's contract now matches what the materializer can actually `INSERT`.

New tests:
- `test_compact_date_string_rejected` (`20260505`)
- `test_week_date_string_rejected` (`2026-W19-2`)
- `test_compact_timestamp_string_rejected` (`20260505T120000Z`)

**2. Permissive-mode entity-segment check** (`graph_validation.py:_validate_edge`)

`allow_external_endpoints=True` previously skipped `wrong_endpoint_entity` entirely. The node-id itself carries the entity segment (`{session}:{entity}:k=v`), so the permissive branch now parses that segment and compares against `expected_entity`. Endpoint-key parse still runs in parallel so missing FK columns also fire per-column.

New test: `test_external_endpoints_catches_wrong_entity_segment` — node-id `sess1:Decision:outcome_id=o1` for a `HasOutcome` edge expecting `Outcome` now fails.

Doc: `validation.md` gains a paragraph on the BigQuery date/timestamp shape narrowing and a note that `wrong_endpoint_entity` still runs in permissive mode.

39 unit tests in `test_graph_validation.py` (was 35); full suite (2207 tests) green; autoformat clean.

caohy1988:
Folded the new findings in `224f8cc`:

**1. Microsecond precision cap on `_BQ_DATETIME_RE`**

Changed `(?:\.\d+)?` → `(?:\.\d{1,6})?`. BigQuery `TIMESTAMP` is microsecond precision, so nanosecond strings like `2026-05-05T12:00:00.123456789Z` would have validated clean but failed at INSERT time. New regressions:
- `test_nanosecond_precision_timestamp_rejected` — 9 fractional digits rejected.
- `test_microsecond_precision_timestamp_accepted` — 6-digit boundary accepted.

**2. Empty entity segment in permissive mode**

The previous `observed_entity and observed_entity != expected_entity` short-circuited on empty strings, so `sess1::outcome_id=o1` slipped through (and `parse_key_segment` happily extracted `outcome_id=o1`, masking it from `missing_endpoint_key` too). Dropped the truthiness guard so any entity-segment mismatch — including empty — emits `wrong_endpoint_entity` with `observed=""`. New regression: `test_external_endpoints_empty_entity_segment_fails`.

**3. PR body refreshed**

Updated counts to current state: 42 unit tests in `test_graph_validation.py` (was 35), 298/298 in the focused suite (was 291/291), full repo suite 2210 passed / 9 skipped. Test breakdown table reorganized to reflect the new TestExternalEndpoints class (5 tests) and the expanded TestFieldScopeCodes (12 tests). Contract-alignment section gains entries for the BigQuery shape-narrowing and the permissive-mode entity check.

Verified at `224f8cc`: focused 298 passed, autoformat clean.

caohy1988:
Folded the `key_mismatch` finding in `66b4eaa`.

**Root cause** confirmed:
- `ontology_materializer.py:295` — node row primary keys come from `node.properties`.
- `ontology_materializer.py:331` — edge FK columns come from `parse_key_segment(edge.from_node_id / to_node_id)`.

Two sources of truth, no cross-check, so a graph with `node_id="sess:Decision:decision_id=d1"` and property `decision_id="d2"` validated clean and produced edges pointing at non-existent rows after materialization.

**Fix**: new `key_mismatch` code (NODE scope, the 12th code). `_validate_node` now parses the node-id key segment once (`parse_key_segment(node.node_id)`), and after the existing `missing_key` check compares each `key_columns` entry's parsed value against the resolved key property value. Mismatch emits `key_mismatch` with `expected=property_value`, `observed=parsed_value`.

This **also catches the unescaped-comma/equals case** the reviewer flagged: `_build_key_string` (`ontology_graph.py:293`) doesn't escape, so a property value containing `,` truncates at the first comma when re-parsed by `parse_key_segment` — the parsed value won't equal the property value, so `key_mismatch` fires.

**Suppression**: short-form node-ids like `d1` produce no parseable keys (the materializer's index-fallback path), so the comparison only runs when the parsed segment actually carries the column. `missing_endpoint_key` continues to cover the edge side.

**Tests** (3 new):
- `test_key_mismatch_between_node_id_and_property` — `decision_id=d1` vs `"d2"`.
- `test_key_mismatch_catches_unescaped_comma_in_value` — `decision_id="a,b"` parses to `"a"`, mismatch fires.
- `test_key_mismatch_does_not_fire_when_node_id_lacks_key_segment` — short-form ids don't trigger.

**Status**: 45 unit tests in `test_graph_validation.py` (was 42), 301/301 focused suite (was 298), full repo suite 2213 / 9 skipped. PR body and `docs/ontology/validation.md` updated to 12 codes; chose dedicated `key_mismatch` over folding into `missing_key` per reviewer's preference for clarity.

Verified at `66b4eaa`: focused 301 passed, autoformat clean.

caohy1988:
Folded all three findings in `ab092ea`.

**1. Validator now routes key values like the materializer**

Confirmed the bug: `_validate_node` was iterating `candidate_names = {key_col, logical_name}` (a `set`) and breaking on the first match — so the validator's chosen value depended on set iteration order, not extraction order. The materializer iterates `node.properties` in extraction order and last-wins on duplicate routing.

Repro:
- Spec: key column `decision_id`, logical name `decisionId`.
- Node: `node_id="sess1:Decision:decision_id=d2"`, properties `[decision_id="d2", decisionId="d1"]`.
- Old behavior: validator checks `decision_id="d2"` first → matches `node_id` → clean. Materializer writes `decision_id="d1"` (last). Edge FK from `parse_key_segment` = `"d2"` → broken.

Fix: `_validate_node` now uses `build_name_to_column` (the same helper the materializer uses) to gather **all** extracted properties routing to each key column, in extraction order. Then:
- If those values are not all equal → emit `key_mismatch` directly with `detail` listing every conflicting property (extractor ambiguity).
- Otherwise compare the **materializer-routed value (last-wins)** against `parse_key_segment(node_id)` and emit `key_mismatch` on disagreement.

**2. CHANGELOG.md**

`CHANGELOG.md:44` updated from "Eleven failure codes ship" to "Twelve failure codes ship" with `key_mismatch` listed in the order it appears alphabetically inside NODE-scope.

**3. Removed misleading `=` wording**

Reviewer is correct — `parse_key_segment` uses `split("=", 1)`, so `key=a=b` → `{"key": "a=b"}` round-trips cleanly. Removed the `or =` from `validation.md:86` and from the `key_mismatch` detail string in `graph_validation.py`. Added a new test `test_key_mismatch_does_not_fire_for_equals_in_value` to lock the round-trip behavior. Module docstring also updated from "Eleven codes" to "Twelve codes".

**Tests** (3 new → 48 total in `test_graph_validation.py`, was 45):
- `test_key_mismatch_catches_duplicate_property_names_with_conflicting_values` — exact reviewer repro.
- `test_key_mismatch_compares_against_materializer_routed_value` — same routing two values, last-wins compared to node_id.
- `test_key_mismatch_does_not_fire_for_equals_in_value` — round-trip verified.

Verified at `ab092ea`: focused suite 304 passed (was 301), full repo 2216 / 9 skipped, autoformat clean.

caohy1988:
Folded the colon-in-value finding in `8cb1375`.

**Root cause**: `parse_key_segment` used `node_id.split(':')` and read `parts[-1]`, so `sess1:Decision:decision_id=a:b` produced `parts = ["sess1", "Decision", "decision_id=a", "b"]`, `parts[-1] = "b"`, and the `=`-check returned `{}`. Same class of bug as the comma case but on the section-delimiter side: a value containing `:` looked indistinguishable from extra node_id sections.

**Fix**: `node_id.split(':', 2)` caps the split at two so `parts[2]` is always the full key segment. The documented `{session}:{entity}:k=v` shape is unchanged — only the parser's tolerance for `:` inside a key value changes. Sanity-checked all four cases:
- `sess1:Decision:decision_id=a:b` → `{"decision_id": "a:b"}`
- `sess1:Decision:decision_id=a:b,other=c` → `{"decision_id": "a:b", "other": "c"}`
- `d1` → `{}` (short-form fallback unchanged)
- `sess1:Decision:decision_id=d1` → `{"decision_id": "d1"}` (plain case unchanged)

**Tests** (2 new):
- `test_parse_key_segment_preserves_colon_in_value` — direct parser regression covering all four cases above.
- `test_key_mismatch_does_not_fire_for_colon_in_value` — end-to-end: node with property `decision_id="a:b"` and matching node_id validates clean.

**Status**: 50 tests in `test_graph_validation.py` (was 48), 306/306 focused suite (was 304), full repo 2218 / 9 skipped. Doc updated to document `=` *and* `:` both round-trip cleanly; only commas truncate.

Verified at `8cb1375`: focused 306 passed, autoformat clean.

caohy1988:
Folded both findings.

**1. Node-id entity segment vs `ExtractedNode.entity_name`** (`8a477e5`)

Confirmed the asymmetry: in-graph edge endpoint resolution uses `referenced.entity_name`, so a node with `node_id="s:Outcome:decision_id=d1"` and `entity_name="Decision"` validated clean and any in-graph edge to it also passed `wrong_endpoint_entity`. The same id seen from a lineage-only batch (`allow_external_endpoints=True`) parses through `parse_key_segment` and fails the permissive-mode entity-segment check. The two paths disagreed on what the id means.

`_validate_node` now splits the node_id on `:` at most twice and emits `key_mismatch` (NODE scope) when the second segment is non-empty and doesn't match `node.entity_name`. Folded under the existing code per your guidance — no 13th code, detail string makes the specific drift explicit.

New test: `test_key_mismatch_catches_node_id_entity_segment_disagreement`.

**2. PR body refreshed**

Now reflects current state at `8a477e5`:
- 51 graph-validation tests (was 48)
- 307/307 focused suite (was 304)
- 2219 full repo / 9 skipped (was 2216)
- contract-alignment bullet for endpoint-key parsing now mentions `:` round-trip
- `key_mismatch` bullet now lists all three triggers (property-vs-segment, conflicting-routes, entity-segment disagreement)
- `TestNodeScopeCodes` test breakdown updated to 14 tests with seven `key_mismatch` cases enumerated

Verified at `8a477e5`: focused suite 307 passed, full repo 2219 passed / 9 skipped, autoformat clean, `git diff --check` clean.

caohy1988:
Folded the empty-entity-segment finding in `f0731bb`.

You're right — the truthiness guard `if observed_entity and observed_entity != node.entity_name` was inconsistent with the permissive-endpoint path I already tightened (which rejects `sess1::outcome_id=o1`). The node-side path now matches: gate on the 3-part shape (`len(nid_parts) >= 3`) instead of 2-part, drop the truthiness guard, compare directly. Empty entity segments in 3-part node_ids now emit `key_mismatch`. Short-form fallback ids (`d1`) and 2-part malformed ids still skip the check — other codes catch those.

Test: `test_key_mismatch_catches_empty_node_id_entity_segment` — `sess1::decision_id=d1` with `entity_name="Decision"` now fails with `observed=""`, `expected="Decision"`.

**Status**: 52 unit tests in `test_graph_validation.py` (was 51), 308/308 focused suite (was 307), full repo 2220 / 9 skipped (was 2219). Autoformat clean.

Verified at `f0731bb`.

PR #114, opened 2026-05-05
feat: add context cache hit rate evaluator

### LINKED ISSUE #32: Improve token efficiency in ADK via BigQuery Agent Analytics
Below is a breakdown of how to implement these three components within the ADK and BQ AA framework.
---

### 1. RFC: Context Prefix Hit Rate for Observation
**Objective:** Track how effectively the Gemini/LLM context cache is being utilized across sessions to reduce TTFT (Time to First Token) and cost.

**Proposed RFC Structure:**
* **Motivation:** Context caching significantly reduces latency and cost. Without tracking the "Hit Rate," we cannot optimize prompt templates or session management.
* **Metric Definition:** $$\text{Context Hit Rate} = \frac{\text{Cached Tokens}}{\text{Total Input Tokens}}$$
* **Implementation in ADK:** * **Instrumentation:** Update the `BigQueryAgentAnalyticsPlugin` to extract `cached_content_token_count` from the Gemini API response metadata (`usage_metadata`).
    * **Schema Change:** Add a `context_cache_metadata` JSON field to the `agent_events` BigQuery table to store hit/miss details per turn.
* **Success Criteria:** Identify "Cold Start" turns where the hit rate is <10% and "Warm" turns where it is >90%.

---

### 2. SDK: Evaluate and Improve Context Prefix Hit Rate
You can extend the `bigquery_agent_analytics` SDK by adding a new evaluator in the `CodeEvaluator` class.

**Implementation Example:**
```python
from bigquery_agent_analytics import CodeEvaluator, Client

# 1. Evaluate: Add a custom metric to your pipeline
def context_cache_hit_rate_metric(trace):
    """Calculates the average cache hit rate across all LLM turns in a trace."""
    llm_calls = [s for s in trace.spans if s.event_type == "LLM_RESPONSE"]
    if not llm_calls: return 1.0
    
    total_hits = sum(s.metadata.get("cached_tokens", 0) for s in llm_calls)
    total_input = sum(s.metadata.get("input_tokens", 1) for s in llm_calls)
    return total_hits / total_input

# 2. Improve: Use the SDK to find 'Cache Breakers'
client = Client(project_id="your-project", dataset_id="your_dataset")
traces = client.list_traces(min_latency_ms=3000) # Find slow traces likely missing cache

for trace in traces:
    hit_rate = context_cache_hit_rate_metric(trace)
    if hit_rate < 0.5:
        # Improvement Logic: Identify which turn changed the prefix significantly
        print(f"Low cache efficiency in Session {trace.session_id}: {hit_rate:.2%}")
```

---

### 3. Context Graph for Tool Call Efficiency
**Yes**, this is the primary use case for the **`context_graph.py`** module in the BQ AA SDK. The "Context Graph" treats agent events as **TechNodes** and your business data/tool results as **BizNodes**.

#### How to do it via BQ AA:
1.  **Plugin Setup:** Ensure the `BigQueryAgentAnalyticsPlugin` is active in your ADK `App`. It will automatically log every tool execution as a span.
2.  **Graph Construction:** Use the SDK to build a property graph that links tool calls to the context available at that moment.
3.  **Efficiency Resolution:** By analyzing the graph, you can detect **Redundant Tool Calls** (e.g., the agent called `get_weather` even though the weather was already in the chat history/context).

**Example Code for Tool Efficiency Graph:**
```python
from bigquery_agent_analytics.context_graph import ContextGraphBuilder

# Initialize the graph builder from your BQ events
builder = ContextGraphBuilder(client)
graph = builder.build_from_session("session-id-123")

# Resolve efficiency: Find tool calls where the 'BizNode' (the data) 
# was already present in the graph before the 'TechNode' (the tool call)
efficiency_report = graph.resolve_tool_efficiency()

for call in efficiency_report.redundant_calls:
    print(f"Inefficient Tool Call: {call.tool_name} was called, " 
          f"but info was already available in node {call.existing_data_node_id}")
```

### Next Steps
1.  **Update Plugin:** If your ADK version doesn't yet log `cached_tokens`, wrap your `Gemini` model call to include that metadata in the plugin's payload.
2.  **GQL Queries:** Use the `context_graph`'s GQL (Graph Query Language) support to query sessions where `(ToolCall)-[:REDUNDANT_TO]->(Context)`.
3.  **Reference:** Follow the [[SDK.md](https://github.com/haiyuan-eng-google/BigQuery-Agent-Analytics-SDK/blob/main/SDK.md#context-graph)](https://github.com/haiyuan-eng-google/BigQuery-Agent-Analytics-SDK/blob/main/SDK.md#context-graph) "4-Pillar Property Graph" section to map your specific tool outputs to `BizNodes`.

sunpan9209:
## Summary
- Add a prebuilt `CodeEvaluator.context_cache_hit_rate()` metric for `cached_tokens / input_tokens`.
- Extend session summary SQL to extract cached-token telemetry and distinguish missing telemetry from true zero cache hits.
- Wire the evaluator into Python, CLI, and remote function dispatch, including strict missing-telemetry mode for operational gates.
- Document aggregate-score caveats and mixed-telemetry dilution behavior.

Closes #32

## Notes
- Missing cache telemetry reports `cache_state="no_cache_telemetry"` and passes by default; Python callers can set `fail_on_missing_telemetry=True`, CLI users can pass `--fail-on-missing-cache-telemetry`, and remote function callers can set `fail_on_missing_telemetry=true`.
- The SDK can flag low cache efficiency but cannot identify the exact cache-breaking prompt segment without prompt hashes, template IDs, or segment-level metadata.

## Tests
- `python3 -m isort --check-only src/bigquery_agent_analytics/evaluators.py src/bigquery_agent_analytics/cli.py deploy/remote_function/dispatch.py tests/test_cli.py tests/test_remote_function.py tests/test_sdk_evaluators.py`
- `python3 -m pyink --check src/bigquery_agent_analytics/evaluators.py src/bigquery_agent_analytics/cli.py deploy/remote_function/dispatch.py tests/test_cli.py tests/test_remote_function.py tests/test_sdk_evaluators.py`
- `python3 -m py_compile src/bigquery_agent_analytics/evaluators.py src/bigquery_agent_analytics/cli.py deploy/remote_function/dispatch.py`
- `git diff --check upstream/main..HEAD`
- `python3 -m pytest tests/test_sdk_evaluators.py::TestContextCacheHitRatePrebuilt tests/test_sdk_evaluators.py::TestSessionSummaryQuery tests/test_cli.py::TestAllEvaluators tests/test_cli.py::TestDefaultThresholds tests/test_cli.py::TestEvaluate::test_emit_evaluate_failures_includes_cache_state tests/test_remote_function.py::TestEvaluate::test_all_code_evaluators tests/test_remote_function.py::TestEvaluate::test_context_cache_hit_rate_strict_missing_telemetry_param`

Focused pytest result: 47 passed.

Full local pytest still has pre-existing failures in this Python 3.9 environment; the same failure classes reproduce on clean upstream HEAD.


caohy1988 (commented):
## Review: context_cache_hit_rate evaluator

Verified locally: pyink check clean, isort check clean, the 43 focused tests in the PR's test plan pass on Python 3.13. CI is fully green. The implementation closes the SDK side of #32 and the cached-tokens / input-tokens math matches the metric definition in the issue body, including the cold-start (<10%) / warm (>90%) thresholds.

A few things worth a second look before merge — none are blockers, but A and B are worth resolving inline.

### A. (Question) Aggregate-score semantics differ from every other prebuilt
Every other `CodeEvaluator` prebuilt (`latency`, `error_rate`, `turn_count`, `token_efficiency`, `ttft`, `cost`) returns binary `0.0 / 1.0` from `_score` with `threshold=1.0`, so `EvaluationReport.aggregate_scores[name]` reads as a **pass rate**.

This metric is the first to return the *raw rate* as the score with `threshold=min_hit_rate`. Two consequences:
- `aggregate_scores["context_cache_hit_rate"]` becomes "average cache hit rate across sessions", not "fraction of sessions that passed". Dashboards that consume `aggregate_scores` uniformly across metrics will plot something qualitatively different here without warning.
- When `fail_on_missing_telemetry=False` (the default), no-telemetry sessions contribute `1.0` to that average, **inflating** the apparent fleet-wide cache hit rate. If 80 of 100 sessions have no cache telemetry, the dashboard shows ~0.8 cache efficiency even when the 20 cache-bearing sessions averaged 0%.

This is salvageable either by (1) returning binary 0/1 like the others and exposing the true rate only via `details["metric_*"]["observed"]`, or (2) calling out in the docstring + design.md that this metric's `aggregate_scores` value is "average rate, with no-telemetry sessions counted as 1.0". Option (2) preserves the current API surface and is probably the right call.

### B. (Low) FAIL line direction is reversed vs other prebuilts
`_emit_evaluate_failures` in `cli.py` prints `observed=X budget=Y`. For latency/cost/etc. the gate is `observed <= budget`, so a fail reads "observed=6000 budget=5000" — naturally readable as "blew past the limit". For this metric the gate is `observed >= budget`, so a fail reads "observed=0.05 budget=0.5" — same word "budget" now means "minimum required". CI log readers won't know which direction without remembering per-metric semantics.

Two cheap mitigations: (a) include the `cache_state` field (already in `details`) on the FAIL line so a `cache_state=cold_start` token disambiguates, or (b) accept that the metric name `context_cache_hit_rate` makes the direction obvious enough. No code change required if you choose (b), but worth noting in design.md.

### C. (Low) `evaluators.py:298` swallows `detail_fn` exceptions without traceback
```python
except Exception:  # pylint: disable=broad-except
    logger.debug("Metric %s detail_fn failed", metric.name)
```
A real bug in a custom `detail_fn` produces a metric with only the base fields and a one-line debug message — no exception class, no stack trace, no traceback even with debug logging on. Suggest `logger.debug("Metric %s detail_fn failed", metric.name, exc_info=True)` or `logger.exception(...)`. Same nit applies to the existing `observed_fn` swallow at line 283 (pre-existing, not this PR's introduction, but worth a follow-up).

### D. (Low) SQL aggregation dilutes the rate when telemetry is partial
`SUM(COALESCE(... 0)) AS cached_tokens` plus `SUM(...) AS input_tokens` (which sums across **all** events including those with no cache telemetry) means: a session with 5 LLM calls, 2 cache-bearing (1000 cached / 1000 input each) and 3 cache-blind (0 cached / 1000 input each), reports `cached=2000 / input=5000 = 40%` even though the cache-aware events were 100% warm.

For a pure ADK/Vertex pipeline this is fine — telemetry is uniform. But for mixed pipelines (fallback path, mid-rollout, multi-provider) the rate underrepresents true prefix reuse. Two options: document it ("rate dilutes when not all LLM events report cache telemetry"), or add a sibling `cache_aware_input_tokens` field summed only when a cache field is present, then divide by that for the rate. The first is cheap; the second is more correct.

### E. (Low) Cache JSON path list duplicated between SUM and COUNTIF
`evaluators.py:951-1027` has 12 JSON paths × 2 (one for `SUM(cached_tokens)`, one for `COUNTIF(cache_telemetry_events)`), and they have to stay in lockstep. Adding a path to `SUM` but forgetting `COUNTIF` produces silently wrong telemetry counts. Consider building both expressions from a shared Python list of paths and `.format()`-ing them in. Not a correctness bug today, just a future-proofing gripe.

### F. (Test gap) Missing edge cases
The `TestContextCacheHitRatePrebuilt` class is solid coverage for the documented states, but a few edge cases aren't asserted:
- `cached_tokens > input_tokens` clamps to `1.0` (the `max(0.0, min(1.0, ...))` line). If a refactor ever drops the clamp, no test catches it.
- Non-numeric inputs in `cached_tokens` / `input_tokens` (string, dict, etc.) hitting the `_number` `TypeError/ValueError` fallback.
- `cache_telemetry_events` key absent but `cached_tokens` present (the legacy shim where `_has_cache_telemetry` falls back to `cached_tokens is not None`). This is a documented compatibility branch but not exercised.

Not blocking — happy paths and the documented `cache_state` enum (warm/partial/cold_start/no_cache_telemetry/no_llm_input) are covered.

### G. (Nit) `cold_start_rate` / `warm_rate` are not validated
Nothing prevents `CodeEvaluator.context_cache_hit_rate(cold_start_rate=0.9, warm_rate=0.1)`. The `_details` if/elif chain just degrades — `rate >= warm_rate` always fires before `rate < cold_start_rate` is checked. Either validate with a `ValueError`, or document the invariant `0 <= cold_start_rate < warm_rate <= 1`.

### H. (Nit) `_rate(s)` is recomputed three times per session
`evaluate_session` calls `metric.fn(s)` (→ `_score` → `_rate`), then `metric.observed_fn(s)` (= `_rate`), then `metric.detail_fn(s)` (→ `_rate`). Cheap dictionary lookups, but for high-volume eval runs this is 3× the work. Could memoize on the dict, or return a single namedtuple from a shared inner helper. Not worth holding the PR.

---

**Verdict:** Implementation is correct, well-tested for the documented contract, and aligns with issue #32. Issues A and B are about presentation/aggregation semantics that other consumers (dashboards, CI logs) will trip over; resolving them in docs or a follow-up PR is fine. C–H are polish.

The plugin schema alignment is solid: I confirmed `attributes.usage_metadata.cached_content_token_count` (the first COALESCE entry) is exactly what the live ADK `BigQueryAgentAnalyticsPlugin` writes today, so this metric will work out-of-the-box on current trace data.

caohy1988:
Fresh-eye follow-up: I found two additional non-blocking comments that would make this PR stronger.

**Medium**

1. Strict missing-telemetry mode is Python-only today.

`SDK.md` documents `CodeEvaluator.context_cache_hit_rate(..., fail_on_missing_telemetry=True)`, but the new CLI and remote-function paths only expose `min_hit_rate` through `--threshold` / `threshold`:

- `src/bigquery_agent_analytics/cli.py:233-235`
- `deploy/remote_function/dispatch.py:146-159`

That means the operational surfaces most likely used for CI gates cannot opt into strict missing-telemetry behavior. Old/non-cache-aware traces will pass silently from CLI/RF even when a user wants to enforce that cache telemetry exists.

Suggested fix: either add a CLI option / RF param such as `fail_on_missing_telemetry`, or explicitly document that strict missing-telemetry mode is Python-only for now.

**Low**

2. The remote-function implementation-plan pseudocode now has a bad default for this metric.

`docs/implementation_plan_remote_function.md:390-408` still shows:

```python
threshold = params.get("threshold", 5000)
...
"context_cache_hit_rate": lambda t: (
    CodeEvaluator.context_cache_hit_rate(min_hit_rate=t)
),
```

If copied literally, `context_cache_hit_rate` defaults to `min_hit_rate=5000`, so every telemetry-bearing session fails. The actual `deploy/remote_function/dispatch.py` implementation avoids this by using the SDK factory default when threshold is absent, but the doc block should be aligned to the real dispatch logic.

No new blocker found beyond the existing review; these are surface/documentation fixes.

sunpan9209:
Thanks for the detailed review. I pushed a follow-up commit (`0fc7cc9`) after rebasing on the latest upstream `main`.

Decision notes:

- Strict missing-telemetry mode: added operational support instead of leaving it Python-only. CLI now has `--fail-on-missing-cache-telemetry`, and the remote function accepts `fail_on_missing_telemetry=true`.
- Remote-function docs default: fixed the implementation-plan pseudocode so `context_cache_hit_rate` no longer inherits a copied `5000` threshold default.
- Aggregate semantics: kept the current raw-rate metric semantics rather than changing this evaluator to a binary pass-rate score. That preserves the API shape and keeps the true observed cache hit rate available in the score/details. I documented that `aggregate_scores["context_cache_hit_rate"]` is an average metric score, with cache-blind sessions contributing `1.0` by default because they pass as unknown.
- CLI fail line direction: added `cache_state` to the failure line so cold/partial/no-telemetry cases are readable in CI logs even though `budget` means a minimum required rate for this metric.
- Partial telemetry dilution: documented the mixed-pipeline caveat instead of changing the denominator to cache-aware input tokens in this PR. That keeps the issue #32 metric definition as `cached_tokens / input_tokens`; a `cache_aware_input_tokens` denominator would be a separate metric/behavior decision.
- Debuggability: added `exc_info=True` for swallowed `observed_fn` / `detail_fn` failures.
- Edge cases: added coverage for cached tokens above input tokens, non-numeric token values, the legacy `cached_tokens`-without-telemetry-count path, and invalid cold/warm threshold configuration.

I intentionally did not refactor the duplicated SQL JSON path list or memoize `_rate` in this PR; those are worthwhile cleanup items, but they are not correctness blockers and would add more churn around the SQL template.

Validation after the follow-up:

- `python3 -m isort --check-only src/bigquery_agent_analytics/evaluators.py src/bigquery_agent_analytics/cli.py deploy/remote_function/dispatch.py tests/test_cli.py tests/test_remote_function.py tests/test_sdk_evaluators.py`
- `python3 -m pyink --check src/bigquery_agent_analytics/evaluators.py src/bigquery_agent_analytics/cli.py deploy/remote_function/dispatch.py tests/test_cli.py tests/test_remote_function.py tests/test_sdk_evaluators.py`
- `python3 -m py_compile src/bigquery_agent_analytics/evaluators.py src/bigquery_agent_analytics/cli.py deploy/remote_function/dispatch.py`
- `git diff --check upstream/main..HEAD`
- focused pytest: `47 passed`

caohy1988 (approved):
## Fresh review — PR #114 head `0fc7cc9`

Re-reviewed top-to-bottom with my prior findings cleared from memory. Six of my eight prior items are addressed; two were knowingly deferred as nits. One new minor finding (dead branches) below. CI is fully green and lint is clean locally.

### Verified locally on `0fc7cc9`
- `pyink --check`: clean (6 files unchanged).
- `isort --check-only`: clean.
- Focused tests: **51 passed** (8 new tests added since the previous head, all green).
- CI rollup: Format check, Test 3.10/3.11/3.12/3.13/3.14, Build package, scan-pr, cla — all SUCCESS.

### Status of the prior review (commit `0cb4d5d` → `0fc7cc9`)

| ID | Prior finding | Status | Evidence |
|---|---|---|---|
| **A** | Aggregate-score semantics differ from binary prebuilts; risk of inflated dashboard numbers when `fail_on_missing_telemetry=False` | ✅ Documented + opt-in fix | SDK.md 144-159 and design.md 425-440 explicitly call out "average metric score rather than fraction of sessions that passed" and the cache-blind-contributes-1.0 dilution. Strict mode is now first-class via `--fail-on-missing-cache-telemetry` (CLI) and `fail_on_missing_telemetry` (remote function dispatch) |
| **B** | FAIL line direction reversed vs other prebuilts | ✅ Fixed | `cli.py:559-561` appends `cache_state={state}` to the FAIL line whenever the metric reports one. New regression test `test_emit_evaluate_failures_includes_cache_state` (test_cli.py:620-651) pins it. So a cold-start failure now reads `... observed=0.05 budget=0.5 cache_state=cold_start score=0.05 threshold=0.5` — the direction ambiguity is resolved by the qualitative state token |
| **C** | `detail_fn` exception swallow lacked traceback | ✅ Fixed (with bonus) | `evaluators.py:296-300` adds `exc_info=True` to `detail_fn`, **and also** to the pre-existing `observed_fn` swallow at `evaluators.py:283-285` that I flagged as a separate follow-up |
| **D** | SQL aggregation dilutes rate when telemetry is partial across LLM events | ✅ Documented (not refactored) | SDK.md 152-159 and design.md 437-440 explicitly note "the rate is also diluted by input tokens from LLM events that do not report cache metadata." Acceptable per the original "Low" severity |
| **E** | 12 JSON paths duplicated between SUM and COUNTIF — drift hazard | ❌ Not addressed | Acceptable — this was a Low/nit, future-proofing concern only |
| **F** | Missing test cases (clamp, non-numeric, legacy) | ✅ Fixed | New tests in test_sdk_evaluators.py: `test_cached_tokens_clamps_above_input_tokens` (line 667), `test_non_numeric_cached_tokens_fall_back_to_zero` (line 683), `test_legacy_cached_tokens_without_telemetry_count_is_observed` (line 699) |
| **G** | `cold_start_rate`/`warm_rate` invariant unvalidated | ✅ Fixed | `evaluators.py:548-552` raises `ValueError` if `not 0.0 <= cold_start_rate < warm_rate <= 1.0`, with a regression test `test_invalid_cache_state_thresholds_raise` (test_sdk_evaluators.py:714) |
| **H** | `_rate(s)` recomputed 3× per session | ❌ Not addressed | Acceptable — this was a Nit |

### One new finding (Low — code hygiene)

**Dead branches in evaluator factory registries.** Both `cli.py:233-236` (`_CODE_EVALUATORS["context_cache_hit_rate"]`) and `dispatch.py:155-157` + `168` (`factories_with_t["context_cache_hit_rate"]` and `factories_default["context_cache_hit_rate"]`) are now unreachable, because both call sites special-case the metric *before* dispatching through the dict:

- `cli.py:431-437` constructs `CodeEvaluator.context_cache_hit_rate(...)` directly with `kwargs`, then returns. The `else: entry = _CODE_EVALUATORS.get(evaluator)` branch on line 438 is never reached for this metric.
- `dispatch.py:179-183` does the same.

Two ways to resolve:
1. **Remove** the `context_cache_hit_rate` entries from `_CODE_EVALUATORS` and the two `factories_*` dicts — they're unreachable, and keeping dead entries invites a future contributor to "fix" the special-case by routing it through the registry, which silently drops `fail_on_missing_telemetry`.
2. **Keep** them but add a comment at the dict entry pointing to the special-case branch, so the next reader knows why the entry exists.

(1) is cleaner; (2) is fine if you'd rather leave the registry shape uniform.

### Other observations from the fresh pass

- **`_bool_param`** (`dispatch.py:120-127`) handles bool/str/int/None correctly, including `"True"` / `"TRUE"` via the explicit lowercase. Solid for JSON-from-BQ-remote-function input shapes.
- **CLI orphan-flag check** (`cli.py:404-413`) — passing `--fail-on-missing-cache-telemetry` with a non-cache evaluator exits 2 with a clear error, regression test at `test_cli.py:947-960`. Good defensive UX.
- **The new docs blocks in SDK.md and design.md are particularly well-written** — they don't just describe the metric, they explicitly flag the two semantic gotchas (aggregate-score interpretation, partial-telemetry dilution) so dashboard authors understand what they're consuming. This is exactly the right level of documentation honesty for a metric whose math diverges from sibling prebuilts.
- **Plugin schema alignment is still solid** — `attributes.usage_metadata.cached_content_token_count` (the first COALESCE entry) matches what the live ADK `BigQueryAgentAnalyticsPlugin` writes today (verified previously).

### Verdict: **APPROVE.**

All meaningful concerns from the prior review are resolved either with code fixes (B, C, F, G) or honest documentation of trade-offs (A, D). The one new finding — dead registry branches — is a polish item that doesn't affect behavior. The two deferred items (E duplicated SQL paths, H recompute) are genuine nits and reasonable to defer.

Ready to merge from a code-review standpoint. The dead-branch cleanup can be a one-line follow-up if you'd rather not roll another commit on this PR.

src/bigquery_agent_analytics/cli.py
```diff
@@ -230,6 +230,10 @@ def _load_spec_from_args(
         lambda t: CodeEvaluator.token_efficiency(max_tokens=int(t)),
         lambda: CodeEvaluator.token_efficiency(),
     ),
+    "context_cache_hit_rate": (
+        lambda t: CodeEvaluator.context_cache_hit_rate(min_hit_rate=t),
+        lambda: CodeEvaluator.context_cache_hit_rate(),
+    ),
```
caohy1988: **Dead entry.** This `_CODE_EVALUATORS["context_cache_hit_rate"]` registration is unreachable: the `evaluate` command special-cases the metric at lines 431-437 and constructs the evaluator directly, then returns. The `else: entry = _CODE_EVALUATORS.get(evaluator)` branch at line 438+ is never taken for this metric.

Keeping the dead entry is a footgun for the next contributor — if someone routes `context_cache_hit_rate` through the dict path "to clean up the special case," they'll silently drop the `fail_on_missing_telemetry` kwarg with no test failure.

Either remove this entry (cleanest) or add a one-line comment pointing to the special-case branch.

deploy/remote_function/dispatch.py
```diff
@@ -153,13 +168,20 @@ def build_evaluator(params):
       "error_rate": CodeEvaluator.error_rate,
```
caohy1988, line 165: Same dead-branch concern as `cli.py:233-236`: this `factories_default["context_cache_hit_rate"]` and the lambda in `factories_with_t` at lines 155-157 are both unreachable since the special-case at line 179-183 returns first. Same fix options apply.

docs/design.md
```diff
@@ -421,9 +421,24 @@ value strictly exceeds the budget.
...
+`cache_state="no_cache_telemetry"` and pass by default unless
+`fail_on_missing_telemetry=True` is set.
+
+Unlike the other prebuilt gates, `aggregate_scores` for this metric is
+the average metric score rather than the fraction of sessions that
+passed. That score is the observed cache hit rate for telemetry-bearing
+sessions, while cache-blind sessions contribute `1.0` by default because
+they pass as unknown. In mixed telemetry pipelines, the rate is also
+diluted by input tokens from LLM events that do not report cache
+metadata.
```
caohy1988, line 440: This block is exactly the right level of disclosure for a metric whose math diverges from sibling prebuilts. Future contributors looking at why `aggregate_scores["context_cache_hit_rate"]` reads as a fleet-wide rate rather than a pass rate will see this comment immediately. Nothing to change here — calling it out as the right pattern to repeat for any future non-binary metric.

caohy1988 (changes requested):
## Amended findings — please address before merge

Following a closer pass on the validation surface, two more items surfaced. The first is a **gate-correctness** issue that should land in this PR rather than as follow-up polish; the second is a doc-vs-code drift. My prior approval stands on the rest of the work, but I'm flipping to request-changes on these two.

### M-1 (Medium — gate correctness): `min_hit_rate` is unvalidated and unnormalized

`src/bigquery_agent_analytics/evaluators.py:522` accepts `min_hit_rate: float = 0.5` but does not range-check or coerce it. Only `cold_start_rate` / `warm_rate` are validated at `evaluators.py:550-554`. Three corner cases this lets through, all silently:

1. **`min_hit_rate=-0.1`** → because `_score()` returns a value clamped to `[0, 1]` and the pass test is `score >= metric.threshold` at `evaluators.py:264`, every telemetry-bearing session passes trivially. The "minimum" gate is effectively disabled.

2. **`min_hit_rate=1.1`** → `score >= 1.1` is impossible (the score is clamped to 1.0 maximum), so even a session with 100% cache reuse fails. The gate becomes a black hole.

3. **`min_hit_rate="0.5"` (string)** → reaches the comparator at `evaluators.py:264`, raises `TypeError: '>=' not supported between instances of 'float' and 'str'`, gets swallowed by the broad-except at `evaluators.py:267-271`, and the metric silently records `score=0.0, passed=False` for every session — the only signal is a `logger.warning` line that operators won't see in normal CLI output. Remote-function callers passing JSON strings (which is allowed by the `params` interface in `dispatch.py`) hit this path.

Suggested fix at the top of `context_cache_hit_rate(...)`, alongside the existing range check:

```python
try:
    min_hit_rate = float(min_hit_rate)
except (TypeError, ValueError) as exc:
    raise ValueError(
        f"min_hit_rate must be a number, got {min_hit_rate!r}"
    ) from exc
if not 0.0 <= min_hit_rate <= 1.0:
    raise ValueError(
        "min_hit_rate must satisfy 0 <= min_hit_rate <= 1, "
        f"got {min_hit_rate}"
    )
```

Plus three regression tests in `tests/test_sdk_evaluators.py::TestContextCacheHitRatePrebuilt`:

```python
def test_invalid_min_hit_rate_negative_raises(self):
    with pytest.raises(ValueError, match="min_hit_rate"):
        CodeEvaluator.context_cache_hit_rate(min_hit_rate=-0.1)

def test_invalid_min_hit_rate_above_one_raises(self):
    with pytest.raises(ValueError, match="min_hit_rate"):
        CodeEvaluator.context_cache_hit_rate(min_hit_rate=1.1)

def test_string_min_hit_rate_is_coerced_or_rejected(self):
    # Either coerce (preferred — remote function passes JSON values that
    # arrive as strings) and accept, or raise a clear ValueError. Silent
    # failure with score=0 for every session is the failure mode to avoid.
    evaluator = CodeEvaluator.context_cache_hit_rate(min_hit_rate="0.5")
    score = evaluator.evaluate_session({
        "session_id": "s1", "input_tokens": 1000,
        "cached_tokens": 950, "cache_telemetry_events": 1,
    })
    assert score.passed is True
```

If the remote function path is the only place strings can show up, alternatively coerce at the boundary in `dispatch.py:_build_evaluator` rather than inside `context_cache_hit_rate(...)` — but coercing at the SDK boundary is more defensive and matches how `_bool_param` already handles `fail_on_missing_telemetry`.

This is the one item I'd not defer. The other six prebuilts don't have this exposure (their thresholds aren't compared against a clamped `[0,1]` score), so it's a context_cache_hit_rate-specific gap.

### L-1 (Low): doc snippet diverges from real boolean parsing

`docs/implementation_plan_remote_function.md:393-395` shows:

```python
fail_on_missing_telemetry = params.get(
    "fail_on_missing_telemetry", False
)
```

But the real implementation in `deploy/remote_function/dispatch.py:142-144` runs the value through `_bool_param()` (defined at `dispatch.py:120-127`), which correctly handles JSON-string `"false"` / `"0"` / `"no"` / `"off"` as falsy.

If a developer reads the implementation plan and copies that snippet into their own dispatch wiring (which is exactly what the doc invites), `"false"` becomes truthy via Python's `bool("false") == True` and missing-telemetry sessions silently start failing. Either:

- Mirror `_bool_param()` in the doc snippet, or
- Add a one-line note: "Treat the value as a JSON boolean; if your transport coerces booleans to strings, parse via `_bool_param()` (see `dispatch.py:120`)."

### Verification on `0fc7cc9`
- pytest focused suite (51 tests): pass — but does not cover the three corner cases above.
- `pyink --check`, `isort --check-only`: clean.
- All 9 CI checks SUCCESS.

### Action requested

@sunpan9209 — please add the `min_hit_rate` validation + tests in this PR rather than as a follow-up. The L-1 doc fix is also worth landing here so the implementation plan stays consistent with `dispatch.py`. Once those are in, I'm happy to re-approve.

The prior approval covered the metric semantics, FAIL-line UX, threshold validation for cold/warm bounds, and the test coverage for the documented happy paths — those all still look right. M-1 is the one correctness gap that wasn't on the prior pass.

src/bigquery_agent_analytics/evaluators.py
```diff
@@ -500,6 +518,105 @@ def _score(s: dict[str, Any]) -> float:
     )
     return evaluator
 
+  @staticmethod
+  def context_cache_hit_rate(
+      min_hit_rate: float = 0.5,
```
caohy1988, line 523: **Gate-correctness gap.** `min_hit_rate` is unvalidated and uncoerced, while `_score()` returns a value clamped to `[0, 1]`. Three silent failure modes:

- `min_hit_rate=-0.1` → trivially passes every telemetry-bearing session (gate disabled).
- `min_hit_rate=1.1` → every session fails, even 100% cache reuse (gate is a black hole).
- `min_hit_rate="0.5"` (string) → raises TypeError inside the broad-except at line 267, silently records `score=0.0, passed=False` for every session. Remote-function callers passing JSON values can hit this.

Fix here, alongside the existing 0..1 check at line 550:

```python
try:
    min_hit_rate = float(min_hit_rate)
except (TypeError, ValueError) as exc:
    raise ValueError(f"min_hit_rate must be a number, got {min_hit_rate!r}") from exc
if not 0.0 <= min_hit_rate <= 1.0:
    raise ValueError("min_hit_rate must satisfy 0 <= min_hit_rate <= 1")
```

Please add this in this PR, plus three regression tests (`-0.1` raises, `1.1` raises, `"0.5"` either coerces or raises clearly). The other six prebuilts don't have this exposure because their thresholds aren't compared against a clamped `[0,1]` score — so it's context_cache_hit_rate-specific.

docs/implementation_plan_remote_function.md
```diff
@@ -387,21 +389,42 @@ def _dispatch(client, operation, params):
 def _build_evaluator(params):
     """Build CodeEvaluator from params dict."""
     metric = params.get("metric", "latency")
-    threshold = params.get("threshold", 5000)
+    threshold = params.get("threshold")
+    fail_on_missing_telemetry = params.get(
+        "fail_on_missing_telemetry", False
+    )
```
caohy1988, line 404: **Drift vs real implementation.** This snippet uses raw `params.get("fail_on_missing_telemetry", False)`, but the actual code at `deploy/remote_function/dispatch.py:142-144` runs the value through `_bool_param()` (`dispatch.py:120-127`) so JSON-string `"false"` / `"0"` / `"no"` are parsed correctly.

A developer copying this snippet into their own wiring will have `"false"` become truthy via `bool("false") == True` and missing-telemetry sessions will silently start failing. Either mirror `_bool_param()` here or add: "Treat as a JSON boolean; if your transport coerces booleans to strings, parse via `_bool_param()` (see `dispatch.py:120`)."

sunpan9209:
Follow-up for the latest requested-changes review landed in `453cc49`.

Addressed items:

- **M-1: `min_hit_rate` validation/coercion**
  - `CodeEvaluator.context_cache_hit_rate()` now coerces `min_hit_rate` to `float` at the SDK boundary.
  - It raises a clear `ValueError` unless `0 <= min_hit_rate <= 1`.
  - Added regression coverage for `-0.1`, `1.1`, and string `"0.5"` inputs. The string case now works as `0.5`, which keeps CLI/RF JSON-string inputs from falling into the previous silent-failure path.

- **L-1: remote-function docs boolean parsing drift**
  - Updated `docs/implementation_plan_remote_function.md` to mirror `_bool_param()` so copied snippets handle string booleans like `"false"`, `"0"`, `"no"`, and `"off"` correctly.

- **Dead registry entries**
  - Removed the unreachable `context_cache_hit_rate` entries from the CLI and remote-function evaluator registries. The metric stays special-cased because it needs `fail_on_missing_telemetry` in addition to the threshold/min-hit-rate value.

Validation on the rebased branch:

- `python3 -m isort --check-only src/bigquery_agent_analytics/evaluators.py src/bigquery_agent_analytics/cli.py deploy/remote_function/dispatch.py tests/test_cli.py tests/test_remote_function.py tests/test_sdk_evaluators.py`
- `python3 -m pyink --check src/bigquery_agent_analytics/evaluators.py src/bigquery_agent_analytics/cli.py deploy/remote_function/dispatch.py tests/test_cli.py tests/test_remote_function.py tests/test_sdk_evaluators.py`
- `python3 -m py_compile src/bigquery_agent_analytics/evaluators.py src/bigquery_agent_analytics/cli.py deploy/remote_function/dispatch.py`
- `git diff --check upstream/main..HEAD`
- Focused pytest: `50 passed`

PR branch is rebased on latest upstream `main` (`36c9acc`) and remains mergeable.

PR #117, opened 2026-05-05
feat: extractor-compilation scaffolding (#75 PR 4b.1)

caohy1988:
Implements **PR 4b.1** in the [#96 working plan](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/96#issuecomment-4363301699) — the deterministic compile-time contract layer the LLM-driven template fill (PR 4b.2) will plug into. Builds on PR 4a's runtime-target decision in [`docs/extractor_compilation_runtime_target.md`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/docs/extractor_compilation_runtime_target.md).

**No LLM call in this PR.** The compile pipeline is hand-fillable for now; PR 4b.2 will replace the hand-authored fixture with LLM-emitted source that has to clear the same gates.

## What's in this PR

New package `bigquery_agent_analytics.extractor_compilation` with:

- **`compute_fingerprint(...)`** — sha256 over the #75 input tuple. Caller-side allowlist order is irrelevant; every named field is hashed.
- **`Manifest`** — bundle provenance dataclass with `to_json` / `from_json`.
- **`validate_source(source) -> AstReport`** — allowlist-based AST safety check. Per-module symbol allowlist; no plain `import x`; no wildcards; no dunder aliases; no decorators; no non-constant defaults; no async/generators/classes/global/nonlocal; no halt/escape constructs (`while`/`raise`/`try`/`with`/`match`); no unbounded `for`/comprehension iterables; **call-target allowlist** for `Call(func=Name)` and **method-name allowlist** for `Call(func=Attribute)`; rejects non-Name/Attribute call targets (lambda, chained, conditional callables) and `ast.Lambda` outright; **rejects shadowing of allowlisted call targets** (no `len = event.get('cb'); len()` smuggling). Stable failure codes: `syntax_error`, `disallowed_import`, `disallowed_name`, `disallowed_attribute`, `disallowed_async`, `disallowed_generator`, `disallowed_class`, `disallowed_scope`, `disallowed_decorator`, `disallowed_default`, `disallowed_while`, `disallowed_for_iter`, `disallowed_raise`, `disallowed_try`, `disallowed_with`, `disallowed_match`, `disallowed_call`, `disallowed_method`, `disallowed_lambda`, `disallowed_shadowing`, `top_level_side_effect`.
- **`run_smoke_test(extractor, ...) -> SmokeTestReport`** — in-process runner. Catches `BaseException`, rejects wrong return types, rejects malformed `StructuredExtractionResult` internals, requires at least one non-empty result by default, gates on the #76 validator. Tracks `nonempty_event_types` so compile-level callers can verify per-event-type coverage.
- **`run_smoke_test_in_subprocess(source_path, ...) -> SmokeTestReport`** — subprocess-isolated runner with `RLIMIT_AS` (POSIX best-effort) + `subprocess.run(timeout=N)`. Returns a report on every path: timeout → `TimeoutError` per event, unpicklable inputs → `PickleError` per event, child crash → `SubprocessFailure` per event. Rejects negative `memory_limit_mb` with `ValueError`.
- **`compile_extractor(...) -> CompileResult`** — end-to-end pipeline. Stages: identifier validation (path-traversal + Python keyword); **`event_types` validation** (non-empty + non-string-rejected + duplicate-rejected + matching sample coverage); fingerprint; cache-hit candidate (with smoke re-run against current inputs; corrupt manifests + missing source fall through); AST gate; staged write; subprocess smoke gate (`isolation=True` default); **declared event_types must produce non-empty smoke output**; staged-replace on success.

Local bundle layout:

```
<parent>/
└── <fingerprint>/
    ├── <module_name>.py
    └── manifest.json
```

Runtime discovery and BQ-table mirror remain explicit C2 concerns per the runtime-target RFC.

## Hand-authored fixture

`tests/fixtures_extractor_compilation/bka_decision_template.py` ships a Python source string equivalent to `extract_bka_decision_event`. The end-to-end test compiles this fixture, asserts every gate passes (including subprocess smoke), re-loads the bundle from disk, and asserts the compiled callable's output matches `extract_bka_decision_event` on the same sample events.

## Out of scope

Per the runtime-target RFC and the PR 4b.1 sizing call:

- LLM/AI prompt generation — PR 4b.2.
- Runtime loader / orchestrator integration — C2.
- BQ-table bundle mirror — C2.
- Per-field / per-node / per-edge / per-event fallback wiring — C2.
- Multiple compiled extractor baselines — later in C1.

## Tests (135 cases in `tests/test_extractor_compilation.py`)

- **`TestFingerprint`** (10) — determinism, every named input hashed.
- **`TestManifest`** (3) — JSON round-trip, deterministic serialization, sorted keys.
- **`TestAstValidator`** (74) — every failure code; allowlists for imports, call targets, method names, for-iter shape, builtins; lambda / chained / conditional / `match` callable rejection; shadowing across all binding shapes (assignment / AugAssign / AnnAssign / for-target / comprehension / walrus / function arg / nested def).
- **`TestSmokeTest`** (15) — empty list / negative floor / `SystemExit` / validator failures / wrong return type / malformed result internals / all-empty fail / `=0` opt-out / clean run / subprocess timeout / subprocess pickle failure / negative memory limit / subprocess happy path.
- **`TestCompileExtractor`** (33) — end-to-end BKA compile through subprocess; equivalence with hand-written extractor; AST/smoke failure leaves nothing on disk; cache hit / re-smoke against current inputs / source-module-event_types divergence misses; corrupt-manifest cache miss; cache-hit import failure fall-through; failed-recompile preservation; **empty / non-string / duplicate / mismatched event_types rejected**; **malformed sample event_type values rejected**; **declared event_types must demonstrate non-empty coverage**; invalid identifiers (path-traversal + Python keywords).

## Verification at `1813379`

- Focused suite: `tests/test_extractor_compilation.py` 135 passed.
- Full repo suite: **2355 passed, 9 skipped**.
- Autoformat clean.
- `git diff --check upstream/main...HEAD` clean.

## Working plan position

```
✅ #76  — extracted-graph validator (merged in #113)
✅ #75 PR 4a — runtime-target RFC (merged in #115)
🔄 #75 PR 4b.1 — compile scaffolding (this PR)
   #75 PR 4b.2 — LLM-driven template fill
   #75 PR 4c — compile extract_bka_decision_event + measurement report
   #75 C2 (gated) — orchestrator integration, runtime discovery, fallback wiring
```


caohy1988:
Folded all six findings in `24dcb3a`.

**P1 — AST allowlist tightened** (`ast_validator.py`)

Switched from a module-level allowlist to a **per-module symbol allowlist**: `__future__: {annotations}`, `extracted_models: {ExtractedNode, ExtractedEdge, ExtractedProperty}`, `structured_extraction: {StructuredExtractionResult, StructuredExtractor}`. Smuggling `__builtins__` through a valid module now fails. Wildcards rejected. Aliases starting with `_` rejected. Decorators rejected. Function default arguments must be primitive constants — closes the `def f(x=open(...).read()): ...` import-time-side-effect smuggling vector.

**P1 — Halt/escape constructs rejected**

New failure codes: `disallowed_while`, `disallowed_raise`, `disallowed_try`, `disallowed_with`, `disallowed_decorator`, `disallowed_default`. Forbidden-name list extended with `breakpoint`, `exit`, `quit`, `__build_class__`. Smoke runner now catches `BaseException`, so `SystemExit` surfaces in the report instead of escaping. `raise SystemExit` now fails at three layers (AST `disallowed_raise` → AST `disallowed_name` on `SystemExit` if it ever got past → smoke runner's `BaseException` catch).

**P1 — Atomic replace + cache hit**

`compile_extractor` writes through a sibling staging directory (`<parent>/.staging-<fp12>-<rand>/`) and atomically replaces the target on success. A failed re-compile leaves any pre-existing valid bundle byte-identical. New: a second `compile_extractor` on identical inputs reads the existing manifest, validates fingerprint + function_name match, and returns `CompileResult(cache_hit=True)` without re-running any gates or re-writing anything. `created_at` is therefore preserved across consecutive compile calls.

**P1 — module_name path-traversal**

Both `module_name` and `function_name` are validated against `str.isidentifier()` up front. `../x`, `foo.bar`, `foo-bar`, `1leading`, empty, and whitespace-containing names fail with `CompileResult.invalid_identifier` before the harness touches the filesystem. Parametrized test confirms no writes happen on rejection.

**P2 — Smoke can no longer pass vacuously**

Wrong return type now fails (`wrong_return_types` field on the report). Empty-result-on-every-event fails by default — new `min_nonempty_results: int = 1` parameter requires at least one non-empty result; callers can opt out with `min_nonempty_results=0` for tests that deliberately exercise the empty path.

**P2 — Byte-stability wording**

Doc + manifest docstring now say: the **cache-hit short-circuit** is what makes the on-disk bundle byte-identical between consecutive compile calls — the manifest's `sort_keys` serialization is deterministic for given fields, and consecutive identical-input compiles don't write (so `created_at` stays the same).

**Tests**: 26 new (14 AST, 4 smoke, 3 compile + parametrized identifier). 64 total in `tests/test_extractor_compilation.py` (was 38). Full repo suite: **2284 passed / 9 skipped**. Autoformat clean.

Verified at `24dcb3a`.

caohy1988:
Folded all four findings in `48d0a10`.

**P1 — Cache hit ignored the actual compile request**

Confirmed the bug: the fingerprint covers the #75 input tuple but NOT `source`, `module_name`, or `event_types`, so a second call with different/broken source could silently return `ok=True, cache_hit=True`. Fixed: `_read_cached_manifest` now requires every one of fingerprint, function_name, `module_filename == f"{module_name}.py"`, `event_types == tuple(event_types)`, AND on-disk source bytes equal to the new `source` argument. Any mismatch falls through to a real compile.

Three new tests cover the divergence cases:
- `test_cache_hit_misses_when_source_differs`
- `test_cache_hit_misses_when_module_name_differs`
- `test_cache_hit_misses_when_event_types_differ`

**P1 — Smoke runner could hang through allowed for loops**

Two layers of defense: `iter` is now in `_FORBIDDEN_NAMES` (blocks `iter(int, 1)` by name); `for` and comprehension iterables that are a Call to a bare Name fail with new `disallowed_for_iter` (blocks `range(...)`, user-defined helpers, etc.). Method calls (`dict.items()`), literal tuples, parameter names, attribute access, and subscripts still pass — so the BKA fixture's `for key in ('outcome', 'confidence', ...):` still validates clean.

Six new tests (parametrized + named):
- `test_iter_name_rejected`
- `test_for_iter_name_call_rejected` (parametrized over `range(10)`, `range(10**100)`, `iter(int, 1)`, `some_helper()`)
- `test_for_iter_tuple_literal_accepted`
- `test_for_iter_method_call_accepted`
- `test_comprehension_with_name_call_iter_rejected`

**P2 — Doc said `while` was accepted control flow**

Updated `docs/extractor_compilation_scaffolding.md`: "Accepted" bullet now reads "if / **bounded** for / comprehensions" with a paragraph on what counts as bounded, and the failure-code table gains a `disallowed_for_iter` row. CHANGELOG codes list refreshed.

**P2 — PR body stale**

Refreshed at `48d0a10`: 75 unit tests (was 64), full repo 2295 passed / 9 skipped (was 2284), failure-code list now includes `disallowed_for_iter`, test breakdown reflects all the new cases.

Verified at `48d0a10`: focused suite 75 passed, full repo 2295 / 9 skipped, autoformat clean.

caohy1988:
Folded all five findings in `743e3ae`.

**P1 — Cache hit re-runs smoke against current inputs**

Fix: on cache match, `compile_extractor` imports the cached callable and runs `run_smoke_test(...)` against the *current* `sample_events` / `resolved_graph` / `min_nonempty_results`. Pass → `cache_hit=True` with fresh `smoke_report`, nothing rewritten on disk. Fail → surface the failure; the on-disk bundle is left intact (rewriting wouldn't help; source hasn't changed). New test exercises a stricter sample set against a previously-cached bundle and confirms the cache hit fails.

**P1 — Call-target allowlist**

Switched to a name-allowlist for `Call(func=Name)`. Allowed: the four constructors (`ExtractedNode`, `ExtractedEdge`, `ExtractedProperty`, `StructuredExtractionResult`), primitive constructors (`str`, `int`, `float`, `bool`, `bytes`), container constructors (`set`, `frozenset`, `dict`, `list`, `tuple`), and `isinstance`, `len`. Anything else fails as `disallowed_call`. Method-style calls (`Call(func=Attribute)`) unchanged — bounded by the receiver. Five parametrized rejection tests (`list(range(10))`, `sum([1,2,3])`, `max(...)`, `some_helper()`, `abs(-5)`); eight parametrized acceptance tests; method-calls-still-allowed regression.

**P2 — "Staged replace" wording**

Renamed `_atomic_replace` → `_staged_replace`. Docstring + scaffolding doc now say "staged replace" with a clear note that the bundle is reproducible from inputs (so a process crash between rmtree and rename costs the next compile a re-creation, no permanent damage). The failure-path guarantee — failed gates leave the existing bundle untouched — is unchanged and accurate.

**P2 — Reject Python keywords as identifiers**

`_is_python_identifier` now also calls `keyword.iskeyword(value)`. Parametrized test rejects `class`, `def`, `for`, `return` as `module_name`. Same rule applies to `function_name`.

**P2 — `min_nonempty_results=-1`**

`run_smoke_test` raises `ValueError` on negative values, message points at `=0` as the explicit opt-out.

**Status**: 95 unit tests in `tests/test_extractor_compilation.py` (was 75); full repo suite 2315 passed / 9 skipped (was 2295). Autoformat clean. Verified at `743e3ae`.

caohy1988:
Folded all four findings in `59716d7`.

**P1 — Subprocess isolation for the smoke gate**

New `subprocess_runner.py` module (invoked as `python -m bigquery_agent_analytics.extractor_compilation.subprocess_runner`). Reads pickled args from stdin, applies `RLIMIT_AS` (POSIX best-effort), imports source via `importlib`, runs per-event extraction, returns per-event outcomes via pickled stdout. New `run_smoke_test_in_subprocess` wrapper in `smoke_test.py` spawns it with `subprocess.run(timeout=N)` for the wallclock cap and runs #76 validation in-process so `ResolvedGraph` doesn't cross the process boundary. `compile_extractor` now defaults `isolation=True` and routes both the main path and the cache-hit re-smoke through the subprocess wrapper.

Verified on the reviewer's repro: `while True: pass` + `subprocess.run(timeout=1.0)` returns a `TimeoutError` exception per event. `bytes(10**100)` blocked at the AST gate (it's a `Call` to `bytes` which is allowed, but the runtime memory cap of 512 MB stops it; for nicer behavior we could add an arg-size check later).

**P1 — Method-name allowlist**

New `_ALLOWED_METHOD_NAMES = {get, items, keys, values, append}`. Anything else → `disallowed_method`. Five parametrized rejection tests (`event.clear()`, `event.repeat_forever()`, `event.pop()`, `event.update()`, `event.fromkeys()`) plus five acceptance tests for the allowed names.

**P2 — `TypeError` from malformed manifest**

`_read_cached_manifest` now also catches `TypeError` so `Manifest.from_json("42")` doesn't crash compile. New test writes `42` as `manifest.json` and confirms compile recovers via fresh compile.

**P2 — Cache-hit import failure falls through**

Restructured the cache-hit path: `None` from `_smoke_test_cached_bundle` (import failure) now falls through to fresh compile (matches the docstring); smoke fail returns `ok=False` (cached source rejects current inputs); smoke ok returns `cache_hit=True`. New test: write a valid bundle, delete the `.py`, recompile — confirms fall-through to a fresh compile.

**Status**: 109 unit tests in `tests/test_extractor_compilation.py` (was 95); full repo 2329 / 9 skipped (was 2315). Subprocess overhead adds ~60s to the focused suite. Autoformat clean. Verified at `59716d7`.

Note on the P1.1 caveat the reviewer raised — "static checks help, but they won't close this class fully": that's now explicit. The AST gate is a pre-filter, the subprocess (with timeout + RLIMIT_AS) is the runtime safety net. Anything that slips past static checks gets caught by the wallclock and memory caps in the child process.

caohy1988:
Folded all six findings in `b09ef50`.

**P1.1 — Subprocess raised on unpicklable spec/events.** Wrapped `pickle.dumps(...)` in a `BaseException` catch; returns a synthesized one-exception-per-event report whose exceptions start with `PickleError:`. Also added explicit `ValueError` on negative `memory_limit_mb` (the child's `setrlimit` was silently no-op'ing). Test: `spec=lambda x: x` now produces a clean failure report.

**P1.2 — Malformed `StructuredExtractionResult` internals.** New `_well_formed_result_error` helper validates each field's runtime type before aggregation. Both in-process and subprocess paths run it. `nodes=[{}]` and `fully_handled_span_ids=("s",)` now land in `wrong_return_types` instead of crashing `merge_extraction_results`. Two new tests cover the dict-node and tuple-span-ids cases.

**P1.3 — AST call allowlist missed non-Name/non-Attribute targets.** Two layers: `_check_call` now emits `disallowed_call` for any `func` that isn't `Name` or `Attribute` (lambda, chained, conditional, etc.); `ast.Lambda` is rejected outright via new `disallowed_lambda` code. Three new tests: lambda-call, chained-call, conditional-callable.

**P2.1 — `event_types` lying about samples.** New stage 1.5 in `compile_extractor`: empty `event_types` rejected; each declared type must have at least one matching sample. New `invalid_event_types` field on `CompileResult`. Two tests: empty list, mismatched coverage.

**P2.2 — Negative `memory_limit_mb` silently disabled cap.** Fixed alongside P1.1 — `ValueError` on negative.

**P3 — Doc test counts.** Updated to actual numbers: TestFingerprint=10, TestManifest=3, TestAstValidator=65, TestSmokeTest=15, TestCompileExtractor=25, **total 118**. Added `disallowed_lambda` row. CHANGELOG codes list refreshed.

**Status**: 118 unit tests in `tests/test_extractor_compilation.py` (was 109); full repo 2338 passed / 9 skipped (was 2329). Verified at `b09ef50`.

caohy1988:
Folded all three findings in `fafc3df`.

**P1.1 — Declared event_types must produce non-empty smoke output.**

`SmokeTestReport` gains `nonempty_event_types: tuple[str, ...]` populated by both the in-process and subprocess aggregators. `compile_extractor` post-smoke gates: every declared event_type must appear in the report. Subprocess pairs per-event entries with the original events list by index. The cache-hit path applies the same gate. New test reproduces the reviewer's exact case (declare `("x",)`, empty x sample + nonempty y sample) and confirms the failure shape.

**P1.2 — Shadowing of allowlisted call targets.**

New `disallowed_shadowing` AST code. Catches any local binding of a name in `_ALLOWED_CALL_TARGETS` via `Name(ctx=Store)`, `ast.arg`, `FunctionDef.name` (top-level and nested defs). Through `Name(Store)`, also catches walrus / for-target / comprehension target / AugAssign / AnnAssign. **Module-level imports of allowlisted constructors by their canonical names are unaffected** — those are the intended source, and the per-module symbol allowlist already controls what's importable from each module. Six parametrized tests cover Assign / AnnAssign / AugAssign / for-target / comprehension target / walrus; two more cover function-arg and nested-def shadowing.

**P2.1 — Malformed sample event_type values.**

`compile_extractor` now validates up front that every sample is a dict with a non-empty string `event_type`. Catches `event_type=1` / `None` / `""` before the missing-coverage check formats its error message (which used to crash on `sorted({1, "a"})`). New test: mixed int+str sample types produce a clear `invalid_event_types` and no partial bundle.

**Status**: 128 unit tests (was 118); full repo 2348 passed / 9 skipped (was 2338). Doc + CHANGELOG list `disallowed_shadowing`. Verified at `fafc3df`.

caohy1988:
Folded all three findings in `1813379`.

**P1.1 — `match` pattern bindings.** New `disallowed_match` AST code rejects `ast.Match` outright. The accepted control-flow set (if / bounded for / comprehensions per the doc) doesn't include `match` anyway, and pattern captures bind names through pattern nodes that the shadowing check can't see. Test reproduces the reviewer's exact bypass (`case {'x': len}` → `len()`).

**P1.2 — Direct `event_types` validation.** Stage 1.5 in `compile_extractor` now validates each declared type up front: non-empty string check (with offending index/value in the failure message) + duplicate rejection. Failures emit `invalid_event_types` before fingerprint computation. Six new tests: 5 parametrized non-string/empty/None rejections + 1 duplicate rejection.

**P3 — PR body refreshed.** Counts now read 135 unit tests / 2355 full repo / `1813379` verification; failure-code list includes `disallowed_match`; class breakdown shows the new rejection categories.

**Status**: 135 unit tests in `tests/test_extractor_compilation.py` (was 128); full repo 2355 / 9 skipped (was 2348). Doc + CHANGELOG list `disallowed_match`. Verified at `1813379`.

caohy1988:
Both polish edits applied in `d64ece9`.

1. **`ast_validator.py` module docstring** — added `match` to the halt/escape line and added distinct bullets for `disallowed_lambda`, `disallowed_call` (non-Name/Attribute targets), `disallowed_method` (method-name allowlist), and `disallowed_shadowing`. The detailed failure-code table in `docs/extractor_compilation_scaffolding.md` is still the source of truth; this just keeps the in-code summary aligned.

2. **`compiler.py` module docstring stage list** — added stage 2 (declared `event_types` validation: non-empty tuple of non-empty unique strings, sample event coverage) and stage 7 (post-smoke "declared types must produce non-empty output" gate). Numbers shifted accordingly so the staged-replace is now stage 9.

No code changes; doc-only diff. Focused suite: 135 passed. Full repo: 2355 / 9 skipped.

Verified at `d64ece9`.

PR #118, opened 2026-05-06
feat: deterministic source generator for compiled extractors (#75 PR 4b.2.1)

caohy1988:
Implements **PR 4b.2.1** in the [#96 working plan](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/96#issuecomment-4363301699) per the user-sharpened slice — the deterministic source generator the LLM-driven template fill (PR 4b.2.2) will plug into. Builds on PR 4b.1's compile scaffolding (`compile_extractor`, AST allowlist, subprocess smoke runner, #76 validator gate).

**No LLM call in this PR.** PR 4b.2.2 owns the LLM step that *resolves* a raw extraction-rule + event-schema pair into a `ResolvedExtractorPlan`; this module is the deterministic boundary the LLM output has to cross before any source generation happens.

## What's in this PR

New module `bigquery_agent_analytics.extractor_compilation.template_renderer` with:

- **`FieldMapping(property_name, source_path)`** — maps an ontology property to a path in the event payload. `source_path` is a tuple of dict keys; missing or wrong-shape intermediates resolve to "field absent" rather than raising (stepwise `isinstance(..., dict)` guards inside each `.get(...)`).
- **`SpanHandlingRule(span_id_path, partial_when_path)`** — optional. Matches the BKA pattern: `partial_when_path` truthy → span is marked partially handled; otherwise fully handled.
- **`ResolvedExtractorPlan`** — deliberately boring dataclass: `event_type`, `target_entity_name`, `function_name`, `key_field`, optional `property_fields`, `session_id_path`, optional `span_handling`. Every field-level decision is already made before the renderer runs.
- **`render_extractor_source(plan: ResolvedExtractorPlan) -> str`** — pure source generation. Strict plan validation up front: every string field is a non-empty string; `target_entity_name` and property names are Python-identifier-shaped (so they're safe to embed directly into generated source); path segments are non-empty strings; duplicate property names rejected; `function_name` is a Python identifier, not a keyword, not in the call-target allowlist.

## Generated-source contract

- Imports only the three symbols actually used (`ExtractedNode`, `ExtractedProperty`, `StructuredExtractionResult`). `ExtractedEdge` isn't imported since the renderer doesn't emit edges yet.
- Calls only allowlisted Names + allowlisted method names (`get`, `items`, `keys`, `values`, `append`).
- No shadowing of allowlisted call targets, no decorators, no non-constant defaults, no halt/escape constructs.
- **Top-of-function `event_type` guard** rejects events whose type doesn't match the plan's declared coverage — layered defense with C2's manifest-driven dispatch so a plan/manifest mismatch can't silently attach an extractor to the wrong event type.
- Returns well-formed `StructuredExtractionResult` (lists for nodes/edges, sets for span ids).
- **Deterministic**: identical plans render byte-identical source.

## BKA equivalence

The BKA-equivalent plan renders source whose runtime output is structurally identical to `extract_bka_decision_event` on the fixture's sample events (which all carry `event_type="bka_decision"`). End-to-end test compiles through `compile_extractor` (subprocess smoke + #76 validator), re-loads the bundle, and asserts equivalence on every BKA sample.

## Out of scope

Per the PR 4b.2.1 sizing call:

- LLM prompt/schema for resolving raw rules → `ResolvedExtractorPlan` — PR 4b.2.2.
- Retry-on-AST/smoke/validator failure — PR 4b.2.2.
- Diagnostics on retry — PR 4b.2.2.
- Edge extractors — only nodes for now.
- Composite property values — ontology v0 explicitly defers these.

## Tests (39 in `tests/test_extractor_compilation_template.py`)

- **`TestPlanValidation`** (27 cases) — valid BKA plan renders; parametrized rejection of bad function names; empty / non-string `event_type` / `target_entity_name` / `key_field`; duplicate `property_name`; `target_entity_name` with a quote rejected; non-identifier-shaped property names rejected (parametrized); non-string top-level fields and path segments rejected (parametrized).
- **`TestGeneratedSourceClearsGates`** (5 cases) — BKA source passes `validate_source`; compiles end-to-end via `compile_extractor`; rendered output matches `extract_bka_decision_event`; **wrong-event-type input returns empty**; render is deterministic.
- **`TestPlanShapeVariations`** (7 cases) — plan with no `property_fields`; plan with no `span_handling`; single-step paths; deep traversal path (length 3) with missing-intermediate negative case; deep optional property / `session_id_path` / `partial_when_path` with non-dict intermediate.

## Verification at `7310f04`

- Focused suite: `tests/test_extractor_compilation_template.py` 39 passed.
- Full repo suite: **2394 passed, 9 skipped**.
- Autoformat clean.
- `git diff --check upstream/main...HEAD` clean.

## Working plan position

```
✅ #76  — extracted-graph validator (merged in #113)
✅ #75 PR 4a — runtime-target RFC (merged in #115)
✅ #75 PR 4b.1 — compile scaffolding (merged in #117)
🔄 #75 PR 4b.2.1 — deterministic source generator (this PR)
   #75 PR 4b.2.2 — LLM-driven plan resolution + retry
   #75 PR 4c — compile extract_bka_decision_event + measurement report
   #75 C2 (gated) — orchestrator integration, runtime discovery, fallback wiring
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,12 +23,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ``ResolvedExtractorPlan`` dataclasses + ``render_extractor_source(plan)
   -> str``. The renderer is the deterministic boundary the LLM
   step in PR 4b.2.2 will plug into; **no LLM call lives here**.
-  Generated source matches ``extract_bka_decision_event``'s
-  runtime behavior on the BKA fixture's sample events, exercised
-  end-to-end by 25 unit tests covering plan validation, the AST
-  gate, the subprocess smoke runner, and plan-shape variations
-  (no property fields, no span handling, single-step paths,
-  deep traversal paths).
+  Generated source carries a top-of-function ``event_type``
+  guard that returns an empty result when the incoming event
+  doesn't match the plan's declared type, layered with the
+  orchestrator's manifest-driven dispatch so a plan/manifest
+  mismatch can't silently attach an extractor to the wrong
+  event type. Output otherwise matches
+  ``extract_bka_decision_event``'s runtime behavior on the BKA
+  fixture's sample events. Exercised end-to-end by 39 unit
+  tests covering plan validation, the AST gate, the subprocess
+  smoke runner, plan-shape variations (no property fields, no
+  span handling, single-step paths, deep traversal paths,
+  non-dict intermediates at every depth-3 traversal site), and
+  wrong-event-type rejection.
 - **`bq-agent-sdk binding-validate` CLI** — pre-flight validator that
   checks whether a binding YAML's referenced BigQuery tables
   physically exist with the columns and types the binding requires,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Deterministic source generator for compiled structured
+  extractors** in
+  `bigquery_agent_analytics.extractor_compilation.template_renderer`
+  and
+  [`docs/extractor_compilation_template_renderer.md`](docs/extractor_compilation_template_renderer.md).
+  Issue [#75](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/75)
+  PR 4b.2.1 — turns a pre-resolved
+  ``ResolvedExtractorPlan`` into a Python source string that 4b.1's
+  ``compile_extractor`` runs through every gate (AST allowlist,
+  smoke runner, #76 validator). Public surface:
+  ``FieldMapping`` / ``SpanHandlingRule`` /
+  ``ResolvedExtractorPlan`` dataclasses + ``render_extractor_source(plan)
+  -> str``. The renderer is the deterministic boundary the LLM
+  step in PR 4b.2.2 will plug into; **no LLM call lives here**.
+  Generated source matches ``extract_bka_decision_event``'s
+  runtime behavior on the BKA fixture's sample events, exercised
+  end-to-end by 25 unit tests covering plan validation, the AST
+  gate, the subprocess smoke runner, and plan-shape variations
+  (no property fields, no span handling, single-step paths,
+  deep traversal paths).
 - **`bq-agent-sdk binding-validate` CLI** — pre-flight validator that
   checks whether a binding YAML's referenced BigQuery tables
   physically exist with the columns and types the binding requires,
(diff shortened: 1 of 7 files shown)


caohy1988:
Folded all four findings in `6bdb908`.

**P1.1 — Deep optional paths now use stepwise guards.** Both `_render_optional_path_get` and `_render_optional_property` emit each `.get(...)` *inside* the previous step's `isinstance(..., dict)` block, so a non-dict intermediate at any depth resolves to "field absent" instead of crashing. Three new tests reproduce the reviewer's exact cases (optional property, session_id_path, partial_when_path with `event["a"] = "notadict"`).

**P1.2 — Plan-string source-injection vectors closed.** Two layers:
- Optional-property helpers now use **index-based variable names** (`_op0_0`, `_op0_1`, ...) instead of property-name-derived (`_op_<name>_0`). Defense-in-depth — even an invalid property name can't pollute generated source identifiers.
- `target_entity_name` and `key_field.property_name` (both embedded directly into the `node_id` f-string) are now validated as Python-identifier-shaped strings up front.

**P2 — Strict string-type validation at the renderer boundary.** New helpers `_require_nonempty_str`, `_require_safe_identifier`, `_require_string_path`. Every plan string field is validated as a non-empty string; identifier-shape required for fields that get embedded raw into source. `event_type=1`, `target_entity_name=42`, `FieldMapping(1, ...)` all raise `ValueError` up front instead of rendering broken source that fails later. Parametrized tests cover the main cases.

**P3 — Doc fix.** Both the module docstring and `docs/extractor_compilation_template_renderer.md` now say three imports (`ExtractedNode`, `ExtractedProperty`, `StructuredExtractionResult`); `ExtractedEdge` isn't imported since the renderer doesn't emit edges yet.

**Status**: 38 unit tests in `tests/test_extractor_compilation_template.py` (was 25); full repo suite 2393 / 9 skipped (was 2380). Verified at `6bdb908`.

caohy1988:
Both findings folded in `7310f04`.

**P2 — event_type guard.** The renderer now emits a top-of-function check as the first executable statement after the docstring:

```python
def f(event, spec):
  """..."""
  if event.get('event_type') != 'bka_decision':
    return StructuredExtractionResult()
  # ...
```

Layered defense with C2's manifest-driven dispatch — even if a plan/manifest mismatch routes the wrong event into this extractor, the extractor itself rejects it. Three deep-path tests adjusted to include `event_type` in the event dicts so they pass the new guard. New `test_wrong_event_type_returns_empty` exercises matching / mismatching / missing event_type cases.

**P3 — metadata refresh.**
- CHANGELOG entry now says 39 tests and mentions the event_type guard.
- Doc test breakdown reflects actual counts (TestPlanValidation=27, TestGeneratedSourceClearsGates=5, TestPlanShapeVariations=7, total=39) and includes wrong-event-type rejection.
- PR body refreshed at `7310f04`: 39 tests / 2394 / 9 skipped, three imports (no `ExtractedEdge`), event_type guard in the contract section.

39 unit tests in `tests/test_extractor_compilation_template.py` (was 38); full repo 2394 passed / 9 skipped (was 2393). Verified at `7310f04`.

PR #124, opened 2026-05-06
feat: LLM-driven plan resolver for compiled extractors (#75 PR 4b.2.2.b)

caohy1988:
Implements **PR 4b.2.2.b** in the [#96 working plan](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/96#issuecomment-4363301699) per the user-sharpened slice — the LLM-driven mapping step that produces JSON for 4b.2.2.a's parser.

**Adapter-free.** No `google-genai` import, no provider-specific glue. The resolver depends only on a minimal `LLMClient` Protocol; concrete adapters and retry orchestration land in PR 4b.2.2.c / PR 4c.

## What's in this PR

New module `bigquery_agent_analytics.extractor_compilation.plan_resolver` with:

- **`LLMClient`** — `typing.Protocol` (structural typing) with one method: `generate_json(prompt: str, schema: dict) -> dict`.
- **`build_resolution_prompt(rule, schema) -> str`** — pure function. Same inputs produce byte-identical output. Embeds the exported `RESOLVED_EXTRACTOR_PLAN_JSON_SCHEMA` and the input rule + schema verbatim. **Strict input contract**: `extraction_rule` and `event_schema` must each be a JSON-serializable `Mapping` at the root, with all mapping keys (recursively) being strings, no `NaN`/`Infinity` values (`allow_nan=False` so the prompt is RFC-8259-compliant). Wrong-shape inputs raise `TypeError` at the boundary with a path-aware message.
- **`PlanResolver(llm_client).resolve(rule, schema) -> ResolvedExtractorPlan`** — wires prompt → LLM call → parser. The schema is **deep-copied** before being handed to the client. `PlanParseError` and any exception the LLM client raises propagate unchanged.

## Out of scope

- **No retry logic** — PR 4b.2.2.c.
- **No provider adapters** — PR 4c.
- **No real LLM tests** — fakes only.

## Tests (35 in `tests/test_extractor_compilation_plan_resolver.py`)

- **`TestBuildResolutionPrompt`** (27) — determinism; dict insertion order; rule + schema grounding; output-contract embed; no-hallucinated-paths rule; identifier-safety rules with the call-target allowlist; non-JSON-serializable `extraction_rule`/`event_schema` (parametrized); wrong-root-type rejection (list/string/int/float/bool, both fields); **non-string mapping keys rejected recursively** (root/nested/inside-list, plus `event_schema`); **non-finite floats rejected** (NaN/Infinity/-Infinity).
[...]

PR #125, opened 2026-05-06
feat: diagnostic builders for retry-prompt feedback (#75 PR 4b.2.2.c.1)

caohy1988:
## Summary

Lands the deterministic, LLM-free **diagnostic builder** layer that converts compile-pipeline failures into compact, actionable strings suitable for embedding in a retry prompt.

This PR is intentionally narrow: **no LLM call, no retry loop, no orchestrator**. It locks the *feedback format* the retry orchestrator (PR 4b.2.2.c.2) will consume, so c.2 can focus on loop semantics rather than arguing about diagnostic wording.

Builds on: PR 4b.2.2.b (`PlanResolver` + parser, merged in #124).
Working plan reference: issue #96.

## What's in this PR

New module `src/bigquery_agent_analytics/extractor_compilation/diagnostics.py`, exposing five public functions:

- `build_plan_parse_diagnostic(error: PlanParseError) -> str`
- `build_ast_diagnostic(report: AstReport) -> str`
- `build_smoke_diagnostic(report: SmokeTestReport) -> str`
- `build_compile_result_diagnostic(result: CompileResult) -> str` — covers the top-level `CompileResult` envelope, including the three failure fields (`invalid_identifier`, `invalid_event_types`, `load_error`) that don't surface through any single gate's report. Falls through to the AST and smoke builders when those gates failed.
- `build_gate_diagnostic(kind, payload) -> str` — dispatcher for `"parse"` / `"ast"` / `"smoke"` / `"compile"`. The retry orchestrator should prefer `kind="compile"`: it has the full `CompileResult` in hand and shouldn't re-derive which gate fired.

Re-exported from `extractor_compilation` and the top-level `bigquery_agent_analytics` lazy-import package.

### Three contracts the output text clears

1. **Actionable.** Stable failure `code` from each gate plus the dotted `path` or source line, so the LLM can grep its own previous output for the offending field/line. Empty path → `<root>` so the LLM never sees blank anchors.
2. **Bounded.** Each section caps at the first ten failures with `... and N more (truncated)`. Tracebacks are reduced to the last informative line (`ExceptionType: message`) so a 30-line stack frame doesn't crowd out the actionable bit.
3. **Deterministic.** Same input report → byte-identical output. Order preservation: `ast.walk` order for AST, event-iteration order for smoke. No sorting that would obscure where the LLM should look.

The per-gate builders never raise — they always return a string, even on degenerate input. The dispatcher `build_gate_diagnostic` is the only function that does raise: `TypeError` for a payload type that doesn't match `kind`, `ValueError` for an unknown `kind`. Both are caller-misuse signals; the retry orchestrator should never see them in practice.

### Output shapes

- **Parser** (single line):
  `PlanParseError [code=missing_required] at extractors[0].entity_extractors[0].id: required field 'id' missing`
- **AST** (multi-line, walk order):
  ```
  AST validation failed (3 issues):
    line 12 col 8: forbidden_call: import statement not allowed
    line 18 col 4: forbidden_attr: __dunder__ access not allowed
    line 24: forbidden_construct: while loop not allowed
  ```
- **Smoke** (four optional sections — exceptions / wrong return types / non-empty floor / #76 graph validator failures, only present when populated):
  ```
  Smoke test failed:

    Per-event exceptions (2 of 5 events):
      - KeyError: 'session_id'
      - TypeError: NoneType is not subscriptable

    #76 graph validator failures (1):
      - [NODE] missing_required_field at events[0].nodes[0]: field 'agent_id' is required by ontology
  ```
- **CompileResult** envelope (single line for the three CompileResult-only fields; falls through to AST / smoke builders for those gates):
  ```
  CompileError [code=invalid_identifier]: <message>
  CompileError [code=invalid_event_types]: <message>
  CompileError [code=load_error]: <message>
  ```

  Check order matches the canonical `compile_extractor` pipeline order (identifier → event_types → AST → load → smoke), so the diagnostic names the **earliest** failed stage even on hand-built `CompileResult` values with multiple fields populated. `invalid_event_types` is the most common compile-level failure mode in retry feedback: parser passes, AST passes, but the LLM declared an `event_type` with no matching sample.

## Tests

34 cases in `tests/test_extractor_compilation_diagnostics.py`:

- `TestPlanParseDiagnostic` (3) — root path → `<root>`, simple path, dotted path.
- `TestAstDiagnostic` (6) — clean passthrough, line+col, line-only, no-line, walk-order preservation, truncation at 10.
- `TestSmokeDiagnostic` (8) — clean passthrough, traceback last-line reduction, each section in isolation, all four sections combined, validator-failure truncation, empty-traceback placeholder.
- `TestBuildGateDiagnostic` (5) — dispatch to each per-gate builder, unknown kind → `ValueError`, payload mismatch → `TypeError`.
- `TestCompileResultDiagnostic` (9) — `ok` passthrough; each of `invalid_identifier` / `invalid_event_types` / `load_error` rendered with the right code prefix; AST and smoke fall-through; AST wins over `load_error` when both populated (locks pipeline-stage ordering); `invalid_event_types` wins over a passing smoke report (post-coverage shape); `[code=unknown]` defensive fallback for the ok=False/no-field shape.
- `TestBuildGateDiagnosticCompileKind` (3) — dispatch via `kind="compile"`; non-`CompileResult` payload raises with `expects CompileResult`; unknown-kind error message advertises `'compile'` so misspellings are self-correcting.

All assertions are exact-string equality on hand-built reports — determinism is enforced byte-for-byte.

## Verification (head `b108465`)

- 34/34 focused tests pass
- Full repo suite: 2541 passed / 9 skipped
- Autoformat clean (pyink)

## Out of scope (deferred to PR 4b.2.2.c.2)

- Retry-prompt builder (`build_retry_prompt(diagnostic, prior_attempt, schema)`)
- Orchestrator (`compile_with_llm(rule, schema, llm_client, max_attempts)`)
- `RetryCompileResult` dataclass
- Per-attempt telemetry / log fields

## Test plan

- [x] Run focused diagnostics test module
- [x] Run full repo test suite
- [x] Verify lazy-import package surface exposes new builders
- [x] Verify byte-identical output for each builder shape (exact-string assertions)
- [x] Verify truncation at 10 entries with summary line
- [x] Verify traceback reduction handles empty-traceback edge case
- [x] Verify `build_compile_result_diagnostic` renders each `CompileResult` failure mode
- [x] Verify earliest-pipeline-stage ordering holds for hand-built CompileResult shapes


caohy1988:
Addressed in 4ad4ef7. The reviewer's P2 was right — `build_gate_diagnostic` covered `parse` / `ast` / `smoke` but had no path for the three `CompileResult`-only failure fields (`invalid_identifier`, `invalid_event_types`, `load_error`). The important one for the retry loop is `invalid_event_types`: parser and AST pass, but the LLM declared an `event_type` with no matching sample, and `compile_extractor` rejects the candidate before smoke runs. c.2 would have had to invent an ad-hoc format.

**What landed:**

1. `build_compile_result_diagnostic(result: CompileResult) -> str` — single entry point for the whole envelope. Check order mirrors the pipeline order so the message names the *earliest* failed stage:

   ```
   CompileError [code=invalid_identifier]: <message>
   CompileError [code=invalid_event_types]: <message>
   CompileError [code=load_error]: <message>
   ```

   AST or smoke failures fall through to the existing `build_ast_diagnostic` / `build_smoke_diagnostic` so c.2 gets one consistent diagnostic regardless of which gate fired.

2. Extended `build_gate_diagnostic` with `kind="compile"`. The retry orchestrator should prefer this — it has the full `CompileResult` in hand and shouldn't re-derive which gate failed.

3. Defensive `CompileError [code=unknown]: ...` covers the hypothetical "ok=False but no field populated" shape — a logic-bug we want labelled in retry feedback rather than silently empty.

**Tests** — 11 new cases in `TestCompileResultDiagnostic` and `TestBuildGateDiagnosticCompileKind`:

- `ok` passthrough returns the success message.
- Each of the three CompileResult-only fields renders with the right `[code=...]` prefix.
- AST and smoke fall-through both pass the equality check against the per-gate builder.
- `invalid_event_types` *plus* a passing smoke report (the post-coverage-check shape) renders the compile-level failure and *omits* the `Smoke test` framing — the actionable bit wins.
- `kind="compile"` dispatches correctly; non-`CompileResult` payload raises with the right `expects CompileResult` message; the unknown-kind error message advertises `'compile'` so a misspelling is self-correcting.

**Verification:**
- Focused module: 33 passed (was 22; +11 new)
- Full repo: 2540 passed / 9 skipped (was 2529; +11)
- Autoformat clean (pyink)

Docs (`docs/extractor_compilation_diagnostics.md`) updated with the new builder, the four `build_gate_diagnostic` kinds, and the rationale for why `compile` is the preferred dispatcher entry point for the retry loop. CHANGELOG updated in place.

caohy1988:
Addressed in b108465 + PR body refresh:

**P2 — "never raises" overpromised.** Narrowed in both the module docstring and `docs/extractor_compilation_diagnostics.md`: per-gate builders never raise (even on degenerate input); the dispatcher `build_gate_diagnostic` validates caller input and is the only function that does raise (`TypeError` on payload-type mismatch, `ValueError` on unknown kind). Both are caller-misuse signals the retry orchestrator should never see in practice.

**P3 — defensive ordering mismatch.** Reordered `build_compile_result_diagnostic` so AST falls through *before* the `load_error` check. The canonical `compile_extractor` pipeline runs AST (stage 4) before in-process load (stage 5), and never sets both fields at once. But a hand-built `CompileResult` with both populated would have rendered `load_error` and silently contradicted the docstring's "earliest stage" contract. New ordering matches the pipeline; the contract holds for any `CompileResult` shape, not just canonical-pipeline output. Added `test_ast_wins_over_load_error_when_both_populated` to lock the ordering. Docstring also calls out the `invalid_event_types`-after-smoke (stage 7 post-coverage check) corner case explicitly so future readers know why the early `invalid_event_types` branch wins over a passing smoke report.

**P3 — stale PR body.** Refreshed: now lists five public functions (incl. `build_compile_result_diagnostic`), the four `build_gate_diagnostic` kinds, the `CompileError` output shape, 34 tests, and head `b108465`.

**Verification:**
- Focused module: 34 passed (was 33; +1 ordering test)
- Full repo: 2541 passed / 9 skipped (was 2540; +1)
- Autoformat clean (pyink)

PR #126, opened 2026-05-06
feat: retry-on-gate-failure orchestrator for compiled extractors (#75 PR 4b.2.2.c.2)

caohy1988:
## Summary

Lands the **retry-on-gate-failure orchestrator** that ties together everything from the previous Phase C PRs into a single loop that turns a raw extraction rule into a compiled extractor.

This is the piece that makes compiled extractors viable in practice: parser, renderer, and compile-pipeline failures are all *recoverable* through prompt feedback if the LLM gets to see what failed. PR 4b.2.2.c.1 locked the diagnostic format; this PR uses it.

Parent epic: #75 — compile-time code generation for structured trace extractors.
Builds on: PR 4b.2.2.c.1 (diagnostic builders, merged in #125), PR 4b.2.2.b (`PlanResolver` + parser, merged in #124), PR 4b.2.1 (renderer), PR 4b.1 (`compile_extractor`).
Working plan: issue #96, Milestone C1 / PR 4b.2.2.c.2.

## What's in this PR

New module `src/bigquery_agent_analytics/extractor_compilation/retry_loop.py`:

- `compile_with_llm(*, extraction_rule, event_schema, llm_client, compile_source, max_attempts) -> RetryCompileResult` — the orchestrator. On each iteration: build prompt → call LLM → parse → render → compile. On any failure, build a diagnostic via the c.1 builders and embed it in the next prompt via `build_retry_prompt`. Stops on success or after `max_attempts`. Each attempt receives a fresh `copy.deepcopy(RESOLVED_EXTRACTOR_PLAN_JSON_SCHEMA)` so a mutating provider adapter can't corrupt the exported module global for later callers.
- `build_retry_prompt(*, original_prompt, prior_response, diagnostic) -> str` — pure prompt-stitching function. Reuses the resolution prompt as grounding; appends the prior response (serialized via `_serialize_prior_response` with a three-tier fallback so degenerate responses don't crash the builder) and the diagnostic.
- `_serialize_prior_response(value)` — internal helper. Tries `json.dumps(sort_keys=True, indent=2)` first for byte-stability, falls back to insertion-order JSON, then `repr()`. Both fallback steps catch `(TypeError, ValueError)` to cover unsortable keys, unserializable types, *and* circular references.
- `AttemptRecord` — one row per loop iteration. For a *failed* attempt, exactly one of `plan_parse_error` / `render_error` / `compile_result` is populated; for the *successful* terminal attempt, both error channels are `None` and `compile_result.ok == True`.
[...]

PR #127, opened 2026-05-07
feat: compile-and-measure utility + BKA-decision end-to-end proof (#75 PR 4c)

caohy1988:
## Summary

The end-to-end proof that the Phase C compile-with-LLM pipeline produces a working compiled extractor for our canonical example (\`extract_bka_decision_event\`), plus a **generic measurement utility** future Phase-C extractor baselines can reuse.

This is the moment of truth for Phase C: does the deterministic pipeline (resolver → renderer → compile_extractor → diagnostics → retry) actually converge on a correct extractor for a real-world example? Yes — at \`n_attempts=1\` with full parity against the handwritten reference. The artifact under \`tests/fixtures_extractor_compilation/bka_decision_measurement_report.json\` captures the numbers.

Builds on: PR 4b.2.2.c.2 (compile_with_llm orchestrator, merged in #126), PR 4b.2.2.c.1 (diagnostic builders, merged in #125), PR 4b.1 (compile_extractor + BKA fixtures).
Working plan: issue #96, Milestone C1 / PR 4c.

### Generic utility — \`extractor_compilation.measurement\`

- \`measure_compile(*, extraction_rule, event_schema, sample_events, reference_extractor, spec, llm_client, compile_source, max_attempts, model_name, source) -> CompileMeasurement\` — wraps \`compile_with_llm\`, loads the resulting bundle's compiled callable, runs both that callable and the reference extractor on the same sample events, and produces a structured record. Loop failure is captured in the record (no exception); a per-event try/except inside the comparator means an extractor crash on any event also lands as a structured divergence rather than an unhandled exception. Callers always route on \`ok\` / \`parity_ok\`.
- \`CompileMeasurement\` (frozen dataclass): \`ok\`, \`n_attempts\`, \`reason\`, \`bundle_fingerprint\`, \`attempt_failures\` (stable codes per failed attempt), \`parity_ok\`, \`n_events\`, per-axis match counts, \`parity_divergences\`, \`captured_at\`, \`model_name\`, \`source\`, \`sample_session_ids\`.
- \`to_json\` / \`from_json\` round-trip is byte-stable. \`from_json\` is the **artifact schema lock**:
  - **Exact keys.** Top-level keys must equal the dataclass field set; unknown fields raise \`TypeError\` (listing names), missing fields raise \`TypeError\` (listing names), both shapes of drift are reported together.
  - **Strict types.** Each field is checked via \`isinstance\` (and \`type(v) is bool\` for bool fields, so \`int 0\` doesn't sneak past as \`False\`). Constructor-style coercion (\`bool(\"false\")\`, \`tuple(\"abc\")\`, \`int(\"3\")\`) is rejected with the field name in the error.
- \`DETERMINISTIC_FAKE_MODEL\` constant for in-test runs.

### Centralized BKA fixtures

\`tests/fixtures_extractor_compilation/bka_decision_inputs.py\` is the single source for everything BKA-shaped both paths consume: extraction rule, event schema, sample events, resolved plan dict, ontology / binding YAML, and \`BKA_FINGERPRINT_INPUTS\`. Both the deterministic test and the gated live test feed the **same** fingerprint inputs into \`compile_extractor\`, so a live-regenerated measurement artifact's \`bundle_fingerprint\` matches the deterministic capture for metadata reasons unrelated to the extractor's behavior.

\`BKA_FINGERPRINT_INPUTS\` covers every field the extractor emits — \`event_schema.bka_decision.content\` lists every payload key the resolver might map (including \`alternatives_considered\`); \`extraction_rules.bka_decision\` mirrors the resolved plan's full structure (key field path, property fields with source paths, span handling, session_id path). This is what makes the bundle fingerprint C2's "fingerprint matches active inputs" contract — a regression test (\`TestBkaFingerprintInputsCoverage\`) proves that changing \`alternatives_considered\` (in \`event_schema\` or \`extraction_rules\`) or re-pathing \`partial_when_path\` each move \`compute_fingerprint\`'s output.

### Parity contract

For each sample event, both extractors are called with \`(event, spec)\`; their \`StructuredExtractionResult\`s are compared on three axes:
- **Node set** — same \`node_id\` values, and per shared \`node_id\` the \`entity_name\`, sorted \`labels\`, and property \`(name, value)\` set are equal.
- **Fully-handled span IDs** — set equality.
- **Partially-handled span IDs** — set equality.

Either extractor crashing on an event is captured as a structured divergence (\`event[i]: <ref|compiled> extractor raised X: msg\`) with a per-event \`try/except\`, never propagated as an exception out of \`measure_compile\`.

Edges aren't compared (the renderer doesn't emit edges yet).

### CI path (deterministic, merge-blocking)

\`tests/test_extractor_compilation_measurement.py\` (26 tests):

- \`TestMeasureCompileBkaHappyPath\` — first-try success with full parity; \`n_attempts=1\`, \`parity_ok=True\` on both events.
- \`TestMeasureCompileLoopFailure\` — exhausted-loop record shape; \`attempt_failures\` carries one stable code per failed attempt.
- \`TestMeasureCompileParityDivergence\` — property-set divergence; reference-extractor exception captured as divergence; compiled-extractor exception captured at the helper level (symmetric branch).
- \`TestCompileMeasurementJson\` — round-trip byte stability; sorted keys; failure-record round-trip.
- \`TestCompileMeasurementFromJsonStrictTypes\` — exact-keys lock (unknown fields rejected, missing fields rejected, both reported together) plus strict type checks (bool / int / str / list-of-str / optional-str rejection cases).
- \`TestBkaFingerprintInputsCoverage\` — three changes to extractor-relevant fingerprint inputs (\`event_schema\`, \`extraction_rules.property_fields\`, \`span_handling.partial_when_path\`) each move \`compute_fingerprint\`'s output.
- \`TestMeasurementAuditFields\` — \`sample_session_ids\` dedup'd in iteration order with empty strings filtered; \`model_name\` / \`source\` pass through; checked-in measurement artifact parses via the strict-keys, strict-types \`from_json\`.

No API key required. Wires the **real** \`compile_extractor\` through the **real** retry loop — only the LLM step is faked.

### Live path (gated, regenerates the artifact)

\`tests/test_extractor_compilation_bka_compile_live.py\` is gated on **both**:
- \`BQAA_RUN_LIVE_TESTS=1\` (project-wide live-test gate)
- \`BQAA_RUN_LIVE_LLM_COMPILE_TESTS=1\` (this test specifically — opting into LLM cost)

Plus runtime: \`PROJECT_ID\`, \`DATASET_ID\`, \`BQAA_LLM_COMPILE_MODEL\` (defaults to \`gemini-2.5-flash\`).

The fixture pulls a 50-row pool of \`bka_decision\` rows, filters out rows without \`content.decision_id\`, partitions by \`content.reasoning_text\` presence, and \`pytest.skip\`s if either branch is empty (skip message names the missing branch). Takes a balanced sample (up to 5/branch), wraps \`google.genai\` inline (adapter choice intentionally out-of-scope for the SDK core, per c.2 docs), runs \`measure_compile\` end-to-end, writes the result to the checked-in artifact path. Asserts only **contract invariants** — \`ok=True\` / \`parity_ok=True\` / \`n_attempts<=3\` / \`n_events>=2\` / sample covers both span-handling branches / bundle exists.

### Measurement artifact

\`tests/fixtures_extractor_compilation/bka_decision_measurement_report.json\`:

\`\`\`json
{
  \"attempt_failures\": [],
  \"bundle_fingerprint\": \"c2e26a8eaf37398b1b9fcc4b6be657f7166f943137341c25525e92224c811c46\",
  \"captured_at\": \"2026-05-07T00:00:00Z\",
  \"model_name\": \"deterministic-fake\",
  \"n_attempts\": 1,
  \"n_events\": 2,
  \"n_events_with_node_match\": 2,
  \"n_events_with_span_match\": 2,
  \"ok\": true,
  \"parity_divergences\": [],
  \"parity_ok\": true,
  \"reason\": \"succeeded\",
  \"sample_session_ids\": [\"sess1\"],
  \"source\": \"deterministic\"
}
\`\`\`

\`test_checked_in_artifact_round_trips_into_compile_measurement\` parses the artifact via the strict-keys, strict-types \`from_json\` and locks the success-record shape so any schema drift is caught at PR review.

## Design choices baked in

- **Generic \`measure_compile\`, BKA as the first concrete consumer.**
- **Deterministic CI + gated live, not live-only.** Merge-blocking proof requires no API key.
- **Loop failure → populated record, not exception.** Same for per-event extractor crashes.
- **\`compile_source\` is a caller-supplied closure** — narrow signature, trivial to stub in tests.
- **No provider adapter in SDK core.** \`google.genai\` lives in the live test only.
- **Centralized fingerprint inputs.** Both paths produce byte-equal \`bundle_fingerprint\` values.
- **Comprehensive fingerprint coverage.** Every extractor-relevant compile input contributes to the hash.
- **Strict \`from_json\`.** Exact keys + strict types. The artifact's schema lock is load-bearing.

## Verification (head \`21361e7\`)

- 26/26 deterministic tests pass; 1/1 live test correctly skips when env vars unset
- Full repo: 2588 passed / 10 skipped

## Out of scope (deferred)

- Multi-extractor measurement suite.
- Cost / token instrumentation.
- Public LLM provider adapter.
- Edge parity (renderer doesn't emit edges yet).

## Test plan

- [x] 26 deterministic tests pass
- [x] Live test skips cleanly without env vars
- [x] Full repo green (2588)
- [x] \`autoformat.sh\` clean
- [x] Measurement artifact JSON parses via strict-keys, strict-types \`from_json\`
- [x] BKA fixture imports unified across deterministic + live paths
- [x] Strict-type rejection covers bool / int / str / list-of-str / optional-str fields
- [x] Strict-keys rejection covers unknown / missing / both-together
- [x] Fingerprint coverage moves with each extractor-relevant compile-input change


caohy1988:
Both findings addressed in 7d6a874.

**P2 #1 — live test didn't actually prove both span-handling branches.** The fixture pulled the latest 10 rows unfiltered, so the live LLM call could happily succeed against a sample where every row had `reasoning_text` (or none did) — passing the `n_events >= 2` invariant while only exercising one branch. Fixed:

- Pool size bumped to 50 rows.
- Filter rows without `content.decision_id` *before* partitioning (the reference extractor returns empty on those — they're dead weight in parity and would burn LLM tokens for nothing).
- Partition remaining rows by `content.reasoning_text` presence; `pytest.skip` if either bucket is empty, with a skip message that names the missing branch and recommends widening the pool.
- Take a balanced sample (up to 5 per branch).
- Defense-in-depth assertion at the start of the test body: a future fixture refactor that loosens the partition guard fails the test rather than silently weakening the live proof.

**P2 #2 — `measure_compile` could raise during parity.** Real concern, especially given that the test helper's `compile_source` closes over `BKA_SAMPLE_EVENTS` while `measure_compile` takes its parity events separately — they can legitimately differ, so a compiled extractor could crash on a measurement-only event the smoke gate never saw. Fixed `_compare_extractors` with per-event try/except (catching `Exception`, not `BaseException`, so Ctrl-C still propagates). On either side's crash:

- Stable divergence string: `event[i]: reference extractor raised X: msg` / `event[i]: compiled extractor raised X: msg`.
- Continue to the next event; this event's per-axis match counters stay at 0.

Two regression tests:

- `test_extractor_exception_becomes_parity_divergence` — reference-side crash through full `measure_compile`; asserts exception is captured, not propagated.
- `test_compiled_extractor_exception_becomes_divergence_directly` — compiled-side crash via the helper (`_compare_extractors`). Crafting a real BKA-shaped event that crashes only the rendered extractor would add plumbing without coverage; the helper's contract is what matters and both try/except branches are now pinned.

Docs page updated to describe the new pool/partition/skip flow and the seventh contract invariant (sample covers both branches).

**Verification:**
- Focused module: 11 deterministic + 1 live (correctly skipping) = 12 (was 9 + 1 = 10; +2)
- Full repo: 2573 passed / 10 skipped (was 2571 / 10; +2)
- `bash autoformat.sh` clean

caohy1988:
All three findings addressed in 69cd4dc + PR body refresh.

**P2 — fingerprint-input divergence between deterministic and live paths.** Lifted the YAML constants and a single \`BKA_FINGERPRINT_INPUTS\` dict into \`tests/fixtures_extractor_compilation/bka_decision_inputs.py\`. Both paths now import from there, so \`compile_extractor.fingerprint_inputs\` receives byte-identical inputs and the resulting \`bundle_fingerprint\` is identical regardless of path. The deterministic capture's fingerprint is preserved (the deterministic path's prior shape *was* the canonical one); the live path now produces a matching fingerprint instead of the divergent flat-content shape.

**P3 — \`from_json\` silent type coercion.** Replaced constructor coercion with strict \`isinstance\`-based readers (\`_require_bool\` / \`_require_int\` / \`_require_str\` / \`_require_optional_str\` / \`_require_tuple_of_str\`). Bool fields use \`type(v) is bool\` — not \`isinstance\` — so \`int 0\` doesn't sneak past as \`False\` (Python treats \`bool\` as a subclass of \`int\`). Each rejection raises \`TypeError\` with the offending field name. Nine new regression tests in \`TestCompileMeasurementFromJsonStrictTypes\` cover root-must-be-object, the \`bool(\"false\")\` case, the \`int 0 → False\` case, the \`int(\"3\")\` case, the \`tuple(\"abc\")\` case, mixed-type list items, int-for-str, int-for-optional-str, and the optional-null sanity check.

**P3 — stale PR body.** Refreshed: 20 deterministic tests, head \`69cd4dc\`, full repo 2582; mentions centralized fingerprint inputs, strict \`from_json\` type checks, branch-balanced live sampling, extractor-exception parity coverage, and the new test classes.

**Verification:**
- Focused module: 20 deterministic + 1 live (correctly skipping) = 21 (was 11 + 1 = 12; +9 from strict-type tests)
- Full repo: 2582 passed / 10 skipped (was 2573 / 10; +9)

caohy1988:
Both findings addressed in 21361e7.

**P2 — fingerprint coverage didn't match the extractor's emitted fields.** Real concern, and reviewer's diagnosis is right: \`BKA_FINGERPRINT_INPUTS\` was missing \`alternatives_considered\` from \`event_schema\` and was keeping \`extraction_rules\` at only \`entity\`/\`key\` while the actual extractor's carry-over loop emits all three properties. Fixed:

- \`event_schema.bka_decision.content\` now lists every payload key the extractor might map (including \`alternatives_considered\`).
- \`extraction_rules.bka_decision\` now mirrors the resolved plan's full structure: \`key_field_path\`, \`property_fields\` with \`source_path\` per field, \`span_handling\` with \`span_id_path\` + \`partial_when_path\`, and \`session_id_path\`.

Three regression tests in \`TestBkaFingerprintInputsCoverage\`:
- Removing \`alternatives_considered\` from \`event_schema\` changes \`compute_fingerprint\`'s output.
- Removing it from \`extraction_rules.property_fields\` changes the output.
- Moving \`span_handling.partial_when_path\` to a different field changes the output.

These are the three real-world compile-input change shapes that would otherwise leave the bundle fingerprint unchanged — exactly C2's "fingerprint matches active inputs" failure mode.

The artifact's \`bundle_fingerprint\` regenerates as \`c2e26a8eaf37398b1b9fcc4b6be657f7166f943137341c25525e92224c811c46\` (was \`9e2b53...\`); the change is the expected consequence of the fuller inputs. PR 4b.1's local \`_fingerprint_inputs\` helper retains its narrower shape, so its tests are unaffected.

**P3 — unknown fields silently ignored.** \`from_json\` only checked types; an extra key parsed cleanly and the schema-lock test still passed. Replaced with exact-keys: JSON top-level keys must equal the dataclass field set. Unknown fields raise \`TypeError\` listing the names; missing fields raise \`TypeError\` listing the names; both shapes of drift are reported together so a single fix-it round suffices. Three regression tests cover unknown-only / missing-only / both-together.

**Verification (head 21361e7):**
- Focused module: 26 deterministic + 1 live (correctly skipping) = 27 (was 20 + 1 = 21; +6)
- Full repo: 2588 passed / 10 skipped (was 2582 / 10; +6)

caohy1988:
Both stale-doc findings addressed in 8896c96:

**P3 — PR body refreshed.** Now reports head \`21361e7\` (then \`8896c96\` after the docs commit), 26 deterministic tests, full repo 2588, with a new "Measurement artifact" block showing the regenerated JSON (\`bundle_fingerprint=c2e26a8e...\`) and bullets for the strict-keys / strict-types lock + fingerprint-coverage tests.

**P3 — \`docs/extractor_compilation_bka_measurement.md\` refreshed.** The CI breakdown now lists \`TestCompileMeasurementFromJsonStrictTypes\` (with the exact-keys + strict-types lock spelled out), \`TestBkaFingerprintInputsCoverage\` (the three regression tests proving the bundle fingerprint moves whenever an extractor-relevant input changes), and the per-event extractor-exception parity branches. Total tests bumped from "(no count)" to "(26 tests)".

Docs-only commit; no code changes. Worktree clean, full repo still 2588 / 10 skipped.

PR #130, opened 2026-05-08
feat: bundle loader + minimal runtime discovery for compiled extractors (#75 PR C2.a)

caohy1988:
## Summary

Lands the **trust boundary** between on-disk compiled bundles and the runtime that's about to import + execute them. Per the runtime-target RFC, compiled extractors run client-side as plain Python callables plugged into the existing \`run_structured_extractors()\` hook. Before any callable is plugged in, this PR is what verifies the bundle on disk matches the runtime's *active inputs* (fingerprint + event_types) and that the imported callable has a usable shape.

C2.a is **loader + discovery only.** Validation/fallback semantics, BQ-table mirror, and the ontology-graph call-site swap are explicitly out of scope and deferred to C2.b/c/d.

Builds on: PR 4c (compile + measure, merged in #127), PR 4b.1 (compile harness + Manifest), the runtime-target RFC (\`docs/extractor_compilation_runtime_target.md\`).
Working plan: issue #96, Milestone C2 / PR C2.a.

## What's in this PR

New module \`src/bigquery_agent_analytics/extractor_compilation/bundle_loader.py\`:

- \`load_bundle(bundle_dir, *, expected_fingerprint, expected_event_types) -> LoadedBundle | LoadFailure\` — eight validation gates short-circuiting in order: manifest exists, parses + has the right shape, fingerprint matches, \`expected_event_types\` (when set) is a subset of manifest's coverage, module file exists, import succeeds, function exists and is callable, signature accepts \`(event, spec)\`. **Never raises.**
- \`discover_bundles(parent_dir, *, expected_fingerprint, event_type_allowlist) -> DiscoveryResult\` — walks the parent, loads each child via \`load_bundle\`, applies optional \`event_type_allowlist\`, detects \`event_type_collision\`s and fails closed.
- \`LoadedBundle\` / \`LoadFailure\` / \`DiscoveryResult\` dataclasses (\`frozen=True\`).

Stable \`LoadFailure.code\` values:

| Code | When |
|---|---|
| \`manifest_missing\` | \`bundle_dir/manifest.json\` doesn't exist |
| \`manifest_unreadable\` | JSON parse error or \`Manifest.from_json\` rejected the payload |
| \`fingerprint_mismatch\` | manifest fingerprint != caller's \`expected_fingerprint\` |
| \`event_types_mismatch\` | \`expected_event_types\` isn't a subset of \`manifest.event_types\` |
| \`module_not_found\` | the module file referenced by the manifest is absent on disk |
| \`import_failed\` | importing the module raised — catches both \`Exception\` and \`BaseException\` (e.g., \`SystemExit\`), so a malicious or buggy bundle can't tear down the loading process |
| \`function_not_found\` | manifest's \`function_name\` isn't defined as a callable |
| \`function_signature_mismatch\` | imported callable can't be called as \`f(event, spec)\` (best-effort introspection via \`inspect.signature\` + \`sig.bind(None, None)\`) |
| \`event_type_collision\` | discovery only: two valid bundles declare coverage of the same event_type. Fail-closed: that event_type is dropped from the registry; both colliding bundles get a failure record. Other event_types from those bundles still register. |

## Design choices baked in

- **Fingerprint check runs before module import.** A bundle with a wrong fingerprint and a broken module fails with \`fingerprint_mismatch\`, not \`import_failed\`. The fingerprint is a real boundary: an attacker can't side-effect via a broken module if their fingerprint doesn't match. \`test_fingerprint_check_runs_before_module_load\` pins this ordering.
- **Loader never raises.** Every defect surfaces as a structured \`LoadFailure\` record. Includes \`BaseException\` capture at import time (e.g., \`sys.exit\` at module top-level). Tested.
- **Multi-event bundles register the same callable.** A manifest with \`event_types=("a", "b")\` produces \`registry["a"] is registry["b"]\` — one bundle, one callable, two registry keys.
- **Subset semantics for \`expected_event_types\`.** \`load_bundle\` accepts a manifest that covers *more* than the caller asked for. The caller's expected event_types just have to be present; extra coverage doesn't fail.
- **Allowlist is a scope filter, not a load gate.** A bundle whose declared coverage falls outside \`event_type_allowlist\` still loads (visible in \`discovery.loaded\`); its event_types just don't reach \`discovery.registry\`. Empty allowlist registers nothing.
- **Collisions fail closed.** Two bundles claiming the same event_type drop that event_type from the registry, with one \`LoadFailure\` per claimant. Other unique event_types from those bundles still register. Silently picking one would make runtime behavior depend on filesystem ordering.
- **Non-bundle entries tolerated.** Loose files in the parent (README, INDEX) are silently skipped (they're files, not subdirectories). Non-bundle subdirectories fail with \`manifest_missing\` — every walked directory is accounted for.

## Tests (30 cases in \`tests/test_extractor_compilation_bundle_loader.py\`)

- **\`TestLoadBundleHappyPath\`** (3) — valid bundle loads; subset check accepts broader manifest; check skipped when \`expected_event_types=None\`.
- **\`TestLoadBundleFailureCodes\`** (13) — every stable code, including all three import-failure varieties (\`SyntaxError\`, \`RuntimeError\` at top-level, \`SystemExit\`); function-not-callable shares \`function_not_found\`'s code; \`*args\` signature accepted, kwargs-only rejected.
- **\`TestLoadBundleGateOrdering\`** (1) — fingerprint check before module import.
- **\`TestDiscoverBundles\`** (6) — empty parent, non-existent parent (single \`manifest_missing\` failure), single bundle, multi-event bundle (same callable identity under both keys), allowlist filtering preserves loaded bundle, empty allowlist.
- **\`TestDiscoverBundlesCollisions\`** (2) — two bundles same event_type fail closed; partial collision preserves unique event_types from each.
- **\`TestDiscoverBundlesNonBundleEntries\`** (2) — loose files tolerated; non-bundle subdirectories fail with \`manifest_missing\`.
- **\`TestBundleLoaderEndToEnd\`** (2) — runs the **real** \`compile_extractor\` to produce a bundle, then loads it via \`load_bundle\` AND \`discover_bundles\`, invokes the loaded callable, and asserts behavioral parity with the handwritten reference. Proves the loader's contract holds for bundles produced by the rest of Phase C, not just hand-built fixtures.

## Verification

- 30/30 focused tests pass
- Full repo: 2618 passed / 10 skipped (was 2588 / 10; +30 new)

## Out of scope (deferred to later C2 PRs)

- **Per-field / per-node / per-edge fallback** through #76's validator. C2.b.
- **Per-event fallback** to handwritten or \`AI.GENERATE\` paths. C2.b.
- **BigQuery-table bundle mirror** for cross-process distribution. C2.c.
- **Ontology-graph call-site swap** — where the discovered registry actually replaces the existing extractors in the orchestrator. Moves once C2.a + C2.b land. C2.c.
- **Revalidation harness** (scheduled / on-demand agreement check). C2.d.

## Test plan

- [x] 30 focused tests pass (every \`LoadFailure\` code has at least one test)
- [x] Loader never raises — including \`BaseException\` capture at import time
- [x] Wrong fingerprint never loads
- [x] Wrong event type never loads
- [x] Multi-event bundle registers all event types
- [x] Duplicate event type across bundles fails closed
- [x] End-to-end real \`compile_extractor\` bundle loads and invokes successfully via both \`load_bundle\` and \`discover_bundles\`
- [x] \`autoformat.sh\` clean


caohy1988:
All four findings addressed in cd154ec.

**P1 #1 — malformed manifests could raise or load nonsense.** Confirmed locally: \`module_filename: 42\` flowed into the dataclass without raising, then \`bundle_dir / 42\` raised TypeError; \`event_types: \"xy\"\` silently became \`(\"x\", \"y\")\`. Replaced the parse step in the loader with \`_parse_manifest_strict\`: rejects unknown/missing fields, type mismatches, empty/duplicate/non-string \`event_types\`, non-identifier \`function_name\`, non-\`<identifier>.py\` \`module_filename\`. The loader is the trust boundary; \`Manifest.from_json\` itself stays lenient because the cache-hit path in \`compile_extractor\` already catches broad exceptions and doesn't need strict semantics. Eighteen-case parametrized regression test covers each malformed shape.

**P1 #2 — \`module_filename\` could escape \`bundle_dir\`.** Two layers of defense now: (a) strict parse rejects any \`module_filename\` that isn't \`<identifier>.py\` — no path separators, no leading dots, no keyword stems, must end in \`.py\`; (b) after path join, the resolved module path must be directly inside \`bundle_dir.resolve()\`. Both checks fire BEFORE any import attempt. Reproducer test (\`test_module_filename_path_traversal_rejected\`) plants an \`escape.py\` sibling that raises \`RuntimeError\` if imported and confirms the loader rejects with \`manifest_unreadable\` without ever touching it.

**P2 — sys.modules leak.** \`_import_module_from_path\` only popped on failure. Moved the pop into a \`finally\` block so successful loads also clean up. The captured callable retains a reference to the module's globals, so runtime invocation is unaffected. Regression test (\`test_repeated_load_does_not_leak_sys_modules\`) runs 5 loads, asserts the \`__loaded_\` entry count is unchanged, AND asserts the captured callable still works after cleanup.

**P3 — \`discover_bundles\` could raise from \`iterdir()\`.** Wrapped \`sorted(parent_dir.iterdir())\` in \`try/except OSError\`; on raise, returns a \`DiscoveryResult\` with a structured \`manifest_missing\` failure naming the underlying \`PermissionError\` / \`OSError\`. Regression test monkeypatches \`pathlib.Path.iterdir\` to raise \`PermissionError\`.

**Verification (head cd154ec):**
- Focused module: 53 passed (was 30; +23 regression cases)
- Full repo: 2641 passed / 10 skipped (was 2618 / 10; +23)

caohy1988:
Both findings addressed in 035004c.

**P2 — invalid-UTF-8 manifest escaped "never raises".** Confirmed locally with the reviewer's exact repro: writing \`b\"\xff\xfe\xff\xfe\"\` to \`manifest.json\` made \`load_bundle\` raise \`UnicodeDecodeError\` instead of returning a structured \`LoadFailure\`. \`Path.read_text(encoding=\"utf-8\")\` raises \`UnicodeDecodeError\` (a subclass of \`UnicodeError\`), which the prior \`OSError\`-only catch missed. Widened to \`(OSError, UnicodeError)\` and added a regression test (\`test_invalid_utf8_manifest_rejected\`) that writes those exact bytes and asserts a structured \`manifest_unreadable\` whose detail names \`UnicodeDecodeError\`.

**P3 — docs test breakdown stale.** Bumped from \"30 cases\" to \"54 cases\" and added explicit sub-sections for each regression group introduced during review: strict manifest validation (20), path-traversal defense (2), \`sys.modules\` cleanup (1), \`iterdir\` failure (1). The original groups (happy path, failure codes, gate ordering, discovery, collisions, non-bundle entries, end-to-end) keep their counts.

**Verification (head 035004c):**
- Focused module: 54 passed (was 53; +1 UTF-8 case)
- Full repo: 2642 passed / 10 skipped (was 2641 / 10; +1)

PR #134, opened 2026-05-09
feat: runtime fallback wiring for compiled structured extractors (#75 PR C2.b)

caohy1988:
## Summary

The runtime safety net for compiled extractors. When a compiled extractor produces output that crashes, doesn't fit the contract, has malformed internals, or violates the ontology in ways that can't be salvaged, this wrapper substitutes the *fallback* extractor (the existing handwritten or \`AI.GENERATE\` path). When the violations are pinpointable to specific nodes / edges, the wrapper drops just those elements **and downgrades the event's span-handling so the AI transcript still sees the source span and can recover the missing pieces.**

C2.b is the wrapper *policy* only. The actual call-site swap inside the orchestrator is C2.c.

Builds on: PR C2.a (bundle loader, merged in #130), the runtime-target RFC, the #76 validator.
Working plan: issue #96, Milestone C2 / PR C2.b.

## What's in this PR

New module \`src/bigquery_agent_analytics/extractor_compilation/runtime_fallback.py\`:

- \`run_with_fallback(*, event, spec, resolved_graph, compiled_extractor, fallback_extractor) -> FallbackOutcome\` — the decision-tree wrapper.
- \`FallbackOutcome\` (\`frozen=True\`): \`result\`, \`decision\`, \`compiled_exception\`, \`dropped_node_ids\`, \`dropped_edge_ids\`, \`validation_failures\`.

## Decision tree

| Step | Condition | Decision |
|------|-----------|----------|
| 1 | Compiled extractor raises (\`Exception\` or \`SystemExit\`) *or* returns a non-\`StructuredExtractionResult\` value | \`fallback_for_event\` |
| 1b | \`fully_handled_span_ids\` / \`partially_handled_span_ids\` aren't a \`set\`/\`frozenset\` of non-empty strings | \`fallback_for_event\` (\`compiled_exception\` starts \`\"MalformedResultInternals:\"\`) |
| 1c | \`nodes\` / \`edges\` raise during \`ExtractedGraph\` construction (e.g., \`nodes=[{}]\`) | \`fallback_for_event\` (\`compiled_exception\` starts \`\"MalformedResultInternals:\"\`) |
| 2 | Validation passes | \`compiled_unchanged\` |
| 3 | Any EVENT-scope failure, or any failure not pinpointable for its scope | \`fallback_for_event\` |
| 4 | Otherwise (every failure is scope-pinpointable) | \`compiled_filtered\` |

Pinpointability is **scope-specific**: NODE iff \`node_id\` set, EDGE iff \`edge_id\` set, FIELD iff *either* is set. Critical for #76's \`missing_endpoint_key\` failures, which populate both \`node_id\` (referenced endpoint) and \`edge_id\` (offending edge) — the wrapper drops the edge, not the endpoint.

\`KeyboardInterrupt\` is intentionally **not** caught so operator cancellation works.

## \`compiled_filtered\` semantics

1. Drops decided by \`failure.scope\`: NODE → \`node_id\`; EDGE → \`edge_id\` (regardless of whether \`node_id\` is also set); FIELD → \`edge_id\`-then-\`node_id\` (the property's actual container).
2. **Orphan-cleans** any edge whose \`from_node_id\` / \`to_node_id\` was dropped.
3. **Downgrades the event's span-handling**: removes \`event[\"span_id\"]\` from \`fully_handled_span_ids\` and adds it to \`partially_handled_span_ids\`. Load-bearing for C2.b — without this the AI transcript would never see source spans for dropped facts.

## Design choices baked in

- **Drop-whole-element on FIELD failures.** Conservative; simpler to reason about than per-property salvage.
- **Don't validate the fallback output.** Fallback is the trusted runtime baseline.
- **Don't catch fallback exceptions.** Tested.
- **EVENT scope wins over per-element failures.**
- **Fail closed on unpinpointable failures.**
- **Strict span-set shape validation.** Bad span containers (\`None\`, raw strings, lists, mixed-type entries) are caught up front rather than leaking into runtime code that expects iterables-of-string.
- **\`SystemExit\` captured, \`KeyboardInterrupt\` propagated.** Matches C2.a's \`BaseException\`-at-import policy.

## Tests (28 cases in \`tests/test_extractor_compilation_runtime_fallback.py\`)

Original groups:
- **\`TestRunWithFallbackCompiledUnchanged\`** (2)
- **\`TestRunWithFallbackForEventTriggers\`** (7) — incl. compiled exception / wrong type / None / EVENT-scope / unpinpointable / mixed / fallback exceptions propagating
- **\`TestRunWithFallbackCompiledFiltered\`** (4) — real validator on ghost-entity NODE; orphan cleanup; EDGE drop; FIELD-with-node_id
- **\`TestRunWithFallbackSpanDowngrade\`** (2) — load-bearing span downgrade + no-span-id no-op
- **\`TestRunWithFallbackEndToEnd\`** (1) — real BKA bundle + handwritten fallback

Review-driven regression groups:
- **\`TestRunWithFallbackMalformedInternals\`** (1) — \`StructuredExtractionResult(nodes=[{}])\` falls back, doesn't raise
- **\`TestRunWithFallbackEdgeFailureWithBothIds\`** (2) — \`missing_endpoint_key\` shape drops edge not node; symmetric NODE-without-\`node_id\` is unpinpointable
- **\`TestRunWithFallbackSystemExit\`** (2) — \`SystemExit\` captured; \`KeyboardInterrupt\` propagates
- **\`TestRunWithFallbackSpanSetShape\`** (7) — \`fully_handled=None\` / \`partially=string\` / list / non-string entry / empty-string entry / frozenset-accepted / combined-NODE-failure-with-bad-spans

## Verification (head \`68e8e9e\`)

- 28/28 focused tests pass
- Full repo: 2670 passed / 10 skipped

## Out of scope (deferred to other C2 sub-PRs)

- **Orchestrator call-site swap** — C2.c.
- **BigQuery-table bundle mirror** — C2.c.
- **Revalidation harness** — C2.d.

## Test plan

- [x] 28 focused tests pass
- [x] Wrapper never raises on compiled-extractor / validator / span-shape failure
- [x] Fallback-extractor exceptions propagate
- [x] EVENT-scope failure triggers whole-event fallback
- [x] Per-element failures drop pinpointable elements only
- [x] EDGE failures with both \`node_id\` and \`edge_id\` drop the edge
- [x] Orphan edges cleaned on dropped nodes
- [x] Span-handling downgraded on \`compiled_filtered\` (load-bearing)
- [x] Bad span containers rejected at the trust boundary
- [x] \`SystemExit\` captured; \`KeyboardInterrupt\` propagates
- [x] End-to-end real BKA bundle + handwritten fallback
- [x] \`autoformat.sh\` clean


caohy1988:
All three findings addressed in 5740b42.

**P1 #1 — malformed result internals raised instead of falling back.** Confirmed reviewer's exact repro locally: \`StructuredExtractionResult(nodes=[{}])\` succeeds (the dataclass doesn't validate types at runtime), then \`ExtractedGraph(nodes=[{}, ...])\` raises \`pydantic.ValidationError\`. Wrapped \`ExtractedGraph(...) + validate_extracted_graph(...)\` in \`try/except Exception\`; on raise, fall back for the event with \`compiled_exception\` starting \`\"MalformedResultInternals:\"\` so logs can route this distinctly from extractor exceptions. Regression test (\`test_compiled_returns_result_with_dict_nodes_falls_back\`) uses the reviewer's exact shape.

**P1 #2 — EDGE failures with both \`node_id\` and \`edge_id\` dropped the node.** Confirmed: #76's \`missing_endpoint_key\` populates both fields (the prior \`elif\` chain prioritized \`node_id\` regardless of scope). Replaced with a scope-aware switch — NODE drops \`node_id\`; EDGE drops \`edge_id\` (always, even when \`node_id\` is also set); FIELD drops \`edge_id\` first then \`node_id\` (the property's actual container). Extracted \`_is_failure_pinpointable(failure)\` helper for the unpinpointable check, also scope-aware: NODE iff \`node_id\` set, EDGE iff \`edge_id\` set, FIELD iff either set. Two regression tests: the missing-endpoint-key shape drops the edge correctly (\`test_edge_failure_with_both_ids_drops_edge_not_node\`); symmetric NODE-without-\`node_id\`-but-with-\`edge_id\` is treated as unpinpointable (\`test_node_failure_without_node_id_is_unpinpointable\`).

**P2 — \`SystemExit\` from the compiled extractor escaped.** Widened the catch from \`except Exception\` to \`except (Exception, SystemExit)\` — same shape C2.a's loader uses at import time. \`KeyboardInterrupt\` is intentionally **not** caught so operator cancellation still works. Two regression tests: \`SystemExit\` captured with \`compiled_exception\` starting \`\"SystemExit:\"\` and \`decision=\"fallback_for_event\"\`; \`KeyboardInterrupt\` propagates through \`pytest.raises\`.

Docs page updated to call out the new \`MalformedResultInternals:\` failure mode, the scope-aware drop policy (with the \`missing_endpoint_key\` example), and the \`SystemExit\` vs \`KeyboardInterrupt\` distinction.

**Verification (head 5740b42):**
- Focused module: 21 passed (was 16; +5 regression cases)
- Full repo: 2663 passed / 10 skipped (was 2658 / 10; +5)

caohy1988:
Both findings addressed in 68e8e9e (rebased on top of the merge commit on the branch).

**P1 — span-handling internals weren't validated.** Confirmed reviewer's exact repros locally:
- \`fully_handled_span_ids=None\` → returned \`compiled_unchanged\` with \`None\` leaking downstream.
- \`partially_handled_span_ids=\"span1\"\` → returned \`compiled_unchanged\` with the raw string.

Added \`_check_span_set_shape\` that requires \`set\` or \`frozenset\` of non-empty strings; runs immediately after the \`isinstance(StructuredExtractionResult)\` check, before the \`ExtractedGraph\` validator stage. Failures route to \`fallback_for_event\` with \`compiled_exception\` starting \`\"MalformedResultInternals:\"\` and naming the offending field. Strings are rejected even though iterable — \`set(\"span1\")\` would silently corrupt to \`{\"s\",\"p\",\"a\",\"n\",\"1\"}\`, exactly the failure mode the reviewer's repro #2 demonstrated. Lists are rejected since the field declaration is \`set[str]\` and downstream code unions via set semantics. \`frozenset\` is accepted (immutable variant of set, equivalent for membership/iteration/union).

Seven regression tests in \`TestRunWithFallbackSpanSetShape\`: \`None\`, string-instead-of-set, list-instead-of-set, non-string entry, empty-string entry, \`frozenset\` accepted, and the combined NODE-failure-AND-bad-spans scenario (which previously would have crashed in the filtered path's \`set(...)\` coercion — now falls back via span-set validation up front).

**P3 — PR body refreshed.** Now reports head \`68e8e9e\`, 28 focused tests, full repo 2670, with the new step 1b in the decision tree, the new test class, and updated design-choices section.

**Verification (head 68e8e9e):**
- Focused module: 28 passed (was 21; +7)
- Full repo: 2670 passed / 10 skipped (was 2663 / 10; +7)

caohy1988:
Fixed in 3241318 — the docs page heading now says `Tests (28 cases...)` to match the section body, the PR body, and the actual test file. Doc-only commit.

PR #137, opened 2026-05-09
feat: runtime extractor-registry adapter for compiled extractors (#75 PR C2.c.1)

caohy1988:
## Summary

The adapter that glues C2.a's bundle loader (\`discover_bundles\`) and C2.b's runtime fallback wrapper (\`run_with_fallback\`) into one call: \`build_runtime_extractor_registry(...)\` returns an \`event_type → extractor\` dict ready to pass into the existing \`run_structured_extractors\` hook.

**This PR ships the adapter, not the orchestrator call-site swap.** Deciding which orchestrator paths actually adopt the registry is C2.c.2.

Builds on: PR C2.a (bundle loader, #130), PR C2.b (runtime fallback, #134).
Working plan: issue #96, Milestone C2 / PR C2.c.1.

## What's in this PR

New module \`src/bigquery_agent_analytics/extractor_compilation/runtime_registry.py\`:

- \`build_runtime_extractor_registry(*, bundles_root, expected_fingerprint, fallback_extractors, resolved_graph, event_type_allowlist, on_outcome) -> WrappedRegistry\` — the main entry point.
- \`WrappedRegistry\` (\`frozen=True\`): \`extractors\`, \`discovery\`, \`bundles_without_fallback\`, \`fallbacks_without_bundle\`.
- \`OutcomeCallback\` type alias (\`(event_type, outcome) -> None\`).

## Wiring matrix

For each event_type in scope (filtered by \`event_type_allowlist\` if set):

| Compiled bundle | Handwritten fallback | Registry entry | Audit |
|---|---|---|---|
| ✓ | ✓ | Wrapped closure → \`run_with_fallback(compiled, fallback)\` + \`on_outcome\` | — |
| ✗ | ✓ | Original fallback callable (identity preserved) | Listed in \`fallbacks_without_bundle\` |
| ✓ | ✗ | **Not registered** — fail-closed | Listed in \`bundles_without_fallback\` |
| ✗ | ✗ | — | — |

The "compiled-only fail-closed" policy is what makes C2's safety contract load-bearing: a compiled extractor without a handwritten fallback would run *without* the validator-driven safety net, which inverts the C2 guarantees. Skip and surface in audit.

## Audit surfaces

- **\`bundles_without_fallback\`** — strict configuration-error signal: a compiled bundle was successfully discovered but had no matching fallback, so it was skipped.
- **\`fallbacks_without_bundle\`** — wider "no *usable* compiled registry entry" signal. Includes both "bundle was never built" *and* "a bundle exists but discovery rejected it" (fingerprint mismatch, \`manifest_unreadable\`, event-type collision, etc.) — with the underlying reason in \`discovery.failures\`. Rollout telemetry that wants to distinguish "bundle never built" from "bundle rejected" should cross-reference \`discovery.failures\` for the failure code rather than treat this as a pure "no coverage yet" count.

Both are sorted tuples for deterministic output; they exclude event_types outside \`event_type_allowlist\` since those are out of the caller's stated scope.

## Build-time validation of \`fallback_extractors\`

Each entry is checked up front; bad entries raise \`TypeError\` naming the offending key, *before* allowlist scoping:

- Key must be a non-empty \`str\` (non-str keys silently never match \`run_structured_extractors\`; mixed key types crash audit-tuple sorting).
- Value must be callable (\`None\` would be silently skipped; other non-callables raise far from the misconfig site).
- Callable signature must accept \`(event, spec)\` — checked via the bundle loader's \`_signature_compatible\` helper so the \`StructuredExtractor\` contract has one source of truth at every trust boundary.

## \`on_outcome\` callback

\`(event_type, outcome) -> None\`. Invoked from inside each wrapped extractor on **every** event including \`compiled_unchanged\` outcomes — denominator metric for compiled-vs-fallback rate analysis. Not invoked for fallback-only registry entries (no \`FallbackOutcome\` to report).

**Callback exceptions propagate.** Telemetry callbacks should be correct; silently swallowing would hide instrumentation bugs.

## Tests (22 cases in \`tests/test_extractor_compilation_runtime_registry.py\`)

- **\`TestRegistryWiringMatrix\`** (4) — compiled+fallback wraps via \`run_with_fallback\`; fallback-only passes through unchanged; compiled-only is skipped and listed; neither-present is empty registry.
- **\`TestRegistryAuditSurfaces\`** (3) — \`fallbacks_without_bundle\` records the "no usable compiled registry entry" cases; audit lists are sorted; mixed three-event scenario lands each event in the right bucket.
- **\`TestRegistryAllowlist\`** (3) — allowlist filters both pools (out-of-scope event_types in NEITHER audit field); empty allowlist registers nothing; \`None\` considers everything.
- **\`TestRegistryOnOutcomeCallback\`** (4) — fires on \`compiled_unchanged\`; fires on \`fallback_for_event\`; one call per invocation (denominator metric); callback exceptions propagate.
- **\`TestRegistryEndToEnd\`** (1) — real BKA bundle + handwritten fallback fed through \`run_structured_extractors\` via the registry; both sample events produce \`compiled_unchanged\` outcomes; callback log shows expected per-event traces.
- **\`TestRegistryFallbackCallableValidation\`** (7) — build-time rejection of every misconfigured shape: \`None\` value, non-callable (int), out-of-allowlist invalid (full dict validated), non-string key, empty-string key, mixed-key-types (rejected before audit-sort crash), one-arg callable (\`_signature_compatible\` check).

## Verification

- 22/22 focused tests pass
- Full repo: 2692 passed / 10 skipped (was 2670 / 10; +22)

## Out of scope (deferred)

- **Orchestrator call-site swap** — **C2.c.2.**
- **BigQuery-table bundle mirror** — **C2.c.3.**
- **Revalidation harness** — **C2.d.**

## Test plan

- [x] 22 focused tests pass
- [x] Wiring matrix covers all four cells
- [x] Audit surfaces sorted; \`fallbacks_without_bundle\` documented as "no usable compiled registry entry"
- [x] Allowlist filters both candidate pools
- [x] \`on_outcome\` fires on \`compiled_unchanged\` (denominator metric)
- [x] Callback exceptions propagate
- [x] Real BKA bundle works through \`run_structured_extractors\`
- [x] Build-time fallback validation: non-str / empty-str keys, non-callable values, signature-incompatible callables
- [x] \`autoformat.sh\` clean


caohy1988:
Addressed in e7a7725 (doc-only).

\`fallbacks_without_bundle\`'s wording now precisely says "no usable compiled registry entry." Both the dataclass docstring and the docs page call out that this set is wider than "no bundle was ever built" — it also includes bundles that exist but were rejected by discovery (fingerprint mismatch, \`manifest_unreadable\`, event-type collision, etc.), with the underlying reason in \`discovery.failures\`. Rollout telemetry that wants to distinguish "bundle never built" from "bundle rejected" should cross-reference \`discovery.failures\` for the failure code rather than treating \`fallbacks_without_bundle\` as a pure "no coverage yet" count.

\`bundles_without_fallback\` retains its strict meaning: compiled bundle successfully discovered but no matching fallback. That's always a configuration-error signal.

Doc-only commit; no code changes. 15 focused tests still pass at \`e7a7725\`.

caohy1988:
Both findings addressed in 06596d8.

**P2 — non-callable fallback values produced a broken registry.** Confirmed reviewer's exact repros locally: \`{\"a\": None}\` and \`{\"a\": 123}\` both flowed through to the registry as-is. \`run_structured_extractors\` then silently skipped \`None\` (treats it as \"no extractor\") or raised \`TypeError\` later when invoking the int — far from the misconfig site.

Added build-time validation: every \`fallback_extractors\` value must be callable. Rejects with \`TypeError\` naming the offending event_type. Validates the **full** input dict *before* allowlist scoping — out-of-allowlist misconfigs still surface, since the caller's input contract has to hold for the whole dict (and the allowlist may widen later). Three regression tests:
- \`test_none_fallback_rejected\`
- \`test_non_callable_fallback_rejected\` (int)
- \`test_invalid_fallback_outside_allowlist_still_rejected\`

**P3 — stale wording scrubbed.** Found three remaining instances of "no compiled coverage yet" beyond the dataclass docstring fix:
- \`runtime_registry.py\` module docstring (line 47)
- docs page test summary (line 94)
- CHANGELOG entry

Plus the README index entry needed an update too. All four now describe \`fallbacks_without_bundle\` consistently as "no usable compiled registry entry" with the explicit note that fingerprint mismatch / collision / \`manifest_unreadable\` also land here, with details in \`discovery.failures\`. Verified via grep — no instances of the stale phrase remain anywhere in the repo.

**Verification (head 06596d8):**
- Focused module: 18 passed (was 15; +3 callable-validation cases)
- Full repo: 2688 passed / 10 skipped (was 2685 / 10; +3)

caohy1988:
Both findings addressed in da1d645.

**P2 — tightened fallback validation beyond just \`callable()\`.** Confirmed all four reviewer repros locally before fixing:
- \`{1: fn}\` → registry contained an int key; \`run_structured_extractors\` keys by str event_type, so silently never matches.
- \`{\"\": fn}\` → empty-string event_type registers an entry no real event can match.
- \`{\"a\": fn, 1: fn}\` → audit-tuple \`sorted(...)\` crashed with the cryptic \`\"'<' not supported between str and int\"\` error, far from the misconfig site.
- \`{\"a\": lambda event: ...}\` → passed \`callable()\` check, then crashed at runtime invocation.

Tightened build-time validation: keys must be \`str\` (TypeError naming the key), keys must be non-empty, values must be callable, **and** the callable signature must accept \`(event, spec)\`. Reused the bundle loader's \`_signature_compatible\` helper so the StructuredExtractor contract check has one source of truth at every trust boundary. Four regression tests:
- \`test_non_string_key_rejected\`
- \`test_empty_string_key_rejected\`
- \`test_mixed_key_types_rejected_before_sort_crash\` (verifies the type check fires before the cryptic sort error)
- \`test_one_arg_callable_rejected\`

**P3 — docs test count bumped.** Docs page now says \"Tests (22 cases…)\" and lists all seven \`TestRegistryFallbackCallableValidation\` rejection cases (including the three pre-existing None / non-callable / out-of-allowlist tests plus the four added here).

**Verification (head da1d645):**
- Focused module: 22 passed (was 18; +4)
- Full repo: 2692 passed / 10 skipped (was 2688 / 10; +4)

caohy1988:
Both hygiene items addressed in 495e93a (rebased atop the latest \`feat/extractor-runtime-registry\` after the main-merge commits that landed in the meantime).

**Raises docstring** in \`build_runtime_extractor_registry\` now enumerates all four rejection shapes (non-str key, empty-string key, non-callable value, signature-incompatible callable), names the offending key, and calls out that validation runs *before* allowlist scoping. API doc matches the stricter contract.

**PR body refreshed**: 22 focused tests / 2692 full repo, the \"no usable compiled registry entry\" wording for \`fallbacks_without_bundle\`, and a new build-time-validation section calling out the four rejection shapes.

Doc-only commit on top of the prior code fixes. 22 focused tests still pass on head \`495e93a\`.

PR #143, opened 2026-05-11
feat: orchestrator call-site swap for compiled extractors (#75 PR C2.c.2)

caohy1988:
## Summary

The actual call-site swap that puts compiled structured extractors on the runtime path. C2.c.1 shipped the registry adapter (\`build_runtime_extractor_registry\`); this PR wires that adapter into \`OntologyGraphManager\` so the existing \`run_structured_extractors\` call inside \`extract_graph\` picks up compiled-with-fallback behavior automatically when the manager is built from a bundle root.

The wire-up is a new classmethod, \`OntologyGraphManager.from_bundles_root(...)\`. The existing \`__init__\` and \`from_ontology_binding\` paths are unchanged — back-compat holds by construction.

Builds on: PR C2.c.1 (registry adapter, #137), PR C2.b (runtime fallback, #134), PR C2.a (bundle loader, #130).
Working plan: issue #96, Milestone C2 / PR C2.c.2.

## What's in this PR

Two changes to \`src/bigquery_agent_analytics/ontology_graph.py\`:

1. **\`OntologyGraphManager.__init__\`** initializes a new attribute \`self.runtime_registry: Optional[WrappedRegistry] = None\`. Direct constructor callers leave this as \`None\` — back-compat preserved.
2. **\`OntologyGraphManager.from_bundles_root(...)\` classmethod** — the bundle-wired construction entry point. Mirrors \`from_ontology_binding\` arg shape; the only differences are the four bundle-specific params (\`bundles_root\`, \`expected_fingerprint\`, \`event_type_allowlist\`, \`on_outcome\`) and that \`extractors=\` becomes \`fallback_extractors=\`.

The classmethod resolves ontology + binding into a spec, calls \`build_runtime_extractor_registry\`, and constructs the manager with \`extractors=registry.extractors\`. Then assigns \`manager.runtime_registry = registry\` as the audit handle.

## Two attributes that always work together

- **\`manager.extractors\`** — the dict the runtime actually invokes (same dict \`run_structured_extractors\` already consumes).
- **\`manager.runtime_registry\`** — the audit handle. \`WrappedRegistry\` when constructed via \`from_bundles_root\`; \`None\` otherwise.

## Inherited \`use_ai_generate\` gate

\`extract_graph\` runs structured extractors only under \`if self.extractors and use_ai_generate\`. This gate pre-dates C2.c.2 — the bundle-wired path inherits it as-is. **When \`use_ai_generate=False\`, the compiled extractors do not run**, even if \`manager.extractors\` is fully populated from \`from_bundles_root\`. C2.c.2 deliberately does not change this gate so that the call-site swap is a pure substitution; a regression test pins the behavior so any future decoupling shows up as a deliberate change. See \`docs/extractor_compilation_orchestrator_swap.md#inherited-use_ai_generate-gate\`.

## Negative-path behavior

A bundle for \`event_x\` with no matching fallback is **not** registered (\`manager.extractors\` doesn't contain \`event_x\`); surfaced in \`manager.runtime_registry.bundles_without_fallback\`. Fail-closed per C2.c.1.

## Tests (8 cases in \`tests/test_ontology_graph_from_bundles_root.py\`)

- **\`TestOntologyGraphManagerDirectInit\`** (1) — direct \`__init__\` leaves \`runtime_registry = None\`; \`extractors\` identity preserved.
- **\`TestFromBundlesRootNoBundles\`** (1) — empty \`bundles_root\` → \`manager.extractors\` identity-preserves the fallback; \`runtime_registry.fallbacks_without_bundle\` lists uncovered event_types.
- **\`TestFromBundlesRootCompiledAndFallback\`** (1) — hand-written bundle + matching fallback → wrapped closure registered; calling it drives \`run_with_fallback\` and \`on_outcome\` fires with \`compiled_unchanged\`.
- **\`TestFromBundlesRootCompiledOnlyNoFallback\`** (1) — **negative case** — bundle without fallback NOT registered; surfaced in audit; behavioral check that \`run_structured_extractors\` yields empty.
- **\`TestFromBundlesRootEndToEnd\`** (1) — real BKA compiled bundle + real \`extract_bka_decision_event\` as fallback, fed through \`run_structured_extractors\` via \`manager.extractors\`. Both events produce \`compiled_unchanged\`; both \`mako_DecisionPoint\` nodes in the merged result.
- **\`TestFromBundlesRootExtractGraphCallSite\`** (3) — **the production call site itself**:
  - Monkeypatches \`_fetch_raw_events\` + \`_extract_via_ai_generate\` and calls \`manager.extract_graph(...)\`; asserts \`on_outcome\` fired. Proves \`extract_graph\` actually invokes the wrapped registry.
  - Same setup with a **non-empty** compiled bundle returning one \`ExtractedNode\`; asserts the structured node propagates through the merge into the final \`ExtractedGraph\`. Pins the compiled-output-is-merged behavior.
  - Pins the inherited \`use_ai_generate=False\` gate: structured extractors don't run, fetch hook isn't called, \`on_outcome\` doesn't fire.

## Verification

- 8/8 focused tests pass
- Full repo: 2708 passed / 10 skipped

## Out of scope (deferred)

- **BigQuery-table bundle mirror** for cross-process distribution. **C2.c.3.**
- **Revalidation harness** — scheduled / on-demand agreement check. **C2.d.**
- **Decoupling structured extraction from \`use_ai_generate\`** — separate scope decision; this PR inherits the pre-existing gate.

## Test plan

- [x] 8 focused tests pass
- [x] Direct \`__init__\` is back-compat (no behavior change; \`runtime_registry\` initialized to \`None\`)
- [x] \`from_bundles_root\` builds a registry and assigns it as audit handle
- [x] \`manager.extractors\` matches the registry's extractors dict
- [x] Compiled-only-no-fallback case fails closed (NOT registered, surfaced in audit)
- [x] End-to-end: real BKA bundle through the manager produces compiled outputs
- [x] \`manager.extract_graph(...)\` invokes the wrapped registry (production call site)
- [x] Non-empty compiled bundle: structured node propagates through extract_graph's merge into the returned ExtractedGraph
- [x] Inherited \`use_ai_generate=False\` gate pinned
- [x] \`on_outcome\` fires per event including \`compiled_unchanged\`
- [x] \`autoformat.sh\` clean


caohy1988:
Both findings addressed in 78df391.

**P2 — call site itself wasn't pinned.** Right — the previous end-to-end test stopped at \`run_structured_extractors(events, manager.extractors, manager.spec)\`, which proves the registry works but not that \`extract_graph\` actually reaches it. Added \`TestFromBundlesRootExtractGraphCallSite\` with two tests:

1. \`test_extract_graph_invokes_wrapped_registry_on_compiled_path\` — monkeypatches \`_fetch_raw_events\` (returns canned events) and \`_extract_via_ai_generate\` (returns empty graph), then calls \`manager.extract_graph(session_ids=[\"sess1\"], use_ai_generate=True)\`. Asserts \`on_outcome\` fired for each event with \`decision=\"compiled_unchanged\"\` — directly pins the production call site at line 655.

2. \`test_extract_graph_skips_structured_when_use_ai_generate_false\` — pins the inherited gate. Asserts \`_fetch_raw_events\` is NOT called and \`on_outcome\` doesn't fire when \`use_ai_generate=False\`.

**P3 — inherited \`use_ai_generate\` gate documented.** The gate is intentional-for-now (pre-dates C2.c.2; decoupling structured extraction from the AI flag is a separate scope decision). Added a new \"Inherited \`use_ai_generate\` gate\" section to \`docs/extractor_compilation_orchestrator_swap.md\` explicitly calling out:

- When \`use_ai_generate=False\`, structured extractors do NOT run, even if \`manager.extractors\` is fully populated from \`from_bundles_root\`.
- C2.c.2 deliberately does not change this gate — the call-site swap is a pure substitution.
- The new regression test pins the behavior so any future decoupling shows up as a deliberate change.

**Verification (head 78df391):**
- Focused module: 7 passed (was 5; +2)
- Full repo: 2707 passed / 10 skipped (was 2705 / 10; +2)

caohy1988:
Both items addressed in b550159.

**P3 — PR body refreshed.** Now reports 8 focused tests / 2708 full repo, the new \`TestFromBundlesRootExtractGraphCallSite\` group (3 cases including the merge-path test), and the "Inherited \`use_ai_generate\` gate" section. Test plan checklist updated.

**P3 (optional) — non-empty compiled bundle merge test.** You said it wouldn't block, but the gap was real, so added \`test_extract_graph_merges_compiled_nodes_into_result\`. Uses a hand-written bundle whose compiled extractor returns one \`ExtractedNode\` and marks the span fully handled; asserts the structured node propagates through \`extract_graph\`'s merge into the final \`ExtractedGraph\` returned to the caller. Pins the compiled-output-is-merged behavior even though the merge itself is pre-existing code.

**Verification (head b550159):**
- Focused module: 8 passed (was 7; +1 merge-path test)
- Full repo: 2708 passed / 10 skipped (was 2707 / 10; +1)

PR #155, opened 2026-05-12
examples(migration_v5): fixture foundation for the four-guarantee MAKO notebook (#107 phase 1)

caohy1988:
**Draft → ready: notebook fixtures + scaffold + live end-to-end run.**

Current HEAD: `e5c7d05` (env-aware install cell + default-dataset snapshots + Beat 4.4 framing soften).

## What this PR ships

1. **Fixture foundation** (`examples/migration_v5/`):
   - `mako_core.ttl` — the real MAKO ontology (authored input).
   - `mako_artifacts.py` — TTL → ontology / binding / DDL / property-graph SQL pipeline.
   - `mako_demo_agent.py` — ADK Gemini agent with 5 MAKO decision-flow tools + `BigQueryAgentAnalyticsPlugin` wiring.
   - `run_agent.py` — driver. Populates `agent_events` for N sessions.
   - `export_events_jsonl.py` — captures a deterministic offline snapshot of `agent_events` (subset of the BQ AA plugin schema; documented in the docstring).
2. **Generated artifacts** (regenerated by `mako_artifacts.regenerate_snapshots(project, dataset)`):
   - `ontology.yaml` — 18 MAKO entities, FILL_IN PKs resolved, dangling cross-namespace relationships dropped, `Candidate.extends: RoleTrait` stripped.
   - `binding.yaml` — 6 demo entities + 7 heterogeneous relationships (self-edges dropped).
   - `table_ddl.sql` — `CREATE TABLE IF NOT EXISTS` for every node + edge table including SDK metadata columns.
   - `property_graph.sql` — `CREATE OR REPLACE PROPERTY GRAPH` over those tables.
   - All three checked in against the default `(test-project-0728-467323, migration_v5_demo)` for reviewer readability; the notebook regenerates against a fresh `migration_v5_demo_<8-hex>` scratch dataset at runtime.
3. **The four-guarantee notebook** (`examples/migration_v5_demo_notebook.ipynb`) — 59 cells covering Sections 0-5. Live outputs included.

## Install cell

The install cell is **environment-aware** so the committed executed notebook stays clean across Colab, venv, and system Python:

- **Colab** → runs `%pip install -q ... "google-adk[vertexai]" ...` via `get_ipython().run_line_magic`. The `"..."` quoting bypasses zsh's `no matches found` glob trap that broke the previous draft.
- **Outside Colab** → skips the install (PEP 668-managed Python refuses `pip install` outside a venv). Checks that the required modules are importable; raises `RuntimeError` with a copy-paste install command if any are missing.

In this repo's homebrew Python the cell prints `"Local kernel — all required packages are importable."` and proceeds. In Colab it installs the deps fresh.

## Live run (scratch `migration_v5_demo_56f9179f`, 3 sessions)

| Beat | Cell evidence (from the committed executed notebook) |
|---|---|
| **0** | Scratch dataset created (TTL 1h); ontology/binding/DDL regenerated against it; DDL applied; agent runs 3 sessions; `agent_events` populated (counts by `event_type` printed). |
| **1 own** | Property graph applied from `property_graph.sql`. `before_skip_build_ts` captured. `ontology-build --skip-property-graph` runs against the 3 discovered sessions. Verify: **(a)** zero `CREATE OR REPLACE PROPERTY GRAPH` jobs after the timestamp, filtered by graph name + `sdk_feature='ontology-gql'` label; **(b)** GQL `DecisionExecution` count **before=0, after=2**. `property_graph_status='skipped:user_requested'`, **`rows_materialized total=38`**. |
| **2 validate** | `ALTER TABLE context_snapshot RENAME COLUMN snapshot_payload TO snapshot_payload_v2`. `binding-validate --format json` → exit 1, `missing_column` failure at `binding.entities[2].properties[1].column`. Restore + re-validate → exit 0, `report.ok=True`. Combined `ontology-build --skip-property-graph --validate-binding` → `property_graph_status='skipped:user_requested'`, `rows_materialized total=19`. |
| **3 cheap extraction** | Cell 3.2 baseline cost table prints per-session transcript chars + estimated tokens from real `agent_events`. Cell 3.6's synthetic `ExtractedGraph` triggers all three `FallbackScope` failures: `NODE + FIELD + EDGE all fired — scope classification works`. 3.3-3.5 + 3.7 print `Skipped: requires MAKO reference extractor at examples/migration_v5/reference_extractor.py — lands in follow-up PR #156`. |
| **4 resolve** | `gm compile --emit-concept-index` emits main + `__meta` DDL pair. Tables applied. `OntologyRuntime.from_files(..., compiler_version=<read from __meta>, ...)` — no hard-coded version; `LabelSynonymResolver.resolve("DecisionExecution")` returns 1 candidate, **`compile_id=ba9c1ff85870`**. **`GRAPH_TABLE count = 3`** against the user-authored property graph. Hub-shape `(DecisionExecution)-[partOfSession]->(AgentSession)` prints a "PR #156 follow-up" note (envelope-side `AgentSession` synthesis lands separately). |
| **5 close** | Four-guarantee recap table + issue pointers + what's-next. |

## Fixture / SDK reconciliation

The live run surfaced several places where the original draft's binding shape didn't survive contact with the materializer:

- **Per-entity PK columns**: `decision_execution_id`, `candidate_id`, `agent_session_id`, …. The materializer's `_relationship_columns` looks up edge FK columns via `src_prop_map[col].sdk_type` — that lookup requires the FK column name to exactly name a column on the source entity. Same convention the original V5 spec (`YMGO_Context_Graph_V3`) and `tests/fixtures/test_binding.yaml` use.
- **Self-edges dropped from the binding**: MAKO's `evolvedFrom` / `supersededBy` are `DecisionExecution → DecisionExecution`. The materializer's FK→PK lookup can't disambiguate them; the ontology still declares them, only the binding scope drops them.
- **Inheritance stripped**: `_strip_inheritance` post-processor removes `Candidate.extends: RoleTrait` from the imported ontology so `gm compile` v0 accepts the binding.
- **Property-graph DDL**: edge `REFERENCES` clauses use node-table **aliases** (not fully-qualified paths); edge tables carry explicit `KEY (src_col, dst_col)`; node `KEY (...)` and `REFERENCES alias (col)` both use the per-entity PK column.

## Notebook hardening

- Cell 1.4 + cell 2.5 subprocess calls surface the child's stderr on failure rather than masking it.
- Drift cells (2.3 / 2.4 / 2.5) have local `import json`.
- Beat 4 cell builds a `_PK_COL` map (entity → PK column) for the GQL `COLUMNS` accessor.
- Beat 4 hub-shape traversal degrades gracefully when zero edges materialize.

## What this PR is NOT

- **MAKO reference extractor** (`extract_mako_decision_event`). Lands in PR #156, flips Beat 3.3-3.5 / 3.7 from gated placeholders to live cells.
- **`AgentSession` node + `partOfSession` edge synthesis** from the plugin envelope. Lands with PR #156.
- `docs/README.md` / `CHANGELOG.md` entries — land alongside PR #156's reference extractor.

## Test plan

- [x] `mako_artifacts.py` regenerates clean: 18 ontology entities, 6 binding entities, 7 relationships. Committed snapshots target default `(test-project-0728-467323, migration_v5_demo)`.
- [x] `gm validate` clean on both ontology and binding.
- [x] `gm compile --emit-concept-index` emits the property-graph DDL + the concept-index `CREATE OR REPLACE TABLE` pair.
- [x] `mako_demo_agent` imports as `LlmAgent` with 5 tools + `BigQueryAgentAnalyticsPlugin`.
- [x] `run_agent.py --help` + `export_events_jsonl.py --help` clean.
- [x] All 24 notebook code cells parse via `ast.parse` (59 cells total — 35 markdown + 24 code).
- [x] **Live end-to-end run** (3 sessions, scratch `migration_v5_demo_56f9179f`) — full evidence above; outputs committed in the executed notebook. Install cell prints a clean "Local kernel — all required packages are importable" message.

Once #156 ships the MAKO reference extractor, the Beat 3 compile / runtime / savings placeholders flip live and the demo's cost story is end-to-end too.


caohy1988:
Updated target after the latest product clarification:

The fixture direction should be **TTL + runnable agent with BQ AA plugin instrumentation**, not **TTL + demo-specific event generator**.

Concretely, the PR should align to this contract before the notebook lands:

1. **Authored inputs**
   - `mako_core.ttl` — the real ontology.
   - A runnable demo agent, e.g. `mako_demo_agent.py`, that uses the BQ AA plugin and emits trace events to BigQuery.
   - Optional small config is fine if it is operational (`project`, `dataset`, model/env flags), but not an alternate hand-authored event corpus.

2. **Event source of truth**
   - Events should be generated by running the agent with the BQ AA plugin enabled.
   - `events.jsonl`, if kept, should be documented as a captured/exported snapshot for deterministic offline tests, not as the primary event generator.
   - `mako_agent.py` should not directly synthesize the final event stream as the main path. If a helper remains, split it into ontology/artifact generation vs. agent runtime.

3. **Ontology-driven behavior**
   - The agent should load or otherwise consume the TTL-derived model so the event content follows MAKO-declared classes/properties.
   - Payload fields inside the plugin trace should still obey the current rule: MAKO-declared data properties plus references to MAKO entities, with the BQ AA envelope supplied by the plugin.

4. **Generated artifacts**
   - `ontology.yaml`, `binding.yaml`, `table_ddl.sql`, `property_graph.sql`, and any exported sample events should be regenerated from `mako_core.ttl` + the runnable agent/artifact helper.
   - The notebook should run the agent against a fresh `migration_v5_demo_*` dataset, then use the plugin-populated `agent_events` table for Beats 0/3.

This means I would not sign off on the current fixture shape yet even after the self-edge/type fixes. The reshape is moving in the right direction for TTL-derived artifacts, but the event path still needs to become “agent writes traces through BQ AA plugin” rather than “fixture Python script returns deterministic events.”

Implementation suggestion:

- Rename/split current `mako_agent.py`:
  - `mako_artifacts.py` or similar: imports TTL, resolves `FILL_IN`, generates ontology/binding/DDL/property-graph SQL.
  - `mako_demo_agent.py`: runnable agent/workflow instrumented with the BQ AA plugin; emits MAKO-shaped events.
- Keep an optional `export_events_jsonl.py` or notebook cell that exports a small deterministic sample from BigQuery for offline revalidation demos.

This keeps the external proof-point honest: the notebook proves the SDK can consume real plugin traces produced by an agent that follows the real MAKO ontology.

PR #156, opened 2026-05-13
feat(scripts): add latency report and scope-aware quality eval

evekhm:
## Summary

- **latency_report.py** — Trace latency analyzer with A2A session stitching, 
  waterfall charts, and multi-trace summary stats (avg, P50, P95)
- **quality_report.py** — Adds `--config` flag for scope-aware evaluation: 
  out-of-scope questions scored as "declined" (correct) instead of "unhelpful" (bug). 
  Also adds `--output-json`, `--session`, per-agent quality breakdown, and A2A response resolution
- **Sample reports** showing actual script output with A2A protocol markers
- **README** expanded with usage examples and inline config file format

## Quality report improvements (latest commit)

- **Multi-turn support** — `get_user_input()` now returns the *last* user
  message in multi-turn sessions, keeping question/response pairs aligned
- **A2A null handling** — `get_a2a_response()` returns `"(no response)"`
  for null/empty A2A content instead of falling through to earlier turns
- **Always-present declined category** — The `response_usefulness` metric
  always includes the `declined` category, letting the LLM judge correctly
  classify legitimate scope declines even without `--config`
- **Auto-discovery** — `_load_agent_config()` now checks
  `eval/data/agent_context.json` in known locations when `--config` is not provided
- **Declined sessions in reports** — Markdown reports now include a
  "Declined Sessions" section alongside unhelpful and partial
- **14 new tests** covering `get_user_input` and `get_a2a_response`

## Consistent CLI flags

Both scripts now share common flags where applicable:

| Flag | quality_report | latency_report |
|------|:-:|:-:|
| `--limit` | ✅ | ✅ |
| `--time-period` | ✅ | ✅ |
| `--app-name` | ✅ | ✅ |
| `--session` | ✅ | ✅ |
| `--report` | ✅ | ✅ |
| `--output-json` | ✅ | ✅ |

## Config-driven scope eval

The quality report supports a `--config` flag pointing to a JSON file with 
`scope_decisions`. The script also auto-discovers `eval/data/agent_context.json`
from the repo root or script directory. No config file is committed — it is
specific to each agent deployment.

Without config: the `declined` category is still available for obvious refusals.
With config: the LLM judge is told exactly which topics are out of scope for
more accurate classification.

## Test plan

- [x] 41 unit tests pass (27 existing + 14 new)
- [x] Multi-turn `get_user_input` returns last message
- [x] A2A null/empty content returns "(no response)"
- [x] quality_report.py --session, --no-eval, --config modes
- [x] latency_report.py --limit, --session, --report modes


caohy1988 (changes requested):
## Comprehensive review — PR #156 head `8ab178a` against current `origin/main`

I checked the PR out locally, ran the tests (41 pass), and inspected each file. Solid PR — the multi-turn fix, A2A null handling, and per-agent breakdown are real value-adds and the helper tests are well-structured. Three real blockers, a handful of high-priority improvements, and a long tail of cleanup items.

---

## Blockers (must fix before merge)

### B1 — Indentation: `latency_report.py` uses 4-space, repo standard is 2-space

The entire 618-line file is at 4-space indent. `quality_report.py` and every other `scripts/*.py` file in the repo is 2-space (per `pyink-indentation = 2` in `pyproject.toml`). Lint confirms:

```
$ pyink --config pyproject.toml --check scripts/latency_report.py scripts/quality_report.py tests/test_quality_report_helpers.py
3 files would be reformatted.

$ isort --settings-path pyproject.toml --check-only scripts/latency_report.py scripts/quality_report.py
ERROR: scripts/latency_report.py Imports are incorrectly sorted and/or formatted.
ERROR: scripts/quality_report.py Imports are incorrectly sorted and/or formatted.
```

**Fix.** One autofix run:

```
pyink --config pyproject.toml scripts/latency_report.py scripts/quality_report.py tests/test_quality_report_helpers.py
isort --settings-path pyproject.toml scripts/latency_report.py scripts/quality_report.py
```

### B2 — `_load_agent_config` returns stale cached config across different `config_path` calls

Verified locally:

```python
c1 = _load_agent_config("config_a.json")   # returns A's content
c2 = _load_agent_config("config_b.json")   # returns A's content (still!) — BUG
```

`scripts/quality_report.py:178-188` uses a module-level `_AGENT_CONFIG = None` as the cache. The first call sets it; subsequent calls return that value regardless of `config_path`. The cache key should be `config_path`:

```python
_AGENT_CONFIG_CACHE: dict[str, dict] = {}

def _load_agent_config(config_path=None):
    key = config_path or "_AUTO_"
    if key in _AGENT_CONFIG_CACHE:
        return _AGENT_CONFIG_CACHE[key]
    ...
    _AGENT_CONFIG_CACHE[key] = result
    return result
```

For the CLI's primary code path (single `--config` per invocation) this is silent — but for tests, REPL exploration, or any long-lived process that loads multiple configs, the bug is real.

### B3 — `_load_agent_config` calls `sys.exit(1)` from a library-style function

`scripts/quality_report.py:184-185` does `logger.error(...) ; sys.exit(1)` when a config file is missing. But this function is called from `get_eval_metrics(config_path=...)` which is in turn called from tests and could be called from other scripts. Killing the interpreter from inside a helper makes the error path untestable and breaks composability.

**Fix.** `raise FileNotFoundError(config_path)` and let `main()` catch and exit. The test suite can then actually test the error path.

---

## High (should fix before merge)

### H1 — `parse_period()` in `latency_report.py:52-64` is dead code

Defined but never called. Confirmed via grep — only definition site, zero call sites. The actual period parsing happens inside the SDK via `TraceFilter.from_cli_args(last=args.time_period)`. Delete the function (and the inline `from datetime import timedelta` it contains).

### H2 — `latency_report.py` has zero tests

The PR adds 14 well-structured tests for `quality_report.py` helpers but the new 618-line `latency_report.py` ships with no coverage. Easy targets that don't need BQ:

- `format_ms` (None, sub-1000, sub-minute, multi-minute)
- `_span_label` (A2A interaction with/without remote, tool spans, plain spans)
- `_extract_text` (text_summary preference, function_call, A2A artifacts)
- `_build_json_output` (summary math: avg/p50/p95 with empty list, single trace, mixed agents)
- `render_summary_table` (same math, no-latency case)

The helper-pattern from `tests/test_quality_report_helpers.py` (fake `_FakeSpan`/`_FakeTrace`) is reusable here.

### H3 — `_load_agent_config` and `_build_scope_context` have no tests

The PR's headline feature is scope-aware eval. The 14 new tests cover `get_user_input` and `get_a2a_response` (both good additions), and skip the actual new code paths. At minimum:

- `_build_scope_context(None)` returns `""`
- `_build_scope_context({"scope_decisions": []})` returns `""`
- `_build_scope_context({"scope_decisions": [{"topic": "foo", "decision": "out_of_scope"}]})` produces a string containing `"foo"`
- `_load_agent_config("nonexistent.json")` raises (after B3 fix) — currently un-testable

### H4 — Python `latency_report.py` is missing `--env` even though the shell wrapper accepts it

`scripts/latency_report.sh` parses `--env path/to/.env`, sources the file, then strips it before calling Python. But `latency_report.py` itself doesn't declare a `--env` arg:

```
$ python scripts/latency_report.py --env path/to/.env
error: unrecognized arguments: --env path/to/.env
```

`scripts/quality_report.py` does support `--env` directly. The README claims "Or run the Python script directly: `python scripts/latency_report.py --limit 5 ...`" which is fine for the simple case but inconsistent if users assume the same flag set as the shell wrapper. Add `--env` to `latency_report.py`'s argparse (mirror `quality_report.py:1383-1391`) and call a `_load_dotenv(env_file=args.env)` helper.

---

## Medium (worth fixing in this PR)

### M1 — Unused import in tests

`tests/test_quality_report_helpers.py:25` does `import pytest` but nothing in the file uses it. Remove.

### M2 — `get_eval_metrics(config_path=None)` is missing a docstring

Sister functions in the same file are documented. The function also has a non-obvious behavior change worth surfacing in the docstring: `declined` is always present in `response_usefulness.categories` regardless of whether `config_path` is provided. The CHANGELOG/PR description says this, but the function signature doesn't.

### M3 — `get_a2a_response` inconsistently shapes its "no response" returns

- `c is None` → `("(no response)", agent)` ✓
- `c == {}` → `("(no response)", agent)` ✓
- `c == "not json"` → `(None, None)` ✗ (different shape)

Then `resolve_trace_responses` interprets the `(None, None)` case as "no A2A response found", which can leave the report missing an entry that should have been classified as "(no response)". Either align the return shape (return `("(no response)", span.agent or "remote_agent")` on JSON parse failure too) or document the asymmetry in the docstring.

### M4 — `resolve_trace_responses` marks `is_a2a=True` for `"(no response)"` cases

Sessions where the remote A2A agent returned nothing now get classified as A2A sessions in the report. That's probably the right semantic (the interaction happened, even if it returned nothing) but it'll show as a remote-agent attribution in per-agent stats. Worth a comment explaining the intent so a future reader doesn't "fix" it.

### M5 — `stitch_a2a_traces` does N serial BQ queries

For each `A2A_INTERACTION` span across all fetched traces, it issues `client.get_session_trace(remote_sid)`. With `--limit 100` and several A2A interactions per session, that's potentially hundreds of small BQ queries serially. Two mitigations:

1. Batch: collect all remote `sid`s into a single `TraceFilter(session_ids=[...])` query
2. Document the cost in the help text + README and consider a `--max-stitch` cap

Not a correctness bug, but a footgun for cost-sensitive users.

### M6 — `stitch_a2a_traces` mutates input traces in place

Extends `trace.spans`, mutates `rs.attributes['_a2a_source']`, reparents `rs.parent_span_id`. Caller may not expect this if they pass the same traces to another function later. Document this with a docstring line ("Mutates traces in place") or return new objects.

### M7 — `run_eval()` re-imports `json` locally

`scripts/quality_report.py:648` does `import json as _json` inside the `if args.session_ids_file:` block, even though `json` is already imported at module level (line 52). Just use the module-level `json`.

### M8 — `--session-ids-file` help references `run_eval.py` ambiguously

Help text says "the output of `run_eval.py`". The file actually lives at `examples/agent_improvement_cycle/eval/run_eval.py` — qualify the reference so users know where to find it.

### M9 — `--config` help advertises a `reason` field that's never read

Help shows `{"scope_decisions": [{"topic": "...", "decision": "out_of_scope", "reason": "..."}]}` but `_build_scope_context` only reads `topic` and `decision`. Either pass `reason` through to the LLM judge as additional context (could help the judge) or note in the help that `reason` is documentation-only.

### M10 — Skip set is a magic literal

`latency_report.py:172-173`:

```python
skip = {'USER_MESSAGE_RECEIVED', 'INVOCATION_STARTING', 'INVOCATION_COMPLETED'}
```

Lift to a module-level `_A2A_NOISE_EVENTS = frozenset({...})` with a comment explaining why these are filtered.

---

## Low / nits

### L1 — Markdown trailing-whitespace in sample reports

`scripts/sample_quality_report.md` and `sample_latency_report.md` use GFM hard-break `  ` (two trailing spaces) on heading metadata lines. `git diff --check` flags ~10 of these. They render as intended in GitHub markdown, but the repo convention is no trailing whitespace. Either add a `.gitattributes` rule for `*.md`, or convert to explicit `<br>` tags.

### L2 — Module-level imports scattered through `latency_report.py`

- Line 54: `from datetime import timedelta` inside `parse_period`
- Line 306: `from collections import Counter` inside `render_waterfall`
- Line 384: `_script_dir = os.path.dirname(...)` defined mid-file

Move all to the top of the file with the other imports.

### L3 — Logging vs print inconsistency between scripts

`quality_report.py` uses `logger.info`/`logger.warning`/`logger.error` throughout. `latency_report.py` uses `print()` for everything (errors, info, progress). Pick one style across both scripts. Suggestion: `latency_report.py` follows `quality_report.py` and uses `logger.*`.

### L4 — `latency_report.py` module docstring is incomplete

The Usage block doesn't list `--no-waterfall`, `--sdk-tree`, `--report`, or `--output-json` — they're only in the argparse help. Either complete the docstring or trim it to "see --help".

### L5 — `render_timing_tree` mutates `span.children`

`span.children = []` is reset on every call, and children are re-attached. Works, but the function has hidden state side effects. A pure function that returns a new `dict[span_id, list[children]]` would be cleaner.

### L6 — `render_waterfall` markers line is fragile

The marker line doesn't account for variable-width `format_ms` outputs. Worth a `# Approximate alignment; format_ms width varies` comment.

### L7 — `_print_eval_results` stores literal `"default (10/5/3)"` string in report details

`result["report"].details["samples"] = args.samples or "default (10/5/3)"` — the placeholder string ends up in the report's structured details. Either store `None` (and let downstream interpret it) or the actual default dict.

### L8 — `--time-period` default differs between scripts

`quality_report.py:1310` defaults to `"all"`, `latency_report.py:497-499` defaults to `None`. The PR's shared-flags story works better if these match.

### L9 — `.gitignore` adds `*.log` globally

Adding `*.log` repo-wide could swallow logs that should be tracked. Suggest scoping to `scripts/**/*.log` or `examples/**/*.log`.

### L10 — `latency_report.sh` env validation duplicated with the Python script

The shell wrapper at `latency_report.sh:50-57` validates `PROJECT_ID`/`DATASET_ID`/etc., and `latency_report.py:536-541` validates the same set. Pick one location (Python is preferred — better error messages, only one place to update).

### L11 — Stale base

`mergeable: MERGEABLE` but `mergeStateStatus: BEHIND` — 12 commits behind `origin/main`, none touching `scripts/`. No conflicts, but the rebase is a polite thing to do before merge.

---

## What looks great (positive reinforcement)

- **The multi-turn bug fix in `get_user_input`** (returning last user message) is exactly the right call and aligns with the existing `reversed()` pattern in `get_responding_agent`. The test `test_multi_turn_returns_last` pins it.
- **A2A null-handling fix in `get_a2a_response`** prevents the question/response misalignment from falling through to earlier turns — subtle bug, surfaced and tested clearly.
- **Auto-discovery of `eval/data/agent_context.json`** with explicit `--config` override is a nice ergonomic touch; the docstring originally described this behavior so the implementation aligns intent with reality.
- **The 41-test helper test file** is the right shape — pure-function tests with lightweight fakes, no BQ dependency. Easy to extend.
- **Per-agent breakdown table** with `[A2A:N/M]` tagging is genuinely useful for triaging "which agent is contributing the unhelpful responses" — turns a quality report from "summary stats" into "actionable signal."
- **Always-present `declined` category** — making the LLM judge able to classify polite refusals without forcing every user to write a config is the right default.
- **Consistent flag set across `quality_report` and `latency_report`** (`--limit`, `--time-period`, `--app-name`, `--session`, `--report`, `--output-json`) — once H4 lands (`--env` on `latency_report.py`), the two will be truly symmetric.
- **Sample reports committed** mean reviewers don't have to provision a project to see what the output looks like. Worth doing.

---

## Verdict

**REQUEST_CHANGES.** Three blockers (B1 indentation, B2 cache bug, B3 sys.exit from library code) are the must-fixes. H1-H4 are strongly recommended to land in this PR rather than as follow-ups. M1-M10 are tighter improvements. The L items are polish — bundle into the same push or skip.

The substantive design is good. Once B1-B3 + H1-H4 land, this is a clean merge.

caohy1988 (commented):
Fresh second-pass review on head `8ab178a`. I filtered out the items already covered in the existing `REQUEST_CHANGES` review and found two additional issues worth fixing.

### 1. Empty `--session-ids-file` silently falls back to recent sessions

`scripts/quality_report.py:647-656` loads the JSON file and can produce `session_ids = []` when the file is empty or when none of the objects contain `session_id`. Then `run_evaluation()` checks `elif session_ids:` at `scripts/quality_report.py:503`, so an empty list is treated the same as “no session filter provided” and the code falls through to the default time/limit query.

That is surprising and potentially expensive: a user intending to evaluate zero selected sessions can accidentally evaluate the latest 100 sessions.

Suggested fix: fail fast after loading the file when `session_ids == []`, or change `run_evaluation()` to distinguish `session_ids is not None` from “no filter provided.”

### 2. `latency_report.py --output-json -` emits mixed human text + JSON on stdout

`scripts/latency_report.py:604-608` supports `--output-json -`, but the script already prints the timing tree, waterfall, summary, and status lines to stdout before dumping the JSON. That makes the stdout stream invalid JSON, so this fails:

```bash
python scripts/latency_report.py --output-json - | jq
```

Suggested fix: when `args.output_json == "-"`, either suppress the human report, send human output to stderr, or add a `--quiet` mode and document that `-` produces clean machine-readable JSON.

evekhm:
## Review fixes applied (commit afcb307)

All review comments addressed. 78 tests pass, autoformat clean (pyink + isort).

### Bug fixes
- **B2** — Config cache now keyed by `config_path` (was stale singleton)
- **B3** — `_load_agent_config` raises `FileNotFoundError` instead of `sys.exit(1)` for missing explicit config path; `run_eval` catches it

### High priority
- **H1** — Removed dead `parse_period()` and unused `import re` from latency_report.py
- **H2** — Added `test_latency_report_helpers.py` (26 tests: `format_ms`, `_span_label`, `_extract_text`, `_build_json_output`, `render_summary_table`)
- **H3** — Added `TestBuildScopeContext` (6 tests) and `TestLoadAgentConfig` (4 tests)
- **H4** — Added `_load_dotenv()`, `_configure_logging()`, and `--env` CLI flag to latency_report.py

### Medium priority
- **M1** — Removed unused top-level `import pytest` from test file
- **M3** — `get_a2a_response` now always returns consistent `(str, str)` tuple (never `(None, None)` on parse failure)
- **M4** — Added comment explaining `is_a2a=True` for "(no response)" sessions
- **M6** — Docstring documents that `stitch_a2a_traces` mutates traces in place
- **M7** — Removed duplicate `import json as _json`
- **M9** — `--config` help text notes that `reason` field is documentation-only
- **M10** — Extracted A2A noise events to module-level `_A2A_NOISE_EVENTS` frozenset

### Low priority
- **L2** — Moved `from collections import Counter` and `_script_dir` to module level
- **L3** — Replaced `print()` with `logger.*` for operational messages in latency_report.py
- **L7** — `details["samples"]` stores `None` instead of misleading default string

### Second review
- **#1** — Fail fast with clear error when `--session-ids-file` has no valid IDs
- **#2** — `--output-json -` sends human output to stderr so stdout is clean JSON

### Formatting
All files pass `autoformat.sh` (pyink 2-space indent + isort google profile).

caohy1988 (changes requested):
## Fresh full review — PR #156 head `afcb307`

Checked locally against `origin/main`. **Excellent fix pass — 16 of 21 prior findings genuinely resolved.** Tests pass (78). The remaining open items are real but mostly small. One still-open blocker plus one new edge-case finding and a few NEW items I found while verifying everything else.

---

## Status of prior findings (verified one-by-one)

### Genuinely fixed

| ID | What was asked | Verification |
|---|---|---|
| B2 | Cache-by-config-path | Verified: `_load_agent_config("a.json")` and `_load_agent_config("b.json")` now return distinct configs |
| B3 | Raise instead of sys.exit | Verified: `FileNotFoundError` now raised; test `test_missing_explicit_path_raises` covers it |
| H1 | Delete dead `parse_period()` | Removed, plus unused `import re` |
| H2 | Tests for `latency_report.py` | 26 new tests in `test_latency_report_helpers.py` (format_ms, _span_label, _extract_text, _build_json_output, render_summary_table) |
| H3 | Tests for scope-context code | 6 + 4 new tests (`TestBuildScopeContext`, `TestLoadAgentConfig`) |
| H4 | `--env` in Python script | `python scripts/latency_report.py --env path/to/.env` now works; mirrors quality_report exactly |
| M1 | Remove unused `pytest` import | Gone from module level in `test_quality_report_helpers.py` |
| M3 | Consistent `get_a2a_response` return shape | Invalid JSON now returns `("(no response)", agent)` matching dict/null cases |
| M4 | Comment on `is_a2a=True` for "(no response)" | Added at `quality_report.py:455-456` |
| M6 | Document `stitch_a2a_traces` mutates | Docstring now says "Mutates *traces* in place." |
| M7 | Remove duplicate `import json as _json` | Gone |
| M9 | Document `reason` field is doc-only | Help text now says `"Only 'topic' and 'decision' are used; 'reason' is documentation-only."` |
| M10 | Extract magic skip set | `_A2A_NOISE_EVENTS = frozenset({...})` at module level |
| L2 | Move scattered imports | `Counter`, `_script_dir`, `datetime` all at top of `latency_report.py` |
| L3 | Logger instead of print | `latency_report.py` now uses `logger.info`/`logger.warning` for operational messages |
| L7 | Don't store literal "default (10/5/3)" | Stores `None` instead |

### Partially fixed

#### B1 — lint (still blocking)

Tests are clean (`tests/test_*.py` both pass pyink). But:

```
$ uv run pyink --check scripts/latency_report.py scripts/quality_report.py tests/test_quality_report_helpers.py tests/test_latency_report_helpers.py
would reformat scripts/latency_report.py
would reformat scripts/quality_report.py
2 files would be reformatted, 2 files would be left unchanged.

$ uv run isort --check-only scripts/latency_report.py scripts/quality_report.py
ERROR: scripts/latency_report.py Imports are incorrectly sorted and/or formatted.
ERROR: scripts/quality_report.py Imports are incorrectly sorted and/or formatted.
```

Root cause: `autoformat.sh` only runs the formatters over `${REPO_ROOT}/src/`, `${REPO_ROOT}/tests/`, `${REPO_ROOT}/examples/` (lines 24, 27). `scripts/` is excluded, so `bash autoformat.sh` only touched the tests. CI is green for the same reason — no workflow runs pyink/isort against `scripts/`.

Concretely, the scripts mix 2-space (newly added functions like `_configure_logging`, `_load_dotenv`) with 4-space (old functions like `_positive_int`, `format_ms`, the entire `main()` body in `latency_report.py`). `pyink --diff` shows the mismatch line-by-line.

**Two-part fix:**

1. `pyink --config pyproject.toml scripts/latency_report.py scripts/quality_report.py` + `isort --settings-path pyproject.toml scripts/*.py` and commit
2. Add `${REPO_ROOT}/scripts/` to `autoformat.sh` lines 24, 27 so this stays caught going forward

#### M8 — `run_eval.py` ambiguous reference

The code comment at `quality_report.py:658-659` was updated to qualify `examples/agent_improvement_cycle/eval/run_eval.py`. Good. But the argparse `--session-ids-file` help text at line 1395 still just says "the output of `run_eval.py`" with no path. Update the help text too.

### Not addressed (acknowledged still open from prior review)

- **M2** — `get_eval_metrics(config_path=None)` still has no docstring.
- **M5** — `stitch_a2a_traces` still does N serial `client.get_session_trace(remote_sid)` queries. Cost footgun for cost-sensitive users with `--limit 100`.
- **L1** — Markdown trailing whitespace in `sample_quality_report.md` and `sample_latency_report.md`. `git diff --check` still flags ~10 lines. These are intentional GFM hard-breaks but breach repo convention.
- **L4** — `latency_report.py` module docstring usage block still missing `--no-waterfall`, `--sdk-tree`, `--report`, `--output-json`, `--env`.
- **L5** — `render_timing_tree` still mutates `span.children`.
- **L6** — `render_waterfall` markers line still doesn't account for variable-width `format_ms` outputs.
- **L8** — `--time-period` default still differs: `quality_report.py` defaults to `"all"`, `latency_report.py` defaults to `None`.
- **L9** — `.gitignore` still adds `*.log` repo-wide rather than scoped.
- **L10** — Env validation still duplicated between `latency_report.sh` and `latency_report.py`.
- **L11** — Still 12 commits behind `origin/main`. `mergeable: MERGEABLE` (no conflicts), but the rebase is polite.

---

## NEW findings (this fresh review)

### N1 — `--output-json -` + `--sdk-tree` still pollutes stdout

`scripts/latency_report.py:639`:

```python
if args.sdk_tree:
    print(file=out)
    print("SDK Trace Tree:", file=out)
    print("─" * 70, file=out)
    trace.render(color=True)   # ← unconditionally prints to stdout
```

Verified in the SDK:

```python
# src/bigquery_agent_analytics/trace.py:733
output = "\n".join(lines)
print(output)        # no file= kwarg — always stdout
return output
```

The rest of the human-output path correctly routes via `print(..., file=out)` where `out = sys.stderr if json_to_stdout else sys.stdout`. But `trace.render()` ignores that and writes directly to stdout. Repro:

```bash
python scripts/latency_report.py --output-json - --sdk-tree | jq
# jq fails: the SDK tree text appears in the JSON stream
```

**Two fixes, pick one:**

- **(a)** Disallow the combination at the argparse layer:
  ```python
  if args.sdk_tree and args.output_json == "-":
      parser.error("--sdk-tree cannot be combined with --output-json - "
                   "(SDK render bypasses the JSON-to-stdout routing)")
  ```
- **(b)** Capture and redirect — `trace.render()` returns the string, so:
  ```python
  if args.sdk_tree:
      import contextlib, io
      buf = io.StringIO()
      with contextlib.redirect_stdout(buf):
          trace.render(color=True)
      print(buf.getvalue(), file=out)
  ```

Option (a) is cleaner. Option (b) is friendlier but requires the `contextlib` wrapper.

### N2 — `Trace.render()` is the API anti-pattern that enables N1

`src/bigquery_agent_analytics/trace.py:690-734`: `Trace.render()` both prints `output` to stdout AND returns it. The docstring acknowledges "Also printed to stdout for notebook/terminal use." This dual-purpose API is exactly what bites N1.

Out of scope for this PR, but worth filing an SDK follow-up to either:
- Add a `file=` keyword arg matching `print()`'s convention
- Split into `render() -> str` (pure) and `print_render(file=sys.stdout)` (side-effectful)

### N3 — README doesn't show `python latency_report.py --env`

H4 added `--env` to the Python script's argparse. But `scripts/README.md:182-196` only shows `--env` via the .sh wrapper (line 189). The direct-Python example at line 195 doesn't mention `--env`. Add one line:

```bash
python scripts/latency_report.py --env path/to/.env --limit 5
```

So users discover the Python-direct path supports `--env` without reading the source.

### N4 — Cache-isolation regression test missing for B2

`TestLoadAgentConfig.test_cache_hit` (test_quality_report_helpers.py:561-574) only verifies that loading the SAME path twice returns the same object (`first is second`). It doesn't prove that loading TWO DIFFERENT paths returns TWO DIFFERENT configs — which is the actual B2 fix.

Add:

```python
def test_cache_isolates_paths(self):
    import json as _json
    paths = []
    for content in [{"a": 1}, {"b": 2}]:
        with tempfile.NamedTemporaryFile(
            mode="w", suffix=".json", delete=False
        ) as f:
            _json.dump(content, f)
            paths.append(f.name)
    try:
        c1 = _load_agent_config(paths[0])
        c2 = _load_agent_config(paths[1])
        assert c1 != c2
        assert c1 == {"a": 1}
        assert c2 == {"b": 2}
    finally:
        for p in paths:
            os.unlink(p)
```

Without this, the B2 regression could come back silently — the existing tests would all still pass with the old `_AGENT_CONFIG = None` cache.

### N5 — `get_a2a_response` warning loses session-ID context

`quality_report.py:422-424`:

```python
logger.warning(
    "Failed to parse A2A payload for session, skipping"
)
```

The session ID is the most useful thing for debugging "which session?" — and the function has `trace` in scope, so `trace.session_id` is available. Update to:

```python
logger.warning(
    "Failed to parse A2A payload for session %s, skipping",
    trace.session_id,
)
```

---

## What looks great in this revision

- **Move of SDK imports into functions.** Both `latency_report.py` and `quality_report.py` now import `bigquery_agent_analytics` symbols inside `main()` / `get_client()` / `get_eval_metrics()`. This makes the module importable for tests without needing the SDK installed — which is what enables the new `test_latency_report_helpers.py` skip-if pattern. Nicely done.
- **`_load_dotenv` + `_configure_logging` extracted in `latency_report.py`** to mirror `quality_report.py` exactly. The two scripts now have a truly symmetric structure.
- **The `--output-json -` design: human output to stderr, JSON to stdout.** Standard Unix idiom. N1 is the one corner you missed; everything else is correct.
- **Test coverage for cache & scope context.** The `setup_method/teardown_method` clearing `_AGENT_CONFIG_CACHE` is the right call for test isolation.
- **`stitch_a2a_traces` mutation now documented** — short, clear, in the right place.
- **Cleaner skip-set extraction (`_A2A_NOISE_EVENTS`)** — module-level frozenset is more idiomatic than the inline literal.
- **`--session-ids-file` empty-file error.** The new `if not session_ids: logger.error(...); sys.exit(1)` at quality_report.py:665-670 turns a confusing later failure into a clear early one. Wasn't asked for in the previous review but a smart catch.

---

## Verdict — REQUEST_CHANGES (one real blocker)

The merge-blocking item is **B1** — `scripts/latency_report.py` and `scripts/quality_report.py` still fail `pyink --check` and `isort --check-only` because `autoformat.sh` doesn't cover `scripts/`. Two-line fix (run the formatters, add `scripts/` to autoformat.sh).

Everything else is improvement-level:
- **N1** is real but narrow (only the `--sdk-tree` + `--output-json -` combination); 5-line argparse guard would close it
- **N3/N4/N5/M8 partial** are 1-2 line fixes; bundle into the same push as B1
- **M2/M5/L1/L4-L11** are still-open from prior review — your call whether to fold them in here or punt to a follow-up

Once B1 lands plus the autoformat scope expansion, this is a clean merge. Substantively, the fix pass is very good — the author addressed the blockers, added genuinely useful tests (including the `setup_method`/`teardown_method` cache discipline), and even caught the empty-session-IDs edge that wasn't on the punch list.

evekhm:
## Second review fixes (commit e06a71a)

79 tests pass, `autoformat.sh` clean. All formatting now consistent 2-space indent.

### Blocking fix
- **B1** — Added `scripts/` to `autoformat.sh` scope (lines 24, 27). Both `scripts/quality_report.py` and `scripts/latency_report.py` now pass `pyink --check` and `isort --check-only`. Consistent 2-space indent throughout.

### New findings addressed
- **N1** — `--sdk-tree` + `--output-json -` now errors at argparse level (SDK's `trace.render()` prints directly to stdout, can't be redirected)
- **N3** — README now shows `python scripts/latency_report.py --env path/to/.env` in the direct-Python usage section
- **N4** — Added `test_cache_isolates_paths` — loads two different config paths, asserts they return distinct configs (regression guard for B2)
- **N5** — A2A JSON parse warning now includes session ID: `"Failed to parse A2A payload for session %s, skipping"`

### Prior findings addressed
- **M2** — `get_eval_metrics(config_path=None)` now has a docstring documenting both metrics and the always-present `declined` category
- **M8** — `--session-ids-file` help now qualifies the path: `examples/agent_improvement_cycle/eval/run_eval.py`
- **L4** — `latency_report.py` module docstring now lists all flags (`--no-waterfall`, `--sdk-tree`, `--report`, `--output-json`, `--env`)

### Still open (acknowledged, not blocking)
- M5 (batch A2A queries) — architecture improvement, follow-up
- L1/L5/L6/L8/L9/L10 — polish items
- L11 — rebase against main (no conflicts)
- N2 — SDK `trace.render()` API improvement (SDK follow-up)

evekhm:
## Third pass — all actionable items resolved (commit f19670e)

Rebased onto latest `origin/main` (0 commits behind). 79 tests pass, autoformat clean.

### Items fixed in this push

| ID | Fix |
|---|---|
| **L1** | Removed GFM trailing whitespace from `sample_quality_report.md` and `sample_latency_report.md` |
| **L6** | Added comment `# approximate alignment; format_ms width varies` to waterfall markers |
| **L8** | Clarified `--time-period` help text in latency_report.py to match quality_report.py pattern |
| **L9** | Scoped `.gitignore` `*.log` to `scripts/**/*.log` and `examples/**/*.log` only |
| **L10** | Removed duplicate env validation from `latency_report.sh` — Python script handles it with `logger.error()` |
| **L11** | Rebased onto `origin/main` — 0 commits behind |

### Items fixed in prior push (e06a71a) — reviewer hasn't seen yet

| ID | Fix |
|---|---|
| **B1** | Added `scripts/` to `autoformat.sh`; all scripts now consistent 2-space indent |
| **N1** | `--sdk-tree` + `--output-json -` errors at argparse level |
| **N3** | README shows `--env` for direct Python usage |
| **N4** | `test_cache_isolates_paths` regression test for B2 |
| **N5** | A2A parse warning includes session ID |
| **M2** | `get_eval_metrics` docstring added |
| **M8** | `--session-ids-file` help qualifies full `examples/agent_improvement_cycle/eval/run_eval.py` path |
| **L4** | `latency_report.py` docstring lists all flags |

### Items NOT fixed (with explanation)

| ID | Why not |
|---|---|
| **L5** | `render_timing_tree` mutates `span.children` — This is an intentional design: the function builds a parent-child tree from a flat span list, which requires setting `span.children`. Refactoring to return a separate dict would add complexity for no functional benefit. The function is only called once per trace in the render path. |
| **M5** | `stitch_a2a_traces` does N serial BQ queries — Correct observation, but batching requires the SDK to support `list_traces(session_ids=[...])` returning individual trace objects (currently it merges them). This is an SDK-level improvement worth a follow-up issue, not a scripts-layer fix. |
| **N2** | `Trace.render()` always prints to stdout — This is an SDK API design issue. The `--sdk-tree` guard (N1) prevents the user-facing problem. Filed mentally as SDK follow-up to add `file=` kwarg to `Trace.render()`. |

caohy1988 (approved):
## Fresh full review — PR #156 head `c872267`

Checked locally. **This PR is essentially ready to merge.** Every prior blocker and high-priority item is verified fixed. One small cosmetic regression worth a 5-minute fix, plus two acknowledged-deferred items the author can punt to follow-ups.

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, 0 commits behind `origin/main`. 79 tests pass.

---

## Status of every prior finding (verified)

### Blockers and high-priority — all fixed

| ID | What was asked | Verification on `c872267` |
|---|---|---|
| **B1** | pyink + isort clean on scripts/, autoformat scope expanded | `pyink --check` and `isort --check-only` both clean on all 4 touched files; `autoformat.sh:24,27` now includes `${REPO_ROOT}/scripts/` |
| **B2** | Cache by `config_path` | Verified — `test_cache_isolates_paths` (new, N4) proves two distinct paths return two distinct configs |
| **B3** | Raise instead of sys.exit | `FileNotFoundError` raised; `test_missing_explicit_path_raises` covers it |
| **H1** | Delete dead `parse_period()` | Gone, plus unused `import re` |
| **H2** | Tests for `latency_report.py` | 26 new tests in `test_latency_report_helpers.py` |
| **H3** | Tests for scope code | `TestBuildScopeContext` (6) + `TestLoadAgentConfig` (5, including the new cache-isolation test) |
| **H4** | `--env` on Python script | Works directly: `python scripts/latency_report.py --env path/to/.env` |

### New findings from last review — all fixed

| ID | What was asked | Verification |
|---|---|---|
| **N1** | Guard `--sdk-tree` + `--output-json -` | Verified at `latency_report.py:614-619`. Repro `python scripts/latency_report.py --output-json - --sdk-tree` now exits with `parser.error()` and a clear message |
| **N3** | README shows `python latency_report.py --env` | Added at `scripts/README.md:196` |
| **N4** | Cache-isolation regression test | `test_cache_isolates_paths` (`test_quality_report_helpers.py:576`) — asserts `c1 != c2` and the actual content match. Passes |
| **N5** | Session ID in A2A parse warning | `quality_report.py:437-438` now logs `"Failed to parse A2A payload for session %s, skipping", getattr(trace, "session_id", "?")` |
| **M2** | Docstring on `get_eval_metrics` | Multi-line docstring at `quality_report.py:241-250` documenting both metrics and the always-present `declined` category |
| **M8** | Qualify `run_eval.py` path in help | `quality_report.py:1476-1477` now says `"examples/agent_improvement_cycle/eval/run_eval.py"` |

### Polish items — fixed

- **L4** — `latency_report.py` module docstring now lists all 11 flags including `--no-waterfall`, `--sdk-tree`, `--report`, `--output-json`, `--env`
- **L6** — Waterfall markers line at `latency_report.py:380` has the comment `# Time axis labels (approximate alignment; format_ms width varies)`
- **L7** — `args.samples or None` instead of the literal `"default (10/5/3)"` string
- **L8** — `--time-period` default is now `"all"` on both scripts (`latency_report.py:560`, `quality_report.py:1405`). The conversion `"all" → None` happens inside the filter builder so runtime behavior is unchanged. Nice touch — the commit message explicitly calls out the alignment
- **L9** — `.gitignore` scoped: `scripts/**/*.log` and `examples/**/*.log` (was `*.log` globally)
- **L10** — `latency_report.sh` env validation removed; replaced with the comment `"Env validation is handled by the Python script with structured logging"`. End-to-end verified — running with empty env vars produces `[ERROR] Required environment variable PROJECT_ID is not set...` from the Python script
- **L11** — PR is rebased onto current `origin/main`, 0 commits behind. `mergeStateStatus: CLEAN`

### Acknowledged still open (author punted, reasonable)

- **M5** — `stitch_a2a_traces` still does N serial `client.get_session_trace(remote_sid)` queries. Documented cost footgun for cost-sensitive users; reasonable follow-up. Worth mentioning in the README's "A2A Session Stitching" section as a known cost characteristic.
- **L5** — `render_timing_tree` still mutates `span.children` on each call. Cosmetic. Not worth blocking on.

---

## Still-open small items found this pass

### L1 — Sample markdown trailing whitespace is partial regression

`f19670e "Address remaining review items"` scrubbed L1, but the next commit `31fe562 "Regenerate sample reports from live data"` regenerated the samples from live script runs and reintroduced two issues:

```
$ git diff --check origin/main...HEAD
scripts/sample_latency_report.md:5: trailing whitespace.
+**Generated:** 2026-05-19 03:50:22  
scripts/sample_latency_report.md:8: trailing whitespace.
+**Location:** us-central1  
scripts/sample_latency_report.md:9: trailing whitespace.
+**Sessions:** 3  
scripts/sample_quality_report.md:138: new blank line at EOF.
```

Three lines (5, 8, 9) of trailing whitespace in `sample_latency_report.md`, plus a trailing blank line at EOF in `sample_quality_report.md`. The author manually scrubbed some metadata lines after regenerating but missed three, and missed the EOF blank line entirely.

**Root cause is systematic, not a one-time miss.** Both `latency_report.py:_write_md_report` (line 496-500) and `quality_report.py:_write_md_report` (line 1070+) emit GFM hard-break metadata headers:

```python
w(f"**Generated:** {timestamp}  ")   # two trailing spaces = GFM <br>
w(f"**Project:** {project_id}  ")
...
```

And `_write_md_report` ends with `w("")` then `f.write("\n".join(lines) + "\n")`, producing a trailing blank line.

Any future "regenerate samples" will reintroduce both problems. Pick one of:

**Option A** — Stop using GFM hard-breaks in `_write_md_report`. The metadata lines are already on their own lines in the rendered output; the `  ` hard-break is decorative. Replace `f"**X:** {v}  "` with `f"**X:** {v}"` and drop the trailing `w("")` before the final write.

**Option B** — Add `.gitattributes`:

```
*.md whitespace=-trailing-space
```

which tells `git diff --check` to skip trailing-whitespace checking for `.md` files.

I'd recommend A — the hard-breaks render fine without them on GitHub because they're already separate paragraphs. But B is one line and forever-future-proof.

---

## What looks great in this revision

- **The `--time-period` alignment commit** (`c872267`) — the author shipped a small, well-scoped commit that aligns `latency_report.py` to default `"all"` matching `quality_report.py`, and explains in the commit message *why* runtime behavior is unchanged ("Convert 'all' back to None before building the TraceFilter"). This kind of clarity is the difference between "rename PR" and "rename PR with a footgun."
- **`autoformat.sh` scope expansion is durable.** Adding `${REPO_ROOT}/scripts/` to both `isort` and `pyink` lines means this can't regress on the next `scripts/*.py` PR. Better fix than just "run the formatter once" — it's a tooling fix not a manual one.
- **N1 guard is the right shape.** `parser.error("--sdk-tree cannot be combined with --output-json - (SDK trace.render() writes directly to stdout, bypassing the JSON-to-stdout routing)")` — names the constraint, explains why, fails fast at argparse layer. No silent stdout pollution possible.
- **L10 cleanup.** Removing duplicated env validation from `latency_report.sh` and leaving a comment pointing to where it now lives is exactly the right cleanup pattern — and the resulting shell wrapper is now 52 lines (was 59), with the comment "Env validation is handled by the Python script with structured logging" so the next reader knows it's intentional.
- **N4 cache-isolation test.** The author didn't just check the box — they wrote a test that proves the B2 fix works AND that catches its regression. `_AGENT_CONFIG_CACHE.clear()` in `setup_method`/`teardown_method` shows good test hygiene.
- **N5 fix uses `getattr(trace, "session_id", "?")`** rather than `trace.session_id` directly — defensive against a malformed fake trace. Small but thoughtful.
- **M8 fully addressed.** Both the code comment AND the argparse help text now qualify `examples/agent_improvement_cycle/eval/run_eval.py`. The previous round only fixed half.
- **The commit sequence is clean.** 7 commits, each with a focused commit message that maps directly to review items. `094ef7d`, `f19670e`, `c872267` each have bullet lists keyed by review ID. This is how to make a multi-round review easy on the reviewer.

---

## Verdict

**APPROVE with one minor non-blocking nit.**

The remaining L1 regression (3 trailing-whitespace lines + 1 EOF blank line in samples) is cosmetic, doesn't break anything, and is a 5-minute fix via Option A or B above. If you want to fold it in before merge, go ahead — otherwise it's safe to merge and clean up in a follow-up.

The substantive design and execution are good. The author's responsiveness across the review rounds (verified test additions, root-cause fixes like autoformat scope expansion rather than one-shot fixes, defensive coding choices like `getattr`) makes this the kind of PR that gets cleaner with each round rather than just shuffling problems.

haiyuan-eng-google (approved):

PR #164, opened 2026-05-15
test: live BigQuery idempotency proof for bqaa-materialize-window (#162 follow-up)

caohy1988:
Closes the explicitly-deferred live integration test from #162. **One linear pytest case** (``test_materialize_window_live_smoke_and_re_run``), gated on three env vars, fully self-contained with defensive teardown and a literal session-set re-discovery proof.

## Gating

Skipped unless ALL of these are set — CI must never accidentally create BigQuery resources:

* ``BQAA_LIVE_BQ=1`` — explicit opt-in (burns BigQuery quota).
* ``BQAA_LIVE_BQ_PROJECT`` — GCP project.
* ``BQAA_LIVE_BQ_SOURCE_DATASET`` — pre-populated ``agent_events`` (read-only).

Optional: ``BQAA_LIVE_BQ_LOCATION`` (default ``US``), ``BQAA_LIVE_BQ_DATASET_PREFIX`` (default ``bqaa_live_test_``).

## Isolation + cleanup

* **Unique scratch dataset per process** — ``<prefix><8-hex>`` so parallel runs don't collide and orphans are unambiguously test artifacts.
* **Module-scope fixture** creates the scratch dataset + bootstraps the entity-table DDL, ``yield``s, drops via ``delete_dataset(delete_contents=True, not_found_ok=True)`` in ``finally``. Tolerates partial DDL creation, streaming-buffer-pinned tables (atomic dataset-level drop bypasses buffer constraint), already-deleted state, and any teardown exception (swallowed + printed, never masks test result).
* **DDL retargeting** replaces the full canonical ``test-project-0728-467323.migration_v5_demo`` prefix with ``{_PROJECT}.{scratch}``, so the test is project-portable.
* **Source dataset untouched** — the test reads ``agent_events`` from ``BQAA_LIVE_BQ_SOURCE_DATASET`` and never writes there.

## Single linear test, two phases

Folded into one test body so the run-1 → run-2 dependency is explicit, not implicit across test files. Survives ``-k <name>`` selective execution and test-order randomization.

**Phase 1 — Run 1 (first materialization):**

* Returns ``ok=True``.
* Sessions discovered > 0 (catches an empty source).
* ``sessions_materialized == sessions_discovered``.
* Checkpoint advanced.
* Core entity tables (``decision_execution``, ``decision_point``, ``candidate``) populated. Empty here = silent no-op = test fails.
* Captures the materialized session **set** (``SELECT DISTINCT session_id FROM decision_execution``) as the run-2 baseline. Non-empty assertion guards against trivially-passing comparisons against empty sets.

**Phase 2 — Run 2 (re-run via wide overlap):**

* Returns ``ok=True``.
* Re-discovery uses ``overlap_minutes = lookback_hours * 60`` so the discovery query is guaranteed to cover any session run-1 saw, independent of seeded-event distribution.
* ``run2.sessions_discovered == run1.sessions_discovered`` — cheap cardinality pre-filter.
* **Materialized session set after run-2 == materialized session set after run-1** — the literal proof. A mutating source dataset that swapped sessions while keeping cardinality would pass count equality but fail set equality.
* Checkpoint does not regress (round-4 P2.1).
* Any table non-zero after run-1 is still non-zero after run-2 — silent zero-out is corruption.
* ``cleanup_status = "delete_failed"`` (if it appears) is surfaced in ``table_statuses`` — operator-visible diagnostic from #162 round-2 P2.2.

## Deferred (NOT proven here)

Strict byte-identical *row-count* equality across two consecutive runs requires waiting ~30-90 min for the BQ streaming buffer to drain. Documented in the test docstring as a deferred long-wait variant — a future nightly job can sleep between runs and assert byte equality. This PR does not imply it is proven.

## Verified run

Against ``test-project-0728-467323`` with ``BQAA_LIVE_BQ_SOURCE_DATASET=migration_v5_idem_43c51d05`` (3 demo sessions, 115 events):

* Test passes in 2:44.
* Run-1: 3/3 sessions, ``decision_execution=2``, ``candidate=7``, ``ok=True``. (The orchestrator processes 3 sessions; 2 produce ``DecisionExecution`` hub rows, 1 session had no hub entity in its extracted graph — surfaced via the hub-projection set rather than masked.)
* Run-1 materialized session set: ``['90161a5c-b4f7-4380-b66e-df6fe9bd4569', 'b2a6e4db-38e6-48bd-9144-1fca727e32b4']``.
* Run-2: 3 sessions re-discovered (== run-1), ``ok=True``, checkpoint stable at 17:55:19. 11 tables surfaced ``cleanup_status=delete_failed`` (streaming buffer pinned).
* Run-2 materialized session set: **identical to run-1's** — literal set-equality assertion passed.
* Scratch dataset auto-dropped on teardown — verified zero orphaned ``bqaa_live_test_*`` datasets remain.

## Test plan

- [x] Test passes with the env-var triplet set.
- [x] Test skips cleanly when any of the three required env vars is missing.
- [x] Scratch dataset is created and dropped within a single test run (verified via ``list_datasets`` post-teardown).
- [x] Source dataset is never modified.
- [x] Materialized session set is byte-identical between run-1 and run-2 — direct query proof, not orchestrator self-report.
- [x] Full focused suite still 68/68 (no regression in unit tests).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,91 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.3.0] - 2026-05-15
+
+### Release highlights
+
+Substantial feature release covering three major workstreams that landed
+between 0.2.3 and 0.3.0:
+
+- **Compiled structured extractors — full Phase C pipeline**
+  ([#75](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/75))
+  — deterministic source generation, JSON-to-plan parsing, LLM-driven plan
+  resolution, retry-on-gate-failure orchestration, runtime fallback wiring,
+  runtime extractor-registry adapter, orchestrator call-site swap,
+  compile-and-measure utility, revalidation harness, BigQuery-table bundle
+  mirror, ``bqaa-revalidate-extractors`` CLI with ``--events-bq-query-file``,
+  and an operational rollout guide. Replaces per-event LLM extraction with
+  deterministic code on the hot path while preserving the LLM fallback for
+  unrecognized event shapes.
+- **Ontology runtime reader**
+  ([#58](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/58)
+  reader follow-on to PR #92's concept-index emission) —
+  ``OntologyRuntime`` façade, ``EntityResolver`` Protocol with two reference
+  implementations (``ExactEntityResolver``, ``LabelSynonymResolver``), and
+  ``ConceptIndexLookup`` with fingerprint-strict verification across three
+  trust points (eager ``verify()`` at construction, explicit re-checks,
+  per-query ``WHERE compile_fingerprint`` defense in depth). Stable failure
+  codes (``FingerprintMismatchError``, ``MetaTableMissingError``,
+  ``MetaTableEmptyError``, ``MetaTableMultipleRowsError``).
+- **Binding + extraction validation toolkit**
+  ([#76](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/76),
+  [#105](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/105))
+  — ``validate_extracted_graph(...)``,
+  ``validate_extracted_graph_from_ontology(...)``,
+  ``validate_binding_against_bigquery(...)`` Python APIs;
+  ``bq-agent-sdk binding-validate`` CLI for pre-flight validation;
+  ``ontology-build --validate-binding`` and ``--location`` flags.
+
+Examples shipped alongside the SDK release: the MAKO four-guarantee notebook
+demonstrating the compiled-extractor pipeline end-to-end
+([#107](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/107)),
+and the A2A joint-lineage demo with auditor projections, the receiver
+``A2A_INTERACTION`` typed view, and an audit-analyst agent that closes the
+BQAA loop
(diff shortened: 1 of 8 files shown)


caohy1988:
Round 2 — addresses all three P2 findings (commit \`bed1607\`).

**P2 — DDL retargeting was project-blind.** Replace the **full** canonical ``test-project-0728-467323.migration_v5_demo`` prefix with ``{_PROJECT}.{scratch}``, not just the dataset segment. The previous version silently created entity tables in the canonical demo project whenever ``BQAA_LIVE_BQ_PROJECT`` differed.

**P2 — implicit run-1 → run-2 coupling between tests.** Merged the two tests into a single linear ``test_materialize_window_live_smoke_and_re_run``. Run-1 → assert → Run-2 → assert, all in one body. No fixture-state coupling across tests, no test-order-randomization fragility, no \`-k second_run\` weak-pass. Run-2 now asserts ``sessions_discovered > 0`` AND ``sessions_materialized == sessions_discovered`` explicitly.

**P2 — overlap window might miss sessions.** Run-2 uses ``overlap_minutes = lookback_hours * 60`` (i.e., overlap covers the whole lookback). And the test asserts ``run2.sessions_discovered == run1.sessions_discovered`` — if the overlap didn't catch the seeded events, this fails loudly instead of degenerating into a meaningless empty re-run.

## Verified run

Against ``test-project-0728-467323`` with ``BQAA_LIVE_BQ_SOURCE_DATASET=migration_v5_idem_43c51d05``:

- Run-1: 3/3 sessions, ``decision_execution=3``, ``decision_point=3``, ``candidate=11``, ``ok=True``.
- Run-2: **3 sessions re-discovered (== run-1's 3)**, ``ok=True``, checkpoint stable at 17:55:19. 11 tables surfaced ``cleanup_status=delete_failed`` (streaming-buffer pinning, expected).
- Row counts after run-2: still non-zero where they were non-zero after run-1 (silent zero-out check passed).
- Scratch dataset auto-dropped on teardown — verified zero orphaned ``bqaa_live_test_*`` datasets remain.

Time: 5:31.

haiyuan-eng-google (approved):

Issue #161, opened 2026-05-15
CLI: bqaa-materialize-window for scheduled graph refresh


caohy1988:
## Problem

Customers running the SDK in production want to keep their MAKO graph fresh on a schedule, but today's `ontology-build` requires them to pass `--session-ids` explicitly. That forces the customer to write their own discovery query, schedule it, and pipe the results — work that should be the SDK's job.

The customer thinks in **time windows** ("materialize the last 6 hours"), not session IDs. We should offer that directly.

This issue is the operational follow-up to #107 (the four-guarantee notebook is now live end-to-end via #155 + #157 + #160; the next product slice is making the demo's pipeline ergonomic to schedule).

## Proposed CLI

```
bqaa-materialize-window \
  --project-id my-project \
  --dataset-id analytics \
  --ontology ontology.yaml \
  --binding binding.yaml \
  --events-table agent_events \
  --lookback-hours 6 \
  --skip-property-graph \
  --validate-binding \
  --format json
```

### Flags (v0 + near-term)

| Flag | Purpose | Default |
|---|---|---|
| `--project-id` / `--dataset-id` | Standard BQ target | required |
| `--ontology` / `--binding` | YAML paths | required |
| `--events-table` | Source telemetry table | `agent_events` |
| `--lookback-hours` | Hours back from now | required (no default — explicit is safer) |
| `--state-table` | Checkpoint table (FQN). Writes `(last_checkpoint, run_id, sessions_materialized, ok)` rows. **Near-term, not optional later** — without it every run repeats the full lookback. | `_bqaa_materialization_state` (in `--dataset-id`) |
| `--overlap-minutes` | Re-process events newer than `(last_checkpoint - overlap_minutes)` to catch late-arriving rows | `15` |
| `--completion-event-type` | Treat sessions as done when this `event_type` appears. **Parameterized** so non-BQ-AA-plugin emitters aren't locked out. | `AGENT_COMPLETED` |
| `--include-active-sessions` | Materialize sessions without a completion event (partial coverage) | off |
| `--skip-property-graph` | Same semantic as `ontology-build` | **on by default** (refresh, not setup) |
| `--validate-binding` | Pre-flight binding-validate before extraction | on by default |
| `--bundles-root` | Compiled bundle directory | env `BQAA_BUNDLES_ROOT` or unset |
| `--reference-extractors-module` | Dotted module path for fallback | unset |
[...]

PR #165, opened 2026-05-15
examples(migration_v5): Cloud Run Job + Scheduler periodic materialization

caohy1988:
Operational wrapper around `bqaa-materialize-window` so a customer can keep their MAKO graph fresh on a cron with **one local command and one deploy command**. Follow-up to #162 (orchestrator) + #164 (live test).

## What ships

| File | Purpose |
|---|---|
| `examples/migration_v5/periodic_materialization/run_job.py` | Cloud Run Job entrypoint. Reads env vars, probe-first checks the graph dataset, retargets the demo binding, bootstraps entity-table DDL, calls `run_materialize_window` with a graph-dataset-co-located state table, emits structured JSON (`severity`/`message`/...) on stdout for Cloud Logging. |
| `deploy_cloud_run_job.sh` | Pre-creates the graph dataset, creates + IAM-scopes the runtime SA, vendors local SDK source into the staging dir, deploys via `gcloud run jobs deploy --source` with `--service-account`, wires Cloud Scheduler + invoker IAM. Optional `--smoke` runs the job once + tails logs. |
| `requirements.txt` | Local-dry-run-only deps. The deploy script generates its own staging-dir `requirements.txt` that installs from vendored SDK source (`./sdk_src`). |
| `periodic_materialization/README.md` | Production-shape diagram, local dry-run + deploy commands, inspection recipes (state table + Cloud Logging filter), full config reference, operational notes, troubleshooting. |
| `examples/migration_v5/README.md` (delta) | New "Run periodic materialization" section linking the new directory. |

## Production shape

```
agent_events            bqaa-materialize-window         graph entity/relationship
(events DS,        ────► (Cloud Run Job, every N hrs) ──► tables + _bqaa_materialization_state
 read-only)               │                              (graph DS, read+write)
                          ▼
                   structured JSON on stdout
                   (Cloud Logging)
```

## Local dry-run

Until a PyPI release containing `materialize_window` (added in #162) ships, install the SDK from local source:

```bash
# From the repo root
pip install -e .
pip install -r examples/migration_v5/periodic_materialization/requirements.txt

BQAA_PROJECT_ID=your-project \
BQAA_EVENTS_DATASET_ID=your_events_dataset \
BQAA_GRAPH_DATASET_ID=your_graph_dataset \
BQAA_LOOKBACK_HOURS=6 \
python examples/migration_v5/periodic_materialization/run_job.py
```

## Deploy

```bash
./examples/migration_v5/periodic_materialization/deploy_cloud_run_job.sh \
  --project your-project \
  --region us-central1 \
  --events-dataset your_events_dataset \
  --graph-dataset your_graph_dataset \
  --schedule "0 */6 * * *" \
  --smoke
```

## Design choices

- **Demo-owned, not SDK-internal.** The wrapper lives in `examples/` so the SDK stays focused on the orchestrator.
- **Local dry-run is first-class.** Same code path, same env-var shape; customers shake out env-var mistakes before paying for a deploy.
- **State table co-located with the graph dataset.** `run_job.py` passes `state_table="{project}.{graph_dataset}._bqaa_materialization_state"` explicitly, so the events dataset stays effectively read-only.
- **Entity-table DDL bootstrap at job startup (`CREATE TABLE IF NOT EXISTS`).** Idempotent — no separate one-time setup step.
- **Dataset existence check is probe-first.** `_ensure_graph_dataset()` calls `get_dataset` first; on success, returns `False` (the production path). Only on `NotFound` does it call `create_dataset`. Production runtime SA has only dataset-level `dataEditor` (which includes read), so it never needs project-level `datasets.create`. Local dry-run hits the create path on first run with the operator's gcloud creds.
- **Vendored SDK source in the deploy bundle.** Today's PyPI 0.3.0 release predates #162; rather than gate deploy on the next release shipping, the deploy script copies `src/bigquery_agent_analytics` + `src/bigquery_ontology` + `pyproject.toml` into the staging dir and the deploy-time `requirements.txt` installs from `./sdk_src`. The deployed image uses the same SDK code as the local dry-run.
- **Narrow IAM for the runtime SA.** Project-level `roles/bigquery.jobUser` (jobs.create only) + **dataset-level** `roles/bigquery.dataViewer` on events + **dataset-level** `roles/bigquery.dataEditor` on graph. The events dataset is genuinely read-only at the IAM layer. Deploy script grants these via `gcloud projects add-iam-policy-binding` (project) and `bq add-iam-policy-binding --dataset` (dataset).
- **Cloud Run Job runtime identity, not the default SA.** `gcloud run jobs deploy --service-account "$SA_EMAIL"` — the default Compute Engine SA wouldn't have the dataset-scoped grants.
- **Structured logging via `severity`.** All `_emit` calls write `{"severity": "INFO" | "ERROR", ...}` per the Cloud Logging spec — the entry's severity is set from the JSON key, not just buried in the payload.
- **Portable Bash.** No Bash 4-only features (`${var,,}` etc.) — runs on macOS's stock Bash 3.2 without "bad substitution".
- **Defensive arg parsing.** `require_arg` validates `${2-}` before assignment; `--project` at end of args produces a clean error, not `bash: $2: unbound variable`.
- **No Terraform.** Shell + `gcloud` is easier to read and copy than IaC for an example.
- **`--smoke` post-deploy verification.** "Did it deploy correctly?" is one command, not a six-hour wait for the first scheduled fire.

## Verified

**Local dry-run** against `test-project-0728-467323:migration_v5_idem_43c51d05` (3 demo sessions, 115 events) into a fresh `periodic_test_*` graph dataset:

```
{"severity": "INFO", "message": "periodic materialization starting", ...}
{"severity": "INFO", "message": "graph dataset created", "created": true}    # first run, NotFound → create
{"severity": "INFO", "message": "entity-table DDL applied", "statements": 13}
{"severity": "INFO", "message": "materialization complete", "sessions_discovered": 3, "sessions_materialized": 3, "ok": true, ...}
```

Re-run against the same dataset:

```
{"severity": "INFO", "message": "graph dataset exists", "created": false}    # subsequent run, get_dataset hit, no create call
```

Post-run table listing confirms `_bqaa_materialization_state` lives in the **graph** dataset (not events), and all 11 entity/relationship tables show `cleanup_status=deleted, insert_status=inserted, idempotent=true` on a fresh dataset.

**Vendored-SDK install** simulated end-to-end: staged `run_job.py` + artifacts + `sdk_src/` + generated `requirements.txt` into a temp dir, built a fresh venv, ran `pip install ./sdk_src`, confirmed `from bigquery_agent_analytics import materialize_window` works and `run_materialize_window` is importable.

**Arg-parse edge cases:**

```
$ ./deploy_cloud_run_job.sh --help; echo $?           # → 0
$ ./deploy_cloud_run_job.sh; echo $?                  # → 1, "--project is required"
$ ./deploy_cloud_run_job.sh --project; echo $?        # → 1, "--project requires a value"
$ ./deploy_cloud_run_job.sh --project --region us-central1; echo $?  # → 1, "--project requires a value"
$ ./deploy_cloud_run_job.sh --bogus-flag; echo $?     # → 1, "Unknown argument"
```

Scratch datasets cleaned up after verification.

## Not in scope

- Terraform / Pulumi templates.
- Compiled-bundle materialization path (`--bundles-root`).
- Backfill mode (out of scope per #161).
- Full live deploy + Scheduler trigger (deferred — quota-burning; the script's correctness is verified syntactically + via the staging simulation above).

## Test plan

- [x] Local dry-run against real BigQuery completes `ok=true` end-to-end.
- [x] DDL bootstrap idempotent (re-runs no-op on existing tables).
- [x] State table created in **graph** dataset, not events dataset.
- [x] `_ensure_graph_dataset` probe-first: `get_dataset` path for existing, `create_dataset` path for missing.
- [x] Cloud Logging entries use `severity` (uppercase).
- [x] Vendored SDK installs cleanly in a fresh venv; `run_materialize_window` importable.
- [x] `bash -n` syntax check passes on `deploy_cloud_run_job.sh`.
- [x] Arg parsing: `--help` exits 0; missing values / unknown flags exit 1 with clean error message.
- [x] No Bash 4-only features (portable to macOS Bash 3.2).
- [x] Missing-env `run_job.py` path returns exit 2 with structured JSON `severity: ERROR`.
- [ ] **(deferred)** Full deploy + Cloud Scheduler trigger against a real project.

### DIFF
--- a/examples/migration_v5/README.md
+++ b/examples/migration_v5/README.md
@@ -168,6 +168,40 @@ A live end-to-end notebook run (`run_agent.py --sessions 3` + Beat 1–4 cells a
 - **Beat 3.6**: synthetic `ExtractedGraph` triggers all three `FallbackScope` failures (`NODE + FIELD + EDGE`).
 - **Beat 4**: concept index emitted + applied; `LabelSynonymResolver.resolve("DecisionExecution")` returns 1 candidate with a 12-hex `compile_id`; `GRAPH_TABLE` count over the user-authored property graph is non-zero. Hub-shape `(DecisionExecution)-[partOfSession]->(AgentSession)` returns at least one row per current session — the compiled extractor wired in Beat 3.5 synthesizes the envelope-side `AgentSession` + `partOfSession`.
 
+## Run periodic materialization (Cloud Run Job + Cloud Scheduler)
+
+The notebook materializes a graph once, ad hoc. Real deployments want the graph kept fresh on a cron. `examples/migration_v5/periodic_materialization/` packages `bqaa-materialize-window` as a Cloud Run Job + Cloud Scheduler trigger, using the v5 demo binding (re-targeted to the customer's project at runtime).
+
+**Local dry-run** — exercise the path against your own BigQuery without deploying:
+
+```bash
+pip install -r examples/migration_v5/periodic_materialization/requirements.txt
+
+BQAA_PROJECT_ID=your-project \
+BQAA_EVENTS_DATASET_ID=your_events_dataset \
+BQAA_GRAPH_DATASET_ID=your_graph_dataset \
+BQAA_LOOKBACK_HOURS=6 \
+python examples/migration_v5/periodic_materialization/run_job.py
+```
+
+**Deploy** — one command, with `--smoke` to verify the deploy by running the job once and tailing logs:
+
+```bash
+./examples/migration_v5/periodic_materialization/deploy_cloud_run_job.sh \
+  --project your-project \
+  --region us-central1 \
+  --events-dataset your_events_dataset \
+  --graph-dataset your_graph_dataset \
+  --schedule "0 */6 * * *" \
+  --smoke
+```
+
+The deploy script bundles `run_job.py` + the demo artifacts (`ontology.yaml`, `binding.yaml`, `table_ddl.sql`) + `requirements.txt` into a staging dir, deploys via `gcloud run jobs deploy --source`, creates a service account, grants `roles/run.invoker`, and wires the Cloud Scheduler HTTP trigger.
+
+The job's JSON report (per run) lands in Cloud Logging as a structured entry. Filter on `resource.labels.job_name` for the materialization audit log. The state table at `<graph_dataset>._bqaa_materialization_state` is a queryable history.
+
+See [`periodic_materialization/README.md`](./periodic_materialization/README.md) for the full operational contract, env-var reference, and troubleshooting notes.
+
 ## What's NOT in this commit
 
 - `docs/README.md` / `CHANGELOG.md` entries — staged for a follow-up PR alongside the user-facing release notes.
--- a/examples/migration_v5/periodic_materialization/README.md
+++ b/examples/migration_v5/periodic_materialization/README.md
@@ -1,0 +1,314 @@
+# Periodic materialization — Cloud Run Job + Cloud Scheduler
+
+Run `bqaa-materialize-window` on a cron, against your own
+BigQuery project, with one local command and one deploy command.
+
+The MAKO demo (`examples/migration_v5/`) ships the ontology,
+binding, and entity-table DDL. This directory wraps them in a
+hands-off scheduled deployment: a Cloud Run Job that fires every
+N hours via Cloud Scheduler, materializes the last N hours of
+events into your graph dataset, and emits a structured JSON
+report to Cloud Logging.
+
+## Production shape
+
+```
+agent_events            bqaa-materialize-window         graph entity/
+(your events DS)  ────► (Cloud Run Job, every N hrs) ──► relationship tables
+                              │                          (your graph DS)
+                              ▼
+                       _bqaa_materialization_state
+                       (checkpoint / state table,
+                        co-located with the graph DS)
+```
+
+Per run, the orchestrator:
+
+1. Reads the prior checkpoint from `_bqaa_materialization_state`.
+2. Scans events in `[checkpoint - overlap_minutes, run_started_at)`,
+   capped at `lookback_hours` worth of history.
+3. Discovers terminal-event sessions (`event_type =
+   'AGENT_COMPLETED'`) and materializes them one at a time.
+4. Advances the checkpoint to the latest successful session's
+   completion timestamp (never past a failure — partial failure
+   leaves a tight high-water mark for the next run).
+5. Writes the JSON report to stdout for Cloud Logging.
+
+State-table semantics, overlap-windowed late-arrival handling,
+and idempotent retries are all in the SDK's design contract — see
+`src/bigquery_agent_analytics/materialize_window.py` for the
+full prose.
+
+## Prerequisites
+
+* GCP project with the BigQuery, Cloud Run, Cloud Scheduler, and
+  Cloud Build APIs enabled.
+* **Events dataset** (`BQAA_EVENTS_DATASET_ID`) already exists
+  with a populated `agent_events` table. The BQ AA plugin writes
+  to this; if you've never run an agent against BQAA, seed one
+  for this demo via `python examples/migration_v5/run_agent.py
+  --project YOUR_PROJECT --dataset YOUR_EVENTS_DS --sessions 3`.
+  This dataset is **read-only** for the periodic job — the
+  job never writes here.
+* **Graph dataset** (`BQAA_GRAPH_DATASET_ID`) — `run_job.py`
+  creates this on first invocation if missing (idempotent), so
+  you don't have to pre-create it. The entity/relationship
+  tables and the state/checkpoint table all live here.
+* `gcloud` authenticated with permissions to deploy Cloud Run
+  Jobs, create scheduler triggers, and grant IAM bindings.
+
+## Local dry-run
+
+Run the job once on your laptop against a real BigQuery project —
+no Cloud Run required. Useful for shaking out the env-var setup
+before paying for a deploy:
+
+```bash
+# From the repo root, install the SDK in editable mode. The
+# example uses bigquery_agent_analytics.materialize_window
+# (added in PR #162); this isn't in the 0.3.0 PyPI release
+# yet, so install from local until 0.4.0 ships.
+pip install -e .
+
+# Then install the example's ancillary deps:
+pip install -r examples/migration_v5/periodic_materialization/requirements.txt
+
+BQAA_PROJECT_ID=your-project \
+BQAA_EVENTS_DATASET_ID=your_events_dataset \
+BQAA_GRAPH_DATASET_ID=your_graph_dataset \
+BQAA_LOOKBACK_HOURS=6 \
+BQAA_OVERLAP_MINUTES=15 \
+python examples/migration_v5/periodic_materialization/run_job.py
+```
+
+Output is a single JSON line on stdout (the materialize-window
+report) — pipe through `jq` for readability:
+
+```bash
+... python run_job.py | jq .
+```
+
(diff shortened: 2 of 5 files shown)


caohy1988:
Round 2 — all four findings addressed (commit \`8d50173\`).

**P1 — state-table co-location.** \`run_job.py\` now passes \`state_table="{project}.{graph_dataset}._bqaa_materialization_state"\` explicitly. The events dataset stays read-only per the README contract. Verified post-run: state table lives in the graph dataset, not the events dataset.

**P1 — Cloud Run Job runtime service account.** Deploy script now creates \`bqaa-periodic-sa\` BEFORE deploy, grants it project-level \`roles/bigquery.user\` + \`roles/bigquery.dataEditor\`, and passes \`--service-account "$SA_EMAIL"\` to \`gcloud run jobs deploy\`. The SA serves both as job runtime identity AND scheduler caller — comment in the script documents the split-into-two-SAs hardening path for production.

**P2 — graph dataset bootstrap.** \`run_job.py\` now calls \`_ensure_graph_dataset()\` at startup (idempotent \`bigquery.create_dataset(exists_ok=False)\` with explicit \`Conflict\` catch so first vs subsequent runs are distinguishable in logs). README's prerequisites section now clarifies: events dataset is a prerequisite (must exist + be populated, read-only), graph dataset is auto-created on first run.

**P3 — severity, not level.** \`_emit\` now writes \`severity\` (the Cloud Logging-recognized key) using uppercase values (\`INFO\` / \`ERROR\`) per https://cloud.google.com/logging/docs/structured-logging. Docstring updated to point at the official severity reference.

## Verified end-to-end

Local dry-run against a fresh dataset \`periodic_test_verify1\` (didn't pre-create it):

\`\`\`
{"severity": "INFO", "message": "periodic materialization starting", ...}
{"severity": "INFO", "message": "graph dataset created", "graph_dataset_id": "periodic_test_verify1", "created": true}
{"severity": "INFO", "message": "entity-table DDL applied", "statements": 13}
{"severity": "INFO", "message": "materialization complete", "sessions_discovered": 3, "sessions_materialized": 3, "ok": true, ...}
\`\`\`

Post-run table listing:

- Graph dataset \`periodic_test_verify1\`: contains \`_bqaa_materialization_state\` ✅ (co-located, per contract).
- 11 entity/relationship tables: all \`cleanup_status=deleted, insert_status=inserted, idempotent=true\`.
- Scratch dataset auto-dropped after verification.

caohy1988:
Round 4 — both findings fixed (commit \`616dcce\`).

**P1 — \`bq add-iam-policy-binding\` needs \`--dataset\`.** Without the flag, \`bq\` parses \`PROJECT:DATASET\` as a malformed table identifier and the dataset-level grants fail. Added \`--dataset\` to both calls (events dataViewer + graph dataEditor) and noted the parsing behavior in a comment so the requirement is visible at the call site.

**P3 — stale staging comment.** Updated to reflect round 3's reality: the committed \`requirements.txt\` lists only the local-dry-run ancillary deps (\`google-cloud-bigquery\` + \`pyyaml\`); the deploy generates its own staging-dir \`requirements.txt\` that installs from \`./sdk_src\`.

\`bash -n\` + \`py_compile\` still clean.

caohy1988:
Round 5 — both findings fixed (commit \`f5b7591\`).

**P1 — \`get_dataset\` first, \`create_dataset\` only on \`NotFound\`.** Round 3 narrowed the runtime SA's IAM to \`dataEditor\` on the graph dataset (no project-level \`datasets.create\`). The previous \`create_dataset(exists_ok=False)\` + \`Conflict\`-catch flow would have failed in production: BigQuery checks IAM **before** existence, so the SA would get \`PermissionDenied\` on every run even though the dataset existed.

Fixed: probe with \`get_dataset\` (read-only, the SA has access). On \`NotFound\`, attempt \`create_dataset\` — which works in local dry-run (operator's gcloud creds, broad perms) and explicitly fails in production if someone skipped the deploy script's pre-create step. Both paths verified end-to-end:

- Fresh dataset name → \`"graph dataset created", "created": true\` (NotFound → create succeeded).
- Re-run on the same name → \`"graph dataset exists", "created": false\` (get_dataset hit, no create call). This is the path the production SA takes every run.

**P3 — \`--help\` now exits 0.** \`usage()\` takes an optional exit-code arg; \`-h|--help\` calls \`usage 0\`, unknown / missing-arg paths call \`usage 1\`. Verified:

\`\`\`
$ ./deploy_cloud_run_job.sh --help; echo $?
... usage text ...
0
$ ./deploy_cloud_run_job.sh --bogus-flag; echo $?
Unknown argument: --bogus-flag
... usage text ...
1
$ ./deploy_cloud_run_job.sh; echo $?
... usage text ...
1
\`\`\`

\`bash -n\` + \`py_compile\` + \`pyink --check\` clean.

caohy1988:
Round 6 — both findings fixed (commit \`e7f69dc\`).

**P2 — \`set -u\` trap on trailing-edge flags.** A user typing \`./deploy_cloud_run_job.sh --project\` (no value) at the end of the args would previously crash with \`bash: \$2: unbound variable\` from the \`PROJECT="\$2"\` assignment. Added a \`require_arg\` helper that reads \`\${2-}\` defensively and fails with a clean \`Error: --project requires a value\` + usage + exit 1.

Implemented as an inline check (not via \`\$(require_arg ...)\`) so the \`exit 1\` inside \`usage 1\` terminates the script, not just a subshell. The check also catches \`--project --region X\` (where the next token is another flag) via the \`value == -*\` guard.

Verified:

\`\`\`
$ ./deploy_cloud_run_job.sh --project; echo $?
Error: --project requires a value.
... usage ...
1
$ ./deploy_cloud_run_job.sh --project --region us-central1; echo $?
Error: --project requires a value.
... usage ...
1
$ ./deploy_cloud_run_job.sh --help; echo $?
... usage ...
0
\`\`\`

**P3 — \`_ensure_graph_dataset\` docstring.** Updated to match the round-5 probe-first implementation: production path is \`get_dataset\` succeeds → return False → no \`create_dataset\` call (so the runtime SA never needs \`datasets.create\`); local dry-run hits \`NotFound\` on a fresh dataset and creates it. The "Conflict" claim was stale.

\`bash -n\` + \`py_compile\` + \`pyink --check\` clean.

caohy1988:
Round 7 — both findings fixed (commit \`accc2b0\`).

**P2 — \`\${var,,}\` is Bash 4-only.** macOS ships Bash 3.2 by default; \`\${VAR,,}\` raises "bad substitution" there. Replaced with portable \`tr '[:upper:]_' '[:lower:]-'\`, which also handles the underscore-to-dash transform (\`EVENTS_DATASET\` → \`--events-dataset\`) in one pass.

Verified the message stays readable across all required-arg cases:

\`\`\`
$ ./deploy_cloud_run_job.sh
Error: --project is required (use --help).
$ ./deploy_cloud_run_job.sh --project foo
Error: --region is required (use --help).
$ ./deploy_cloud_run_job.sh --project foo --region us-central1 --graph-dataset bar --schedule "* * * * *"
Error: --events-dataset is required (use --help).
\`\`\`

**P3 — stale PR body bullet.** Rewrote the "Dataset auto-creation" bullet to match the round-5/6 probe-first reality: \`get_dataset\` first → \`False\` for existing (production path), only \`NotFound\` falls through to \`create_dataset\` (local dry-run path). Also added two new design-choice bullets covering the Bash 3.2 portability + \`require_arg\` defensive parsing fixes, and a verified arg-parse edge-cases section.

\`bash -n\` + \`py_compile\` + \`pyink --check\` clean.

haiyuan-eng-google (approved):

PR #166, opened 2026-05-16
examples(migration_v5): live Cloud Run deploy verification + IAM fixes (#165 follow-up)

caohy1988:
Follow-up to #165. Actually deployed the periodic-materialization Cloud Run Job + Cloud Scheduler trigger against `test-project-0728-467323`, captured the evidence, and fixed four real-world issues the live run surfaced.

## The four bugs the live verification caught

| # | Issue | Symptom | Fix |
|---|---|---|---|
| 1 | IAM propagation race | `gcloud iam service-accounts create` returns success, `gcloud projects add-iam-policy-binding` fails with `INVALID_ARGUMENT: Service account ... does not exist` ~1s later | `_retry_iam` helper retries on the "does not exist" error class with 3s backoff, 20 attempts |
| 2 | `bq add-iam-policy-binding` needs project allowlisting | `This feature requires allowlisting` on some projects | Switched dataset-level grants to a Python heredoc using `google-cloud-bigquery`'s `AccessEntry` API. Idempotent + portable |
| 3 | Buildpacks build / runtime entrypoint mismatch | (a) `web process not found in Procfile` with `job:` only. (b) `provide a main.py or app.py file` without a Procfile. (c) `Application failed to start: container exited abnormally` (no Python output) when `--command/--args` skipped Buildpacks' venv-activation wrapper | `web: python run_job.py` Procfile (the label is a Buildpacks convention, doesn't imply HTTP) + no `--command/--args` override |
| 4 | Runtime SA missing `roles/aiplatform.user` | The MAKO demo calls `AI.GENERATE`. Without the grant: `user does not have the permission to access resources used by AI.GENERATE` and the orchestrator silently extracts empty graphs | Added project-level `roles/aiplatform.user` grant to the deploy script |

## What the verification proved

* Deploy script ran clean from `bq mk` through Cloud Scheduler creation.
* Cloud Run Job built from the vendored SDK source (`Building bigquery-agent-analytics @ file:///workspace/sdk_src`).
* Cloud Scheduler created with the correct OAuth-authenticated HTTP target pointing at the Cloud Run Job's `:run` endpoint.
* Runtime SA can read `agent_events` (READER on events dataset, verified via IAM policy inspection).
* Runtime SA can write graph tables (WRITER on graph dataset).
* Runtime SA cannot write to events dataset (zero WRITE/OWNER bindings — read-only contract holds at the IAM layer).
* Cloud Logging structured output works: `severity`, `message`, plus the full `materialization complete` JSON payload.
* Cloud Scheduler trigger fires the cron successfully: a real cron firing at `2026-05-16T06:00:00Z` produced a third state-table row with the same materialization shape — proving end-to-end without manual intervention.
* State table (`_bqaa_materialization_state`) lives in the **graph** dataset, not events. Audit log shows 3 runs: deploy-script's `--smoke` (pre-aiplatform fix), manual re-execute (post-fix, full materialization), and the cron-triggered execution.

## Known issue (out of scope — SDK follow-up)

The orchestrator reports `sessions_materialized == sessions_discovered` and `ok=true` even when every per-event `AI.GENERATE` call failed. The first state-table row from the verification proves this: AI.GENERATE failed for every session yet `ok=true, sessions_materialized=3`. `rows_materialized` is empty in that case, but the top-level `ok` doesn't reflect the silent failure.

Documented as a known issue in the README with a workaround: alert on `jsonPayload.rows_materialized == {}` in Cloud Logging as a second-line check.

## Test plan

- [x] Live deploy + Scheduler trigger against `test-project-0728-467323` end-to-end.
- [x] `--smoke` execution completes ok.
- [x] Real Cloud Scheduler cron firing succeeds (06:02 UTC firing on 2026-05-16).
- [x] IAM policies match documented contract (events READER, graph WRITER, zero events WRITE).
- [x] State-table audit log shows expected rows.
- [x] All GCP resources cleaned up after verification (job, scheduler, graph dataset).
- [x] `bash -n` syntax clean.

## Not in scope

* Terraform / Pulumi.
* SDK-side fix for the silent AI.GENERATE failure (flagged in README, not fixed here).
* Backfill mode.
* Compiled-bundle materialization path.

### DIFF
--- a/examples/migration_v5/periodic_materialization/README.md
+++ b/examples/migration_v5/periodic_materialization/README.md
@@ -56,6 +56,15 @@ full prose.
   tables and the state/checkpoint table all live here.
 * `gcloud` authenticated with permissions to deploy Cloud Run
   Jobs, create scheduler triggers, and grant IAM bindings.
+* `python3` on PATH. The deploy script uses Python to apply
+  dataset-level IAM via the BigQuery client's `AccessEntry`
+  API (since `bq add-iam-policy-binding` requires project
+  allowlisting in some environments). If your `python3` doesn't
+  have `google-cloud-bigquery` installed, the script
+  transparently creates a one-shot temp venv with it — no
+  manual install required. If it does (e.g., you ran
+  `pip install -e .` from the repo root), the script reuses
+  that directly.
 
 ## Local dry-run
 
@@ -127,6 +136,14 @@ The script:
 3. **Grants narrow IAM** to the SA:
    * Project-level `roles/bigquery.jobUser` —
      `bigquery.jobs.create` only.
+   * Project-level `roles/aiplatform.user` — required because
+     the MAKO demo's extraction path calls BigQuery's
+     `AI.GENERATE` function (Gemini-backed entity extraction).
+     Without this grant, the AI call returns "user does not
+     have the permission to access resources used by
+     AI.GENERATE" and the orchestrator silently extracts an
+     empty graph for every session. Surfaced by the live
+     verification in PR #166.
    * Dataset-level `roles/bigquery.dataViewer` on
      **events** — read-only access. The events dataset stays
      effectively read-only per the contract above.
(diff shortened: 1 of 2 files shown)


caohy1988:
Round 2 — P1 fixed (commit \`f7e054e\`).

**P1 — \`python3\` dependency on \`google-cloud-bigquery\` was undocumented.** The deploy script's Python heredoc for dataset-level IAM imports \`google.cloud.bigquery\`, but a clean local \`python3\` doesn't have it. Without a preflight, the script either fails opaquely or makes the "one command" promise from the README false.

Fix: new **section 0 preflight** runs BEFORE any IAM mutation. Logic:

\`\`\`bash
PY_CMD="python3"
if ! python3 -c "import google.cloud.bigquery" >/dev/null 2>&1; then
  echo "==> creating temp venv with google-cloud-bigquery (for dataset IAM)"
  IAM_VENV="$(mktemp -d -t bqaa-iam-venv-XXXXXXXX)"
  python3 -m venv "$IAM_VENV" >/dev/null
  "$IAM_VENV/bin/pip" install --quiet google-cloud-bigquery
  PY_CMD="$IAM_VENV/bin/python"
fi
\`\`\`

\`_grant_dataset_iam\` invokes \`"$PY_CMD" - <<EOF\` so it uses either the operator's existing \`python3\` (fast path, if they ran \`pip install -e .\`) or the temp venv (one-shot ~15s install on first deploy from a clean shell).

The unified \`_cleanup\` trap removes both the staging dir AND the IAM venv on any exit path — no orphans.

**Verified end-to-end with a clean python3.** Simulated the "import fails" case via a PATH-shimmed fake python3:

\`\`\`
==> creating temp venv with google-cloud-bigquery (for dataset IAM)
venv python has bigquery: ok
\`\`\`

Updated **prerequisites in the README** to call this out: only \`gcloud\` + \`python3\` are required from the operator; \`google-cloud-bigquery\` install is automatic if missing.

\`bash -n\` clean.

PR #167, opened 2026-05-16
fix(materialize_window): empty-extraction sessions are not ok (#166 follow-up)

caohy1988:
Closes the silent-failure mode #166's live deploy surfaced. When `AI.GENERATE` failed per-event (e.g., runtime SA missing `roles/aiplatform.user`), the SDK swallowed the error and returned an empty graph for every session. The orchestrator then reported `sessions_materialized == sessions_discovered`, `ok=true`, and an EMPTY `rows_materialized` dict. Operators monitoring on `jsonPayload.ok` saw "all good" while the entity tables stayed empty.

## The fix

After `materialize_with_status` succeeds, check the total rows materialized. If zero, classify the failure mode via `table_statuses` and break the loop (same shape as an exception — advance checkpoint only to the prior successful session):

* **`empty_extraction`** — `table_statuses` empty or all `rows_attempted == 0`. Extraction (AI.GENERATE or compiled bundle) returned an empty graph; nothing was inserted. Operator action: check `roles/aiplatform.user`, AI.GENERATE quotas, or whether the session's events legitimately have any MAKO content.

* **`materialization_failed`** — any `rows_attempted > 0` with `insert_status == "insert_failed"`. Extraction produced rows but every insert returned an error. `error_detail` names the specific tables (e.g., `DecisionExecution: rows_attempted=3, insert_status='insert_failed', cleanup_status='deleted'`). Operator action: dataset write-perm regression on the runtime SA, schema drift the binding-validate pre-flight missed, or streaming-buffer pinning.

Operators chasing the wrong failure mode (AI vs schema) is exactly what the classifier prevents.

## Failed-session table_statuses now surface in the aggregate report

`_build_result` previously aggregated `table_statuses` only from `r.ok` sessions. A `materialization_failed` session's per-table diagnostic was getting dropped from `result.table_statuses` and was only visible buried in `error_detail`. Fixed: aggregate `table_statuses` from EVERY session (the status surface is operator-visible regardless of session success). `rows_materialized` still aggregates only `r.ok` sessions — correct semantics for "rows that actually landed".

## What does NOT change

The empty-window heartbeat path (`sessions_discovered == 0`) is untouched. The empty-extraction guard is per-session and doesn't fire when no sessions were processed. A cron firing into an idle window still reports `ok=true`.

## Test cases (4 total)

| Test | Scenario | Expected |
|---|---|---|
| `test_all_sessions_zero_rows_reports_not_ok` | Every discovered session extracts to empty (e.g., AI permission missing). `row_counts={}`, `table_statuses={}` | `ok=false`, first session reported as `empty_extraction`, loop breaks (sessions_failed=1, not N — no waste of BQ quota), error_code matches |
| `test_partial_extraction_partial_failure_conservative_checkpoint` | Session 1 extracts rows, session 2 empty, session 3 never reached | `ok=false`, `sessions_materialized=1`, `sessions_failed=1`, checkpoint advances ONLY to session 1's timestamp, materializer called exactly twice |
| `test_insert_failure_classified_as_materialization_failed` | Empty `row_counts` but `table_statuses` shows `rows_attempted=3, insert_status='insert_failed'` (insert-failure shape) | `error_code="materialization_failed"` (NOT empty_extraction). `error_detail` names DecisionExecution + Candidate + `insert_failed`. Failed session's `table_statuses` appear in `result.table_statuses` at the top level |
| `test_empty_window_remains_ok` | `sessions_discovered == 0` (idle window) | `ok=true`, `failures == []` — the per-session guard does NOT flip the heartbeat case |

## #166 README cleanup

The "Known issue surfaced by this verification" section in `periodic_materialization/README.md` documented the now-fixed silent-failure mode and recommended a `rows_materialized == {}` workaround. Replaced with a "Failure-mode surface (post-#167)" section that documents both `error_code` values, what each means, and what operator action to take. The primary signal is now `jsonPayload.ok=false` plus `jsonPayload.failures[].error_code` — no second-line check needed.

## Verification

* `pytest tests/test_materialize_window.py` — **72/72 pass** (68 prior + 4 new).
* Full focused suite — **2931 pass**, 15 skipped (15th being the existing live-deploy test from #164, unchanged).
* `pyink --check` + `isort --check-only` clean.
* No existing test broke.

## Operator-visible change

**Before:** `jsonPayload.ok = true` even when every AI.GENERATE call failed silently OR every insert failed. Operators had to alert on `rows_materialized == {}` as a second-line check, and couldn't tell whether the failure was AI/IAM or schema/write-perm without digging into log payloads.

**After:**

* `jsonPayload.ok = false` is the unmistakable red signal.
* `jsonPayload.failures[].error_code` is either `empty_extraction` or `materialization_failed` — direct operator-actionable classification.
* `jsonPayload.failures[].error_detail` names the specific failing tables for the materialization-failed case.
* `jsonPayload.table_statuses[<table>]` carries the per-table diagnostic at the top level even for failed sessions.
* CLI exits 1.

## Out of scope (per reviewer)

* `--check-deploy-inputs` doctor command — separate later PR.
* v5 demo README flow + customer deployment checklist — next PR per reviewer's sequencing.
* Configuration knob to disable the guard for use cases where empty extraction is expected — hold until a real customer asks.

## Test plan

- [x] All 4 new test cases pass.
- [x] All 68 prior `test_materialize_window.py` tests still pass.
- [x] Full focused suite 2931 pass, 15 skipped.
- [x] `pyink --check` + `isort --check-only` clean.
- [x] #166 README updated to reflect the new contract.

### DIFF
--- a/examples/migration_v5/periodic_materialization/README.md
+++ b/examples/migration_v5/periodic_materialization/README.md
@@ -397,23 +397,38 @@ ff1e956df8b8    2026-05-16 04:38:59    3 / 3 / 0                       true
 
 (Row 1 is the deploy script's `--smoke` execution, which ran
 BEFORE the verification added `roles/aiplatform.user` to the
-deploy. AI.GENERATE failed for every session there, but the
-orchestrator still reported `ok=true` with empty
-`rows_materialized` — see the known issue below.)
+deploy. AI.GENERATE failed for every session there. At the time
+of #166, the orchestrator reported `ok=true` with empty
+`rows_materialized` — the silent-failure mode that PR #167
+fixed. Today, the same situation would produce `ok=false` with
+`failures[0].error_code = "empty_extraction"`.)
 
-### Known issue surfaced by this verification
+### Failure-mode surface (post-#167)
 
-The orchestrator currently reports `sessions_materialized ==
-sessions_discovered` and `ok=true` even when every per-event
-`AI.GENERATE` call failed. The `rows_materialized` dict is
-empty in that case, but `sessions_materialized` doesn't
-reflect the underlying failure. Operators monitoring on
-`jsonPayload.ok` would miss the silent extraction failure.
+The orchestrator distinguishes two zero-row session outcomes
+that look identical from `rows_materialized` alone:
 
-Tracked for SDK follow-up — out of scope for this example PR.
-Workaround: alert on
-`jsonPayload.rows_materialized == {}` in Cloud Logging /
-Monitoring as a second-line check.
+* **`empty_extraction`** — extraction (AI.GENERATE or compiled
+  bundle) returned an empty graph; no inserts attempted.
+  Diagnose by checking the runtime SA's `roles/aiplatform.user`
+  grant, AI.GENERATE quotas, or whether the session's events
+  legitimately had any MAKO content.
+
+* **`materialization_failed`** — extraction produced rows but
+  every insert returned an error. The `failures[].error_detail`
+  names the specific tables (e.g.,
+  `DecisionExecution: rows_attempted=3, insert_status='insert_failed'`),
+  and the aggregate `table_statuses` carries the per-table
+  diagnostic at the top level of the report. Diagnose by
+  checking the SA's dataset write perm on the graph dataset,
+  schema drift the binding-validate pre-flight missed, or
+  streaming-buffer pinning.
+
+In both cases: `ok=false`, CLI exit 1, the cron run shows up
+as a failed execution in Cloud Monitoring. Alert directly on
+`jsonPayload.ok=false` plus `jsonPayload.failures[].error_code`
+for the failure-mode breakdown — no second-line
+`rows_materialized == {}` check needed.
 
 ## Not in scope here
 
--- a/src/bigquery_agent_analytics/materialize_window.py
+++ b/src/bigquery_agent_analytics/materialize_window.py
@@ -921,6 +921,81 @@ def run_materialize_window(
             "insert_status": ts.insert_status,
             "idempotent": ts.idempotent,
         }
+      # Zero-rows guard. ``row_counts`` only includes tables
+      # whose insert SUCCEEDED (per ``ontology_materializer.py``
+      # — see the ``inserted_tables`` filter at the build site).
+      # So ``total_rows == 0`` collapses two distinct failure
+      # modes that operators must triage differently:
+      #
+      #   * **empty_extraction** — the graph itself was empty.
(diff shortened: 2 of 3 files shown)


caohy1988:
Round 2 — both findings fixed (commit \`ac1393d\`).

**P2 — \`empty_extraction\` conflated with \`materialization_failed\`.** \`OntologyMaterializer\` only populates \`row_counts\` for tables whose insert succeeded (see \`ontology_materializer.py:803\`), so \`total_rows == 0\` collapses two distinct failure modes. Classify them via \`table_statuses\`:

* \`empty_extraction\` — \`table_statuses\` empty or all \`rows_attempted == 0\`. Operator action: AI/IAM / extraction issue.
* \`materialization_failed\` — any \`rows_attempted > 0\` with \`insert_status == "insert_failed"\`. Operator action: dataset write-perm / schema mismatch / streaming-buffer pinning.

The \`error_detail\` for \`materialization_failed\` names the specific tables (e.g. \`DecisionExecution: rows_attempted=3, insert_status='insert_failed', cleanup_status='deleted'\`) so operators don't chase the wrong root cause.

**Failed-session \`table_statuses\` now surface in the aggregate report.** \`_build_result\` was only aggregating from \`r.ok\` sessions, so a \`materialization_failed\` session's per-table diagnostic was getting dropped from \`result.table_statuses\` and was only visible buried in \`error_detail\`. Fixed: aggregate \`table_statuses\` from EVERY session (the status surface is operator-visible regardless of session success); \`rows_materialized\` still aggregates only successful sessions (correct semantics for "rows that actually landed").

**New regression \`test_insert_failure_classified_as_materialization_failed\`** — mocks \`materialize_with_status\` returning a result with empty \`row_counts\` AND \`table_statuses\` showing \`rows_attempted=3, insert_status='insert_failed'\`. Asserts:

- \`error_code == "materialization_failed"\` (not the wrong \`empty_extraction\` classification).
- \`error_detail\` names the specific tables that failed.
- The failed session's \`table_statuses\` appear in the aggregate \`result.table_statuses\` — at the top level of the report, not just buried in error_detail.

**P2 — #166 README known-issue section.** Replaced the now-fixed "alert on \`rows_materialized == {}\`" workaround with the new "Failure-mode surface (post-#167)" section that documents both error codes, what they mean, what operator action to take, and confirms the primary signal is now \`jsonPayload.ok=false\` plus \`failures[].error_code\`. Updated the inline reference in the smoke-execution evidence to point at the new section.

72/72 focused tests pass (71 prior + 1 new). Full suite 2931 pass.

Issue #187, opened 2026-05-19
Tracking: MAKO production rollout — Yahoo design-partner asks for v6


mikulbhatt:
This is a tracking issue for ten production-readiness asks raised by Yahoo Monetization Platform as a design partner for `migration_v5` and `periodic_materialization`. Filed ahead of the joint Yahoo / Google Cloud blog post (target: on or before June 15, 2026).

All ten issues were identified during Yahoo's internal review of PRs #166, #167, #168, and #170. None are blocking; the priority tiers below reflect the June 15 window.

---

### File now — strongest engineering substance, broadest community benefit

| Issue | Title | Key source file |
|---|---|---|
| #177 | `--backfill --from/--to` mode with `--state-key-suffix` | [`run_job.py`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/examples/migration_v5/periodic_materialization/run_job.py) |
| #178 | Compiled extractor as default; `AI.GENERATE` as fallback | [`reference_extractor.py`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/examples/migration_v5/reference_extractor.py) |
| #179 | Self-edge support in `ontology_materializer` | [`ontology_materializer.py`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/src/bigquery_agent_analytics/ontology_materializer.py) |

### This week — pairs naturally with existing PRs

| Issue | Title | Key source file |
|---|---|---|
| #180 | Orphaned-session watchdog for runs without `AGENT_COMPLETED` | [`run_job.py`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/examples/migration_v5/periodic_materialization/run_job.py) |
| #181 | Add `RejectionReason` + `hasRejectionReason` to demo | [`mako_core.ttl`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/examples/migration_v5/mako_core.ttl) |

### Before publish — deploy-script hardening

| Issue | Title | Key source file |
|---|---|---|
[...]

PR #193, opened 2026-05-20
feat(deploy): compiled-only mode on Cloud Run deploy path (#187)

caohy1988:
## Summary

Closes the gap that PR #192 explicitly carved out: the SDK accepted
`--extraction-mode=compiled-only` but `deploy_cloud_run_job.sh` rejected
it because no path staged a structured-extractor module into the Cloud
Run image. This change ships the deploy path end-to-end, so the
customer-facing story is finally true: **one deploy path, no AI calls,
no Vertex AI IAM**.

Part of the #187 (Yahoo design-partner asks) sequencing:
A (#188) → B1 (#189) → C1 (#191) → B2 (#192) → **this** → C2 → D/E/F/G.

### SDK (`materialize_window._build_manager` + boundary check)

Adds a reference-only sub-path to `_build_manager`:

| `bundles_root` | `reference_extractors_module` | Behavior |
|---|---|---|
| None | None | AI-only (existing legacy) |
| None | dotted path | **NEW**: module's `EXTRACTORS` dict registers on the manager via `from_ontology_binding(extractors=...)` — no compiled bundles, no AI |
| set | dotted path | Compiled bundles + reference fallback (existing) |
| set | None | `ValueError` at boundary (existing) |

`run_materialize_window` now also boundary-rejects
`extraction_mode="compiled-only"` when **both** `bundles_root` and
`reference_extractors_module` are `None` — otherwise compiled-only
would silently emit empty graphs (no extractors registered, no AI
fallback) instead of the typed `empty_extraction` failures the mode
guarantees.

### Cloud Run wrapper (`run_job.py`)

Reads `BQAA_REFERENCE_EXTRACTORS_MODULE` and `BQAA_BUNDLES_ROOT` env
vars and threads them through to `run_materialize_window`.

### Deploy script (`deploy_cloud_run_job.sh`)

* Stages `reference_extractor.py` next to `run_job.py` in **both**
  modes — so flipping an existing ai-fallback deploy to compiled-only
  doesn't need a separate code-staging step.
* Sets `BQAA_REFERENCE_EXTRACTORS_MODULE=reference_extractor` on the
  env-var array when `--extraction-mode=compiled-only`.
* Drops the compiled-only reject; the validator now accepts both modes
  and only rejects operator typos.
* Makes `roles/aiplatform.user` conditional on `--extraction-mode`:
  * `ai-fallback`: `add-iam-policy-binding` (unchanged behavior).
  * `compiled-only`: skip the add **and** defer the
    `remove-iam-policy-binding` until **after** the new revision is
    deployed AND (if `--smoke`) the smoke execution passes. A failed
    deploy / failed smoke leaves the existing schedule's IAM intact so
    the cron stays recoverable. ⟶ addresses review P1 #1.
  * The remove itself distinguishes the binding-absent case
    (gcloud's stable `Policy binding not found` stderr) — treated as
    success — from real failures (`PERMISSION_DENIED`, transient
    gcloud errors, org-policy rejects), which fail loudly with the
    gcloud exit code, indented stderr, and a remediation hint
    pointing at the manual `remove-iam-policy-binding` invocation.
    ⟶ addresses review P2.

### Docs

* `--help` text + `README.md` rewritten to describe the conditional
  IAM grant and drop the "follow-up will lift this" language.

## Tests

`tests/test_materialize_window.py`: **109 passing** (was 86 before
this PR; +23 new). `tests/test_extract_graph_diagnostics_modes.py`
+ `tests/test_binding_explicit_fk_mapping.py` unchanged and still
pass (54 tests).

### New: `TestBuildManagerReferenceOnly` (3 tests)
* `test_reference_only_mode_loads_extractors_from_module` — pins
  the new sub-path: stub a module with `EXTRACTORS = {...}`,
  assert it's threaded into `from_ontology_binding(extractors=...)`.
* `test_reference_only_mode_rejects_empty_extractors` — pins the
  boundary validator: empty `EXTRACTORS` raises `ValueError`.
* `test_legacy_no_extractors_path_still_works` — pins that the
  default `bundles_root=None, reference_extractors_module=None`
  path still constructs the manager with no extractors (AI-only
  path).

### New on `TestExtractionModeFlag`
* `test_compiled_only_without_extractor_source_rejected` — pins
  the SDK boundary rejection (review P1 #2).
* `test_compiled_only_with_reference_module_accepted` — mirror
  case proving the boundary check only rejects the both-missing
  combination.

### `TestDeployScriptExtractionModeBoundary` (rewritten)
* **Dropped** the obsolete `test_compiled_only_rejected_with_actionable_error`.
* **Added** `test_compiled_only_now_accepted_by_validator` (the direct
  inverse).
* **Added** `test_compiled_only_wires_reference_extractor_env_var`
  (static: env var is behind the `EXTRACTION_MODE == "compiled-only"`
  guard).
* **Added** `test_compiled_only_skips_aiplatform_user_grant`
  (static: ai-fallback guard + `remove-iam-policy-binding`).
* **Added** `test_reference_extractor_present_for_compiled_only_staging`
  (the file must exist at the path the deploy script copies from).
* **Added** `test_compiled_only_iam_remove_is_deferred_until_after_deploy`
  (source-order check: remove line is below the
  `gcloud run jobs deploy` line) — review P1 #1.
* **Added** `test_smoke_failure_propagates_before_iam_remove`
  (static: `SMOKE_RC` capture + `exit "$SMOKE_RC"` propagation) —
  review P1 #1.
* **Added** `test_compiled_only_iam_remove_distinguishes_absent_from_failure`
  (static: `REMOVE_RC` capture, `Policy binding not found` match,
  `exit "$REMOVE_RC"` propagation, negative assertion that the
  legacy `--quiet 2>/dev/null || true` swallow is gone) — review P2.
* Retained the `bash -n` syntax check and the catch-all typo-reject
  test.

### Updated: `TestCompiledOnlyMakesZeroLLMCalls`
Docstring updated to drop "until the follow-up" language now that
the follow-up has landed. Existing 6 compiled-only test callsites
now pass `reference_extractors_module=_TEST_REF_MODULE` (a stub
module registered at test-module import time) so they satisfy the
new SDK boundary check.

## Review history

| Commit | Findings addressed |
|---|---|
| `0dd53b3` | Initial implementation |
| `6a95bcc` | P1 #1 (defer IAM remove until after deploy / smoke succeeds) + P1 #2 (SDK boundary check on compiled-only without extractor source) |
| `f25faed` | P2 (IAM remove fails loud on real errors; only the absent-binding case is treated as success) |

## Test plan
- [x] `python -m pytest tests/test_materialize_window.py` — 109 pass
- [x] `python -m pytest tests/test_extract_graph_diagnostics_modes.py tests/test_binding_explicit_fk_mapping.py` — 54 pass
- [x] `bash -n examples/migration_v5/periodic_materialization/deploy_cloud_run_job.sh` — clean
- [x] `python -m pyink --check` + `python -m isort --check-only` — clean
- [ ] Live Cloud Run smoke (`--extraction-mode=compiled-only --smoke`) on a real GCP project (reviewer / merger-side; the deploy-script tests cover the boundary contract but can't exercise real `gcloud`)

### DIFF
--- a/examples/migration_v5/periodic_materialization/README.md
+++ b/examples/migration_v5/periodic_materialization/README.md
@@ -101,7 +101,7 @@ makes it a small edit.
 | Scope | Role | Why |
 |---|---|---|
 | Project | `roles/bigquery.jobUser` | Run BQ jobs (DDL, discovery, state writes) |
-| Project | `roles/aiplatform.user` | Call `AI.GENERATE` for entity extraction |
+| Project | `roles/aiplatform.user` | Call `AI.GENERATE` for entity extraction. Granted in `--extraction-mode=ai-fallback` (default); skipped (and idempotently removed if pre-existing) in `--extraction-mode=compiled-only` |
 | Cloud Run Job | `roles/run.invoker` | Cloud Scheduler invokes the job. Granted on the specific job resource, not project-wide |
 | Events DS | `roles/bigquery.dataViewer` | Read `agent_events`; events stay read-only |
 | Graph DS | `roles/bigquery.dataEditor` | Write entity/relationship tables + `_bqaa_materialization_state` |
@@ -260,14 +260,21 @@ The script:
 3. **Grants narrow IAM** to the SA:
    * Project-level `roles/bigquery.jobUser` —
      `bigquery.jobs.create` only.
-   * Project-level `roles/aiplatform.user` — required because
-     the demo's extraction path calls BigQuery's
-     `AI.GENERATE` function (Gemini-backed entity extraction).
-     Without this grant, the AI call returns "user does not
-     have the permission to access resources used by
-     AI.GENERATE" and the orchestrator silently extracts an
-     empty graph for every session. Surfaced by the live
-     verification in PR #166.
+   * Project-level `roles/aiplatform.user` — required in the
+     default `--extraction-mode=ai-fallback` mode because the
+     demo's extraction path calls BigQuery's `AI.GENERATE`
+     function (Gemini-backed entity extraction). Without this
+     grant, the AI call returns "user does not have the
+     permission to access resources used by AI.GENERATE" and
+     the orchestrator silently extracts an empty graph for every
+     session. Surfaced by the live verification in PR #166. In
+     `--extraction-mode=compiled-only` the SDK never calls
+     `AI.GENERATE` (proven by `TestCompiledOnlyMakesZeroLLMCalls`),
+     so the deploy script skips this grant **and**
+     idempotently removes any pre-existing
+     `roles/aiplatform.user` binding from the same SA, so a
+     customer who flips an existing deploy from ai-fallback to
+     compiled-only ends up with no Vertex AI dependency at all.
    * Dataset-level `roles/bigquery.dataViewer` on
      **events** — read-only access. The events dataset stays
      effectively read-only per the contract above.
--- a/examples/migration_v5/periodic_materialization/deploy_cloud_run_job.sh
+++ b/examples/migration_v5/periodic_materialization/deploy_cloud_run_job.sh
@@ -693,19 +693,57 @@ fi
 # ``roles/aiplatform.user`` and its schedule keeps working
 # during a botched transition.
 #
-# ``remove-iam-policy-binding`` is idempotent only when the
-# binding exists; we route stderr to ``/dev/null`` and
-# ``|| true`` so a never-granted SA (the common first-deploy
-# case) doesn't tank the script.
+# ``remove-iam-policy-binding`` is naturally non-idempotent:
+# gcloud returns non-zero with "Policy binding not found" if the
+# binding doesn't exist. We treat *that specific error* as
+# success (the common first-deploy case where the SA was never
+# granted the role), but surface every other failure
+# (``PERMISSION_DENIED``, org-policy reject, transient gcloud
+# errors, malformed condition) so the script's "Done." doesn't
+# claim victory while the role silently stays attached. The
+# previous ``2>/dev/null || true`` swallowed all of those.
 if [[ "$EXTRACTION_MODE" == "compiled-only" ]]; then
   echo
   echo "==> compiled-only deploy succeeded; idempotently removing any"
   echo "    pre-existing roles/aiplatform.user grant from $SA_EMAIL"
+  REMOVE_STDERR="$(mktemp -t bqaa-iam-remove-stderr-XXXXXXXX)"
+  set +e
   gcloud projects remove-iam-policy-binding "$PROJECT" \
     --member "serviceAccount:${SA_EMAIL}" \
     --role roles/aiplatform.user \
     --condition=None \
-    --quiet 2>/dev/null || true
+    --quiet 2>"$REMOVE_STDERR"
+  REMOVE_RC=$?
+  set -e
+  if [[ $REMOVE_RC -ne 0 ]]; then
+    # gcloud's "binding doesn't exist" error wording is stable
+    # across the current release channels: "Policy binding not
+    # found" (most cases). We match that exact substring so a
+    # future gcloud wording change becomes a loud failure, not a
+    # silent regression. Other failures (permission, network,
+    # org policy) carry different wording and fall through to
+    # the propagate-and-exit branch.
+    if grep -q "Policy binding not found" "$REMOVE_STDERR"; then
+      echo "    (no pre-existing roles/aiplatform.user grant — nothing to remove)"
+    else
+      echo "Error: failed to remove roles/aiplatform.user from $SA_EMAIL" >&2
+      echo "  gcloud exit code: $REMOVE_RC" >&2
+      echo "  gcloud stderr:" >&2
+      sed 's/^/    /' "$REMOVE_STDERR" >&2
+      echo "" >&2
+      echo "The new compiled-only revision is deployed and scheduled," >&2
+      echo "but $SA_EMAIL may still hold roles/aiplatform.user, which" >&2
+      echo "contradicts the 'no Vertex AI IAM' guarantee for" >&2
+      echo "--extraction-mode=compiled-only. Resolve the cause above," >&2
+      echo "then re-run this script or remove the role manually:" >&2
+      echo "  gcloud projects remove-iam-policy-binding $PROJECT \\" >&2
+      echo "    --member 'serviceAccount:$SA_EMAIL' \\" >&2
+      echo "    --role roles/aiplatform.user --condition=None" >&2
+      rm -f "$REMOVE_STDERR"
+      exit "$REMOVE_RC"
+    fi
+  fi
+  rm -f "$REMOVE_STDERR"
 fi
 
 echo
--- a/examples/migration_v5/periodic_materialization/deploy_cloud_run_job.sh
+++ b/examples/migration_v5/periodic_materialization/deploy_cloud_run_job.sh
@@ -117,17 +117,17 @@ Optional:
   --job-name NAME              Cloud Run Job name
                                (default: bqaa-periodic-materialization).
   --smoke                      After deploy, run the job once + tail logs.
-  --extraction-mode MODE       'ai-fallback' (default; only supported value
-                               on this deploy path today). The SDK also
-                               supports 'compiled-only' (structured
-                               extractors only, no AI.GENERATE) — accessible
-                               via the CLI when the caller wires bundles
-                               themselves — but this deploy script does
-                               not yet stage compiled-bundle artifacts into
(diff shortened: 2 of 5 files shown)


caohy1988:
Addressed both P1 findings in 6a95bcc:

## P1 #1 — compiled-only IAM remove ran before the new deploy

Moved the compiled-only \`remove-iam-policy-binding\` from the pre-deploy IAM block to a new section 8 that runs **after** \`gcloud run jobs deploy\` succeeds and (if \`--smoke\` is set) after the smoke execution passes.

* \`set +e\` / \`set -e\` around the smoke \`execute\` so we can capture \`SMOKE_RC\`.
* Non-zero \`SMOKE_RC\` → \`exit \"\$SMOKE_RC\"\` before the remove block, so a smoke failure leaves the existing schedule's IAM untouched.
* \`set -e\` at the top of the script propagates any earlier failure (staging, buildpacks, run-jobs-deploy, scheduler wiring) — none of those paths reach section 8.

Net effect: the existing ai-fallback deploy keeps \`roles/aiplatform.user\` and stays scheduled through every failure-of-the-new-deploy path. Only a fully-successful (and, with \`--smoke\`, fully-validated) new deploy strips the role.

New regression tests:
* \`TestDeployScriptExtractionModeBoundary.test_compiled_only_iam_remove_is_deferred_until_after_deploy\` — source-order check that \`remove-iam-policy-binding\` is below \`gcloud run jobs deploy\`.
* \`TestDeployScriptExtractionModeBoundary.test_smoke_failure_propagates_before_iam_remove\` — static check for \`SMOKE_RC\` capture + \`exit \"\$SMOKE_RC\"\`.

## P1 #2 — SDK accepted compiled-only with no extractor source

Added a boundary check in \`run_materialize_window\`: \`extraction_mode=\"compiled-only\"\` now requires either \`bundles_root\` or \`reference_extractors_module\`. The error message names both fix paths (the simple \`reference_extractors_module=reference_extractor\` for the demo, or the advanced bundles path).

New regression tests:
* \`TestExtractionModeFlag.test_compiled_only_without_extractor_source_rejected\` — pins the new boundary check.
* \`TestExtractionModeFlag.test_compiled_only_with_reference_module_accepted\` — mirror case proving the check only rejects the both-missing combination.

The existing 6 \`compiled-only\` test callsites now pass \`reference_extractors_module=_TEST_REF_MODULE\` (a stub module registered at test-module import time, so any future refactor that lets the import path run still works — but \`_build_manager\` is patched in those tests, so the import never actually fires).

## Verification

* \`python -m pytest tests/test_materialize_window.py\` — **108 passed** (was 104; +4 new).
* \`python -m pyink --check\` + \`python -m isort --check-only\` — clean.
* \`bash -n examples/migration_v5/periodic_materialization/deploy_cloud_run_job.sh\` — clean.

Ready for re-review.

caohy1988:
Addressed P2 in f25faed and refreshed the PR body (P3).

## P2 — compiled-only IAM remove now fails loud on real errors

Previously: \`gcloud projects remove-iam-policy-binding ... --quiet 2>/dev/null || true\` swallowed everything — PERMISSION_DENIED, transient gcloud errors, org-policy rejects all printed \"Done.\" while the role silently stayed attached.

Now:

* Capture stderr to a temp file and capture \`REMOVE_RC\`.
* If \`REMOVE_RC == 0\`: clean success.
* If \`REMOVE_RC != 0\` and stderr contains \`Policy binding not found\` (gcloud's stable absent-binding signature): treat as the first-deploy idempotent case, print \"(no pre-existing grant — nothing to remove)\", continue.
* Otherwise: print the gcloud exit code + indented stderr, print a remediation hint with the exact manual \`remove-iam-policy-binding\` command, then \`exit \"\$REMOVE_RC\"\`.

The new revision is still deployed and scheduled either way — the script only fails after telling the operator that the role may still be attached and what to do about it.

New regression test: \`test_compiled_only_iam_remove_distinguishes_absent_from_failure\` covers \`REMOVE_RC\` capture, the \`Policy binding not found\` match, \`exit \"\$REMOVE_RC\"\` propagation, and a negative assertion that the legacy \`--quiet 2>/dev/null || true\` swallow idiom is gone from the remove block.

## P3 — PR body refreshed

Updated to:
* Cite the new test count (**109 passing**, was 104 originally).
* Document the deferred + error-aware IAM remove as the final contract (replacing the old \`2>/dev/null || true\` description).
* Add a review-history table tracking the three commits and which findings each addresses.

## Verification

* \`python -m pytest tests/test_materialize_window.py\` — **109 passed** (was 108; +1 new for P2).
* \`python -m pyink --check\` + \`python -m isort --check-only\` — clean.
* \`bash -n examples/migration_v5/periodic_materialization/deploy_cloud_run_job.sh\` — clean.

Ready for re-review.

caohy1988:
## Live verification ✅

Deployed the compiled-only path end-to-end against \`test-project-0728-467323\` (the same project as PR #166's ai-fallback verification). Reused the existing \`bqaa-periodic-sa@\` service account, which carried \`roles/aiplatform.user\` from PR #166's deploy — so this exercised the actual **ai-fallback → compiled-only transition** the PR is designed for, not just a clean first deploy.

### Deploy

\`\`\`
./examples/migration_v5/periodic_materialization/deploy_cloud_run_job.sh \\
  --project test-project-0728-467323 \\
  --region us-central1 \\
  --events-dataset agent_analytics_demo \\
  --graph-dataset migration_v5_compiled_only_verify \\
  --schedule "0 */6 * * *" \\
  --extraction-mode compiled-only \\
  --job-name bqaa-compiled-only-verify
\`\`\`

Script ran to completion (\`Done.\`). Verified post-deploy state:

| Check | Result |
|---|---|
| Cloud Run Job env vars | \`BQAA_EXTRACTION_MODE=compiled-only\`, \`BQAA_REFERENCE_EXTRACTORS_MODULE=reference_extractor\` both set |
| Scheduler trigger | \`bqaa-compiled-only-verify-cron\` ENABLED on \`0 */6 * * *\` |
| SA pre-deploy IAM | \`roles/aiplatform.user\` + \`roles/bigquery.jobUser\` |
| SA post-deploy IAM | only \`roles/bigquery.jobUser\` — **\`aiplatform.user\` removed by the deferred-remove block** |

The deferred remove fired (the verbose policy dump in the script's tail output is gcloud printing the post-remove policy), and crucially **only after \`gcloud run jobs deploy\` succeeded** — exactly the ordering P1 #1 fixed.

### Execution proof (backfill against April events to actually exercise extraction)

Updated the job to backfill the populated window (events were 4 weeks old; the default 6-hour lookback found 0 sessions) and re-executed:

\`\`\`
sessions_discovered=37
sessions_failed=1
sessions_materialized=0
ok=False
exit code=1   (expected: typed empty_extraction failures, not a crash)
\`\`\`

Failure record from the result JSON (one verbatim sample):

\`\`\`
error_code: empty_extraction
error_detail: compiled-only extraction-mode: 6 structured_unhandled + 0 extractor_exception
  diagnostic(s) fired on this session. The compiled extractor set does not cover every
  event type seen here. To resolve, either extend the compiled extractors (preferred) or
  re-run with \`--extraction-mode=ai-fallback\` to let AI.GENERATE fill the gaps. Sample
  diagnostics: structured_unhandled span_id='1d212fa1...' event_type='USER_MESSAGE_RECEIVED';
  structured_unhandled span_id='a0f96ea6...' event_type='AGENT_STARTING';
  structured_unhandled span_id='2a6575c7...' event_type='LLM_REQUEST'; ...
\`\`\`

The \`calendar_assistant\` events aren't MAKO-shaped, so the reference extractor legitimately returned empty results for every span — which is the **happy path for this verification**: it surfaces the typed \`empty_extraction\` failure mode end-to-end, with the diagnostic prose, sample span_ids, event types, and remediation hint exactly as designed.

### Zero-AI proof

The SA had **no** \`roles/aiplatform.user\` for the entire run. If the compiled-only path had attempted any \`AI.GENERATE\` call, the BigQuery client would have surfaced \`PermissionDenied\`; instead the job exited cleanly with the typed empty_extraction failure. Combined with the existing \`TestCompiledOnlyMakesZeroLLMCalls\` unit-test pin, that's mechanical + live proof of zero LLM calls.

### Teardown

\`\`\`
gcloud scheduler jobs delete bqaa-compiled-only-verify-cron       ✓ Deleted
gcloud run jobs delete bqaa-compiled-only-verify                  ✓ Deleted
bq rm -r migration_v5_compiled_only_verify                        ✓ Deleted
gcloud projects add-iam-policy-binding ... roles/aiplatform.user  ✓ Restored on bqaa-periodic-sa@
\`\`\`

Final SA state matches pre-test state, so PR #166's ai-fallback deploy fixtures continue to work for future verifications.

**Ready to merge.**

Issue #179, opened 2026-05-19
SDK: self-edge support in ontology_materializer (unblocks evolvedFrom / supersededBy)


mikulbhatt:
## Background

The BigQuery Agent Analytics SDK materializes agent decision traces into a **property graph** stored in BigQuery. It reads an ontology (a `.ttl` / Turtle RDF file) and a binding config (mapping ontology properties to BigQuery table columns), then writes nodes and edges into BigQuery tables that can be queried with GQL (BigQuery's graph query language).

A **self-edge** is a relationship where both the source and destination are entities of the *same type* — for example, a `DecisionExecution` that evolved from an earlier `DecisionExecution` in a previous session. This is common in temporal lineage modelling (cf. `prov:wasDerivedFrom` in W3C PROV).

## Problem

[`_relationship_columns` (L167–L191)](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/src/bigquery_agent_analytics/ontology_materializer.py#L167) builds the `{col_name: DDL_type}` dict for a relationship's BigQuery DDL. It looks up `from_columns` against the source entity's property map and `to_columns` against the target entity's property map:

```python
# ontology_materializer.py L167–L191
def _relationship_columns(rel, spec):
    entity_map = {e.name: e for e in spec.entities}
    src = entity_map[rel.from_entity]          # L173
    tgt = entity_map[rel.to_entity]            # L174
    src_prop_map = {p.column: p for p in src.properties}   # L175
    tgt_prop_map = {p.column: p for p in tgt.properties}   # L176

    cols = {}
    from_cols = rel.from_columns or src.key_columns
    for col in from_cols:
        cols[col] = _ddl_type(src_prop_map[col].sdk_type)    # L181
    to_cols = rel.to_columns or tgt.key_columns
    for col in to_cols:
        if col not in cols:                                   # L184 ← the bug
            cols[col] = _ddl_type(tgt_prop_map[col].sdk_type)
```

For a self-edge (`evolvedFrom`: `DecisionExecution → DecisionExecution`):
- `src` and `tgt` are the **same entity** — same key columns, same property map
- `from_cols` and `to_cols` are both `["decision_execution_id"]`
- The `if col not in cols` guard on L184 **silently drops the destination column** because it was already written by the source pass
- Result: the relationship DDL has one `decision_execution_id` column instead of two — it cannot encode a self-edge

[...]

PR #222, opened 2026-05-20
feat(relationship): consume canonical FK→PK mapping; re-add migration v5 self-edges (#179 follow-up)

caohy1988:
## Summary

Closes the loop on #179: C1 (#191) added the canonical
``ResolvedRelationship.from_column_mapping`` / ``to_column_mapping``
fields and the list-view shim. C2 wires those into the four downstream
surfaces that needed them for self-edge support — the materializer's
edge-row router + DDL column resolver, the PG DDL compiler, and the
extracted-graph validator's endpoint-key check — and re-adds the
migration v5 self-edge relationships (``evolvedFrom`` / ``supersededBy``)
that were previously dropped from the binding.

Part of #187 sequencing: A (#188) → B1 (#189) → C1 (#191) → B2 (#192)
→ deploy-path (#193) → **this** → D/E/F/G.

#### ``ontology_materializer._route_edge`` + ``_relationship_columns``

Self-edges need the edge-table FK column name (``src_decision_execution_id``)
to differ from the endpoint's PK column name (``decision_execution_id``) —
otherwise the DDL has duplicate columns. The canonical mapping pairs
``(edge_column, target_property)``, so the materializer:

* In ``_route_edge``: bridges through ``ResolvedProperty.logical_name → column``
  to translate ``target_property`` (logical) into the endpoint's physical
  column, then reads ``parsed_keys.get(physical_col, '')`` and writes
  ``row[edge_col]``.
* In ``_relationship_columns``: looks up the SDK type by the endpoint's
  ``logical_name`` and keys the result dict by the edge column.

The renamed-PK case (binding-time property rename, e.g. property ``id`` →
column ``decision_execution_id``) is covered by the same logical→physical
bridge.

#### ``graph_validation`` endpoint-key check

Same translation. The reported failure ``path`` still names the edge-table
FK column (what the user reads in the binding), but the lookup keys by
the endpoint's physical PK column (what ``_build_node_id`` writes).

#### ``bigquery_ontology/graph_ddl_compiler``

New ``_ordered_edge_columns`` helper: runs ``normalize_relationship_columns``
and re-projects the ``(edge_col, target_property)`` pairs into the
endpoint's PK-property declaration order, so the emitted
``SOURCE KEY (...) REFERENCES Alias (...)`` positional contract holds for
both legacy and dict-shape bindings. The C1 ``require_legacy_column_shape``
guards in ``_resolve_edge_table`` and ``_resolve_edge_key_columns`` are
removed.

#### ``runtime_spec.graph_spec_from_ontology_binding``

Guard **intentionally retained**: the legacy ``GraphSpec``/``BindingSpec``
world still uses ``from_columns: list[str]`` so widening it is a separate
refactor. The actionable error message remains pointing at the canonical
resolve path.

### Migration v5

* ``ontology_artifacts.make_binding`` no longer drops self-edges. It emits
  the dict-shape ``[{src_<col>_id: <pk_prop>}, {dst_<col>_id: <pk_prop>}]``
  for ``rel.from_ == rel.to`` relationships; heterogeneous edges keep the
  legacy ``list[str]`` shape.
* ``make_table_ddl`` / ``make_property_graph_sql`` route through
  ``edge_column_names`` so the positional ``from_columns[0]`` /
  ``to_columns[0]`` indexing works for both shapes.
* Regenerated artifacts: ``binding.yaml`` (now 9 relationships, was 7),
  ``table_ddl.sql`` (new ``evolved_from`` / ``superseded_by`` edge tables),
  ``property_graph.sql`` (new PG DDL with correct ``SOURCE KEY (src_...)
  REFERENCES decision_execution (decision_execution_id)`` clauses).
* README design decision 9 rewritten from \"self-edges dropped\" to
  \"self-edges via explicit FK→PK mapping\".

## Verification

* ``python -m pytest tests/`` — **3082 passed**, 15 skipped (was 3077; +5 new).
* ``python -m pyink --check`` — clean.
* ``python -m isort --check-only`` — clean.
* Live verification on this branch: migration v5 \`compile_property_graph\`
  emits valid PG DDL for both self-edges, and the materializer round-trips
  a synthetic decision-flow with self-edge endpoints (no \`KeyError\`,
  no false \`missing_endpoint_key\` failures).

## Test plan
- [x] ``python -m pytest tests/`` — 3082 pass
- [x] ``python -m pyink --check`` + ``python -m isort --check-only`` — clean
- [x] Migration v5 artifacts regenerate to expected shape (verified via diff)
- [ ] Live Cloud Run run against an event stream containing
  ``evolvedFrom`` / ``supersededBy`` payloads (reviewer / merger-side; the
  unit + DDL tests cover correctness, but the migration v5 demo's
  reference extractors don't emit decision-history edges yet)


caohy1988:
Addressed both findings in 0443efa.

## P1 — Public PG compiler now consumes the canonical mapping

\`bigquery_agent_analytics.ontology_property_graph._compile_edge_table_clause\` now mirrors \`bigquery_ontology.graph_ddl_compiler\`: when \`rel.from_column_mapping\` / \`to_column_mapping\` is populated (which it always is for bindings that go through \`resolve()\`), it projects the \`(edge_col, target_prop_logical)\` pairs into the endpoint's PK-column declaration order via the \`ResolvedProperty.column → logical_name\` reverse lookup. The emitted \`SOURCE KEY (...)\` then positionally references \`src.key_columns\` — the PG DDL contract holds.

The legacy strict-match check stays in place for the fallback path (\`from_column_mapping is None\` — older callers that construct \`ResolvedRelationship\` manually).

**Reproducer**: loaded the committed \`examples/migration_v5/ontology.yaml\` + \`binding.yaml\`, called \`compile_property_graph_ddl\`. Before the fix: \`ValueError: Relationship 'evolvedFrom': from_columns ['src_decision_execution_id'] do not match DecisionExecution primary key ['decision_execution_id']\`. After the fix:

\`\`\`
\`...evolved_from\` AS evolvedFrom
  SOURCE KEY (src_decision_execution_id, session_id) REFERENCES DecisionExecution (decision_execution_id, session_id)
  DESTINATION KEY (dst_decision_execution_id, session_id) REFERENCES DecisionExecution (decision_execution_id, session_id)
\`\`\`

Three new regression tests in \`tests/test_ontology_property_graph.py::TestCompilePropertyGraphSelfEdge\`:
* \`test_self_edge_dict_shape_compiles\` — hand-built self-edge spec.
* \`test_migration_v5_binding_compiles_end_to_end\` — your exact reproducer pinned.
* \`test_legacy_mismatch_still_raises\` — strict check preserved for the fallback path.

## P3 — Stale docs/comments fixed

* \`examples/migration_v5/README.md\` design decision 8 now says heterogeneous edges keep the legacy \`list[str]\` shape (so per-entity PK names give them a clean \`{src_entity}_id, {dst_entity}_id\` shape and avoid the \`id, id\` duplicate-column collision), and self-edges go through the dict-shape FK→PK mapping.
* \`examples/migration_v5/ontology_artifacts.py\` PK-column comment block rewritten with the same explanation.

## Verification

* \`python -m pytest tests/\` — **3085 passed**, 15 skipped (was 3082; +3 new).
* \`python -m pyink --check\` + \`python -m isort --check-only\` — clean.

Ready for re-review.

caohy1988:
Addressed P3 in 5ffd04a.

The docstring and error message on \`require_legacy_column_shape\` were stale after C2 lifted the dict-shape guard on \`compile_graph\` and the public PG compiler. The docstring now names the one consumer the guard still protects — \`runtime_spec.graph_spec_from_ontology_binding\`, which feeds the legacy \`BindingSpec.from_columns: list[str]\` shape — and the error message drops the \"C2 will lift this restriction\" promise (this *is* C2) and instead recommends the canonical \`resolve()\` → \`ResolvedRelationship.from_column_mapping\` path, with a parenthetical listing the modern consumers that already read it.

No behavior change. 3085 tests pass, lint clean.

## Residual gap (acknowledged)

The reviewer's last note about no live \`CREATE PROPERTY GRAPH\` / GQL smoke is correct. The unit + bridge tests cover correctness of the emitted DDL byte-for-byte; the live-BQ verification belongs to a follow-up that also exercises the self-edge edge tables (\`evolved_from\`, \`superseded_by\`) end-to-end via the migration v5 demo. Happy to do that after merge alongside the live smoke for the migration v5 notebook's Beat 4 cells.

Issue #180, opened 2026-05-19
materialize_window: orphaned-session watchdog for runs without AGENT_COMPLETED


mikulbhatt:
## Background

The SDK's `materialize_window` orchestrator scans the `agent_events` BigQuery table for completed agent sessions and materializes them into the property graph. It identifies completed sessions by looking for rows where `event_type = 'AGENT_COMPLETED'` — the terminal event emitted when an agent finishes cleanly.

Sessions are tracked in `_bqaa_materialization_state`. A session only appears in the state table after it has been picked up and processed. See [`run_job.py`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/examples/migration_v5/periodic_materialization/run_job.py) for the orchestrator entry point.

## Problem

Sessions that fail, time out, or crash without emitting `AGENT_COMPLETED` are **never materialized and never recorded as failures**. They sit in `agent_events` indefinitely — un-extracted, invisible to the state table, and invisible to Cloud Monitoring alerts.

For audit-critical deployments where the requirement is "every decision must be auditable," this is a silent gap: a failed session looks identical to a session that hasn't happened yet.

## Proposed solution

New flag `--max-session-age-hours <N>` (default: unset / disabled).

When set, the orchestrator additionally scans for sessions whose **first event is older than N hours** but that have **no terminal event row**. Two sub-modes for what to do with them:

| Sub-mode | Behaviour |
|---|---|
| `--orphan-mode=fail-explicit` *(recommended default)* | Surface each orphan as a `session_orphaned` failure in the JSON report; flip `ok=false`; do not attempt extraction |
| `--orphan-mode=force-materialize` | Attempt extraction against the partial event set; mark resulting graph rows with `from_orphan = TRUE` |

The `session_orphaned` code slots into the existing failure-mode classifier (PR #167) — no new alert wiring required.

## Source files

| File | Relevance |
|---|---|
| [`periodic_materialization/run_job.py`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/examples/migration_v5/periodic_materialization/run_job.py) | Orchestrator entry point — where the watchdog scan and `--max-session-age-hours` flag would be added |
[...]

PR #224, opened 2026-05-20
feat(materialize_window): orphan-session watchdog (#180)

caohy1988:
## Summary

Closes the audit gap that #180 surfaces: sessions that fail, time out, or
crash without emitting \`AGENT_COMPLETED\` are never materialized AND
never recorded as failures. For audit-critical deployments where the
contract is \"every decision must be auditable,\" this is a silent gap — a
failed session looks identical to one that hasn't happened yet.

PR D adds an opt-in orphan watchdog that runs alongside each cron pass
when \`--max-session-age-hours <N>\` is set. New orphans surface as
typed \`session_orphaned\` failures in the existing #167 failure-mode
classifier, so Cloud Monitoring alerts on \`ok=false\` keep working
without new wiring. The state table grows two new row kinds for the
operator-visible audit trail.

Part of #187 sequencing: A (#188) → B1 (#189) → C1 (#191) → B2 (#192) →
deploy-path (#193) → C2 (#222) → **this** → demo/content expansion (Beat 5,
MAKO coverage, Terraform/hardening).

### SDK — schema additions

| Column | Purpose |
|---|---|
| \`orphan_watermark TIMESTAMP\` | Upper bound of the most recent scan window. Next scan uses strict \`>\` on this column so the boundary session isn't reconsidered. |
| \`flagged_session_ids ARRAY<STRING>\` | Cumulative ledger of unresolved orphan session_ids per state_key, or per-scan delta on the audit row. |

Two idempotent \`ALTER TABLE ... ADD COLUMN IF NOT EXISTS\` migrations
alongside the existing \`mode\` migration so pre-D state tables get
patched without a destructive migration.

### SDK — state-row modes

* \`orphan_scan\` — per-pass audit row. \`flagged_session_ids\` carries only
  the **new** orphans flagged in this scan (the delta); \`orphan_watermark\`
  carries this scan's cutoff.
* \`orphan_ledger\` — cumulative ledger row written alongside every
  \`orphan_scan\`. \`flagged_session_ids\` carries the full running set
  after the resolve sweep. Append-only by run; reads pick the most recent
  row per state_key (mirroring the steady-checkpoint pattern).

### SDK — scan logic (\`run_orphan_watchdog\`)

1. Read the most recent \`orphan_ledger\` row for \`state_key\` (or empty).
2. Discover sessions whose \`MIN(timestamp)\` falls in
   \`(orphan_watermark, cutoff_at]\`, have \`terminal_event_count = 0\`,
   and aren't already in the cumulative ledger. \`cutoff_at\` =
   \`run_started_at - max_session_age_hours\`.
3. Probe for resolved orphans: previously-flagged session_ids that now
   have \`AGENT_COMPLETED\`. Drop them from the cumulative ledger.
4. Write \`orphan_scan\` (delta) + \`orphan_ledger\` (cumulative after
   resolve). Watermark advances unconditionally so the next scan's \`>\`
   boundary moves forward even on no-op scans.

### Orchestrator + CLI integration

* New optional kwarg \`max_session_age_hours: Optional[float]\` on
  \`run_materialize_window\`. \`None\` disables the watchdog (default);
  positive values enable it. Boundary check rejects \`<= 0\` at the
  same level as \`--lookback-hours\`.
* Skipped automatically in \`backfill=True\` mode.
* The watchdog runs **after** the steady pass writes its row, so a
  partial steady failure doesn't block the watchdog signal and the
  per-mode audit-row stream stays cleanly partitioned.
* New orphans get threaded into \`MaterializeWindowResult.failures[]\`
  as \`error_code='session_orphaned'\` entries with a remediation hint
  pointing at the ledger.
* \`MaterializeWindowResult.to_json()\` adds an \`orphan\` key when the
  watchdog ran so operators can pivot Cloud Logging filters on
  \`jsonPayload.orphan.cumulative_flagged\`.
* \`bqaa materialize-window --max-session-age-hours N\` CLI flag.

### Deploy path (migration v5)

* \`run_job.py\` reads \`BQAA_MAX_SESSION_AGE_HOURS\`.
* \`deploy_cloud_run_job.sh\` adds \`--max-session-age-hours N\`; the
  env var is wired only when the operator opts in (default deploys
  don't pay the extra discovery query).
* \`README.md\`'s Failure-mode surface section grows a third bullet
  (\`session_orphaned\`) plus a paragraph on the state-table audit
  trail.

## Tests (3097 passing; +12 new on \`materialize_window\`)

* \`TestOrphanWatchdogBuilders\` (5) — SQL shape: strict \`>\` on
  watermark, \`terminal_event_count = 0\` filter,
  \`NOT IN UNNEST(@already_flagged)\` exclusion, ledger read filtered
  on \`mode='orphan_ledger'\`, resolved-orphan probe bounded by the
  \`previously_flagged\` set.
* \`TestOrphanWatchdogScan\` (4) — end-to-end via stub client: first
  scan flags new orphans + writes both rows; second scan carries
  forward, drops resolved, appends new; empty scan still writes
  ledger so running set persists across no-op scans; \`<= 0\`
  rejected at boundary.
* \`TestOrphanWatchdogOrchestratorIntegration\` (3) — \`None\` skips
  the watchdog (no \`orphan\` key in payload); backfill mode skips
  even when set; invalid values fail fast at the orchestrator
  boundary.
* \`TestEnsureStateTableMigration\` grew to assert the two new
  \`ADD COLUMN IF NOT EXISTS\` migrations + the new columns in the
  initial DDL.

## Test plan
- [x] \`python -m pytest tests/\` — 3097 pass, 15 skipped
- [x] \`python -m pyink --check\` + \`python -m isort --check-only\` — clean
- [x] \`bash -n examples/migration_v5/periodic_materialization/deploy_cloud_run_job.sh\` — clean
- [ ] Live Cloud Run run with \`--max-session-age-hours\` set against a real events stream that has unfinished sessions (reviewer / merger-side; the unit + integration tests cover the SQL shape and orchestrator wiring, but the cross-run watermark + ledger persistence story belongs to a live verification)

### DIFF
--- a/examples/migration_v5/periodic_materialization/README.md
+++ b/examples/migration_v5/periodic_materialization/README.md
@@ -644,7 +644,7 @@ fixed. Today, the same situation would produce `ok=false` with
 
 ### Failure-mode surface (post-#167)
 
-The orchestrator distinguishes two zero-row session outcomes
+The orchestrator distinguishes three session-level outcomes
 that look identical from `rows_materialized` alone:
 
 * **`empty_extraction`** — extraction (AI.GENERATE or compiled
@@ -663,10 +663,32 @@ that look identical from `rows_materialized` alone:
   schema drift the binding-validate pre-flight missed, or
   streaming-buffer pinning.
 
-In both cases: `ok=false`, CLI exit 1, the cron run shows up
-as a failed execution in Cloud Monitoring. Alert directly on
+* **`session_orphaned`** *(only when `--max-session-age-hours`
+  is set, issue #180)* — a session's first event landed before
+  the configured cutoff but the session never emitted
+  `AGENT_COMPLETED`. The orphan-watchdog scan flagged it as
+  in-flight-too-long. `failures[].error_detail` names the
+  session_id, the cutoff timestamp, and the missing terminal
+  event_type, plus a remediation hint pointing at the
+  cumulative ledger row (`mode='orphan_ledger'`) in the state
+  table. Diagnose by triaging the underlying agent failure or
+  by emitting the missing terminal event (the watchdog drops
+  the session from the ledger on its next pass once
+  `AGENT_COMPLETED` appears).
+
+In all three cases: `ok=false`, CLI exit 1, the cron run shows
+up as a failed execution in Cloud Monitoring. Alert directly on
 `jsonPayload.ok=false` plus `jsonPayload.failures[].error_code`
 for the failure-mode breakdown — no second-line check needed.
+
+The orphan watchdog also persists a structured audit trail in
+the state table: `mode='orphan_scan'` for the per-cron-pass
+delta (`flagged_session_ids` = sessions newly flagged this
+scan, `orphan_watermark` = the scan's cutoff that becomes the
+next scan's exclusive `>` boundary) and `mode='orphan_ledger'`
+for the cumulative running set after the resolved-orphan
+sweep. Operators reading the ledger see "what's currently
+unresolved" without having to diff across runs.
 
 ## Cleanup and redeploy
 
--- a/examples/migration_v5/periodic_materialization/deploy_cloud_run_job.sh
+++ b/examples/migration_v5/periodic_materialization/deploy_cloud_run_job.sh
@@ -92,6 +92,7 @@ MAX_SESSIONS=""
 JOB_NAME="bqaa-periodic-materialization"
 SMOKE=false
 EXTRACTION_MODE="ai-fallback"
+MAX_SESSION_AGE_HOURS=""
 
 # Print usage. Exit code is the caller's choice: ``usage 0``
 # for ``--help`` (success), ``usage 1`` for parse / required-arg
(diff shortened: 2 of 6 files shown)


caohy1988:
Addressed both findings in f331407.

## P1 — Watchdog now respects \`--completion-event-type\`

Both SQL builders parameter-bind \`@completion_event_type\` instead of hardcoding \`'AGENT_COMPLETED'\`. The orchestrator forwards the same value it already passes to the discovery query, so the watchdog and the steady cron stay aligned by construction. Deploys using \`--completion-event-type=MY_TERMINAL\` would have seen the watchdog falsely flag sessions the steady cron just materialized — that's fixed.

Test \`test_custom_completion_event_type_is_query_bound\` walks every \`@completion_event_type\` binding across the discovery + resolve calls and asserts they all carry the operator's value. The existing builder tests now negatively assert that the \`'AGENT_COMPLETED'\` literal is gone.

## P2 — Resolve probe now partition-bounded

Added \`timestamp > @resolve_lower_bound AND timestamp <= @resolve_upper_bound\` to \`build_resolved_orphan_sql\`. The orchestrator passes:
* \`resolve_lower_bound\` = previous \`orphan_watermark\` (advances every scan)
* \`resolve_upper_bound\` = \`run_started_at\`

On the plugin's timestamp-partitioned events table this prunes to a small fixed number of partitions regardless of history length, restoring the bounded-cost claim the docstring promised.

The docstring grew a note: late terminal events with backdated \`timestamp\` values (e.g. agent restart with a stale emission time more than \`max_session_age_hours\` in the past) won't be picked up by the tight bound. Operators with that pattern can widen it, but the common case stays cheap.

New test \`test_resolved_orphan_sql_has_timestamp_partition_predicate\` pins the SQL shape; \`test_resolved_probe_passes_timestamp_partition_bounds\` pins the live parameter values (\`previous_orphan_watermark\` → \`run_started_at\`).

## Verification

* \`python -m pytest tests/test_materialize_window.py\` — **124 passed** (+3 new for this round)
* \`python -m pytest tests/\` — 3100 passed, 15 skipped
* \`python -m pyink --check\` + \`python -m isort --check-only\` — clean
* \`bash -n examples/migration_v5/periodic_materialization/deploy_cloud_run_job.sh\` — clean

Ready for re-review.

PR #225, opened 2026-05-21
fix(materialize_window): orphan watchdog crashes on every cron pass — empty-array insert rejected by BigQuery (#180 hotfix)

caohy1988:
## Summary

**Critical**: live verification of PR #224 against \`test-project-0728-467323\` surfaced an insert-time crash on the very first cron pass. BigQuery's streaming-insert API rejects empty \`ARRAY<STRING>\` values with:

\`\`\`
Field value of flagged_session_ids cannot be empty.
\`\`\`

The previous \`append_state_row\` serialization sent \`\"flagged_session_ids\": []\` whenever the \`StateRow\`'s tuple was empty — i.e. on every steady-state row, every backfill row, and every zero-orphan watchdog scan. Net effect: every cron pass after PR #224 merged would crash at insert time and report exit code 2 (\"unexpected internal error\") in Cloud Logging.

## Fix

Build the \`insert_rows_json\` payload incrementally: nullable columns (\`last_completion_at\`, \`error_detail\`, \`report_json\`, \`orphan_watermark\`, \`flagged_session_ids\`) are only added to the payload when they have a value. Missing fields default to NULL on the BQ side — the right \"this row kind doesn't populate that column\" semantic — and avoid the empty-array rejection.

The truthy check on \`flagged_session_ids\` folds \`None\` and \`()\` into the same omit branch (both store as NULL). The distinction between \"field didn't apply on this row kind\" and \"scan ran with zero orphans\" is carried by the \`mode\` column.

## Tests (+3 new)

New class \`TestAppendStateRowOmitsEmptyArrays\`:
* \`test_steady_row_omits_array_field_when_none\` — pins the pre-#224 steady-row payload shape (nullable columns absent).
* \`test_orphan_scan_with_zero_orphans_omits_array\` — the unit-level reproducer: a \`StateRow(mode=STATE_MODE_ORPHAN_SCAN, flagged_session_ids=())\` produces a payload **without** the field; \`orphan_watermark\` (which IS populated on orphan rows) is still sent.
* \`test_orphan_row_with_flagged_sessions_sends_array\` — the non-empty path stays unchanged.

Full suite: **3103 passed**, 15 skipped.

## Live verification (post-fix)

Re-deployed against the same repro environment, with two synthesized orphans (\`129ea426-...\`, \`12f9922c-...\`) plus four pre-existing incomplete sessions in \`agent_analytics_demo\`. After dropping the prior state table to start clean:

| Signal | Value |
|---|---|
| Run exit code | 1 (expected: \`ok=false\` with orphan failures) |
| \`failures[]\` entries | 6 × \`error_code='session_orphaned'\` with the exact PR #224 prose, cutoff timestamp, and ledger pointer |
| \`orphan\` JSON payload | \`new_orphans=[6 ids], cumulative_flagged=[same 6], previous_watermark=None, resolved_orphans=[]\` |
| State-table rows from this pass | \`orphan_scan\` (delta, 6), \`orphan_ledger\` (cumulative, 6), \`steady\` (0) |

Failure prose verbatim:
\`\`\`
session 'fa596607-...' has first event older than 24.0 hours
(cutoff 2026-05-20T07:03:11.423669+00:00) but never emitted
'AGENT_COMPLETED'. The orphan watchdog flagged it on this scan
and recorded it in the cumulative ledger (state_key='...',
mode='orphan_ledger'). Resolve by either emitting the missing
terminal event (the watchdog will drop the session from the
ledger on its next pass) or triaging the underlying agent failure.
\`\`\`

## Test plan
- [x] \`python -m pytest tests/\` — 3103 pass
- [x] \`python -m pyink --check\` + \`python -m isort --check-only\` — clean
- [x] Live Cloud Run smoke against synthesized orphans — passes end-to-end
- [x] Live state-table inspection — \`orphan_scan\` + \`orphan_ledger\` + \`steady\` rows written cleanly

This is a no-grace-period fix — without it, every deployed cron pass against \`_bqaa_materialization_state\` crashes. Merging this before any further work on D follow-ups (or any new PRs that touch the state table).

### DIFF
--- a/src/bigquery_agent_analytics/materialize_window.py
+++ b/src/bigquery_agent_analytics/materialize_window.py
@@ -584,32 +584,52 @@ def append_state_row(
     bq_client: Any, state_table_ref: str, row: StateRow
 ) -> None:
   """Append a state row. Uses ``insert_rows_json`` (streaming
-  insert) — cheaper than an INSERT job for tiny single-row writes."""
-  payload = {
+  insert) — cheaper than an INSERT job for tiny single-row writes.
+
+  Two BigQuery streaming-insert quirks shape the payload assembly:
+
+  * **Empty ARRAY columns must be omitted, not sent as ``[]``.**
+    Sending ``{"flagged_session_ids": []}`` triggers BigQuery's
+    "Field value of flagged_session_ids cannot be empty" rejection
+    (surfaced as the very first failure when the orphan-watchdog
+    PR #224 went live — every scan with zero new/cumulative
+    orphans crashed). Omitting the field stores NULL, which is
+    the right "scan ran, no orphans" semantic anyway: the
+    ``mode='orphan_scan'`` row's other fields already convey the
+    "scan ran" signal.
+  * **NULL TIMESTAMP fields are safe to send as ``None``** —
+    ``insert_rows_json`` translates that into a NULL cell. But
+    we already special-case via ``_iso_optional`` to keep the
+    omitted-field pattern consistent across both column kinds.
+  """
+  payload: dict[str, Any] = {
       "state_key": row.state_key,
       "run_id": row.run_id,
       "run_started_at": _iso(row.run_started_at),
       "scan_start": _iso(row.scan_start),
       "scan_end": _iso(row.scan_end),
-      "last_completion_at": _iso_optional(row.last_completion_at),
       "sessions_discovered": row.sessions_discovered,
       "sessions_materialized": row.sessions_materialized,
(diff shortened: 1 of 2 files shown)


caohy1988:
P3 polish addressed in 8589757 — docstring bullet rewritten to match the actual omit-when-None behavior (the previous wording referenced \`_iso_optional\`, which the hotfix no longer calls).

No behavior change. 127 \`materialize_window\` tests pass; lint clean. Ready when you are.

Issue #234, opened 2026-05-22
Distribute tracing package and Claude Code plugin


caohy1988:
## Context

PR #232 adds the shared `producers/` tracing package scaffold and writer/drainer. The next distribution milestone is to make this usable by Claude Code users: publish the Python tracing package and ship a Claude Code plugin artifact that can run hooks and drain rows without requiring a separate tracing-wheel install in the hook runtime.

This tracks the work after #232 lands. It is a follow-up to #229.

## Goals

- Publish `bigquery-agent-analytics-tracing` from `producers/` as its own package.
- Add the Claude Code hook producer surface to the tracing package.
- Build a Claude Code plugin artifact that vendors the tracing package source at release time.
- Ensure the vendored plugin can execute hook -> spool -> drainer -> BigQuery without `bigquery-agent-analytics-tracing` installed on `BQAA_PYTHON`.
- Prepare the plugin for marketplace submission.

## Proposed work

1. Add `bigquery_agent_analytics_tracing.claude_code`
   - Port the existing Claude Code hook event parsing and entrypoint.
   - Preserve current default agent/source semantics.
   - Keep `attributes.writer` package-owned and preserve caller metadata under non-reserved keys.
   - Preserve environment controls including `BQAA_TRACE_ENABLED`, `BQAA_PYTHON`, BigQuery destination vars, dry-run, auto-create flags, and writer label.

2. Add plugin artifact build
   - Generate the plugin artifact from source rather than maintaining a second vendored copy.
   - Copy `src/bigquery_agent_analytics_tracing/` into the plugin artifact during build.
   - Propagate `importlib.metadata.version("bigquery-agent-analytics-tracing")` into plugin metadata where applicable, with `0.0.0+local` fallback for dev/vendored runs.

3. Make vendored drainer invocation explicit
   - For wheel installs: use `python -m bigquery_agent_analytics_tracing.drain`.
   - For vendored plugin installs: either inject the vendored package root into `PYTHONPATH` on the spawned drainer env, or ship a tiny `bqaa_drain_entry.py` wrapper inside the plugin artifact.
   - Test that the drainer succeeds from a hook running under the vendored plugin with no tracing wheel installed.

4. Add distribution workflows
   - Add a producer-package publish workflow separate from the root SDK release workflow.
   - Use package-specific tags, for example `tracing-v0.1.0`, or another maintainer-approved versioning trigger.
[...]

Issue #229, opened 2026-05-21
Move tracing producers (OpenAI Agents SDK, Codex CLI, Claude Code plugin) into this repo and publish PyPI + plugin artifacts


caohy1988:
## Summary

Move the tracing producer surfaces from
[`Erroration2022/bigquery_agent_analytics_skill`](https://github.com/Erroration2022/bigquery_agent_analytics_skill/tree/main/plugins/bigquery-agent-analytics-tracing)
into this repo as first-class BQAA producers, and publish them through the
install channels their users actually use:

- Python package: `bigquery-agent-analytics-tracing`
- Claude Code plugin marketplace entry: `bigquery-agent-analytics-tracing`
- Codex CLI wrapper console script: `bqaa-codex`

These producers already emit rows compatible with the canonical BQAA
`agent_events` shape consumed by this SDK. The move should preserve row shape
and producer labels so existing queries, dashboards, and SDK consumers keep
working.

## What exists today in the source repo

Verified against the current code under
`plugins/bigquery-agent-analytics-tracing/`:

| Surface | Current implementation | Notes for port |
|---|---|---|
| Shared Python writer | `sdk/python/bqaa_tracing.py` | Defines `BQAAConfig`, `BigQueryAgentAnalyticsLogger`, `bq_schema()`, Claude hook adapter, dry-run, direct-write, local spool, writer attribution, and state locking. |
| Async drainer | `sdk/python/bqaa_drain.py` plus `scripts/bqaa_drain.py` wrapper | Uses a pidfile `flock`, batches spool files, writes through BigQuery Storage Write API with `pyarrow`, falls back to `insert_rows_json`, retries transient failures, and dead-letters failed rows. |
| OpenAI Agents SDK | `sdk/python/bqaa_openai_agents.py` | Optional `openai-agents` dependency. Implements `BQAAOpenAIAgentsProcessor` and `add_bqaa_trace_processor()`. Maps `generation` spans to `LLM_REQUEST`/`LLM_RESPONSE`, `function` spans to `TOOL_STARTING`/`TOOL_COMPLETED`, and other spans to `STATE_DELTA`. |
[...]

PR #237, opened 2026-05-22
producers: add Claude Code plugin artifact build (vendor + tar)

caohy1988:
Maps to **steps 2 + 3 of #234** — adds the plugin source tree, the vendoring build script, and the CI wiring that produces the marketplace-installable artifact. Release-to-PyPI plumbing stays out of scope (PR C / step 4).

Refs #234, #229.

## Plugin tree (`plugins/claude_code/`)

| Path | Purpose |
|---|---|
| `.claude-plugin/plugin.json` | Native Claude Code manifest declaring all nine hooks. Version is a placeholder (`0.0.0+local`) that the build script stamps with the resolved value at release time. |
| `hooks/common.sh` | Shared shell entry. Sets `PYTHONPATH` to the vendored package root, then execs `python -m bigquery_agent_analytics_tracing.claude_code <HookName>`. The drainer subprocess inherits the same `PYTHONPATH` via `os.environ.copy()` — covers the failure mode #234 step 3 calls out. Respects `BQAA_TRACE_ENABLED=false` as a clean short-circuit before spawning Python. |
| `hooks/<hook>.sh` × 9 | One-line stubs that exec common.sh with the hook name. |
| `.gitignore` | Excludes `vendor/` so the canonical source is always `producers/src/` — no drift between the wheel and the plugin. |
| `README.md` | Runtime model, layout, build instructions, required runtime deps on `BQAA_PYTHON`. |

## Build script (`producers/scripts/build_claude_plugin.py`)

| Function | Behavior |
|---|---|
| `resolve_version()` | Prefers installed distribution metadata (matches what `__version__` resolves to at runtime), falls back to reading `producers/pyproject.toml` for fresh checkouts. |
| `vendor_package()` | Copies `producers/src/bigquery_agent_analytics_tracing/` into `plugins/claude_code/vendor/`, skipping `__pycache__` / `*.pyc`. |
| `stamp_manifest_version()` | Writes the resolved version into `.claude-plugin/plugin.json` so wheel and plugin always ship the same version label. |
| `build_tarball()` | Tars the plugin tree into `producers/dist/bigquery-agent-analytics-tracing-claude-code-<version>.tar.gz`, excluding `.gitignore` and `__pycache__` so the artifact is exactly what marketplace install sees. |
| `main()` | argparse CLI (`--dist-dir`, `--skip-tar`). Emits a JSON summary on stdout for CI to log. |

## CI (`.github/workflows/producers-ci.yml`)

- `plugins/claude_code/**` added to `paths:` triggers.
- Build job: installs the wheel produced by `python -m build`, then invokes `python scripts/build_claude_plugin.py`. Installing the wheel first makes `importlib.metadata.version` resolve correctly so the manifest gets a real version stamped, not the pyproject fallback.
- The existing `producers-dist` upload artifact now carries both the wheel/sdist and the plugin tarball.

## Tests — 10 new, 72 total, all green

- `resolve_version` — returns non-empty string; pyproject fallback exercised when `importlib.metadata.version` is forced to fail
- `vendor_package` — copies every expected module file; skips `__pycache__` via a temp fake-source fixture
- `stamp_manifest_version` — version writes correctly, other manifest fields untouched. Snapshot-restore fixture protects the source-controlled manifest from test drift
- `build` end-to-end — tarball name has version baked in, contains manifest + hooks + vendored module, excludes `__pycache__` and `.gitignore`
- `build(skip_tar=True)` — only vendors and stamps, no tarball
- **Vendored plugin end-to-end** — fires `SessionStart` and `UserPromptSubmit` via `python -m bigquery_agent_analytics_tracing.claude_code` with `PYTHONPATH` pointing only at `vendor/` (no `producers/src/` on path). Asserts the SessionStart state file landed and the LLM_REQUEST row hit the dry-run log. This is exactly the runtime model marketplace users will experience.
- `BQAA_TRACE_ENABLED=false` shell short-circuit — invokes `hooks/session_start.sh` with the kill switch on plus a deliberately invalid `BQAA_PYTHON`; asserts exit 0 and no log file written, proving the shell exits before trying to launch Python.

## Acceptance criteria from #234 covered here

- [x] Hook → spool → drainer works from the vendored plugin path with `BQAA_TRACE_ENABLED=true` (vendored plugin smoke tests)
- [x] `BQAA_TRACE_ENABLED=false` cleanly disables emission (shell short-circuit test)
- [x] Dry-run writes expected rows without BigQuery access (carried over from PR A + exercised in vendored plugin tests)
- [x] CI builds the wheel/sdist and plugin artifact
- [x] Plugin artifact is generated from source rather than maintained as a second vendored copy

Deferred to follow-up PRs per #234:
- [ ] `pip install bigquery-agent-analytics-tracing` from PyPI — PR C (step 4)
- [ ] TestPyPI smoke install — PR C (step 4)
- [ ] Marketplace metadata polish + install docs + `/bqaa-setup` interactive flow — PR D (step 5)

## Test plan

- [x] `pytest tests/ -q` — 72/72 pass on Python 3.13.5
- [x] `pyink src/ tests/ scripts/ --check` — clean
- [x] `isort src/ tests/ scripts/` — clean
- [x] `python scripts/build_claude_plugin.py` produces wheel + sdist + plugin tarball
- [x] Vendored plugin: `python -m bigquery_agent_analytics_tracing.claude_code SessionStart` with `PYTHONPATH=plugins/claude_code/vendor` exits 0
- [x] `BQAA_TRACE_ENABLED=false bash plugins/claude_code/hooks/session_start.sh` exits 0 without spawning Python
- [ ] Producers CI — green after maintainer approves the first-time-fork gate
- [ ] Live BigQuery round-trip with the vendored plugin — recommended manual smoke before merging

### DIFF
--- a/.github/workflows/producers-ci.yml
+++ b/.github/workflows/producers-ci.yml
@@ -19,11 +19,13 @@ on:
     branches: [main]
     paths:
       - 'producers/**'
+      - 'plugins/claude_code/**'
       - '.github/workflows/producers-ci.yml'
   pull_request:
     branches: [main]
     paths:
       - 'producers/**'
+      - 'plugins/claude_code/**'
       - '.github/workflows/producers-ci.yml'
 
 concurrency:
@@ -91,7 +93,19 @@ jobs:
       - name: Build sdist and wheel
         run: python -m build
 
-      - name: List artifacts
+      - name: List wheel/sdist
+        run: ls -la dist/
+
+      - name: Install built wheel for plugin build
+        # Make `importlib.metadata.version` resolve correctly so the
+        # plugin manifest gets a real version stamped in, not the
+        # pyproject fallback.
+        run: pip install dist/bigquery_agent_analytics_tracing-*.whl
+
+      - name: Build Claude Code plugin artifact
+        run: python scripts/build_claude_plugin.py
+
+      - name: List plugin artifact
         run: ls -la dist/
 
       - uses: actions/upload-artifact@v4
--- a/.github/workflows/producers-ci.yml
+++ b/.github/workflows/producers-ci.yml
@@ -51,10 +51,10 @@ jobs:
         run: pip install "pyink>=24.3.0" "isort>=5.0"
 
       - name: isort --check
-        run: isort --check-only src/ tests/
+        run: isort --check-only src/ tests/ scripts/
 
       - name: pyink --check
-        run: pyink --check src/ tests/
+        run: pyink --check src/ tests/ scripts/
 
   test:
     name: Producers test (Python ${{ matrix.python-version }})
--- a/plugins/claude_code/.claude-plugin/plugin.json
+++ b/plugins/claude_code/.claude-plugin/plugin.json
@@ -1,0 +1,109 @@
+{
+  "name": "bigquery-agent-analytics-tracing",
+  "description": "Stream Claude Code hook traces to Google BigQuery Agent Analytics agent_events.",
+  "version": "0.0.0+local",
+  "author": {
+    "name": "Google LLC"
+  },
+  "repository": "https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK",
+  "license": "Apache-2.0",
+  "keywords": [
+    "bigquery",
+    "agent-analytics",
+    "observability",
+    "tracing",
+    "claude-code"
+  ],
+  "hooks": {
+    "SessionStart": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/session_start.sh\""
+          }
+        ]
+      }
+    ],
+    "UserPromptSubmit": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/user_prompt_submit.sh\""
+          }
+        ]
+      }
+    ],
+    "PreToolUse": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/pre_tool_use.sh\""
+          }
+        ]
+      }
+    ],
+    "PostToolUse": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/post_tool_use.sh\""
+          }
+        ]
+      }
+    ],
+    "Stop": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/stop.sh\""
+          }
+        ]
+      }
+    ],
+    "SubagentStop": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/subagent_stop.sh\""
+          }
+        ]
+      }
+    ],
+    "Notification": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/notification.sh\""
+          }
+        ]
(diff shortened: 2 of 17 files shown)


caohy1988:
Addressed all three findings in [d1ca11c](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/237/commits/d1ca11c):

**1. Python 3.10 CI failure (high).** `build_claude_plugin.py` and `test_build_claude_plugin.py` both imported `tomllib` unconditionally; that module is 3.11+. Added a `tomllib` → `tomli` fallback in both files and pinned `tomli>=1.1; python_version < '3.11'` as a conditional dev dep. Both 3.10 (via fallback) and 3.11+ (via stdlib) now resolve the same way.

**2. Real vendored drainer test (medium).** The prior vendored smoke tests used `BQAA_DRY_RUN=true`, which short-circuits before spool/drainer — so they only proved the hook module could be imported via `PYTHONPATH`. New test `test_vendored_plugin_spools_and_spawns_drainer_with_inherited_pythonpath`:
- Runs the hook with `BQAA_DRY_RUN=false`, dummy `BQAA_PROJECT_ID` / `BQAA_DATASET`, and `PYTHONPATH=plugins/claude_code/vendor`.
- Sets `BQAA_PYTHON` to a fake recorder shell script — drainer spawn never has to talk to BigQuery.
- Asserts a real spool envelope landed on disk (`LLM_REQUEST` row, correct `project_id`).
- Asserts the recorder fired with argv containing `-m bigquery_agent_analytics_tracing.drain`.
- Asserts the recorder's inherited `PYTHONPATH` matches the vendored package root.

That's the marketplace runtime contract end-to-end: hook → spool → drainer, no wheel install on `BQAA_PYTHON`, `os.environ.copy()` propagating the import path.

**3. `scripts/` format-check (low).** Added `scripts/` to both `isort --check-only` and `pyink --check` invocations in `producers-ci.yml`. Applied pyink formatting; 73/73 tests still green.

Should re-trigger the first-time-fork approval gate again since this is a new head sha. Heads up when you've approved.

caohy1988:
Fixed in [18b226e](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/237/commits/18b226e).

**Root cause:** the build script copied the package source into `vendor/` and stamped the *plugin manifest* with the resolved version, but the *package itself* on `PYTHONPATH` had no installed distribution metadata — so `_writer_identity.get_writer_version()` hit `importlib.metadata.PackageNotFoundError`, fell through to `"0.0.0+local"`, and every row's `attributes.writer.version` came out as the dev sentinel.

**Fix:** the build script now also writes a minimal PEP 376 `.dist-info/METADATA` alongside the vendored package:

```
plugins/claude_code/vendor/
├── bigquery_agent_analytics_tracing/...
└── bigquery_agent_analytics_tracing-<version>.dist-info/
    └── METADATA          # Metadata-Version + Name + Version
```

With this file on `PYTHONPATH`, `importlib.metadata.version()` resolves the real version on first import — the same code path wheel installs already use. **No runtime change** to `_writer_identity.py`; the `PackageNotFoundError` fallback just stops being reached.

Also tightened `vendor_package()` to wipe `vendor/` entirely (not just the package subdir) before re-copying, so stale `.dist-info` from older builds can't accumulate (two distributions with the same name on `sys.path` is undefined behavior).

**Tests (75 total, all green):**

- New: `test_write_vendor_dist_info_creates_pep_376_metadata` — direct coverage of the new helper.
- New: `test_vendor_package_wipes_stale_dist_info` — proves a stale `X-0.0.1-old.dist-info` is removed on re-vendor.
- Extended: `test_build_produces_vendor_tree_and_tarball` — asserts the `.dist-info/METADATA` file is included in the marketplace tarball.
- **Regression guard:** `test_vendored_plugin_spools_and_spawns_drainer_with_inherited_pythonpath` now asserts the spool envelope's row carries `attributes.writer.version == resolve_version()` (and is NOT `"0.0.0+local"`), and `attributes.writer.label` ends with the resolved version. This is the test that would have caught the bug originally.

The fix surface is build-script-only; the runtime `_writer_identity.get_writer_version()` is unchanged because the lookup mechanism is the same — only the inputs improved.

PR #244, opened 2026-05-26
Add per_session_context to categorical evaluator

evekhm:
## Summary
- Adds optional `per_session_context: dict[str, str]` parameter to `classify_sessions_via_api()`
- Allows injecting per-session context (e.g. golden eval expected answers) into the judge prompt before the transcript
- Enables targeted evaluation without modifying the global prompt

## Test plan
- [ ] Verify `classify_sessions_via_api()` works without the new parameter (backward compatible)
- [ ] Pass `per_session_context` with matching session IDs and verify context appears in prompt
- [ ] Pass `per_session_context` with non-matching session IDs and verify no effect


caohy1988 (approved):
## Review — PR #244 head `20e68c3`

This is the clean split-out of the `per_session_context` change from PR #174 — I reviewed this exact diff as part of #174's June 5 pass and re-verified it here.

### The change is correct

- Additive, optional parameter (`per_session_context: dict[str, str] | None = None`) — fully backward-compatible; existing callers unaffected.
- Injection point is right: between the global `prompt_prefix` and the `Transcript:` block, so per-session context (e.g. a golden expected answer) reads as judge instruction, not as transcript content.
- Defensive lookup (`if per_session_context and sid in per_session_context`) — non-matching session IDs are a clean no-op, matching the test plan's third bullet.
- `dict[str, str] | None` bare-union annotation is fine: `pyproject.toml` pins `requires-python = ">=3.10"`.
- Splitting the SDK-surface change out of the scripts-heavy #174 is the right call — if this lands first, #174's identical hunk merges cleanly.

### One ask before merge — add the unit test the PR's own test plan describes

The test plan lists three behaviors and all three checkboxes are unchecked; none are covered anywhere in `tests/` (`grep -rn per_session_context tests/` → empty). All three are cheap to pin with a mocked `generate_content`:

```python
async def test_per_session_context_injected_before_transcript():
    # capture the prompt passed to the mocked client for sid "s1"
    assert "EXPECTED ANSWER..." in captured_prompt
    assert captured_prompt.index("EXPECTED ANSWER") < captured_prompt.index("Transcript:")

async def test_per_session_context_non_matching_sid_is_noop(): ...
async def test_classify_without_per_session_context_unchanged(): ...
```

This parameter is the programmatic hook that #174's golden-Q&A grounding hangs off — worth locking its contract at the SDK layer where it lives, rather than only indirectly via #174's script tests.

### Housekeeping

`mergeStateStatus: BEHIND` — needs a branch update (no conflicts).

## Verdict — APPROVE

The change itself is verified correct. Please add the three-case unit test (it's your own test plan, mechanized) and update the branch before merging.

haiyuan-eng-google (approved):

haiyuan-eng-google:
Closing as redundant: the only substantive change here (per_session_context in classify_sessions_via_api) is already present in main, so this PR is effectively a no-op. Closing rather than merging an empty change.

Issue #257, opened 2026-05-28
Make the Context Graph Colab notebook product-ready


caohy1988:
PR #243 added a runnable Colab notebook for the periodic-materialization codelab. TL feedback after merge asks for a higher-polish, less technical first-reader experience. This issue tracks the Colab-specific work in detail, separate from the broader Context Graph user guide (#255) and the markdown/codelab copy polish (#256).

## Goal

Make the Colab feel like a product demo a new customer can run and understand quickly:

- The first screen should communicate the business outcome: trace an AI agent decision through a BigQuery context graph.
- The notebook should be runnable end-to-end, but markdown cells should stay short and user-facing.
- Advanced mechanics should be explained as tips or notes, not dense paragraphs.
- The notebook should show the value visually, not only through raw tables.

## Validated TL feedback

Source: PR #243 comments 10-16 and final Conversational Analytics comment.

- The current CLI name is too implementation-shaped. Tracked in #245, but Colab should eventually use the final product command, likely `bqaa context-graph` or an alias.
- The notebook should mention lower-cost / no-LLM extraction, not only audit framing, when explaining compiled-only mode.
- Important concepts should use tip blocks so readers can scan.
- Cron / scheduled refresh needs a plain-English definition before use.
- Backfill, state table, checkpoint, and state-key details are too dense for first-time users.
- The notebook should not assume highly technical readers.
- Conversational Analytics should be the hero experience in the larger guide. For this Colab, at minimum, make the CA path visible and screenshot-ready.

### 1. Opening section

- Reframe the first markdown cell around the outcome:
  - Raw agent events land in BigQuery.
  - The SDK builds a context graph.
  - BigQuery lets users ask why an agent made a decision.
- Keep the setup list, but reduce jargon.
- Mention that GQL is shown for transparency and repeatability, while Conversational Analytics is the natural-language consumption path.

### 2. Runtime setup cells

- Keep auth and project setup as-is unless commands are broken.
- Add a short note for Colab Enterprise vs public Colab vs local Jupyter.
- Avoid long paragraphs before setup commands.

### 3. Artifact download cell

- After #248 lands, update any `migration_v5` or periodic-materialization path leaks.
- If #246 lands, replace `curl seed_events.py` with the official SDK demo-data command.
- Keep the pre-merge `BASE` override only while PR branches need it. Remove once artifacts are on main.

### 4. Event generation section

[...]

PR #268, opened 2026-06-01
docs(codelab): add one-row corpus preview query (#257)

caohy1988:
## Summary

Adds the one remaining Colab checklist item from #257: a small **preview query** after the seed step that shows **sessions, total events, and the time range** in a single readable row.

```sql
SELECT COUNT(DISTINCT session_id) AS sessions, COUNT(*) AS events,
       MIN(timestamp) AS earliest_event, MAX(timestamp) AS latest_event
FROM `$PROJECT_ID.$DATASET.agent_events`
```

For the default 5-session run it shows 5 sessions / 30 events over a few minutes; with the realistic scenario it shows ~100 sessions over ~3 days.

SQL is single-line so the generated Colab cell is valid (the generator joins multi-line bash with `&&`). Notebook regenerated; drift `--check` passes.

Closes the last open Colab-specific item tracked in #257.

## Test plan

- [x] Notebook regenerated; `--check` in sync
- [x] Preview cell is a single valid `bq query` (no `&&` mangling)
- [x] `git diff --check` clean

### DIFF
--- a/docs/codelabs/periodic_materialization.md
+++ b/docs/codelabs/periodic_materialization.md
@@ -269,6 +269,16 @@ bqaa seed-events \
 
 The command prints a JSON report. For 5 sessions you should see `"events_generated": 30`, `"events_inserted": 30`, and `"ok": true`.
 
+Preview the corpus at a glance — how many sessions, how many events, and the time range they span — in one row:
+
+<!-- colab:code bash -->
+```bash
+bq query --use_legacy_sql=false \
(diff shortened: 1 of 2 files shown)


caohy1988:
Pre-landing review: LGTM. Reviewed the final #257 corpus-preview diff. Scope is clean: one markdown insertion plus regenerated Colab notebook. The preview query is placed immediately after seeding, uses single-line Standard SQL so the generated Colab bash cell stays valid, reports sessions/events/time range in one row, and the notebook is in sync with the markdown generator. CI is green across Format, Build, and Python 3.10-3.14; merge state is CLEAN.\n\nNote: GitHub does not allow me to submit a formal approval on an own-authored PR, so leaving this as the merge stamp.

Issue #277, opened 2026-06-03
Feat: derive ontology + binding from the property-graph DDL so `bqaa context-graph` no longer requires ontology.yaml/binding.yaml


caohy1988:
## Summary

Today `bqaa context-graph` (the materializer) **requires** `--ontology ontology.yaml` and `--binding binding.yaml` (both are Typer-required; missing either exits 2 before any code runs — `cli.py:1778-1783`). For the codelab and the common product path this means a user authors **three** declarative artifacts — `property_graph.sql`, `ontology.yaml`, `binding.yaml` — that overlap heavily.

**Hard requirement raised:** materialization should run **without** hand-written `ontology.yaml` + `binding.yaml`.

**Feasibility verdict: doable.** The `CREATE PROPERTY GRAPH` DDL (which the user authors anyway as the query surface) plus the physical table schemas (`INFORMATION_SCHEMA.COLUMNS`) together reconstruct everything the materializer actually consumes, with a small set of well-understood, non-blocking gaps. The clean design is an **internal inverse compiler**: derive in-memory `Ontology` + `Binding` objects from `(property-graph DDL + table schemas)` and feed the **existing** `resolve()` path — no downstream changes.

This collapses 3 authored artifacts → 1 (`property_graph.sql`), and 2 hand-written semantic files → 0.

---

## What the materializer actually consumes

The materialize path runs on a frozen `ResolvedGraph` (`resolved_spec.py:124`) built by `resolve(ontology, binding)` (`resolved_spec.py:222`). Field-by-field, here is where each input comes from today and whether it is recoverable without the YAML:

| Resolved-spec input | Today's source | Recoverable from DDL? | Recoverable from `INFORMATION_SCHEMA`? |
|---|---|---|---|
| Graph name | `ontology.ontology` | **Yes** — the `CREATE PROPERTY GRAPH {name}` object name | — |
| Entity / relationship **names** | ontology | **Yes** — node/edge `LABEL` + alias | — |
| Entity/edge → **physical table** | binding `source` | **Yes** — node/edge table `source` in DDL | — |
| Property → **physical column** | binding | **Yes** — `PROPERTIES (col AS name)` | — |
| Entity **primary key columns** | ontology keys → binding cols | **Yes** — node `KEY (...)` | — |
[...]

PR #278, opened 2026-06-03
feat(ontology): offline CREATE PROPERTY GRAPH DDL parser (issue #277, PR 1)

caohy1988:
First building block of **issue #277, Option A** — derive the materialization spec from the property graph alone so `bqaa context-graph` no longer requires hand-written `ontology.yaml` / `binding.yaml`.

## What this PR adds

`bigquery_ontology.graph_ddl_parser` — a pure, **no-I/O** tokenizer + recursive-descent parser for the supported `CREATE PROPERTY GRAPH` DDL. It is the inverse of **both** emitters in the repo and parses the output of each:

- **`bigquery_ontology.graph_ddl_compiler`** (upstream v0): **bare** graph name (`CREATE PROPERTY GRAPH finance`), **bare/dotted** table refs (`raw.accounts AS Account`), property **renames** (`acct_id AS account_id`), and **derived expressions** (`(given_name || ' ' || family_name) AS full_name`).
- **`bigquery_agent_analytics.ontology_property_graph`** (SDK transpiler): **backtick-quoted** refs (`` `proj.ds.t` AS T ``) with `session_id` / `extracted_at` injected into `KEY` / `PROPERTIES`.

References may be backtick-quoted **or** bare/dotted; `--`/`/* */` comments are ignored; keywords are case-insensitive.

## Design notes

- **Types-free AST by design.** BigQuery property-graph DDL declares no property *types* — a `PROPERTIES` clause lists bare column names. Types live in the underlying table columns and are recovered from `INFORMATION_SCHEMA.COLUMNS` in **PR 2**. Inventing a type at parse time would be lossy, so the AST (`ParsedPropertyGraph` / `ParsedNodeTable` / `ParsedEdgeTable` / `ParsedProperty`) carries only what the DDL states. (This is why it targets a dedicated AST rather than `graph_ddl_models.ResolvedGraph`, whose `ResolvedProperty.type` is required.)
- **Derived properties are captured, not dropped.** A `(expr) AS name` property is parsed (string- and nested-paren-aware) into `ParsedProperty(expression=…, derived=True)` for later degradation/diagnostics, since derived properties are not materializable in schema-derived mode.
- **Faithful, not normalizing.** `session_id`/`extracted_at` and `${VAR}` placeholders are captured verbatim; stripping/resolving them is a later step (PR 3).
- Unsupported constructs raise `GraphDDLParseError` with a `line:col` position.

## Scope / what's next (per #277)

- **PR 2** — join the parsed graph with table schemas (`INFORMATION_SCHEMA.COLUMNS`) for property types.
- **PR 3** — build the inferred in-memory ontology+binding / resolved spec.
- **PR 4** — CLI: `bqaa context-graph --property-graph property_graph.sql …`, keeping `--ontology`/`--binding` as overrides.

This PR is standalone and imported by nothing yet, so it cannot affect existing behavior.

## Tests

`tests/test_graph_ddl_parser.py` — 17 tests (pytest-collected), fully offline. Covers:
- the real codelab `property_graph.sql` and the SDK transpiler form;
- bare/dotted refs, `col AS name`, multi-label nodes, comments + case-insensitivity, no-edge graphs;
- **compatibility with the upstream compiler**: parses the exact `compile_graph` finance golden DDL (renames + derived `full_name`), a derived-expression-with-strings/nested-parens case, and a **live `compile_graph` round-trip** that compiles a real ontology+binding and parses the output (stays in sync with the compiler automatically);
- error paths (empty, missing `NODE TABLES`, label-less node, missing `DESTINATION`, unterminated paren).

All green; the broader ontology suite (344 tests) passes; pyink/isort clean; no warnings.


caohy1988:
Thanks for the review — both points addressed in 5beca14.

**P1 (compatibility gap) — fixed.** The parser rejected the upstream `graph_ddl_compiler` output. It now inverts **both** emitters:
- accepts **bare/dotted** graph and table references (`CREATE PROPERTY GRAPH finance`, `raw.accounts AS Account`) in addition to backtick-quoted ones;
- parses **derived** properties, capturing the raw `(expr) AS name` text (string- and nested-paren-aware) into `ParsedProperty(expression=..., derived=True)` for later degradation/diagnostics instead of choking.

New regression tests: the compiler's exact **finance golden DDL** (renames + derived `full_name`), a derived-expression-with-strings/nested-parens case, and a **live `compile_graph` round-trip** (compiles a real ontology+binding, then parses the output — stays in sync with the compiler automatically). The old "unquoted ref raises" test is flipped to assert bare refs parse. 344 ontology tests pass; pyink/isort clean; no warnings.

**P2 (stale tracker) — fixed** on #277: PR 1 produces a **types-free `ParsedPropertyGraph` AST** (not `graph_ddl_models.ResolvedGraph`); types come from `INFORMATION_SCHEMA.COLUMNS` in PR 2. The PR body already documents that rationale.

PR #280, opened 2026-06-03
feat(ontology): synthesize ontology+binding from a property graph (issue #277, PR 3)

caohy1988:
**PR 3 of 4** for **issue #277, Option A**. Branched off fresh `main` after #279 (PR 2) merged — **not stacked**.

## What this PR adds

The keystone of deriving a materialization spec from the property graph alone: `bigquery_ontology.graph_to_spec.derive_ontology_binding()`. It combines the parsed graph (PR 1) with the schema-recovered column types (PR 2) into the upstream **`Ontology` + `Binding`** objects that the SDK's existing **`resolve()`** already consumes — so nothing downstream changes.

```
ParsedPropertyGraph + GraphColumnTypes  ->  derive_ontology_binding()  ->  (Ontology, Binding)  ->  resolve()  ->  materializer
```

## Mapping rules

- **Entity** = one node table; name is the node `LABEL`. The alias is used only to wire edge `REFERENCES`.
- **Relationship** = one edge table; name is the edge `LABEL`; endpoints resolved from referenced node alias → entity name.
- **Property types** come from the joined table schema (PR 2), never a hand-written `type:` field.
- **Primary key** = node `KEY` columns mapped to property names; a key column with no stored property gets a **passthrough** property so the PK stays bindable.
- **SDK metadata columns** (`session_id` / `extracted_at`) are stripped from properties, keys, and edge FK columns — the resolver re-injects them as `metadata_columns`.
- **Derived `(expr) AS name` properties are skipped** (no physical column; `resolve()` rejects `expr` for materialization).

## Limitations (raise, don't guess)

Multi-label nodes (label inheritance), missing primary key, and dangling edge endpoints raise `GraphSpecSynthesisError` — keep an explicit ontology/binding for those. Descriptions/synonyms aren't recoverable from DDL+schema; derived mode pairs best with `--extraction-mode=compiled-only`, where descriptions are unused.

## Tests

`tests/test_graph_to_spec.py` — 11 offline tests: entity name = LABEL + types from schema, metadata-stripped primary key mapped to property names, relationship endpoint resolution + FK stripping, binding target/sources, renamed property + passthrough key, derived-property skipping, multi-label rejection, and — the payoff — an **end-to-end test that feeds the synthesised pair through the SDK's real `resolve()`** and asserts the materializer spec (key columns, re-injected `session_id`/`extracted_at`, relationship endpoints). Full ontology + derive suite (384 tests) green; pyink/isort clean.

## Scope / what's next

Produces the `(Ontology, Binding)` pair. **PR 4** wires it into the CLI: `bqaa context-graph --property-graph property_graph.sql …` (deriving from the parsed graph + live `BigQuerySchemaProvider`), keeping `--ontology`/`--binding` as overrides. Imported by nothing yet → zero behavior change.

### DIFF
--- a/src/bigquery_ontology/graph_to_spec.py
+++ b/src/bigquery_ontology/graph_to_spec.py
@@ -1,0 +1,282 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Synthesise an in-memory ontology + binding from a property graph.
+
+Building block 3 of deriving a materialization spec from the property graph
+alone (GitHub issue #277). Given the types-free
+:class:`~bigquery_ontology.graph_ddl_parser.ParsedPropertyGraph` (PR 1) and the
+:class:`~bigquery_ontology.graph_schema_join.GraphColumnTypes` recovered from
+the table schemas (PR 2), this module produces the upstream
+:class:`~bigquery_ontology.ontology_models.Ontology` and
+:class:`~bigquery_ontology.binding_models.Binding` objects that the SDK's
+existing ``resolve()`` already consumes -- so nothing downstream changes.
+
+Mapping rules
+-------------
+
+* **Entity** = one node table; its name is the node ``LABEL``. The node alias
+  (the DDL-local table handle) is used only to wire edges (``REFERENCES``).
+* **Relationship** = one edge table; its name is the edge ``LABEL``; its
+  endpoints are the entity names of the referenced node aliases.
+* **Property types** come from the joined table schema (PR 2), never from a
+  hand-written ``type:`` field.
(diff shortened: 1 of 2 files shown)


caohy1988:
Thanks — both findings were real and are fixed in b1bf2c0.

**P1 (FK→PK mapping) — fixed.** Relationship bindings now preserve the `REFERENCES` side. Instead of bare positional `from_columns`/`to_columns` (edge FK cols only), `_endpoint_column_refs()` zips each edge `KEY` column with its paired `REFERENCES` column and emits **explicit dict-shape ColumnRefs** `{edge_fk_column: endpoint_pk_property}`, resolving the referenced physical PK column to the endpoint entity's PK *property name* (via the node's column→name map, now threaded out of `_build_entity`). Metadata columns are stripped as **paired** entries, and a `KEY`/`REFERENCES` arity mismatch raises.

Regression added: `SOURCE KEY (fk_b, fk_a) REFERENCES Thing (pk_b, pk_a)` on a node declared `KEY (pk_a, pk_b)` with renamed PK properties now asserts `from_columns == [{"fk_b": "key_b"}, {"fk_a": "key_a"}]` — i.e. `fk_b`→`key_b`, **not** the positional `fk_b`→`key_a` the old code produced.

**P2 (duplicate labels) — fixed.** Now tracking synthesized entity names and relationship names; two node tables (or edges) with distinct aliases but the same `LABEL` raise `GraphSpecSynthesisError` ("Duplicate entity name" / "Duplicate relationship name") instead of constructing invalid models that bypass the loaders' uniqueness checks. Tests added for both.

Full ontology + derive suite (384) green; the end-to-end `resolve()` test still passes on the dict-shape mapping. (FYI the resolver-backed test imports `bigquery_agent_analytics.resolved_spec`; it passes in CI and locally here.)

PR #281, opened 2026-06-03
feat(cli): bqaa context-graph --property-graph derives ontology+binding (issue #277, PR 4)

caohy1988:
**PR 4 of 4** — the capstone of **issue #277, Option A**. With the three offline building blocks merged (#278 parser, #279 schema join, #280 synthesis), this wires them into the CLI so `bqaa context-graph` can materialize from the `CREATE PROPERTY GRAPH` DDL the user already authors — **no hand-written `ontology.yaml` / `binding.yaml`**.

```
bqaa context-graph --project-id P --dataset-id D --property-graph property_graph.sql --lookback-hours 24
```

## What this PR adds

- **`bigquery_agent_analytics.property_graph_spec.derive_ontology_binding_from_ddl()`** — ties the blocks together: resolve `${VAR}` placeholders (envsubst-style, `PROJECT_ID`/`DATASET` defaults) → parse the DDL (PR 1) → read column types from `INFORMATION_SCHEMA.COLUMNS` via `BigQuerySchemaProvider` (PR 2) → synthesize `Ontology` + `Binding` (PR 3).
- **`run_materialize_window()`** — `--ontology`/`--binding` are now optional; added `property_graph_path`. The load step branches (derive from DDL + live schema, or load YAML) behind a guard that exactly one mode is supplied. **Everything downstream — `resolve()`, extraction, materialization, state, validation — is unchanged**; both paths converge on the same `(Ontology, Binding)` objects.
- **CLI** — `bqaa context-graph` (and the `materialize-window` alias) gain `--property-graph`; `--ontology`/`--binding` made optional; the input mode is validated at the boundary with a clear `BadParameter` (exactly one of the two modes).

## Why this is safe

The derived `(Ontology, Binding)` is the *same* shape the YAML loaders produce and is fed to the *same* `resolve()`. Property types come from the live table schema (authoritative), and the binding is schema-consistent by construction, so the existing binding pre-flight passes trivially. `--ontology`/`--binding` remain fully supported as the override for advanced cases (descriptions, renames, derived properties, inheritance).

## Tests

- `tests/test_property_graph_spec.py` — derive-from-DDL end-to-end against a **table-keyed fake BigQuery client** (no live BQ): placeholder resolution, recovered types, binding target/sources, relationship endpoints, unresolved-placeholder surfacing, **and an orchestrator-seam test** that drives `_resolve_ontology_binding` (the exact branch `run_materialize_window` uses) with a real DDL file — proving the `--property-graph` path reads the DDL, derives, and converges on the same `(Ontology, Binding)` the YAML path feeds downstream.
- `tests/test_cli_context_graph_input_modes.py` — the input-mode rule is a **pure helper** (`_validate_context_graph_input_mode`) asserted directly on the raised `BadParameter` message (stable across CI's non-TTY rendering), plus a `CliRunner` smoke that only checks a non-zero exit.

**549 tests** across the materializer + ontology + derive suites pass; pyink/isort clean.

## Follow-up

A separate docs PR will update the codelab to teach the one-artifact flow (this PR keeps to the code path, per the agreed PR breakdown). After this lands, #277's hard requirement — materialize without `ontology.yaml` / `binding.yaml` — is met.

### DIFF
--- a/src/bigquery_agent_analytics/cli.py
+++ b/src/bigquery_agent_analytics/cli.py
@@ -1775,11 +1775,28 @@ def materialize_window(
     dataset_id: str = typer.Option(
         ..., envvar="BQ_AGENT_DATASET", help="BigQuery dataset."
     ),
-    ontology_path: str = typer.Option(
-        ..., "--ontology", help="Path to ontology YAML file."
+    ontology_path: Optional[str] = typer.Option(
+        None,
+        "--ontology",
+        help=(
+            "Path to ontology YAML file. Use with --binding, or omit both and"
+            " pass --property-graph to derive them from the graph DDL."
+        ),
     ),
-    binding_path: str = typer.Option(
-        ..., "--binding", help="Path to binding YAML file."
+    binding_path: Optional[str] = typer.Option(
+        None,
+        "--binding",
+        help="Path to binding YAML file (use with --ontology).",
+    ),
+    property_graph_path: Optional[str] = typer.Option(
+        None,
+        "--property-graph",
+        help=(
+            "Path to a CREATE PROPERTY GRAPH .sql file. Derives the ontology"
+            " and binding from the graph definition + table schemas, so no"
+            " hand-written ontology.yaml/binding.yaml is needed. Mutually"
+            " exclusive with --ontology/--binding."
+        ),
     ),
     events_table: str = typer.Option(
         "agent_events",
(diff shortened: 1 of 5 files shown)


caohy1988:
All three addressed in 0025713.

**P1 (CI red) — fixed.** The failure was asserting on Rich-rendered CLI output, whose wrapping is unstable under CI's non-TTY width (same pattern as the earlier help-test). The input-mode rule is now a pure helper `_validate_context_graph_input_mode()`, asserted directly on the raised `typer.BadParameter` message; the `CliRunner` tests only check a non-zero exit. No more substring assertions on rendered output.

**P2 (orchestrator coverage) — added.** Extracted the load branch into `_resolve_ontology_binding()` (the exact seam `run_materialize_window` calls) and added a test that drives it with a real `property_graph.sql` file + a table-keyed fake BigQuery client, asserting the `--property-graph` path reads the DDL, derives, and converges on the same `(Ontology, Binding)` the YAML path feeds downstream — plus both/neither guard tests at that seam.

**P3 (docs) — fixed.** `run_materialize_window` docstring now documents `property_graph_path` and the mutually-exclusive input rule; PR body refreshed (test section + count).

549 tests across the materializer + ontology + derive suites pass locally; pyink/isort clean. Pushed, so CI should go green on the new head.

Issue #286, opened 2026-06-04
Support --property-graph in scheduled Cloud Run and Terraform deploys


caohy1988:
## Problem

The local/product CLI path now supports the one-artifact Context Graph flow:

```bash
bqaa context-graph --project-id P --dataset-id D --property-graph property_graph.sql --lookback-hours 24
```

That completed the #277 local materialization path: `property_graph.sql` + live table schemas can derive the in-memory ontology and binding, so readers no longer need to hand-author or render `ontology.yaml` / `binding.yaml` for the common, rename-free path.

The scheduled production deploy path has not caught up yet. The Cloud Run Job image/deploy helpers still stage and consume the explicit YAML pair:

- `examples/migration_v5/periodic_materialization/build_image.sh` stages `ontology.yaml` and `binding.yaml`.
- `examples/migration_v5/periodic_materialization/deploy_cloud_run_job.sh` stages `ontology.yaml` and `binding.yaml`.
- `examples/migration_v5/periodic_materialization/run_job.py` retargets `binding.yaml` and calls `run_materialize_window(..., ontology_path=..., binding_path=...)`.
- The Terraform deployment mirrors that container/runtime shape, so production users still see the older artifact contract.

This creates a product mismatch: local codelab/blog path is one-artifact, but scheduled Cloud Run/Terraform deployment still implies ontology/binding are required.

### Retargeting is required

The migration-v5 production wrapper currently retargets `binding.yaml` and `table_ddl.sql` from the canonical demo dataset to the customer project/dataset. `property_graph.sql` is currently hardcoded to `test-project-0728-467323.migration_v5_demo.*`, not `${PROJECT_ID}` / `${DATASET}` placeholdered.

A production `--property-graph` deploy must therefore retarget or placeholderize both:

- `table_ddl.sql` before applying table DDL; and
- `property_graph.sql` before `run_materialize_window(property_graph_path=...)` reads it and queries `INFORMATION_SCHEMA.COLUMNS`.

Acceptance must include a deploy against a project/dataset different from the canonical demo values, proving schema lookup and writes hit the customer dataset.

### Migration v5 should stay explicit YAML by default

[...]

PR #288, opened 2026-06-05
feat(materialize): derive mode can target a separate graph dataset (issue #286 enabler)

caohy1988:
**PR 1 for #286** (scheduled-deploy parity for `--property-graph`). This is the SDK **enabler** the deploy wiring needs; the deploy-script / `run_job.py` / Terraform / docs changes follow in subsequent PRs.

## Why this is needed (a real finding while scoping #286)

Derive mode (`--property-graph`) used a single `dataset_id` for **both**:
- the **events** read (`{project}.{dataset_id}.agent_events`), and
- the **derived graph target** — `${PROJECT_ID}`/`${DATASET}` resolution in `property_graph.sql`, the `INFORMATION_SCHEMA.COLUMNS` lookup, and the synthesized binding's target.

That's fine for the codelab's single-dataset shape. But the production Cloud Run wrapper (`examples/migration_v5/periodic_materialization/run_job.py`) deliberately uses a **read-only events dataset** and a **separate graph dataset** (the README's read-only-source contract; today it decouples them by *retargeting* `binding.yaml`). With one `dataset_id`, a naive `--property-graph` deploy would read schema from — and write the graph into — the **events** dataset. So deploy parity can't just pass `property_graph_path`; it needs to target the graph dataset separately.

## Change

Add optional `graph_project_id` / `graph_dataset_id` to `run_materialize_window` (and the `_resolve_ontology_binding` seam). In schema-derived mode they target the derived graph's project/dataset; they **default to `project_id` / `dataset_id`**, so single-dataset callers (codelab, the `bqaa context-graph` CLI) are completely unchanged. Events are still read from `dataset_id`.

`run_job.py` (a later PR) will set `graph_dataset_id` = the graph dataset so events stay read-only and the graph lands in the right place.

## Tests

- `test_orchestrator_separate_graph_dataset`: with `graph_*` set, the derived binding target **and** sources resolve to the graph dataset (`graph-proj.graph_ds.*`), not the events dataset.
- `test_orchestrator_graph_dataset_defaults_to_events_dataset`: omitting `graph_*` targets `dataset_id` (single-dataset unchanged).

167 tests pass across the materializer + derive + CLI suites; backward-compatible (new params default to current behavior). Called by nothing new yet — pure enabler.

Refs #286.

### DIFF
--- a/src/bigquery_agent_analytics/materialize_window.py
+++ b/src/bigquery_agent_analytics/materialize_window.py
@@ -1238,6 +1238,8 @@ def _resolve_ontology_binding(
     project_id: str,
     dataset_id: str,
     bq_client: Any,
+    graph_project_id: Optional[str] = None,
+    graph_dataset_id: Optional[str] = None,
 ):
   """Resolve the ``(Ontology, Binding)`` pair from one of two input modes.
 
@@ -1246,6 +1248,13 @@ def _resolve_ontology_binding(
     file) -- parse it, read column types from ``INFORMATION_SCHEMA.COLUMNS``
     via ``bq_client``, and synthesise the pair in memory (#277). No
     hand-written ontology/binding required.
+
+  ``graph_project_id`` / ``graph_dataset_id`` target the derived graph (its
+  ``${PROJECT_ID}`` / ``${DATASET}`` placeholder resolution, schema lookup, and
+  binding target) at a project/dataset distinct from the events ``project_id`` /
+  ``dataset_id``. They default to ``project_id`` / ``dataset_id`` (single-dataset
+  shape, e.g. the codelab). A two-dataset deploy -- a read-only events dataset
+  and a separate graph dataset -- sets them to the graph dataset.
 
   Exactly one mode must be supplied; both or neither raises ``ValueError``.
   This is the single seam the orchestrator uses to obtain its spec, so both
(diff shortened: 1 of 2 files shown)


caohy1988:
Good catch — fixed in the latest commit.

The enabler moved the derived target, schema lookup, and graph writes to `graph_dataset_id`, but the per-run **state/checkpoint table** still defaulted to `project_id`/`dataset_id` (the read-only events dataset). That's exactly the half-fixed contract you flagged.

**Fix:** the state table now **defaults to the graph target** (`graph_project_id`/`graph_dataset_id`) when split-dataset mode is in use, via a small `_state_table_defaults()` seam. An explicit `state_table` still overrides, and single-dataset callers (`graph_*` unset) are unchanged — so the per-run checkpoint never lands in a read-only events dataset.

Tests added per your request: `test_state_table_defaults_to_graph_dataset_in_split_mode` (graph_dataset_id + no state_table → state ref resolves to the graph dataset) and `test_state_table_defaults_to_events_in_single_dataset`.

The full split-dataset contract now holds: events read from `project_id.dataset_id.agent_events`; graph schema + materialized tables + **state table** in `graph_project_id.graph_dataset_id`.

caohy1988:
Fresh re-review at a9626cd.

Findings: none.

The state-table fix closes the remaining split-dataset foot-gun. The current contract is now coherent end to end: event discovery still reads from project_id.dataset_id.agent_events, derived DDL placeholder resolution + INFORMATION_SCHEMA schema reads + synthesized binding targets the graph_project_id/graph_dataset_id pair, graph writes follow that derived binding, and the default state/checkpoint table now follows the writable graph target unless an explicit state_table override is supplied. Single-dataset callers keep the previous default.

I verified the patch structure and the new tests cover the two important cases: split mode with no explicit state_table defaults to the graph dataset, and single-dataset mode defaults to the events dataset. GitHub CI is green across Format, Build, Python 3.10-3.14, scan, and CLA.

Ready to merge. Recommendation after merge: build PR 2 unstacked from fresh main, wiring run_job.py to pass graph_dataset_id in the --property-graph branch and keeping the explicit YAML branch for migration_v5.

PR #289, opened 2026-06-05
feat(deploy): run_job.py --property-graph mode for scheduled materialization (issue #286, PR 2)

caohy1988:
**PR 2 for #286** — teaches the Cloud Run Job **runtime wrapper** (`run_job.py`) to materialize in schema-derived mode. Built on the #288 split-dataset enabler; branched **unstacked** off fresh `main`.

## What changes

- **New env var `BQAA_PROPERTY_GRAPH`** — when set to a staged `CREATE PROPERTY GRAPH` `.sql` filename (e.g. `property_graph.sql`), the wrapper runs in schema-derived mode: ontology + binding are derived from the property graph + table schemas, so **no `ontology.yaml`/`binding.yaml` is staged and no binding retargeting happens**. Unset ⇒ the explicit ontology+binding path (migration-v5 / compiled extractor), unchanged.
- **Split-dataset wiring (the #288 contract):** the property-graph branch calls `run_materialize_window(property_graph_path=…, graph_project_id=project_id, graph_dataset_id=graph_dataset_id, …)`. Events stay **read-only** in `BQAA_EVENTS_DATASET_ID`; schema lookup, materialized tables, and the state table all target the **writable** `BQAA_GRAPH_DATASET_ID`. The `${PROJECT_ID}`/`${DATASET}` placeholders in `property_graph.sql` are resolved by the materializer from `project_id`/`graph_dataset_id`.
- **Table bootstrap handles placeholdered artifacts.** `_bootstrap_entity_tables` now resolves `${PROJECT_ID}`/`${DATASET}` (in addition to the existing canonical-prefix rewrite), so the codelab-style placeholdered `table_ddl.sql` is retargeted to the graph dataset before the first run. This is a **hard prerequisite**: derive mode reads `INFORMATION_SCHEMA` and can't create the tables itself.
- **No rewriter** (per the agreed simplification): property-graph mode just requires placeholdered, rename-free artifacts and reuses existing `${}` resolution. `reference_extractors_module`/`bundles_root` are passed **only** in the explicit branch (compiled extraction stays an explicit-mode concern).

## Tests

`tests/test_run_job_property_graph.py` loads `run_job.py` and drives `main()` with BigQuery, dataset bootstrap, and the orchestrator all faked:
[...]

PR #290, opened 2026-06-05
feat(deploy): --property-graph mode for the Cloud Run deploy + image scripts (issue #286, PR 3)

caohy1988:
**PR 3 for #286** — schema-derived mode for the **deploy + image scripts**, on the #288/#289 runtime. Unstacked off `main`.

## Operator UX (one-artifact contract)

```bash
./deploy_cloud_run_job.sh --project P --region R --events-dataset E --graph-dataset G \
  --schedule "0 */6 * * *" --property-graph path/to/property_graph.sql
```

`--property-graph` points at the `CREATE PROPERTY GRAPH` `.sql`; the deploy stages `table_ddl.sql` from the **same directory** (codelab bundles ship both together). No `ontology.yaml`/`binding.yaml` to author.

## `deploy_cloud_run_job.sh`

- **New `--property-graph PATH`**, validated at the deploy boundary (before any gcloud):
  - **Rejects `--extraction-mode=compiled-only`** — derived mode stages no reference extractors, so it would `empty_extract` at runtime (per review note on #289).
  - **Requires a sibling `table_ddl.sql`** — derive mode reads `INFORMATION_SCHEMA`, so the graph tables must be bootstrappable before the first run; missing file fails with an actionable error.
- **Staging branch:** property-graph mode stages only `property_graph.sql` + `table_ddl.sql` (no ontology/binding/reference_extractor); explicit mode unchanged.
- **Env:** sets `BQAA_PROPERTY_GRAPH=property_graph.sql` so the #289 runtime derives the spec (and targets the graph dataset via the #288 enabler).
- Explicit ontology+binding (migration-v5 / compiled-extractor) path is the default when `--property-graph` is omitted.

## `build_image.sh`

Mirrors the same `--property-graph` staging branch (graph + sibling table DDL vs. the explicit ontology/binding/extractor set), so bash-built and Terraform-built images stay consistent.

## Tests (`tests/test_deploy_property_graph_mode.py`)

- **Shell:** invoke the deploy script and assert the two boundary rejections (`compiled-only`; missing sibling `table_ddl.sql`) and the missing-file case exit non-zero with actionable messages — all **before any gcloud call**.
- **Static:** pin the staging + `BQAA_PROPERTY_GRAPH` contract for both scripts.

5 tests pass; `bash -n` clean on both scripts.

## Scope / next

Deploy + image scripts only. **PR 4**: Terraform property-graph mode (matching artifact/input mode + env). **PR 5**: deploy README + Terraform README docs.

Refs #286. Builds on #288 + #289 (merged).

### DIFF
--- a/examples/migration_v5/periodic_materialization/build_image.sh
+++ b/examples/migration_v5/periodic_materialization/build_image.sh
@@ -56,6 +56,11 @@ Optional:
   --tag TAG              Image tag (default: short git SHA of HEAD, or
                          a timestamp if HEAD isn't a git ref).
   --create-repo          Create the AR repo if it doesn't exist.
+  --property-graph PATH   Schema-derived mode (#286): stage this CREATE
+                         PROPERTY GRAPH .sql plus a sibling table_ddl.sql
+                         instead of ontology.yaml/binding.yaml/
+                         reference_extractor.py. The runtime derives the
+                         ontology + binding from them.
   -h | --help            Show this help.
 
 Outputs (last line of stdout):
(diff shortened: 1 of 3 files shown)


caohy1988:
Fresh review at e25addf.

[P1] Enforce placeholdered property-graph artifacts before staging.

`deploy_cloud_run_job.sh` currently validates only that `--property-graph` exists and that the sibling `table_ddl.sql` exists (`deploy_cloud_run_job.sh:279-288`); `build_image.sh` has the same existence-only validation (`build_image.sh:95-105`). That leaves the core #286 safety bug open: an operator can pass a hardcoded `property_graph.sql` (for example the migration-v5 snapshot with `test-project-0728-467323.migration_v5_demo.*`) and the deploy will stage it successfully. At runtime, the derive helper only resolves `${PROJECT_ID}` / `${DATASET}` if they are present; it does not rewrite arbitrary hardcoded refs. The graph schema lookup/materialized binding can therefore point at the canonical demo dataset or any stale authoring dataset instead of the requested graph dataset.

This is exactly the failure mode #286 decided to avoid by requiring placeholdered artifacts for property-graph mode. Please make both scripts fail before any build/deploy when `--property-graph` mode is selected and either `property_graph.sql` or its sibling `table_ddl.sql` does not contain the required `${PROJECT_ID}` and `${DATASET}` markers. Add shell tests for the deploy script rejection, and static/build-helper coverage for `build_image.sh` so Terraform-built images cannot silently bake the wrong artifacts either.

Everything else in the PR looks aligned with the selected UX: single `--property-graph PATH`, sibling table DDL, compiled-only rejected at deploy boundary, property-graph mode stages only graph + table DDL, explicit migration-v5 staging remains unchanged, and `BQAA_PROPERTY_GRAPH=property_graph.sql` is wired into the job env. CI was almost green when reviewed; only Python 3.12 was still in progress.

caohy1988:
Good catch — fixed in the latest commit.

You're right: validating *existence* left the original #286 failure mode open (a hardcoded graph DDL, e.g. the migration-v5 snapshot, could be staged and derive against the wrong dataset).

**Fix:** both `deploy_cloud_run_job.sh` and `build_image.sh` now **reject property-graph mode unless both `property_graph.sql` and the sibling `table_ddl.sql` contain `${PROJECT_ID}` and `${DATASET}`** (`grep -qF`), with an actionable error naming the offending file. Enforced at the deploy/build boundary — before any gcloud or image bake — so Terraform-built images can't bake hardcoded artifacts either.

**Tests added:** a hardcoded `property_graph.sql` and a hardcoded `table_ddl.sql` are each rejected with the placeholder message; static assertions pin the `grep` checks in both scripts. The earlier rejections (compiled-only, missing sibling, missing file) still fire ahead of the placeholder check. 7 tests pass; `bash -n` clean.

caohy1988:
Fresh re-review at fbbbc87.

Findings: none.

The P1 is fixed. Both deploy surfaces now enforce the exact placeholder contract #286 relies on:

- `deploy_cloud_run_job.sh` rejects property-graph mode unless both `property_graph.sql` and the sibling `table_ddl.sql` contain `${PROJECT_ID}` and `${DATASET}`, before any gcloud/build side effects.
- `build_image.sh` applies the same check before baking a Terraform-consumed image, so hardcoded artifacts cannot slip in through the image-build path.
- The error messages name the offending file and explain the wrong-dataset risk / explicit-mode fallback.
- Tests now cover hardcoded `property_graph.sql`, hardcoded `table_ddl.sql`, missing sibling, missing graph file, compiled-only rejection, and static staging/env/grep contracts.

The rest of the PR still looks right: single `--property-graph PATH`, sibling `table_ddl.sql`, property-graph mode stages only graph + table DDL, explicit migration-v5 ontology/binding/reference-extractor staging is unchanged, and `BQAA_PROPERTY_GRAPH=property_graph.sql` is wired for the #289 runtime.

CI is still running the Python matrix at the time of this review; Format/check/scan/CLA are green. Ready to merge once the remaining checks are green, then proceed to PR 4 (Terraform) unstacked off fresh main.

PR #291, opened 2026-06-05
feat(deploy): Terraform property-graph mode parity (issue #286, PR 4)

caohy1988:
**PR 4 for #286** — Terraform parity for schema-derived mode, so the bash deploy and the Terraform module expose the **same artifact-mode contract**. Unstacked off `main`, on #288/#289/#290.

## What changes

- **New variable `property_graph`** (`bool`, default `false`). When `true`, the module sets `BQAA_PROPERTY_GRAPH=property_graph.sql` on the Cloud Run Job, so the #289 runtime derives the ontology + binding (targeting the graph dataset via the #288 enabler) instead of an explicit ontology/binding pair. The published image must be built with `build_image.sh --property-graph` (PR 3) so the placeholdered `property_graph.sql` + `table_ddl.sql` are staged.
- **Env wiring** mirrors the bash deploy's `ENV_VARS`: the `BQAA_PROPERTY_GRAPH` entry is merged in only when `var.property_graph` is true.
- **Plan-time precondition** rejects `property_graph = true` + `extraction_mode = "compiled-only"` (derived mode stages no reference extractors → would `empty_extract`), mirroring the bash deploy's boundary rejection.
- **Explicit ontology/binding** (compiled-extractor) path unchanged when `property_graph = false`. `terraform.tfvars.example` documents the toggle.

## Validation

`terraform validate` → **"The configuration is valid."**; `terraform fmt` clean (verified locally with the real binary).

## Tests (`tests/test_terraform_property_graph_mode.py`)

Static HCL assertions pin the variable, the **conditional** `BQAA_PROPERTY_GRAPH` env wiring, and the compiled-only **precondition** — so the bash and Terraform surfaces can't silently drift — plus a `terraform fmt -check` gate when the binary is present. (Full `terraform validate` needs provider init/network, so it's run in dev rather than CI.) 12 tests pass (TF + deploy-script).

## Scope / next

Terraform module only. **PR 5 (last):** deploy README + Terraform README — show `property_graph.sql` as the default rename-free production path, with a short explicit-ontology/binding advanced note (migration-v5 compiled path).

Refs #286.

### DIFF
--- a/examples/migration_v5/periodic_materialization/terraform/main.tf
+++ b/examples/migration_v5/periodic_materialization/terraform/main.tf
@@ -66,6 +66,12 @@ locals {
     var.extraction_mode == "compiled-only" ? {
       BQAA_REFERENCE_EXTRACTORS_MODULE = "reference_extractor"
     } : {},
+    # Schema-derived mode (#286): tell the runtime to derive the spec from the
+    # staged ``property_graph.sql`` instead of the explicit ontology/binding
+    # pair. Mirrors the bash deploy's ``BQAA_PROPERTY_GRAPH`` wiring.
+    var.property_graph ? {
+      BQAA_PROPERTY_GRAPH = "property_graph.sql"
+    } : {},
   )
 }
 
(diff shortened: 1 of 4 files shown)


caohy1988:
Fresh review at b924040.

Findings: none.

The Terraform parity change matches the bash/runtime contract:

- `var.property_graph` is a narrow bool toggle, defaulting false, so existing explicit ontology/binding deployments remain unchanged.
- When true, the module conditionally wires `BQAA_PROPERTY_GRAPH=property_graph.sql`, matching the image layout produced by `build_image.sh --property-graph`.
- The plan-time precondition rejects `property_graph = true` with `extraction_mode = "compiled-only"`, which mirrors the bash deploy boundary and prevents the no-reference-extractor empty-extraction path.
- Vertex AI API/IAM still follows `ai-fallback`; that is correct because property-graph mode is currently ai-fallback only.
- Tests pin the variable, conditional env wiring, precondition, tfvars example, and fmt gate.

Operationally, the remaining contract is documentation: users must build the image with `build_image.sh --property-graph path/to/property_graph.sql` before applying Terraform with `property_graph = true`. That is already called out in the variable/tfvars text and should be expanded in PR 5.

CI is still running the Python matrix at review time; Format/check/scan/CLA are green. Ready to merge once the remaining checks are green, then proceed to PR 5 (deploy + Terraform README docs) unstacked off fresh main.

PR #292, opened 2026-06-05
docs(deploy): document --property-graph schema-derived deploy mode (issue #286, PR 5)

caohy1988:
**PR 5 for #286 (the last one)** — docs closeout. The bash deploy and Terraform module both support schema-derived mode now (#288–#291); this documents it as the simple **rename-free production path**, with the explicit ontology/binding (migration-v5 compiled) path as the advanced override. Docs only.

## Deployment README (`examples/migration_v5/periodic_materialization/README.md`)

New **"Schema-derived deploy (`--property-graph`)"** section after the deploy command:
- the one-artifact flow — `--property-graph path/to/property_graph.sql` with a sibling placeholdered `table_ddl.sql`;
- the **placeholder** requirement and the **no-`compiled-only`** rule (both rejected at the deploy boundary);
- the read-only-events / writable-graph split (events stay in the events dataset; graph + state in the graph dataset);
- an **advanced note** for explicit ontology + binding (descriptions, inheritance, derived properties, renames, compiled extractor — the migration-v5 path).

## Terraform README (`.../terraform/README.md`)

- New **`property_graph = true`** subsection (build the image with `build_image.sh --property-graph`; the module sets `BQAA_PROPERTY_GRAPH`; plan-time precondition rejects `compiled-only`).
- Added `property_graph` to the tfvars override list, the Cloud Run Job env-var list, and the build-image note.

## #286 — complete

| PR | Surface | State |
|----|---------|-------|
| #288 | SDK enabler (split-dataset events/graph + state table) | merged |
| #289 | `run_job.py` derive mode | merged |
| #290 | deploy + image scripts | merged |
| #291 | Terraform module | merged |
| this | deploy + Terraform docs | — |

With this, the local one-artifact flow (#277) reaches the **scheduled production path**: a Cloud Run Job / Terraform deploy can run on `property_graph.sql` alone, no hand-written `ontology.yaml`/`binding.yaml`, with the explicit pair preserved as the advanced override.

Refs #286.

PR #294, opened 2026-06-08
fix(deploy): run_job table-DDL bootstrap mis-split on comment semicolons (found by --property-graph smoke)

caohy1988:
Found by the **live `--property-graph` deploy smoke** run after #292 merged (the deploy-parity arc, #286). Bugfix + CHANGELOG.

## The bug

`run_job.py::_bootstrap_entity_tables` applies the table DDL by splitting on `;` and running each fragment. It split **without stripping `--` comments**, so a comment that contains `;` — e.g. the codelab `table_ddl.sql`'s `-- materializer fills automatically; they are required` — became a **comment-only fragment** that BigQuery rejects with `Syntax error: Unexpected end of statement`. The graph-table bootstrap then failed for *any* `--property-graph` deploy using codelab-style artifacts (the placeholdered ones the new mode requires).

```
{"severity":"ERROR","message":"unexpected error...","error_type":"BadRequest",
 "error":"400 Syntax error: Unexpected end of statement at [5:36]..."}
```

## The fix

`_split_sql_statements()` is a **quote- and comment-aware** scanner: it removes `--` line comments and `/* */` block comments and splits on `;` **only at the top level** — never inside single quotes, double quotes, or backtick identifiers (honoring backslash escapes). This is safe both for the codelab comment-with-`;` case and for customer DDL with `;`/`--` inside `OPTIONS(description=...)`, `DEFAULT` string literals, or quoted identifiers. Regression tests cover all of these.

## Live smoke (post-fix) — clean end-to-end

`run_job.py` in property-graph mode, scratch **split** datasets on a real project:

- `spec mode = property-graph`, `BQAA_PROPERTY_GRAPH=property_graph.sql` — no ontology/binding staged ✓
- table DDL applied: **5 statements** (bootstrap fixed) ✓
- **5/5 sessions materialized, 0 failed**; `DecisionRequest 5 / DecisionOption 15 / DecisionOutcome 5` + edges ✓
- graph tables **and** `_bqaa_materialization_state` in the **graph** dataset ✓
- **events dataset read-only** (only `agent_events`) ✓

## CHANGELOG

Added the `Unreleased` entry for the schema-derived (`--property-graph`) local→production materialization (#277 + #286 chains) and this fix.

25 tests pass across the `run_job` / deploy-script / Terraform / derive suites. Refs #286.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,38 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Schema-derived (`--property-graph`) materialization, local → production**
+  (local: [#277](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/277)
+  via [#278](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/278)–[#281](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/281),
+  [#285](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/285),
+  [#287](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/287);
+  deploy: [#286](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/286)
+  via [#288](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/288)–[#292](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/292))
+  — `bqaa context-graph --property-graph property_graph.sql` derives the ontology
+  + binding from the property graph + live table schemas, so no hand-written
+  `ontology.yaml` / `binding.yaml` is needed for rename-free graphs. Scheduled
+  production reaches parity: the Cloud Run deploy script, image builder,
+  `run_job.py` (`BQAA_PROPERTY_GRAPH`), and the Terraform module all accept the
+  one-artifact flow (placeholdered `property_graph.sql` + sibling
+  `table_ddl.sql`), with a split read-only-events / writable-graph contract
+  (events read-only; graph tables + `_bqaa_materialization_state` land in the
+  graph dataset). Explicit `--ontology`/`--binding` is preserved as the advanced
+  override (descriptions, inheritance, derived properties, renames, the
+  migration-v5 compiled extractor).
+
+### Fixed
+
+- **`run_job.py` table-DDL bootstrap mis-split on comment semicolons**
+  ([#286](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/286))
+  — `_bootstrap_entity_tables` split the DDL on `;` without stripping `--`
+  comments, so a comment containing `;` (e.g. the codelab `table_ddl.sql`'s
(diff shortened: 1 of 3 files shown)


caohy1988:
[P2] Make `_split_sql_statements` quote-aware before using it for customer `table_ddl.sql`.

The live-smoke bug is real, but the current fix at `run_job.py:298-309` strips `--` comments globally and then splits on every `;`. That solves the codelab comment case, but it still corrupts valid BigQuery DDL that this production deploy path can reasonably receive from a user. For example, a table/column `OPTIONS(description="review score; chosen by policy")`, a `DEFAULT "a;b"`, or any quoted/backticked text containing `--` will be modified or split into bogus statements before BigQuery ever sees it. The new test even asserts every returned statement contains no `;` (`tests/test_run_job_property_graph.py:119-122`), which would reject valid semicolons inside string literals rather than just removing statement terminators.

Because `--property-graph` deploy mode now accepts operator-provided, placeholdered `table_ddl.sql`, this is not just an internal committed-artifact limitation. We should either implement a small scanner that ignores `--` comments and statement-terminating `;` only when outside single quotes, double quotes, and backtick identifiers, or explicitly reject unsupported DDL forms before deploy. The scanner is small and safer: preserve quoted content verbatim, skip line comments only outside quotes, and split only on semicolons outside quotes. Add regression coverage for `OPTIONS(description="has; semicolon")` and a quoted `--` sequence so this does not regress again.

Everything else in the PR looks right: the smoke caught a real bug, `_bootstrap_entity_tables` is the right boundary, and the CHANGELOG entry accurately records the #277/#286 local-to-production schema-derived flow.

caohy1988:
Good call — fixed.

The regex approach (strip `--` globally, split on every `;`) was unsafe for customer DDL with `;` / `--` inside string literals or backticks. Replaced it with a **quote- and comment-aware char scanner**: removes `--` line comments and `/* */` block comments, and splits on `;` **only at the top level** — never inside single quotes, double quotes, or backtick identifiers, honoring backslash escapes. (Also dropped the now-unused `re` import.)

Tests added per your request:
- `;` and `--` inside `OPTIONS(description="has; semicolon and -- dashes")` and `DEFAULT 'x;y'` → one statement, contents preserved.
- block comment `/* inline; comment */` removed without gluing tokens; escaped quote then `;` inside a string (`'a\';b'`) handled.
- the original comment-with-`;` regression still passes.

178 tests pass across run_job / deploy-script / Terraform / derive / materializer suites.

caohy1988:
Fresh re-review at 23a0430.

Findings: none on code.

The P2 is fixed. `_split_sql_statements` is now a quote/comment-aware scanner: it preserves semicolons and `--` inside single-quoted strings, double-quoted strings, and backtick identifiers; removes line and block comments outside quotes; and splits only on top-level statement terminators. The new tests cover the exact risky customer-DDL cases from the review: `OPTIONS(description="has; ... -- ...")`, `DEFAULT 'x;y'`, escaped quotes with in-string semicolons, block comments, and the original semicolon-in-comment regression.

One small non-blocking wording cleanup before merge would be good: the PR body and the CHANGELOG entry still describe the fix as "strips line comments before splitting." That was true for the first fix, but the final implementation is stronger: quote-aware line/block comment handling plus top-level semicolon splitting. Updating that wording would prevent future readers from thinking the deploy path is still regex/quote-unaware.

CI is still running the Python matrix at review time; Format/check/scan/CLA are green. Ready to merge once remaining checks are green, with the optional wording refresh above.

caohy1988:
Wording updated — thanks. The CHANGELOG `Fixed` entry and the PR body now describe the actual fix: a quote- and comment-aware splitter (strips `--` line + `/* */` block comments, top-level-only `;` splitting that ignores string literals and backtick identifiers), not just "strips line comments."

PR #295, opened 2026-06-08
chore(release): 0.3.3 — schema-derived (--property-graph) Context Graph, local → production

caohy1988:
Cuts **0.3.3**, packaging the schema-derived (`--property-graph`) Context Graph work now that it's complete and live-smoke-validated.

## What's in 0.3.3

- **Local** ([#277](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/277)): `bqaa context-graph --property-graph property_graph.sql` derives the ontology + binding from the property graph + live table schemas — no hand-written `ontology.yaml`/`binding.yaml` for rename-free graphs (#278–#281, #285, #287).
- **Production** ([#286](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/286)): the same one-artifact flow on the scheduled path — deploy script, image builder, `run_job.py` (`BQAA_PROPERTY_GRAPH`), and Terraform — with a split read-only-events / writable-graph dataset contract (#288–#292).
- **Fix** (#294): quote/comment-aware table-DDL bootstrap splitter (found by a live `--property-graph` deploy smoke).

Explicit `--ontology`/`--binding` is retained as the advanced override (descriptions, inheritance, derived properties, renames, migration-v5 compiled extractor).

## Changes

- `pyproject.toml`: `0.3.2` → `0.3.3`.
- `CHANGELOG.md`: `[Unreleased]` → dated `[0.3.3]` with a release-highlights summary.
- migration-v5 deploy README: PyPI pin `>=0.3.2` → `>=0.3.3` (the documented `--property-graph` deploy path requires 0.3.3) + a note on what 0.3.3 adds.

Patch bump (continues the 0.3.x migration-v5 production track); backward-compatible — `--property-graph` is additive, the explicit YAML path is unchanged.

PR #296, opened 2026-06-08
docs(guide): customer runbook for scheduled --property-graph Context Graph deploy

caohy1988:
A **customer-followable runbook** that takes the codelab (rename-free) Context Graph from a local run to a **scheduled Cloud Run + Cloud Scheduler deploy** using the one-artifact `--property-graph` flow — the "live customer-style example" of the production path.

Grounded in the exact flow a live smoke validated: `run_job.py` in `--property-graph` mode against scratch split datasets materialized 5/5 sessions, graph tables + `_bqaa_materialization_state` in the graph dataset, events dataset read-only.

## `docs/guides/scheduled-context-graph-deploy.md`

End-to-end, with copy-paste commands:
1. **Prereqs** — APIs, the read-only-events / writable-graph two-dataset shape, seed events.
2. **Two artifacts** — placeholdered `property_graph.sql` + sibling `table_ddl.sql` (the codelab ships both).
3. **Local dry-run** — run `run_job.py` with the env vars to validate the placeholder + split-dataset wiring **before paying for Cloud Run** (same code path the Job runs).
4. **Build** — `build_image.sh --property-graph`.
5. **Deploy** — `deploy_cloud_run_job.sh --property-graph … --smoke` (Option A) or Terraform `property_graph = true` (Option B).
6. **Verify** — graph tables + state table in the graph dataset; events dataset only `agent_events`; the JSON log's `mode = "property-graph"`.
7. **When to use explicit `--ontology`/`--binding`** — descriptions, inheritance, derived props, renames, compiled extractor → migration-v5 deep reference.

## Discoverability

- Linked from the codelab's "run this every six hours" aside (GitHub URL, claat-safe).
- The Colab notebook is regenerated (`--check` passes); `claat export` builds clean; 10 doc tests pass.

Docs only; no code change. Complements the migration-v5 deploy README (the deep IAM/monitoring/teardown reference for the explicit path). Follows the 0.3.3 release (#295).

### DIFF
--- a/docs/codelabs/periodic_materialization.md
+++ b/docs/codelabs/periodic_materialization.md
@@ -639,7 +639,7 @@ The local run you completed in *Materialize the decision graph* uses default beh
 * **Bound the per-run batch size** (`--max-sessions`). Useful when an upstream event spike threatens to overwhelm a single scan.
 
 > aside positive
-> **From "run this once" to "run this every six hours":** The SDK ships a deploy script and a Terraform module that wrap `bqaa context-graph` as a Cloud Run Job triggered by Cloud Scheduler, with least-privilege service accounts and the IAM grants the job needs. See the [periodic-materialization deployment guide](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/tree/main/examples/migration_v5/periodic_materialization) for the worked example, the IAM matrix, and the recommended schedules.
+> **From "run this once" to "run this every six hours":** The SDK ships a deploy script and a Terraform module that wrap `bqaa context-graph` as a Cloud Run Job triggered by Cloud Scheduler, with least-privilege service accounts and the IAM grants the job needs. Follow the [scheduled Context Graph deploy runbook](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/docs/guides/scheduled-context-graph-deploy.md) to take *this* `--property-graph` graph to a scheduled deploy with the same one-artifact flow, or the [periodic-materialization deployment guide](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/tree/main/examples/migration_v5/periodic_materialization) for the full IAM matrix and the explicit-ontology/binding path.
 
 ## Clean up
 
--- a/docs/guides/scheduled-context-graph-deploy.md
+++ b/docs/guides/scheduled-context-graph-deploy.md
@@ -101,21 +101,12 @@ Expect a JSON report with `"mode": "property-graph"` and
 `"sessions_materialized" > 0`. This is the cheapest way to confirm the
 placeholder + split-dataset wiring before deploying.
 
-## 3. Build the image
+## 3. Deploy on a schedule
 
-The image build is outside Terraform; the helper stages your artifacts + the SDK
-source and runs Cloud Build:
-
-```bash
-IMAGE_URI="$(./examples/migration_v5/periodic_materialization/build_image.sh \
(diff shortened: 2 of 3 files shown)


caohy1988:
Runbook review at 1882044.

[P2] Make the image-build step conditional on the Terraform path; the bash deploy path builds its own image.

The runbook currently has `## 3. Build the image` before both deploy options, then Option A runs `deploy_cloud_run_job.sh --property-graph ... --smoke`. That bash script does its own Cloud Build from a staging directory via the deploy flow; it never consumes the `IMAGE_URI` produced by `build_image.sh`. So a reader following Option A will pay for/build an Artifact Registry image that is unused, then immediately pay for/build a second image through the bash deploy.

Please restructure the deploy section so the flows are distinct:

- Option A — bash: run `deploy_cloud_run_job.sh --property-graph ... --smoke` directly. No prior `build_image.sh` step.
- Option B — Terraform: first run `build_image.sh --property-graph path/to/property_graph.sql`, then set `image_uri` and `property_graph = true` in `terraform.tfvars`, then `terraform apply`.

This is only a docs-flow issue; the underlying commands are correct. The rest of the runbook is strong: it explains the split read-only-events / writable-graph shape, placeholder contract, local runtime dry-run, verify queries, and when to fall back to explicit ontology/binding. CI is fully green.

Answer on the live Cloud Run demo question: I would not do a persistent scheduled deploy by default. The run_job.py split-dataset property-graph smoke already validated the runtime bug-prone path, and a Cloud Run + Scheduler + Artifact Registry demo creates billable infrastructure. If you want a one-off demonstration, do it only after this runbook flow is fixed, with `--smoke`, immediate teardown, and a short evidence table; otherwise skip it.

caohy1988:
Good catch — fixed.

You're right: the bash deploy (`deploy_cloud_run_job.sh`) builds + publishes its own image inline via Cloud Buildpacks and never consumes a prebuilt `IMAGE_URI`, so running `build_image.sh` before Option A was a wasted build. Split the flows:

- **Option A (bash):** run `deploy_cloud_run_job.sh --property-graph … --smoke` **directly** — it builds its own image. No `build_image.sh` step.
- **Option B (Terraform):** build the image first with `build_image.sh --property-graph` (Terraform takes a published image), then `terraform apply -var image_uri=$IMAGE_URI`.

Dropped the standalone 'Build the image' section, folded it into Option B, wired `$IMAGE_URI` into the `terraform apply`, and renumbered Verify. Docs-only; notebook unaffected.

caohy1988:
Fresh re-review at bc8dd38.

Findings: none.

The P2 is fixed. The runbook now separates the deploy paths correctly:

- Bash path: `deploy_cloud_run_job.sh --property-graph ... --smoke` runs directly and builds/publishes its own image inline. No wasted prebuild.
- Terraform path: `build_image.sh --property-graph ...` is scoped only to Terraform, then `terraform apply -var "image_uri=$IMAGE_URI"` consumes that image.
- Verify is renumbered and still checks the right outcomes: graph tables + `_bqaa_materialization_state` in the graph dataset, events dataset read-only.
- The local runtime dry-run, placeholder contract, split-dataset explanation, and explicit ontology/binding fallback remain accurate.

CI is fully green and the branch is CLEAN. Ready to merge.

Answer on the live e2e scope: if you still want a real-agent run after this merges, choose the lower-risk Option 1 first: real ADK agent + plugin → BigQuery events → explicit MAKO materialization → GQL query, then clean up. Skip the Cloud Run one-off unless you specifically need production-infra evidence; it adds billable AR/Cloud Run/Scheduler resources and the deploy path has already been smoke-validated.

PR #298, opened 2026-06-08
feat(materialize): let run_materialize_window pick the AI.GENERATE model

caohy1988:
Let callers and operators choose the AI.GENERATE model for context-graph materialization, across all three surfaces (Python API, CLI, scheduled deploy).

## Why

The `AI.GENERATE` extraction fallback was pinned to the `gemini-2.5-flash` default with no way to select the model. `_resolve_endpoint` already builds a `locations/global` publisher URL, so Gemini 3.x names (e.g. `gemini-3.5-flash`) resolve correctly — the only missing piece was a parameter to set it, plumbed through every entry point operators actually use.

## What

**Commit 1 — Python API:**
- `run_materialize_window(..., endpoint="gemini-2.5-flash")` — new keyword, default unchanged.
- Forwarded to `_build_manager`, which passes it to both manager factories (`from_ontology_binding`, `from_bundles_root`).

**Commit 2 — operator paths (addresses review P2):**
- `bqaa context-graph --endpoint` (default `gemini-2.5-flash`) → forwarded to `run_materialize_window`.
- `run_job.py` reads `BQAA_ENDPOINT` (default `gemini-2.5-flash`) → forwarded through `common_kwargs`; documented in the env header and the startup log line.

**No behavior change for existing callers** — omitting the param / env / flag keeps `gemini-2.5-flash`. Short names resolve to `locations/global`, so `gemini-3.5-flash` works on all three paths.

## Tests

- `tests/test_materialize_window.py::TestEndpointThreading` — endpoint reaches `from_ontology_binding` through the **real** `_build_manager` (not mocked) + unchanged default.
- `tests/test_cli_context_graph_input_modes.py` — `--endpoint` forwarding + default.
- `tests/test_run_job_property_graph.py` — `BQAA_ENDPOINT` forwarding + default.

183 passed across the touched suites.

## Validation (live, real ADK agent + plugin)

Two end-to-end runs on a live project, agent and `AI.GENERATE` both on `gemini-3.5-flash` (Vertex `global` location):

- **Codelab decision graph (DDL-only / `--property-graph`):** `ok=true`, 4/4 sessions; GQL traversal returned real request→option→outcome traces.
- **Explicit MAKO graph (`ontology.yaml` + `binding.yaml` + reference extractor):** `ok=true`, 3/3 sessions, full 11-entity / 12-relationship hub (89 rows). GQL reward loop returned real `reward_value` floats (0.90 / 0.88 / 0.85) from the agent's `compute_reward`.

The Vertex 404 some users may hit on `gemini-3.5-flash` is a location issue, not access: Gemini 3.x lives in the `global` location, which `_resolve_endpoint` already targets.

### DIFF
--- a/examples/migration_v5/periodic_materialization/run_job.py
+++ b/examples/migration_v5/periodic_materialization/run_job.py
@@ -88,6 +88,11 @@ Env vars (all set by the deploy script via
   gaps). ``compiled-only`` skips ``AI.GENERATE`` entirely; any
   span the compiled extractors don't cover surfaces as a typed
   ``empty_extraction`` failure with sample diagnostics.
+* ``BQAA_ENDPOINT`` (default ``gemini-2.5-flash``) — Vertex AI
+  model for the ``ai-fallback`` AI.GENERATE call. Short names
+  resolve to a ``locations/global`` publisher URL, so Gemini 3.x
+  names (e.g. ``gemini-3.5-flash``) work. Ignored under
+  ``compiled-only`` (no AI call is made).
 * ``BQAA_REFERENCE_EXTRACTORS_MODULE`` (default unset, set to
   ``reference_extractor`` by the deploy script in
   compiled-only mode) — dotted module path whose ``EXTRACTORS``
@@ -458,6 +463,11 @@ def main() -> int:
   extraction_mode = (
       os.environ.get("BQAA_EXTRACTION_MODE") or "ai-fallback"
   ).strip()
+  # AI.GENERATE model for the ai-fallback path. Short names resolve to a
+  # locations/global publisher URL inside the SDK, so Gemini 3.x names
+  # (e.g. ``gemini-3.5-flash``) work. Default matches the SDK default so
+  # an unset env var is a no-op. Ignored under ``compiled-only`` (no AI call).
+  endpoint = (os.environ.get("BQAA_ENDPOINT") or "gemini-2.5-flash").strip()
   # Reference module + bundles root. Both default to None;
   # ``compiled-only`` mode requires at least one of them to be set
   # at the SDK boundary (else the manager has no structured
@@ -507,6 +517,7 @@ def main() -> int:
       to_time=to_time_raw,
       state_key_suffix=state_key_suffix,
       extraction_mode=extraction_mode,
+      endpoint=endpoint,
       bundles_root=bundles_root,
       reference_extractors_module=reference_extractors_module,
       max_session_age_hours=max_session_age_hours,
@@ -581,6 +592,7 @@ def main() -> int:
         state_key_suffix=state_key_suffix,
         extraction_mode=extraction_mode,
         max_session_age_hours=max_session_age_hours,
+        endpoint=endpoint,
     )
 
     if property_graph_name:
--- a/src/bigquery_agent_analytics/cli.py
+++ b/src/bigquery_agent_analytics/cli.py
@@ -1996,6 +1996,18 @@ def materialize_window(
             "mode."
         ),
     ),
+    endpoint: str = typer.Option(
+        "gemini-2.5-flash",
+        "--endpoint",
+        help=(
+            "Vertex AI model for the AI.GENERATE extraction "
+            "fallback (used in 'ai-fallback' mode). Short names "
+            "resolve to a locations/global publisher URL, so "
+            "Gemini 3.x models such as 'gemini-3.5-flash' work "
+            "here. Ignored under --extraction-mode=compiled-only "
+            "(no AI call is made)."
+        ),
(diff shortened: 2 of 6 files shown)


caohy1988:
Independent review/investigation at 461277b.

[P2] Endpoint selection is still unreachable from the CLI and scheduled runtime paths.

The code change correctly adds `endpoint` to the Python API (`run_materialize_window`) and threads it into `_build_manager` → `OntologyGraphManager.from_ontology_binding` / `.from_bundles_root`; the test captures that factory call and preserves the old default. That proves the Python API path is fixed.

But the user-facing materialization surfaces still cannot use it:

- `bqaa context-graph` has no `--endpoint` option and does not pass `endpoint=` into `run_materialize_window` (`cli.py` around the context-graph command still calls the function without the new kwarg).
- `examples/migration_v5/periodic_materialization/run_job.py` has no `BQAA_ENDPOINT` / `BQAA_AI_ENDPOINT` env var and both runtime branches call `run_materialize_window` without `endpoint=`.

So after this PR, a Python caller can select `gemini-3.5-flash`, but a customer using the documented CLI or scheduled deploy path remains pinned to the default `gemini-2.5-flash`. If the intent is only to expose a Python SDK knob, the code is internally consistent. If the intent is to fix the real public materialization path that operators use (`bqaa context-graph` / Cloud Run job), this PR is incomplete.

Recommended fix: add `--endpoint` to `bqaa context-graph`, pass it through to `run_materialize_window`, and add a small CLI param-inspection/forwarding test. For deploy parity, add an optional `BQAA_ENDPOINT` env var in `run_job.py` and pass it through; the deploy/Terraform env wiring can be a follow-up if you want to keep this PR scoped.

Root-cause conclusion independent of the previous description: the manager already supports custom endpoints and resolves short model names to `locations/global`; the materializer did not expose that knob. #298 fixes that only for direct Python calls.

caohy1988:
Agreed — the intended scope is the public operator path, not just the Python API. Pushed a follow-up commit closing the gap you identified:

- **`bqaa context-graph`** now has `--endpoint` (default `gemini-2.5-flash`), forwarded to `run_materialize_window`.
- **`run_job.py`** now reads `BQAA_ENDPOINT` (default `gemini-2.5-flash`), forwarded through `common_kwargs`; documented in the env header and the startup log line.

Both default to the SDK default, so unset = no behavior change for existing CLI/cron users. Short names still resolve to the `locations/global` publisher URL, so `gemini-3.5-flash` works on both paths.

**Tests added:**
- `test_cli_context_graph_input_modes.py`: `--endpoint` forwarding + default.
- `test_run_job_property_graph.py`: `BQAA_ENDPOINT` forwarding + default.

183 passed across the touched suites (`test_cli_context_graph_input_modes`, `test_run_job_property_graph`, `test_materialize_window`, `test_cli_bqaa_app`).

**Live validation:** ran the explicit MAKO graph end to end with the new knob — real ADK agent on `gemini-3.5-flash` (Vertex `global`) → `agent_events` → `run_materialize_window(endpoint="gemini-3.5-flash")` with the reference extractor (`TOOL_COMPLETED`) + AI.GENERATE fallback for the rest: `ok=true`, 3/3 sessions, full 11-entity / 12-relationship hub (89 rows). GQL reward-loop traversal returned real `reward_value` floats (0.90 / 0.88 / 0.85) from the agent's `compute_reward`.

caohy1988:
Fresh re-review at 4339822.

Findings: none on the code.

The earlier P2 is resolved across the real operator surfaces:
- Python API: `run_materialize_window(endpoint=...)` reaches `_build_manager` and both manager factories; default stays `gemini-2.5-flash`.
- CLI: `bqaa context-graph --endpoint ...` forwards to `run_materialize_window`; tests pin explicit forwarding and the default.
- Scheduled runtime: `run_job.py` reads `BQAA_ENDPOINT`, includes it in `common_kwargs`, and logs/documents it; tests pin explicit env forwarding and unset-default behavior.

CI is fully green now: Format, Build, Python 3.10-3.14, scan, check-changes, and CLA all pass. I attempted the touched local suites too, but this workspace hits the known stale `.venv` package-metadata issue during collection before any PR code runs, so CI is the authoritative signal here.

Non-blocking hygiene: the PR body still mostly describes the first commit/Python API path and its older `test_materialize_window.py`-only validation. If you want the merge record to be accurate, refresh the body to mention the CLI + `BQAA_ENDPOINT` follow-up and the full green CI. Not a code blocker.

Merge recommendation: ready after handling the branch state. GitHub reports `mergeStateStatus=BEHIND`; if the repo requires up-to-date branches, sync with main and let checks rerun, then merge. If that requirement is not enforced, this is good to merge now.

PR #301, opened 2026-06-08
Add demo and implementation: an agent that writes its own skill

evekhm:
## What

A runnable, self-contained example — **`examples/skill_evolution_lab/`** — of an agent that **rewrites its own versioned `SKILL.md` from its conversation traces, with no teacher model**, plus the reusable engine that powers it (**`scripts/skill_evolution.py`**).

A company-policy Q&A agent ships with a deliberately **flawed V0 skill** ("answer only from the baked summary, else contact HR") that suppresses a tool which already knows every answer. The engine reads the scored traces — successes *and* failures — and evolves a tool-first **V1**. The model, tool, and questions are identical across V0/V1, so any quality delta is attributable to the skill.

## Contents

- **`scripts/skill_evolution.py`** — the evolution engine. Partitions scored conversations into successes/failures, runs a **parallel analyst fleet** (one analyst per trajectory, each on a frozen copy of the skill), and **consolidates recurring rules** into a new version (best-of-N, size compaction, anti-parroting reclassification). Standalone — consumes a scored report dict and returns skill text, so it composes with the scorer without importing it.
- **`examples/skill_evolution_lab/`** — the agent (`agent/`), a golden-Q&A eval spec (`eval/eval_spec.json`, 50 pairs), evolve + disjoint held-out question sets (~50 each, plus multi-turn anti-parroting cases), a one-command `run_e2e_demo.sh`, a committed `sample_run/`, `README.md`, and `VERIFICATION.md`. (The narrative write-up lives in the companion Medium post, not in this repo.)
- **`tests/test_skill_evolution.py`** — 19 engine unit tests.
- `examples/README.md` — a "Skill Evolution Lab" section; `scripts/README.md` — an engine section.

## Models (all GA except the internal analyst)

- **Agent under test:** `gemini-3.5-flash` (default); also swept on `gemini-3.1-flash-lite` and `gemini-2.5-pro` (all GA), plus `gemini-3.1-pro-preview`.
- **Evolution analyst/consolidator:** `gemini-3.1-pro-preview` (internal to the engine; on this task it produces cleaner, tool-first skills than the GA pro model, which tended to bake facts).
- **Judge:** `gemini-2.5-flash` (GA). Gemini 3.x and `gemini-2.5-pro` route to the Vertex `global` endpoint automatically.

## Verified (55-question held-out set, golden-grounded; see `VERIFICATION.md`)

Four agent models × 3 seeds. V1 correctness as **mean [range]**:

| Model | V0 → V1 correctness | V1 grounding |
|---|---|---|
| `gemini-3.5-flash` | 17% → **99% [98–100]** | 91% |
| `gemini-3.1-flash-lite` | 16% → **90% [71–100]** | 74% |
| `gemini-2.5-pro` | 53% → **95% [93–96]** | 82% |
| `gemini-3.1-pro-preview` | 19% → **99% [96–100]** | 84% |

Recorded default run (`gemini-3.5-flash`): overall **16.4% → 98.2%**, anti-parroting corrections **0% → 100%**, evolved skill ~2.6KB — tool-first, with a learned "when corrected, re-verify with a tool, don't just agree" rule. Every number is graded against the golden Q&A answer key (not a no-ground-truth "usefulness" estimate).

Built on the golden-Q&A / `--eval-spec` / `--tag-turns` scoring from #174 (merged).


caohy1988 (commented):
Thorough, well-documented PR — the engine is cleanly decoupled from the agent/registry (genuinely imported, not copied), the verification doc honestly reports run-to-run variance instead of cherry-picked bests, the arXiv citations check out, and there's real unit coverage. Close to mergeable. A few findings below, in priority order; full context in each inline comment.

**High (worth blocking on):** `format_trajectory` reads trace keys the report never emits, silently dropping the parrot/recover evidence the README sells.

**Medium:** sanitize/validate ordering discards candidates that sanitize would save; the e2e demo conflates the judge region with the Skill Registry region.

**Low/nits:** registry token re-fetched on every LRO poll; engine vs. demo default-model mismatch; `print_rate.py` leaked file handle; the six committed sample-run JSON files are most of the diff size — consider trimming to the reports + RESULT.

Things I verified are fine: arXiv 2603.25158 / 2603.01145 are real with matching titles; the 0.92 golden threshold in VERIFICATION.md matches `_DEFAULT_GOLDEN_THRESHOLD`; `partition_trajectories` correctly reads `sub_trajectories` (the bug is only in `format_trajectory`); `quality_scores`/`metrics` keys do match; license headers, `.gitignore`, V0 auto-restore trap, and CLI/env precedence are all correct.

scripts/skill_evolution.py
```diff
@@ -0,0 +1,762 @@
...
+        "scope_compliance",
+        "first_time_right",
+    ):
+      score_data = quality.get(dim, {})
+      if score_data:
+        result += (
+            f"{dim}: {score_data.get('score', '?')}/2 -"
+            f" {score_data.get('reason', '')}\n"
+        )
+    for seg in session.get("execution_sub_trajectories", []) or []:
```
caohy1988: **High — reads a key the report never emits (confidence 9/10).** This loop reads `execution_sub_trajectories`, but `quality_report.py` and the committed `sample_run/v0_evolve_report.json` only emit `sub_trajectories`. The real session key union is `answered_by, conversation, corrections, golden_eval, is_a2a, latency_s, metrics, quality_scores, question, response, session_id, sub_trajectories, tool_calls, user_turns, verifications` — no `execution_sub_trajectories` and no `execution_trace`.

So this block (and the `execution_trace` block just below at L269-270) never fires on a real report. Evolution still works because `conversation`/`corrections`/`quality_scores` *do* match, but the **per-segment parrot/recover outcome labels never reach the Error Analyst** — exactly the PARROTING evidence its prompt is told to use. Note `_has_parroted_recovery` (L193) correctly checks `sub_trajectories`, so the inconsistency is within this file.

Also: real `sub_trajectories` entries have keys `label, outcome, start_turn, end_turn` and **no `trace` field**, so `seg.get('trace')` would be empty even after a rename.

**Fix:** read `sub_trajectories`; render `outcome`/`label`/`start_turn`-`end_turn` instead of `seg.get('trace')`; drop or rewire the `execution_trace` block. Add a unit test feeding a real `sub_trajectories` entry through `format_trajectory` and asserting the outcome label appears — the current tests don't cover this path, which is why it slipped.

scripts/skill_evolution.py
```diff
@@ -0,0 +1,762 @@
...
+        )
+        for _ in range(candidates)
+    ]
+    for fut in as_completed(futures):
+      try:
+        cands.append(fut.result())
+      except Exception as e:  # noqa: BLE001
+        logger.warning("Candidate consolidation failed: %s", e)
+
+  viable = [
```
caohy1988: **Medium — candidates that sanitize would save are discarded (confidence 8/10).** `validate_evolved_skill` rejects any candidate with an unescaped `{identifier}` (the ADK context-variable collision check at L406), but you call it on `c` *before* `sanitize_adk_vars`. So a candidate whose only flaw is a stray `{requested_topic}` picked up from a tool error — the exact case `sanitize_adk_vars` exists for — is thrown out instead of cleaned and kept. The `sanitize_adk_vars(c)` here only ever runs on candidates that already passed, making it a no-op in this path.

**Fix:** sanitize first, then validate the sanitized text:
```python
viable = []
for c in cands:
    if not c or c == current_skill:
        continue
    c = sanitize_adk_vars(c)
    if not validate_evolved_skill(c, current_skill):
        viable.append(c)
```

examples/skill_evolution_lab/run_e2e_demo.sh
```diff
@@ -0,0 +1,125 @@
...
+run_agent "$SKILL" "$RUN/v0_test_traffic.json" "$TEST" "$CORR_HO"
+score    "$RUN/v0_test_traffic.json" "$RUN/v0_test_report.json"
+echo "     V0 evolve: $(rate "$RUN/v0_evolve_report.json")"
+echo "     V0 test:   $(rate "$RUN/v0_test_report.json")"
+
+# --- Evolve: extract a tool-first V1 skill from the V0 failures ---
+echo "[evolve] analyst=$ANALYST_MODEL (this is the slow step) ..."
+REG_ARGS=()
+if [ "${WITH_REGISTRY:-0}" = "1" ]; then
+  REG_ARGS=(--registry-update --skill-id "${SKILL_ID:?set SKILL_ID with WITH_REGISTRY=1}" --location "$JUDGE_LOCATION")
```
caohy1988: **Medium — Skill Registry region conflated with judge region (confidence 7/10).** Registry ops are passed `--location "$JUDGE_LOCATION"`, but the judge region and the Skill Registry region are unrelated, and the registry only supports `us-central1`, `europe-west4`, `us-east5` (per your own `skill_registry.py` docstring). Default `us-central1` happens to work, but the moment someone sets `JUDGE_LOCATION` to a judge-only region (e.g. `us-east4`), `WITH_REGISTRY=1` breaks with a confusing error. `setup.sh` already uses `REGION` for the registry — the demo (and `reset.sh`) should too. **Fix:** use `REGION` or a dedicated `REGISTRY_LOCATION` for registry ops; keep `JUDGE_LOCATION` for scoring only.

examples/skill_evolution_lab/agent/skill_registry.py
```diff
@@ -0,0 +1,186 @@
...
+  def __init__(self, project: str, location: str = DEFAULT_LOCATION):
+    self.project = project
+    self.location = location
+    self.base = (
+        f"{_API.format(loc=location)}/projects/{project}/locations/{location}"
+    )
+
+  # -- auth ---------------------------------------------------------------
+
+  def _token(self) -> str:
```
caohy1988, line 67: **Low — token re-fetched on every request, including each poll (confidence 8/10).** `_headers()` calls `_token()` per request, and `_wait()` polls an LRO every 3s for up to 300s — so a multi-minute create/update spawns dozens of `gcloud ... print-access-token` subprocesses for a token valid ~1 hour. Not a correctness issue, just wasteful and slow. Cache the token on the instance with a short TTL.

examples/skill_evolution_lab/print_rate.py
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Print the golden-matched meaningful rate from a quality report (one line)."""
+
+import json
+import sys
+
+g = json.load(open(sys.argv[1]))["summary"].get("golden_eval_summary", {})
```
caohy1988: **Nit.** `json.load(open(sys.argv[1]))` leaks the file handle — use a `with` block to match the house style used everywhere else in this PR.

examples/skill_evolution_lab/run_e2e_demo.sh
```diff
@@ -0,0 +1,125 @@
...
+CORR_HO=eval/questions_corrections_heldout.json
+
+TS="$(date +%Y%m%d_%H%M%S)"
+RUN="runs/${TS}_${AGENT_MODEL//[^a-zA-Z0-9]/_}"
+mkdir -p "$RUN"
+echo "== Skill lab E2E  agent=$AGENT_MODEL  analyst=$ANALYST_MODEL  judge=$JUDGE_MODEL"
+echo "   run=$RUN   log=$RUN/run.log"
+
+# Always leave the working copy on V0 (repo sits at V0).
+restore() { cp "$V0" "$SKILL"; }
```
caohy1988, line 75: **P2 — registry state is not restored when `WITH_REGISTRY=1` (confidence 8/10).** `restore()` only copies `SKILL.v0.md` back to the local working copy, but this script can update the remote Skill Registry to V1 later via `--registry-update`. That leaves users with local V0 and remote V1, which conflicts with the script/header/docs saying the demo restores V0 after compare.

Fix: either make the EXIT trap call the registry reset/update path when `WITH_REGISTRY=1`, or narrow the docs/comments to say only the local working copy is auto-restored and the registry requires `./reset.sh`.

caohy1988 (commented):
Fresh follow-up review: posting only new findings that were not already mentioned in the existing review comments.

examples/skill_evolution_lab/run_e2e_demo.sh
```diff
@@ -0,0 +1,125 @@
...
+trap restore EXIT
+
+run_agent() {  # run_agent <skill> <out> <qfile...>
+  local skill="$1" out="$2"; shift 2
+  local qargs=(); local q
+  for q in "$@"; do qargs+=(--questions "$q"); done
+  uv run python run_agent.py --skill "$skill" "${qargs[@]}" \
+    --model "$AGENT_MODEL" --concurrency 6 -o "$out" >>"$RUN/run.log" 2>&1
```
caohy1988: **P3 — `CONCURRENCY` does not throttle agent traffic generation (confidence 8/10).** The script advertises `CONCURRENCY` as an env override and resolves it into `CONC`, but `run_agent()` hard-codes `--concurrency 6`. So `CONCURRENCY=1 ./run_e2e_demo.sh` only throttles judging, not the V0/V1 traffic-generation phases where users can also hit Gemini quota/429s.

Fix: pass `--concurrency "$CONC"` here too, or add a separate `AGENT_CONCURRENCY` if judge and agent concurrency should be independent.

examples/skill_evolution_lab/sample_run/RESULT.md
```diff
@@ -0,0 +1,12 @@
...
+Golden-grounded correctness (matched & meaningful) on the held-out set.
+
+| Metric | V0 (flawed) | V1 (evolved) | Delta |
+| --- | --- | --- | --- |
+| Overall | 16.4% (9/55) | 98.2% (54/55) | +81.8pp |
+| Single-turn | 18.0% (9/50) | 98.0% (49/50) | +80.0pp |
+| Corrections (anti-parrot) | 0.0% (0/5) | 100.0% (5/5) | +100.0pp |
+
+Parroted sub-trajectories: V0=0  V1=0 (lower is better -- the agent re-verified instead of caving).
+
```
caohy1988: **P3 — `git diff --check` fails on this committed sample artifact (confidence 9/10).** This file has a new blank line at EOF, and `examples/skill_evolution_lab/sample_run/v1_skill.md:30` also has trailing whitespace. CI format passed, but `git diff --check upstream/main...HEAD` still fails.

Fix: remove the extra EOF blank line here and strip the trailing space in `v1_skill.md`.

scripts/skill_evolution.py
evekhm: Fixed in d61f957. `format_trajectory` now reads `sub_trajectories` and renders `outcome`/`label`/turn-span (no `trace` field); the dead `execution_trace` block is removed. Also made `_has_parroted_recovery` read only `sub_trajectories` (dropped the `execution_sub_trajectories` fallback) and fixed the partition test's key, so the file is now fully consistent on `sub_trajectories`. Added `test_format_renders_subtrajectory_outcomes` asserting the parroted outcome + `[~]` icon reach the analyst text.

scripts/skill_evolution.py
evekhm: Fixed in d61f957. Reordered exactly as suggested — `sanitize_adk_vars` runs first, then `validate_evolved_skill` on the sanitized text, so a candidate whose only flaw is an unescaped `{context_var}` is cleaned and kept instead of discarded.

examples/skill_evolution_lab/run_e2e_demo.sh
evekhm: Fixed in d61f957. Added a dedicated `REGISTRY_LOCATION` (defaults to `REGION` from .env, else us-central1) used for all registry ops in both `run_e2e_demo.sh` and `reset.sh`; `JUDGE_LOCATION` is now scoring-only.

examples/skill_evolution_lab/agent/skill_registry.py
evekhm: Fixed in d61f957. `SkillRegistry` now caches the access token on the instance with a 50-min TTL, so a multi-minute LRO poll reuses one token instead of spawning a `gcloud print-access-token` subprocess every 3s. (`SKILL_REGISTRY_TOKEN` env still short-circuits.)

examples/skill_evolution_lab/print_rate.py
evekhm: Fixed in d61f957 — now uses a `with open(...)` block, matching the house style.

examples/skill_evolution_lab/run_e2e_demo.sh
evekhm: Addressed in d61f957 by narrowing the docs (your second option): the EXIT trap comment, the header step list, and the file header now state it restores only the **local** working copy and that the Skill Registry is reverted by `./reset.sh` (WITH_REGISTRY=1). Chose docs over auto-reverting in the trap because a registry LRO in an EXIT handler can hang/fail.

examples/skill_evolution_lab/run_e2e_demo.sh
evekhm: Fixed in d61f957. `run_agent()` now passes `--concurrency "$CONC"`, so `CONCURRENCY` throttles the V0/V1 traffic-generation phases as well as judging.

examples/skill_evolution_lab/sample_run/RESULT.md
evekhm: Fixed in d61f957. Stripped the trailing EOF blank line in `RESULT.md` and the trailing whitespace in `v1_skill.md`, and fixed the generators so it stays clean: `compare_runs.py` writes `out.rstrip() + "\n"`, and `analyze_and_evolve.py` rstrips each skill line before writing. `git diff --check` is now clean.

evekhm:
Thanks for the thorough review — all findings addressed in d61f957 (replies on each thread). Summary:

**High** — `format_trajectory` read `execution_sub_trajectories`/`execution_trace` (never emitted), so parrot/recover labels never reached the Error Analyst. Now reads `sub_trajectories` (outcome/label/turn-span); added a unit test; and removed the matching dead `execution_sub_trajectories` fallback in `_has_parroted_recovery` so the file is consistent.

**Medium** — sanitize-then-validate reorder; Skill Registry region decoupled from judge region (`REGISTRY_LOCATION`) in `run_e2e_demo.sh` + `reset.sh`.

**P2/P3/Low** — trap-restore scope documented (local only; registry via `./reset.sh`); `CONCURRENCY` now throttles agent traffic; registry token cached (50-min TTL); `print_rate.py` uses `with`; trailing-whitespace fixed in the sample artifacts **and** their generators (`compare_runs.py`, `analyze_and_evolve.py`) so `git diff --check` stays clean.

Two summary items handled as judgment calls:

- **Engine vs. demo default-model mismatch** — intentional, now documented in `analyze_and_evolve.py`'s `--model` help: the demo pins the analyst to `gemini-3.1-pro-preview` rather than the engine's generic `gemini-2.5-pro` default because, on this task, `gemini-2.5-pro` baked specific (wrong) figures into the skill, while 3.1-pro-preview produces clean tool-first skills (a 4×3-seed sweep in VERIFICATION.md shows this).
- **Committed sample-run JSON size** — keeping these intentionally: a referenceable, recorded end-to-end run (scored reports **and** raw traffic) is a deliberate feature of the example, documented in `sample_run/README.md` and cited by `VERIFICATION.md` (the before/after and trajectory excerpts are drawn from these exact files). Happy to trim to reports + RESULT if you'd prefer, but they're the artifact the example is built to ship.

Verification: 20 unit tests pass, pyink/isort clean, `bash -n` clean, `git diff --check` clean.

scripts/skill_evolution.py
```diff
@@ -0,0 +1,770 @@
...
+    return selected
+
+  incumbent = score_fn(current_skill)
+  best, best_score = None, float("-inf")
+  for cand in viable:
+    score = score_fn(cand)
+    logger.info("Candidate scored %.3f (incumbent %.3f).", score, incumbent)
+    if score > best_score:
+      best, best_score = cand, score
+  if best_score < incumbent + min_improvement:
```
caohy1988: **Suggestion — add a negative control that exercises this gate.** This `min_improvement` incumbent gate (and the "no viable candidate → keep base" path above) is exactly the right safety property for a self-modifying system: it should leave an already-good skill alone. But nothing in the demo or the tests ever exercises it — every shown run is a big improvement from the crippled V0, so "the engine knows when **not** to change" is asserted by code but never demonstrated.

A cheap, high-trust addition: feed the evolved V1 back in (or any already-tool-first skill) and assert the engine returns it unchanged. Even a unit test with a stubbed `score_fn` where the best candidate doesn't clear `incumbent + min_improvement`, asserting `evolve_skill` returns `current_skill`, would lock in the no-op behavior and document it. For a "self-evolving" demo, *restraint* is as important to show as improvement.

examples/skill_evolution_lab/VERIFICATION.md
```diff
@@ -0,0 +1,123 @@
...
+The agent model, tools, and questions are identical for V0 and V1 — **only the
+skill file changes** — so the delta is attributable to the skill.
+
+## Result (recorded run, gemini-3.5-flash, held-out)
+
+| Metric | V0 (flawed) | V1 (evolved) | Delta |
+| --- | --- | --- | --- |
+| Overall | 16.4% (9/55) | 98.2% (54/55) | +81.8pp |
+| Single-turn | 18.0% (9/50) | 98.0% (49/50) | +80.0pp |
+| Corrections (anti-parrot) | 0.0% (0/5) | 100.0% (5/5) | +100.0pp |
```
caohy1988, line 41: **Honesty — this row is n=5.** The 0% → 100% anti-parroting result is 5 multi-turn cases. That's an illustrative anecdote, not a statistic, and 100.0% reads as stronger evidence than 5 cases can carry. This doc is already admirably honest about variance on the main number (you report ranges and even call out the flash-lite unlucky seed) — extend that same honesty here: either expand the corrections set to ~15–20 cases so the rate means something, or explicitly label this row as illustrative (n=5) rather than a measured rate. A one-line caveat is enough; it costs nothing and protects the credibility of the rest of the table.

caohy1988 (commented):
Re-reviewed at d61f957 — all five findings from my earlier review are fixed and verified (the `sub_trajectories` key fix with a new unit test, sanitize-then-validate reordering, `REGISTRY_LOCATION` split, token TTL cache, `print_rate.py` `with` block). I ran the suite in a clean worktree: 20 passed (was 19), and CI's `pytest -q` over `tests/` will pick up the new test. The eval files back every doc claim (50+50+5+5 = 55 held-out, 50 golden Q&A, `corr_` ids make the correction split correct). No code-level issues remain — this is mergeable as engineering.

The remaining feedback is about **claims vs. what's demonstrated**, not code. Three points, inline where they anchor:

**1. The headline delta rests on a deliberately crippled baseline.** The 16% → 98% jump is dominated by the fact that V0 is *told to ignore* a tool that already holds every answer — so the "evolution" mostly learns one rule: stop deflecting to HR, use the tool. The number measures how crippled V0 was, not how strong the loop is. A reader who takes "+82pp from self-evolution" at face value has been over-sold. Two ways to make the framing match the evidence:
  - **(a) Lead with the loop, not the delta.** The genuinely strong, defensible story here is that the SDK closes the quality loop end-to-end: trace → golden-grounded score → evolve → re-score, all attributable because only the skill file changes. Headline *that*. The delta becomes an illustration, not the claim.
  - **(b) Add a fair-baseline arm.** Keep the crippled V0 as the legible teaching case, but add one plausibly-written-but-incomplete V0 and show evolution still helps from a non-sabotaged start. That converts a layup into a result.

  Neither blocks merge — but if the goal is a claim people remember and trust, the current setup proves less than the title implies. This is a judgment call; happy to defer to your read on the narrative.

Detail on points 2 and 3 in the inline comments.

examples/skill_evolution_lab/agent/skill_registry.py
```diff
@@ -0,0 +1,196 @@
...
+        timeout=60,
+    )
+    resp.raise_for_status()
+    return resp.json().get("skillRevisions", [])
+
+  def exists(self, skill_id: str) -> bool:
+    resp = requests.get(
+        f"{self.base}/skills/{skill_id}", headers=self._headers(), timeout=60
+    )
+    return resp.status_code == 200
```
caohy1988: [P3] Treat only 404 as “skill missing” in `exists()`

`exists()` currently returns `resp.status_code == 200`, so 401/403/5xx all collapse into `False`. That is misleading in the CLI paths: `delete` can report “does not exist; nothing to delete” when ADC lacks permission or the API is failing, and `create` falls through to a later POST failure. Please return `False` only for 404 and call `resp.raise_for_status()` for other non-200 responses.

scripts/skill_evolution.py
evekhm: Done in 06db6b0. Extracted the selection/gate logic into a pure `select_candidate()` helper and added a negative-control unit test: with a stub `score_fn` where the best candidate doesn't clear `incumbent + min_improvement`, `select_candidate` returns `current_skill` unchanged (plus empty-viable and clears-margin cases). Restraint is now demonstrated, not just asserted. 24 tests pass.

examples/skill_evolution_lab/VERIFICATION.md
evekhm: Done in 06db6b0. Expanded the held-out anti-parroting set from 5 to **15** cases (disjoint-by-design from the teach set) so the row is a statistic, not an anecdote. A full 4-model x 3-seed re-run on the post-fix engine is in progress; docs will report the n=15 result with the same range/variance honesty as the main number.

examples/skill_evolution_lab/agent/skill_registry.py
evekhm: Done in 06db6b0. `exists()` returns `False` only on 404 and calls `raise_for_status()` for other non-2xx, so 401/403/5xx surface as real errors instead of a misleading 'does not exist'.

haiyuan-eng-google (approved):

Issue #293, opened 2026-06-05
adk-2.0 producer: minimum plan to unblock mid-June customer cutover


caohy1988:
> **Revision history**
> - v5: References-section cleanup — added #200/#201/#202 (the existing emit sub-issues for C4/C5/C6, parallel to how #196/#197/#198 are cited for C1/C2/C3) and inline `→ maps to **#NNN**` pointers on the C4/C5/C6 bullets. The #199 reference was rewritten to make explicit that #293 covers **only the pair-key subset of #199**; the pause registry and `pause_orphan` semantics defer to #206. No contract changes.
> - v4: wording-only corrections from the v3 review — (1) C7 prose no longer claims the future registry will *retroactively* mark mid-June orphans; replaced with an honest "future rows only; historical rows remain pair-key-only unless a separate backfill is designed" framing. (2) v3 revision-history bullet rewritten to stop calling #206 a C7 blocker — the dependency graph already correctly says `NOT blocked by #206`; the bullet now matches that, framing #206 as the deferred home for orphan semantics rather than a blocker. No contract changes.
[...]

Issue #220, opened 2026-05-20
adk-2.0 docs: document attributes.adk.* contract + schema_version in SDK.md


caohy1988:
Parent tracker: #190 (v18 contract).
Wave: 6 — Docs.
Blocked by: Wave 1 (#194, #195), Wave 2 producer-emission issues (#196-#205), **the design sub-issues that may add or shape fields in the `attributes.adk.*` contract: #203 (route/render_ui_widgets/rewind_before_invocation_id capture-or-defer decision) and #208 (oversized AGENT_STATE_CHECKPOINT offload field shapes), and #211 (the existing `TOOL_COMPLETED` typed-view shape decision lands here)**. Docs cannot close until those land or are explicitly deferred.

## Contract

Document the `attributes.adk.*` contract as the supported producer-consumer interface. Include:

- The full field inventory. **Inventory must be re-read against #203, #208, and #211 outcomes before docs close** — those issues can add `attributes.adk.route`, `attributes.adk.render_ui_widgets`, `attributes.adk.rewind_before_invocation_id`, `attributes.adk.agent_state_uri`, `attributes.adk.agent_state_sha256`, and the inline `content.agent_state` stub shape, and can decide whether the existing `TOOL_COMPLETED` typed view exposes `attributes.adk.{function_call_id, pause_kind, pause_orphan}` or routes consumers to `long_running_tool_durations` / base-table JSON. Today's known fields: `schema_version`, `app_name`, `source_event_id`, `node.{path, run_id, parent_path, output_for, message_as_output}`, `branch`, `scope`, `pause_kind`, `pause_orphan`, `function_call_id`, `otel_span_id`.
- Per-field semantics, types, and which rows carry them (every enriched / Event-originating / long-running-path).
- The `attributes.adk.schema_version` deprecation policy.
- Explicit statements that raw ADK `isolation_scope` is internal and not part of the public contract — consumers join on `attributes.adk.scope`.
- The `pause_orphan` drift-quantification contract and how `long_running_tool_durations` surfaces it (healthy pairs exclude orphans; orphan rows surface separately).
- The #211 decision for the existing `TOOL_COMPLETED` typed view: either the new ADK long-running columns are documented as part of the typed view, or the doc says explicitly that consumers must read those fields from `long_running_tool_durations` / base-table JSON.

## Acceptance

[...]

Issue #211, opened 2026-05-20
adk-2.0 consumer: register six new event types across all SDK type surfaces


caohy1988:
Parent tracker: #190 (v19 contract).
Wave: 4 — Consumer foundation.
Design dependency: #207 (workflow node-boundary derivation) — the typed-view column shapes for `WORKFLOW_NODE_STARTING` and `WORKFLOW_NODE_COMPLETED` are blocked until #207 picks OTel-span consumption vs event-observation derivation.

## Contract

This issue covers two SDK-surface changes that must land together so the typed consumer surface matches the producer contract:

### A. Register the six new event types across **every** event-type allowlist / categorization surface

Surfaces to audit and update (concrete list, not a "minimum" — implementers must audit the SDK and add to this list if they find more):

- `src/bigquery_agent_analytics/views.py::_EVENT_VIEW_DEFS` — **typed per-event-type views** for each new type, with extra columns extracting the documented `content` / `attributes` fields. Status per type:
  - `AGENT_TRANSFER` — typed view from `content.{from_agent, to_agent}` (and optional `reason`). **Specifiable now from #200.**
  - `EVENT_COMPACTION` — typed view from `content.{start_timestamp, end_timestamp, compacted_content}`. **Specifiable now from #201.**
  - `AGENT_STATE_CHECKPOINT` — typed view from `content.{end_of_agent, agent_state}` (and offload fields from #208 once they land). **Specifiable now from #202 for the inline case; offload fields wait on #208.**
  - `TOOL_PAUSED` — typed view from `attributes.adk.{function_call_id, pause_kind}`. **Specifiable now from #199.** `pause_orphan` is **not** a `TOOL_PAUSED` field — it lives on the long-running-path `TOOL_COMPLETED` row, addressed by Section B and by `long_running_tool_durations` (#215).
  - `WORKFLOW_NODE_STARTING` / `WORKFLOW_NODE_COMPLETED` — **blocked by #207.** The producer-derivation strategy (OTel span consumption vs deterministic event-observation) determines the payload shape, so typed columns can't be specified yet. Until #207 lands, these two types are either registered with base-header-only views (no extra columns) or this issue's `_EVENT_VIEW_DEFS` work is split so #207's decision can extend `_EVENT_VIEW_DEFS` in a follow-up PR.
- `src/bigquery_agent_analytics/trace.py::EventType` enum.
[...]

Issue #208, opened 2026-05-20
adk-2.0 design: oversized AGENT_STATE_CHECKPOINT offload contract


caohy1988:
Parent tracker: #190 (v15 contract).
Wave: 3 — Design blocker.
Blocks: large-snapshot handling in #202 (AGENT_STATE_CHECKPOINT).

## Problem

`AGENT_STATE_CHECKPOINT` can carry large `agent_state` dicts. The plugin already has GCS offload for content parts via `HybridContentParser`, but that path is content-part/text-shaped, **not directly reusable** for a structured-dict state snapshot.

## Resolve

Define the explicit offload contract:

- (a) **Where the GCS URI lives**: `attributes.adk.agent_state_uri`?
- (b) **Where the SHA-256 lives**: `attributes.adk.agent_state_sha256`?
- (c) **Whether the inline `content.agent_state` is removed or replaced with a stub** (e.g. `{"_offloaded": true, "uri": ..., "sha256": ...}`).
- (d) **Whether to share parser code with `HybridContentParser` or add a dedicated helper.**
- (e) **Size cap env var**: `BQAA_AGENT_STATE_MAX_BYTES` default value.

## Acceptance

- [ ] All five questions answered in this issue.
- [ ] Decisions reflected in #202 (AGENT_STATE_CHECKPOINT) acceptance.
- [ ] Test: a state snapshot above the cap is offloaded; the inline content shape matches the chosen stub format; the SHA and URI are queryable.
- [ ] Test: a state snapshot below the cap remains inline (no GCS write).

## References

- #190 (v15 risk section; v6 offload sub-issue carve-out).
- `src/google/adk/plugins/bigquery_agent_analytics_plugin.py` (`HybridContentParser`).

---

## Selected approach (v4)

This section captures the decision reached in the #297 v3 review thread.

### Strategy

Dedicated structured-state offloader. Do NOT reuse `HybridContentParser` — that helper is content-part / text-shaped and not directly reusable for a structured `agent_state` dict.

### Canonical contract

- **Env var:** `BQAA_AGENT_STATE_MAX_BYTES` (default **524288** / 512 KB).
  - Recommended default keeps inline rows well under BQ Storage Write API's per-row hard limit (10 MiB) and the practical streaming-window soft limit, with envelope headroom.
- **Inline behavior (below threshold):** unchanged from #202. `content.agent_state` holds the dict; no GCS write.
[...]

PR #311, opened 2026-06-15
feat(views): register ADK 2.0 event types across SDK type surfaces (#211)

caohy1988:
Implements the "now" slice of #211 (the v4 selected approach): register the four #293-shipped event types with full typed views and the two workflow-boundary types as header-only, across **every** consumer type surface, and extend `TOOL_COMPLETED` with the long-running pair keys.

Producer state: #293 merged into `google/adk-python` main (`e2676fcb`), so these rows are already being emitted. This PR gives consumers typed access to them. All column SQL was verified against the **exact keys the producer writes** in `origin/main` (not the issue text), so the views won't silently extract nulls.

## Surfaces updated
- `trace.py::EventType` — six new enum members.
- `event_semantics.py` — `AGENT_TRANSFER` / `AGENT_STATE_CHECKPOINT` join the `agent` family, `TOOL_PAUSED` the `tool` family (and `is_tool_event`), plus new `compaction` and `workflow` families. `ALL_KNOWN_EVENT_TYPES` stays complete (derived).
- `trace_evaluator.py::_DEFAULT_EVENT_TYPES` — six new types.
- `udf_kernels.py::_EVENT_LABEL_MAP` — `agent` / `tool` / `compaction` / `workflow` labels.
- `views.py::_EVENT_VIEW_DEFS`:
  - `AGENT_TRANSFER` → `content.{from_agent, to_agent}`
  - `EVENT_COMPACTION` → `content.{start_timestamp, end_timestamp, compacted_content}`, fractional float-epoch preserved via `TIMESTAMP_MICROS(CAST(... AS INT64))`
  - `AGENT_STATE_CHECKPOINT` → `content.{end_of_agent, agent_state}` (inline; offload columns are a #208 follow-up)
  - `TOOL_PAUSED` → `attributes.adk.{function_call_id, pause_kind}` (no `pause_orphan` — that's a `TOOL_COMPLETED` field)
  - `WORKFLOW_NODE_STARTING/COMPLETED` → base-header-only, `TODO(#207)`
  - **`TOOL_COMPLETED` extended** with `attributes.adk.{function_call_id, pause_kind, pause_orphan}` — null on ordinary completions; `pause_orphan` reserved for #206.
- `SDK.md` — documents the new typed views and the null semantics (esp. "null `pause_orphan` = not yet determined, not 'not an orphan'").

## Tests
`tests/test_adk2_event_types.py` (40 cases): cross-surface registration for all six types (enum / families / evaluator allowlist / UDF map / view defs), the exact typed-column SQL per type, `WORKFLOW_NODE_*` header-only, and the `TOOL_COMPLETED` long-running columns. Full suite: **3421 passed, 15 skipped**.

## Deferred (tracked)
- `WORKFLOW_NODE_*` typed columns → #207 (workflow-boundary derivation).
- `AGENT_STATE_CHECKPOINT` offload columns → #208.
- The `source_event_id`/long-running join recipes in `SDK.md` → #220.


caohy1988:
Review finding:

[P2] Preserve the AGENT_STATE_CHECKPOINT state discriminator in the SDK typed view.
`src/bigquery_agent_analytics/views.py:268` defines the SDK `agent_state_checkpoints` view with only `end_of_agent` and `agent_state`, but the merged ADK producer's corresponding typed view also emits `JSON_TYPE(JSON_QUERY(content, '$.agent_state')) AS agent_state_type`. That discriminator is not cosmetic: the producer comment documents the `{agent_state: null, end_of_agent: true}` shape, where consumers need to distinguish an absent key / SQL NULL from an explicit JSON null state. Without `agent_state_type`, downstream SDK users lose that distinction as soon as they use the SDK-managed view. Please mirror the producer column and pin it in `test_agent_state_checkpoint_columns_inline`.

Everything else I checked looks consistent: the six event types are registered across `EventType`, semantics, evaluator defaults, UDF labels, and views; `TOOL_COMPLETED` carries the long-running pair keys with the documented `pause_orphan` null semantics; workflow views are correctly header-only pending #207; CI is green.

Verification: checked PR head `7696e70`, compared the view SQL against `google/adk-python` `origin/main` producer view definitions, `git diff --check` clean, and all GitHub required checks passing. Local targeted pytest collection is blocked by stale `.venv` package metadata on this machine, so I relied on CI for execution.

caohy1988:
P2 addressed in 4d3d764. The `AGENT_STATE_CHECKPOINT` view now exposes `agent_state_type` = `JSON_TYPE(JSON_QUERY(content, '$.agent_state'))`, mirroring the producer's own view in `google/adk-python` `origin/main`. Consumers can now distinguish:

- **SQL NULL** → `agent_state` key absent
- **`'null'`** → the explicit `{agent_state: null, end_of_agent: true}` shape
- **anything else** (`'object'`, …) → a real state object

Test asserts the column; SDK.md documents the discriminator. Full affected suite green.

haiyuan-eng-google (approved):

PR #319, opened 2026-06-25
skill_evolution_lab: trace-grounded skill evolution with tool selection, anti-parroting, declines, and reproducible sweeps

evekhm:
## Summary

Evolves the `examples/skill_evolution_lab` demo (merged in #301) into a
**multi-tool, golden-grounded, self-improving skill loop**, with a reproducible
multi-model sweep and a recorded end-to-end run. All changes are within
`examples/skill_evolution_lab/` plus the shared engine (`scripts/skill_evolution.py`)
and its tests.

### Multi-tool demo (tool selection)
- Added a **2nd tool** (`calculate_disability_pay`) alongside `lookup_company_policy`,
  so the agent must learn *tool selection* (look up a fact vs. compute a payout).
- Made the evolution **analysts tool-aware** (they see a tool description) so
  deflection failures yield real "use the tool" patches instead of `NO_PATCH`.
- Dropped the hardcoded synonym/keyword mappings — the consolidator now produces a
  tool-first skill that lets the tool resolve wording. Renamed the error category
  `KEYWORD_GAP → WRONG_TOOL`.

### Engine (`scripts/skill_evolution.py`) + tests
- `strip_code_fences` now also removes an orphan ``` fence left after YAML
  frontmatter (it would otherwise render verbatim in the deployed skill).
- Tool-aware analyst/consolidator path. **26 unit tests, all green.**

### Verified results (recorded run + 4×3 sweep, held-out, golden-grounded)
Recorded `gemini-3.5-flash` run: **V0 20% (14/70) → V1 100% (70/70)**, evolved
skill ~1.9KB. Four models × 3 seeds on the multi-tool demo:

| Model | V0 | V1 mean [range] | V1 grounding [range] |
|---|---|---|---|
| gemini-3.5-flash | 21% | 100% [100-100] | 89% [81-96] |
| gemini-3.1-flash-lite | 34% | 98% [96-100] | 89% [83-100] |
| gemini-2.5-pro | 51% | 93% [89-97] | 84% [83-84] |
| gemini-3.1-pro-preview | 47% | 100% [100-100] | 83% [81-86] |

The pro models reach for the tool even under the flawed V0 skill (higher V0), so
they start higher and have less headroom; the evolved skill lifts every model to
93–100%.

### Tooling / DX
- `run_e2e_demo.sh` mirrors `examples/agent_improvement_cycle/run_cycle.sh`:
  `--flag` CLI + `--help`, one `run.log` via the same `tee | sed` ANSI-strip,
  colored stage banners with per-step timing, config banner, final summary.
- `run_sweep.sh`: detachable long-running multi-model sweep — self-logs to
  `runs/SWEEP_<ts>.log`, continues past a single failed run, aggregates whatever
  completed.
- **Bug fix:** env-passed model overrides are now captured before `.env` is sourced
  (precedence: CLI flag > env > .env > default), and `run_sweep.sh` passes the model
  via `--agent-model`. Previously every sweep run silently fell back to the `.env`
  default — the sweep now genuinely runs all four models.
- `run_agent.py` / `analyze_and_evolve.py`: silence noisy `google_genai` / `httpx`
  loggers so the demo flow is readable.

## Testing
- `pytest tests/test_skill_evolution.py` → **26 passed**.
- Full 4-model × 3-seed sweep run end-to-end on the multi-tool code; per-run model
  identity verified (each model actually ran). `sample_run/` is the committed
  artifact set for the recorded run; `VERIFICATION.md` documents both.

---

### Update — out-of-scope / `declined` (latest commit)

The demo now genuinely exercises the `declined` verdict. Added out-of-scope question
slices (`eval/questions_oos.json`, `eval/questions_oos_heldout.json`), wired them into
`run_e2e_demo.sh`, and taught `compare_runs.py` to credit a correct decline (new
**Out-of-scope (declined)** RESULT row). No scorer change — the judge derives `declined`
from the eval_spec `scope` field.

Refreshed recorded run (`gemini-3.5-flash`, 80-question held-out = 55 single + 15
anti-parrot + 10 out-of-scope):

| Metric | V0 (flawed) | V1 (evolved) |
|---|---|---|
| Overall | 28.7% (23/80) | 100.0% (80/80) |
| Single-turn (in-scope) | 18.2% (10/55) | 100.0% (55/55) |
| Corrections (anti-parrot) | 26.7% (4/15) | 100.0% (15/15) |
| Out-of-scope (declined) | 90.0% (9/10) | 100.0% (10/10) |

V0 already declines out-of-scope well (deflecting everything is *correct* for a genuinely
out-of-scope question); the evolved V1 (~2.2KB) declines deliberately via an engine-written
**Out-of-Scope Handling** section (route IT to IT, refuse unrelated). `sample_run/`,
`README.md`, and `VERIFICATION.md` updated. The 4-model sweep above predates this slice
(in-scope only). Tests: **27 passed**.


evekhm:
Pushed `c63468d`: removed all hardcoded keyword/synonym mappings.

- **Tool** (`agent/tools.py`): dropped `_TOPIC_ALIASES`/`_resolve_topic`; `lookup_company_policy` is topic-keyed (exact + substring) and the model maps the user's wording itself.
- **Engine** (`scripts/skill_evolution.py`): stopped emitting Keyword/Terminology Mapping; strengthened the consolidator to be tool-first (rewrite "contact HR" deflection into a tool-first rule), ban fact-baking, and generalize per-topic lists.
- Refreshed `sample_run/`, `VERIFICATION.md`, `README.md` from a clean gemini-3.5-flash run: **V0 20% → V1 100%**, 2.4KB tool-first skill, no synonym table. 26 unit tests pass.

4-model × 3-seed sweep is running now to refresh the multi-model table.

evekhm:
Pushed `c0f4610`: made the analysts **tool-aware** -- fixes a NO_PATCH bug that made the demo's core claim hollow.

**The bug:** the error analysts never knew a lookup tool existed (the V0 skill doesn't mention one), so every single-turn deflection ("contact HR") came back `NO_PATCH` -- *can't invent the missing fact.* The tool-first rule was only leaking into V1 from the handful of tool-using successes plus the consolidator, **not** from the failures. It stayed hidden because V1 still scored ~100%.

**The fix:** thread a tool **description** (not access) into the analysts via the eval spec's `tools` field. A deflection on a tool-covered topic is now a fixable `TOOL_USAGE` patch (*call the tool first*), never NO_PATCH and never a baked fact.

**Verified end-to-end** (gemini-3.5-flash): V0 18.5% -> **V1 100%**, a **1.5KB** tool-first skill, and **46/46 patches pass the quality gate** (previously the deflection failures were dropped). `sample_run/`, `VERIFICATION.md`, `README.md` refreshed from that run; 26 unit tests pass.

Files: `scripts/skill_evolution.py` (tools threaded through evolve_skill -> collect_patches -> run_analyst; error-analyst prompt), `eval/eval_spec.json` (+`tools`), `analyze_and_evolve.py` (+`--eval-spec`), `run_e2e_demo.sh`.

evekhm:
Pushed `94a7119`: **added a second tool so the demo shows tool *selection*** (the single-tool skill was too thin).

- New tool `calculate_disability_pay(annual_salary, weeks_out)` -- *computes* a personalized short-term-disability payout the lookup can't produce. The agent now has to choose: **lookup for facts, calculator for personalized math.**
- 10 STD-payout calc questions (5 evolve + 5 held-out with **distinct** salaries/weeks so the split stays disjoint), each in `golden_qa` with its computed answer; `eval_spec.tools` describes both tools + defines a wrong-tool-selection failure.
- Verified (gemini-3.5-flash): **V0 20% (14/70) → V1 100%**, 51/51 patches pass, and V1 evolves a real **routing** rule:

```
## Tool Usage
- lookup_company_policy: retrieve authoritative facts for any policy/benefit...
- calculate_disability_pay: compute personalized dollar amounts when a user gives salary/duration...
```

A V0 that deflects on *"what's my STD payout?"* becomes a V1 that picks the calculator and shows the breakdown -- a richer, more realistic result than the single-tool skill. `sample_run/`, `VERIFICATION.md`, `README.md` refreshed; 26 tests pass. (The 4-model sweep table in VERIFICATION still reflects the single-tool run -- flagged inline, will refresh on the next sweep.)

caohy1988:
I reviewed this PR with two lenses: code/reproducibility, and whether the accompanying story makes "skill evolution from BQAA traces" feel like a powerful, reusable pattern.

Overall: the core is strong. The PR demonstrates a real closed loop:

```text
trace -> golden-graded score -> analyst diagnosis -> behavioral skill patch -> held-out re-score -> versioned skill
```

The biggest opportunity is to make the trace evidence more literal. Right now the accuracy story is clear, but the tool-selection story is partly inferred. If the PR shows observed tool names/args and reports "right tool selected" directly, this becomes much harder to challenge and much easier for other BQAA users to copy.

## Comments

1. `examples/skill_evolution_lab/run_agent.py:117`

   `_count_tool_calls()` only records a count, but the PR's central claim is that V1 learns tool selection. Please persist structured tool calls too, at least `{name, args}`. Then the result can show `lookup_company_policy` vs `calculate_disability_pay` directly.

2. `scripts/skill_evolution.py:255`

   `format_trajectory()` does not include tool names, args, or results. For a trace-to-skill story, analysts should see the actual tool behavior, not only the judge's grounding category. This would make `WRONG_TOOL` and `TOOL_USAGE` diagnoses much stronger.

3. `scripts/skill_evolution.py:841`

   Artifacts are written from raw `cands`, but selection happens after `sanitize_adk_vars()`. If sanitization changes the selected candidate, `_SELECTED` may not be tagged correctly, and the saved candidate may differ from returned `v1_skill.md`. Pass the sanitized viable candidates into `_write_evolution_artifacts()`.

4. `examples/skill_evolution_lab/run_sweep.sh:60`

   The loop calls these "seeds," but no seed is passed into the run. Either rename this to "runs" / "repetitions," or thread a real seed through the agent/evolution calls where supported.

5. `examples/skill_evolution_lab/VERIFICATION.md:56`

   The 4x3 sweep table says it predates the out-of-scope slice, but current `run_sweep.sh` now runs through `run_e2e_demo.sh`, which includes OOS. The reproduction instructions will not reproduce the same table. Either rerun the sweep with OOS or label this as a historical in-scope-only sweep.

6. `examples/README.md:95`

   This top-level summary is stale: it says 70-question held-out, `V0 20% -> V1 100%`, and ~1.9KB. Current sample run says 80-question held-out, `28.7% -> 100%`, and ~2.2KB.

7. `examples/skill_evolution_lab/analyze_and_evolve.py:141`

   If `--eval-spec` is provided but missing or has no `tools`, the demo silently becomes non-tool-aware. Since tool-aware analysts are part of the core claim, this should fail fast with a clear error.

8. `examples/skill_evolution_lab/agent/tools.py:203`

   The STD calculator mentions a 7-day waiting period, but payout examples count every requested week. Either document that the waiting period is informational only in this demo, or make the calculator/golden answers apply it consistently.

9. `scripts/skill_evolution.py:853`

   The reusable engine CLI does not expose `--tools`, `--eval-spec`, or `--artifacts-dir`, even though the lab wrapper depends on those to make evolution useful and inspectable. Exposing them would make this easier to reuse outside the example.

10. `examples/skill_evolution_lab/README.md:3`

   "No teacher model" is a little confusing because the evolution loop uses analyst/consolidator LLMs. Consider "no managed optimizer" or "no hand-written patch" instead.

## Story suggestions

Suggested PR title:

```text
skill_evolution_lab: trace-grounded skill evolution with tool selection, declines, and reproducible sweeps
```

Add one explicit "why BQAA matters" paragraph:

```markdown
BQAA is what makes this pattern reusable: it captures the conversation, tool behavior, corrections, and outcome labels. The evolution engine turns those traces into behavioral skill rules, then validates them on held-out traffic before creating a new skill revision.
```

Add one result table that proves behavior, not just accuracy:

```text
Behavior                     V0        V1
In-scope answered correctly   10/70     70/70
Calculator tool selected      X/5       Y/5
Lookup tool selected          X/65      Y/65
OOS cleanly declined          9/10      10/10
```

That would make this land as a reusable BQAA pattern, not just a good demo. The PR already has the right ingredients: trace capture, golden grading, analyst patches, candidate artifacts, held-out validation, and registry revisioning. The last mile is making the trace-to-skill evidence visible enough that readers immediately understand how to apply the pattern to their own agents.

caohy1988:
High-level take: yes, this is a useful and potentially very impactful demo.

The reason is not just the `V0 -> V1` number. The important pattern is that BQAA traces become a learning substrate:

```text
user conversations + tool behavior + corrections + golden scoring
  -> recurring failure modes
  -> behavioral skill rules
  -> held-out validation
  -> versioned skill revision
```

That is a strong demo because it shows a concrete path from observability to agent improvement. Most tracing demos stop at "look, we can inspect failures." This one goes further: it turns failures into a better skill and proves the change on held-out traffic. That is the real product value.

What makes it impactful:

- It makes BQAA feel actionable, not just diagnostic.
- It gives users a repeatable pattern for improving agents without hand-writing every patch.
- It shows the right abstraction boundary: tools own facts, skills own behavior.
- It makes skill changes inspectable and reversible through versioned artifacts.
- It includes hard cases beyond simple Q&A: tool selection, anti-parroting, and out-of-scope decline.

The one thing I would tighten before presenting it as a flagship pattern: make the trace evidence more visible. The demo should show not only that V1 answered correctly, but that it selected the right tool, re-verified after correction, and declined OOS for the right reason. A small behavior table would make this land immediately:

```text
Behavior                     V0        V1
In-scope answered correctly   10/70     70/70
Calculator tool selected      X/5       Y/5
Lookup tool selected          X/65      Y/65
OOS cleanly declined          9/10      10/10
```

If that is added, the demo becomes more than "an agent got better." It becomes "BQAA provides the traces needed to evolve skills safely." That is the powerful and easy-to-leverage pattern worth centering.

examples/skill_evolution_lab/agent/tools.py
```diff
@@ -198,44 +162,74 @@
...
-  topic_key, result = _resolve_topic(topic)
-  if result is None:
-    return {
-        "error": (
-            f"'{topic}' did not match a known policy. Valid topics: "
-            + ", ".join(sorted(COMPANY_POLICIES))
-        )
-    }
-  return result
+  covered_weeks = min(weeks_out, _STD_MAX_WEEKS)
```
caohy1988, line 227: **[P2]** `calculate_disability_pay` returns `waiting_period_days: 7` but never applies it. The docstring says "60% of salary for up to 12 weeks *after a 7-day waiting period*," yet the payout is `weekly_benefit * min(weeks_out, 12)` with no waiting-week deduction. The tool and the golden answers agree with each other, so the demo scores fine -- but this is a demo about grounding answers in correct tools, and the showcase tool's math quietly contradicts its own policy text. A careful reader (or a strong agent model) will compute "4 weeks out = 3 paid weeks = $1,800", get marked wrong against the $2,400 golden, and the judge is then penalizing the *more* correct reading.

Fix: either apply the waiting period in the math and regenerate the goldens, or state explicitly (docstring + the STD entry in `lookup_company_policy`) that the waiting period is unpaid time before benefits start and does not reduce the number of payable weeks.

examples/skill_evolution_lab/run_e2e_demo.sh
```diff
@@ -12,121 +12,357 @@
...
+      ;;
+  esac
+done
+
+REPORTS_DIR="$SCRIPT_DIR/runs/${RUN_TIMESTAMP}_${AGENT_MODEL//[^a-zA-Z0-9]/_}"
+mkdir -p "$REPORTS_DIR"
+
+# Tee all output: terminal gets colour, log file gets plain text.
+RUN_LOG="$REPORTS_DIR/run.log"
+exec > >(tee >(sed 's/\x1b\[[0-9;]*m//g' >> "$RUN_LOG")) 2>&1
```
caohy1988: **[P2]** `sed 's/\x1b\[[0-9;]*m//g'` doesn't strip ANSI on macOS: BSD sed doesn't interpret `\x1b`, so the pattern matches the literal characters `x1b[...m` and every color code lands in `run.log` verbatim. The committed sample log is clean because it was recorded on Linux; a Mac user reproducing the run gets a garbage log.

Fix: let bash produce the ESC byte instead of sed:
```bash
exec > >(tee >(sed $'s/\033\[[0-9;]*m//g' >> "$RUN_LOG")) 2>&1
```
Same pattern exists in `agent_improvement_cycle/run_cycle.sh` (pre-existing, worth a follow-up since this PR copies it deliberately).

examples/skill_evolution_lab/print_rate.py
```diff
@@ -12,14 +12,30 @@
...
-print(
-    f"{g.get('matched_meaningful_rate')}% "
-    f"({g.get('matched_meaningful')}/{g.get('matched')} golden-matched)"
-)
+
+rate = g.get("matched_meaningful_rate")
+matched_meaningful = g.get("matched_meaningful")
+matched = g.get("matched")
+total = g.get("total_sessions")
+out_of_scope = g.get("unmatched")
```
caohy1988: **[P2]** This labels *all* `unmatched` sessions as "out-of-scope". An in-scope question that fails the cosine >= 0.92 golden match would be silently mislabeled out-of-scope *and* silently dropped from the correctness denominator, inflating the rate. It happens to be true in the recorded run (70/70 in-scope matched), but the label bakes in an assumption the code doesn't check.

Fix: count sessions whose `session_id` starts with `oos_` (the same convention `compare_runs.py` uses) and print any remainder as "unmatched" so a matching failure is visible instead of flattering.

examples/skill_evolution_lab/run_sweep.sh
```diff
@@ -13,37 +13,73 @@
...
+#   cat runs/SWEEP_*.md               # final mean [range] table when done
+#
+#   # subset / fewer seeds:
+#   MODELS="gemini-3.5-flash gemini-2.5-pro" SEEDS=2 ./run_sweep.sh
+#
+# A single failed run (API blip, quota) is logged and skipped; the sweep keeps
+# going and still aggregates whatever completed. To re-read a finished sweep
+# without re-running, point aggregate_sweep.py at its manifest:
+#   uv run python aggregate_sweep.py --manifest runs/sweep_<ts>.tsv
+set -uo pipefail
```
caohy1988: **[P3]** Dropping `set -e` entirely is broader than the goal. The `if ./run_e2e_demo.sh ...` guard already suppresses `-e` for that command by POSIX rules, so one failed run wouldn't kill the sweep anyway. Without `-e`, failures in `mkdir`, `cd`, and the manifest writes are now silently tolerated too. Recommend restoring `set -euo pipefail`; the `if` guard alone gives the continue-past-failure behavior.

scripts/skill_evolution.py
```diff
@@ -391,6 +434,57 @@ def compute_prevalence_summary(patches: list[str]) -> str:
   return "\n".join(lines)
 
 
+def _write_evolution_artifacts(
```
caohy1988, line 465: **[P3]** Two things that will confuse the first V1->V2 run:

1. Artifact names are hardcoded `v1_*` regardless of which round produced them.
2. When `select_candidate` keeps the incumbent, no candidate gets the `_SELECTED` tag and nothing records why.

Cheap fix: accept a version label, and write a one-line `selection.txt` (e.g. "kept incumbent: no candidate beat score by 0.5") when the guard fires. That file is also great story material -- it's the audit trail of the safety mechanism the README advertises.

examples/skill_evolution_lab/run_agent.py
```diff
@@ -40,6 +40,33 @@
...
+
+  warnings.filterwarnings(
+      "ignore", category=authlib.deprecate.AuthlibDeprecationWarning
+  )
+except ImportError:
+  pass
+
+# Suppress noisy SDK loggers before any google imports ("AFC is enabled",
+# "...will take precedence", "HTTP Request: POST ...").
+for _noisy in (
```
caohy1988: **[P3]** This ~20-line warning/logger-suppression block is duplicated verbatim in `analyze_and_evolve.py` (and belt-and-suspenders'd again via `PYTHONWARNINGS` in `run_e2e_demo.sh`). Third copy incoming the next time someone adds a script here. Suggest moving it to a tiny `_quiet.py` in the lab and importing it from both.

examples/skill_evolution_lab/run_e2e_demo.sh
```diff
@@ -12,121 +12,357 @@
...
-echo "   run=$RUN   log=$RUN/run.log"
-
-# Always leave the LOCAL working copy on V0 (repo sits at V0). Note: with
-# WITH_REGISTRY=1 this does NOT revert the remote Skill Registry to V0 -- run
-# `./reset.sh` (WITH_REGISTRY=1) for that.
+
+# Parse arguments
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --agent-model)
```
caohy1988, line 111: **[P3]** A flag missing its value dies with a raw `set -u` "unbound variable" error, e.g. `./run_e2e_demo.sh --agent-model`. A one-line guard per case (or a small `need_val()` helper) turns that into a readable usage error.

Also: `jq` is now a soft dependency for the golden-count banner and isn't in the README prerequisites; the `?` fallback works, but a one-line mention avoids a confused first run.

caohy1988 (commented):
This is a strong PR. The engineering is honest (the env-precedence bug fix, a real 4-model sweep, committed artifacts, disclosed caveats), the engine changes are well-tested, and the tool-aware analyst fix repairs a hole that made the demo's core claim hollow. No blocking bugs found; I've left a handful of P2/P3 inline comments. The bigger opportunity is the story, below.

## Verification & metrics framing

1. **The sweep predates the demo it's published next to.** The 4x3 table was run before the OOS slice and before the current engine. It's disclosed, but the disclosure means VERIFICATION.md documents two different experiments. Since this PR made detached sweeps easy (~3-4h wall time), re-running it on the final code is cheap relative to how much credibility one consistent artifact buys.
2. **Lead with the in-scope number.** "Overall 28.7% -> 100%" includes 10 out-of-scope questions where V0 already scores 90% by accident. The honest headline is single-turn in-scope 18.2% -> 100% and corrections 26.7% -> 100%; keep Overall in the table, but don't make it the opening claim.
3. **The held-out set is saturated.** V1 scores 100% on every slice and a perfect 2.00 on all five quality dimensions. That reads as ceiling, not excellence: the eval can no longer distinguish a good evolved skill from a great one, and it makes best-of-N / `score_fn` look unnecessary (they only matter when candidates differ measurably). A harder slice (ambiguous phrasing, multi-topic questions, questions needing lookup + calculation combined) would land V1 at 85-95% and give deltas room to breathe.

## Making "skill evolves from traces" a first-class BQAA pattern

**The flagship self-improvement demo of the BigQuery Agent Analytics SDK never touches BigQuery.** `setup.sh` says it outright: "No BigQuery is required -- scoring runs offline via --conversations-file." Meanwhile the README's research table says traces are "logged to BigQuery by the Agent Analytics plugin" -- aspirational, not what the lab does. And the machinery already exists in this repo: `scripts/quality_report.py` is natively a BigQuery tool (PROJECT_ID/DATASET_ID/TABLE_ID, `bigquery_agent_analytics.Client`), and `examples/agent_improvement_cycle` already demonstrates the full path: traffic -> `BigQueryAgentAnalyticsPlugin` -> `agent_events` -> quality report from BigQuery.

So today the repo splits the story in half: `agent_improvement_cycle` has the BQAA data path but improves a *prompt* with a different mechanism; `skill_evolution_lab` has the skill-evolution engine but sidesteps BigQuery. Neither example shows the sentence this SDK should own: **"Your agent already logs its failures to `agent_events`. Here's the command that turns them into a better skill."**

Recommendations, in priority order:

1. **Add a `--from-bigquery` mode to the lab (or at minimum document the command chain).** `run_agent.py` logs via the plugin; the score step drops `--conversations-file` and reads from `agent_events`; `evolve_skill()` consumes the same scored report. The delta is small because `quality_report.py` already supports both modes -- this is wiring, not new machinery. Keep the offline mode as the zero-setup on-ramp, but make the BigQuery path the headline.
2. **Ship a generic entrypoint, not just a lab.** Adopting the pattern today means copying a demo directory and editing HR-specific files. The engine deserves a front door (`skill_evolution.py --report r.json --skill SKILL.md --eval-spec spec.json`, or a `bqaa skill-evolve` subcommand from dataset + app_name + date range). Then document the adoption contract that is currently implicit: `session_id` prefixes (`corr_`, `oos_`) drive metric slicing, `eval_spec.tools` is a freeform prose string, `golden_qa` is optional-but-recommended, root-cause categories are fixed. A "point this at your own agent" section with those four bullets is worth more than another hundred golden questions.
3. **Show round 2.** The story sells versioned, continuous evolution (registry revisions, AutoSkill merge semantics), but the demo runs exactly one round. Run V1's traffic back through the loop. Even a null result is powerful -- "V2 = V1, the engine kept the incumbent, no bloat, no regression" is exactly the stability claim a practitioner needs before trusting this near production, and it exercises the incumbent guard for real instead of in prose.
4. **Put receipts behind `score_fn`.** Best-of-N with an incumbent gate is the differentiator over Trace2Skill/AutoSkill (the README says so), but the recorded run skips scoring and takes the median-size candidate -- the safety feature the docs highlight is switched off in the artifact that proves the system works. One recorded run with `score_fn` (score candidates on the evolve set), plus a small `selection.txt` audit artifact, makes the claim demonstrable.
5. **Add an honest-baseline arm.** The crippled-V0 hedging is genuinely well done, but skeptics discount a demo where the defect was planted. One extra recorded run with a *plausible* V0 showing a modest, real gain (even "+8pp on corrections, no regression elsewhere") is the number people will quote in a design review. It also naturally de-saturates the eval.
6. **Resolve the baked-facts tension in the showcase artifact.** The story's moral is "skills are behavioral; facts live in tools" -- yet the evolved V1 still opens with V0's hardcoded fact block ("PTO: 20 days per year...") because consolidator rule 8 preserves base facts and rule 2 forbids dropping content. Rule 2 already has an exception for deflections; consider the symmetric exception: base facts that a tool can supply may be migrated to a lookup rule. Watching the skill get *smaller and more general* as it evolves would be the most persuasive artifact this demo could produce.

One-sentence version: solid and mergeable after the P2s; the pattern becomes powerful the day the loop starts and ends in BigQuery, has a generic front door, and shows a second round.

caohy1988:
## Review of the companion blog post ("Your Agent Can Learn From Its Own Conversations")

Since this PR backs the blog post's numbers, leaving the blog review here so it lives next to the code. Overall: strongest post in the series -- the NO_PATCH analyst-blindness story and the parroting trace-walkthrough are excellent technical writing, and I fact-checked the numbers against `sample_run/` (prevalence 51/53 + 2/53, the 17/51 partition, 68/80 splits, sweep figures, 0.6 -> 2.00 dimensions): all clean. A few things worth fixing before it circulates widely.

### Accuracy (worth fixing first)

1. **"Keep it only if it's better" doesn't match what the demo does.** Loop step 5 says the new skill is kept only when it wins, otherwise fall back, and "mirror each *kept* version to the Skill Registry." But in `run_e2e_demo.sh` the registry mirror happens in STEP 2 (`analyze_and_evolve.py --registry-update`), *before* V1 is measured in STEP 3, and nothing gates on the STEP 4 comparison -- `compare_runs.py` prints a table, there's no fall-back path, and the incumbent guard (`score_fn`) is off by default. With `--with-registry`, V1 lands in the registry even if it had lost. Fix in the post ("the demo mirrors unconditionally; gating is the `score_fn` path") or better, in this PR: move the registry push after the compare. See inline comment.

2. **A copy-edit bug garbles the core section.** In "How we measure quality", the `ground_truth` bullet is truncated ("...trust the tool's value over any figure") and its tail is spliced into the `golden_qa` bullet ("...grades against. a user asserts; this is what catches..."). Two-minute fix, but it's the section explaining the central mechanism.

3. **Loop Step 1 says the plugin logs to BigQuery -- the runnable demo never touches BigQuery.** `run_agent.py` writes local JSON, scoring runs `--conversations-file`, and `setup.sh` says "No BigQuery is required." The post discloses this later, but the loop description states the BQ path as fact, in a series named after BigQuery. One sentence at the loop diagram fixes it ("in the demo, traces are local JSON in the same schema the plugin logs; the BigQuery read is the production wiring") -- and the durable fix is a `--from-bigquery` mode in this lab (see my code review on this PR).

4. **The sweep table sits next to the OOS narrative without the caveat that the sweep predates the OOS slice.** VERIFICATION.md discloses it; the post should too -- half a sentence.

### Credibility

5. **"No teacher model" needs a footnote.** The analyst is `gemini-3.1-pro-preview` while the default agent is `gemini-3.5-flash`; a skeptic will call the analyst a teacher. The defense is real (the analyst diagnoses, it never supplies answers, and `gemini-2.5-pro` as agent improves too) -- state it explicitly. Even better: one run with analyst = the same flash model would make the claim bulletproof.

6. **Own the 100%.** All slices at 100% with every dimension pinned at 2.00 means the held-out set is saturated for this defect. Saying so (and that a harder set would land V1 at 90-something) reads *more* credible than a perfect score, not less. The crippled-V0 honesty box is already the best part of the framing -- this is its missing other half.

### Structure

7. **27 minutes is two posts.** Natural seam: through "The results" = the loop + the proof; the parroting deep-dive + failure attribution can carry its own headline. If it stays single, a TL;DR box with the results table + jump links up top -- the best material (the NO_PATCH bug) is currently buried at minute ~15.
8. **No hero diagram.** The five-step loop exists only as prose and ASCII. One image (trace -> score -> analyst fleet -> consolidate -> re-score -> keep/registry) is what gets shared when the post is linked.
9. **Broken promise on compaction.** The intro says the four additions are "each unpacked in its own section later"; compaction gets one sentence. Unpack it in three sentences or drop the promise.

### Nits

- "held-out" and "held-back" both used for the same set -- pick one.
- The evolved-skill section doesn't note the tension that V1 still opens with V0's baked fact block while the post's moral is "facts live in tools." One sentence acknowledging it (migrating base facts to tool-first = future work) preempts the sharpest reader comment.
- "Going further" is where the adoption contract belongs: `session_id` prefixes drive metric slicing, `eval_spec.tools` is freeform prose, the root-cause taxonomy is fixed. Three bullets save every adopter an hour of source-reading.

### What's working

The NO_PATCH story (a green end metric hiding a hollow middle step, caught only by reading the patches), Alex the skeptic, the filled-in `t04_pto` scorecard, and the skill/knowledge/tool attribution ending -- "the difference between a quality score and a quality loop" is the takeaway line of the series. Priority if only three things get fixed: #2, #1, #3.

caohy1988:
Blog post review, high-level narrative and credibility feedback.

This is a strong post. The core idea is genuinely useful: BQAA should not stop at observability. It can become the feedback loop that improves an agent.

The strongest framing is:

```text
BQAA traces -> golden-graded scorecards -> analyst patches -> consolidated SKILL.md -> held-out re-score -> versioned skill revision
```

That is a real and impactful story. Most observability demos end at "we found the failure." This one goes further: it turns traces into a reviewable skill diff and proves the diff on held-out traffic.

## Main feedback

1. Clarify "no teacher model."

   The post says "no teacher model," but the loop still uses judge, analyst, and consolidator LLM calls. A skeptical reader may read that as overclaiming. I would change the claim to:

   > No hand-written fixes. No per-failure human labels. The agent improves from its own scored traces.

   Or:

   > No managed optimizer and no manually written patch; the learning signal comes from the agent's own scored conversations.

2. Make BQAA's unique value sharper.

   The key value is not just that conversations are stored in BigQuery. It is that BQAA captures the full improvement substrate:

   > conversations, tool calls, corrections, outcomes, and skill/version attribution in one analyzable place.

   This should appear near the top. The reader should understand early that BQAA is what makes trace-to-skill evolution practical.

3. Move the "deliberately crippled V0" caveat earlier.

   The caveat is good and honest, but it appears too late. Put it in the first third:

   > This demo intentionally starts from a flawed skill so the loop has a clear defect to find. The size of the gain is illustrative; the reusable pattern is the contribution.

   That will prevent readers from feeling like the post is hiding the setup until after the big numbers.

4. Be more precise about anti-parroting.

   The post says the trace gives "mechanical" proof because a post-correction tool call means recovery. That is a little too strong. A tool call can be unrelated, ignored, or followed by a wrong answer.

   Better:

   > A genuine recovery requires both: the post-correction answer matches the golden answer, and the trace shows the agent re-verified with a relevant tool before answering.

   This preserves the point while making the claim technically defensible.

5. Resolve the 70 vs 80 denominator confusion.

   The post uses both `70/70 golden-matched` and `80/80 overall`. Add one clear sentence before the results:

   > The held-out set has 80 conversations: 70 in-scope conversations matched to golden Q&A, plus 10 out-of-scope conversations scored as clean declines.

   This will save readers from thinking the denominator changed mid-post.

6. Tighten the research section.

   The Trace2Skill / AutoSkill context is useful, but it slows the opening. I would keep the research explanation to 2-3 short paragraphs and push the full mapping to the lab README. The blog post should lead with the product pattern, not the paper lineage.

7. Make demo vs production explicit.

   The post sometimes reads like the demo consumes real BigQuery production traffic, but the lab generates local conversations in the BQAA-compatible schema. Add this distinction:

   > The demo generates local conversations in the same schema; in production, those conversations come directly from BQAA tables in BigQuery.

8. Avoid "the agent read all of them."

   "All conversations" sounds risky and unrealistic. Better:

   > The engine reads a filtered batch of recent conversations for the deployed skill version.

   That also strengthens the production story because skill-version filtering matters.

9. Center the tool/fact boundary.

   One of the best ideas in the post is:

   > tools own facts, skills own behavior.

   This should be promoted earlier and repeated. It explains why the evolved skill is not a stale knowledge base, and why BQAA traces can produce durable behavioral rules.

10. End with a sharper takeaway.

   Suggested final line:

   > The demo is not that one HR agent got better. The demo is that BQAA traces can become reviewable, versioned skill improvements.

## Suggested opening rewrite

```markdown
Most agent observability stops at dashboards: what failed, where, and how often.

This demo closes the loop.

The BigQuery Agent Analytics Plugin captures conversations, tool calls, corrections, and outcomes. The SDK scores those traces against a golden Q&A, turns recurring failures into behavioral skill rules, writes a new SKILL.md, and validates it on held-out traffic.

That is skill evolution: not a model magically improving itself, but a reviewable loop from trace -> score -> patch -> re-score -> versioned skill.
```

## Suggested structure

1. The closed-loop claim: BQAA traces can become skill improvements.
2. The demo setup: deliberately flawed skill, two tools, golden Q&A, held-out set.
3. The loop: generate/collect traces, score, evolve, re-score, version.
4. What the engine learned: tool-first behavior, tool selection, anti-parroting, clean declines.
5. Results, with the V0 caveat and denominator explanation.
6. Production pattern: filter by skill version, run scheduled, open PR, file knowledge/tool issues.
7. Takeaway: trace observability becomes a quality-improvement loop.

## Verdict

This is impactful and worth publishing. The post just needs to be a little more disciplined about what is automatic, what is judged, what is trace-observed, and what is intentionally demo-shaped.

The claim to center:

> BQAA turns agent traces into a reusable improvement loop.

Everything else should support that claim.

examples/skill_evolution_lab/run_e2e_demo.sh
```diff
@@ -12,121 +12,357 @@
...
+echo "  Method:  Analyst fleet reads the scored evolve set; best-of-3 candidates"
+echo "  Model:   analyst=$ANALYST_MODEL  (this is the slow step)"
+echo ""
+step_start
+
 REG_ARGS=()
-if [ "${WITH_REGISTRY:-0}" = "1" ]; then
-  REG_ARGS=(--registry-update --skill-id "${SKILL_ID:?set SKILL_ID with WITH_REGISTRY=1}" --location "$REGISTRY_LOCATION")
+if [[ "$WITH_REGISTRY" == "1" ]]; then
+  REG_ARGS=(--registry-update --skill-id "${SKILL_ID:?--skill-id is required with --with-registry}" --location "$REGISTRY_LOCATION")
```
caohy1988: **[P2]** Registry mirroring happens here in STEP 2 -- *before* V1 is measured (STEP 3) and compared (STEP 4), and nothing gates on the comparison result. With `--with-registry`, V1 is pushed to the Skill Registry even if it loses on the held-out set. That contradicts the loop's step 5 as described in the blog post and README ("keep the new skill only when it wins; mirror each *kept* version").

Suggest moving the registry push after the STEP 4 comparison and gating it on the V1 >= V0 result (or on the engine's `score_fn` incumbent guard). As-is, the demo's stated safety property isn't enforced anywhere in the recorded path.

caohy1988 (commented):
Follow-up to my earlier review: one more finding surfaced while reviewing the companion blog post (full blog review in the PR conversation).

examples/skill_evolution_lab/agent/tools.py
evekhm: Fixed in 0eef3c8 (option 2): the waiting period is now explicitly documented as unpaid time before benefits begin that does **not** reduce the 12 payable weeks -- in the calculator docstring, the `weeks_out` arg description, the STD entry in `lookup_company_policy`, and a `waiting_period_note` field in the tool's return value (so a strong agent model reading the tool result can't misread `waiting_period_days: 7` as a deduction). The math and the policy text now agree, and the goldens stay valid.

examples/skill_evolution_lab/run_e2e_demo.sh
evekhm: Fixed in 0eef3c8: `sed $'s/\033\[[0-9;]*m//g'` -- bash produces the ESC byte, so BSD sed never sees an escape sequence it can't interpret. Verified the strip works. Agreed on `agent_improvement_cycle/run_cycle.sh` -- leaving that for a follow-up since it's pre-existing outside this PR's paths.

examples/skill_evolution_lab/print_rate.py
evekhm: Fixed in 0eef3c8: out-of-scope is now counted by the `oos_` session-id prefix (same convention as `compare_runs.py`), and any in-scope session that fails the golden match prints as `unmatched`. This immediately made the evolve-set line honest: it now shows `8 out-of-scope; 34 unmatched` (the study-set questions deliberately differ from the held-out answer key) where it previously flattered as `42 out-of-scope`. The held-out test lines are unchanged (70/70 in-scope matched).

examples/skill_evolution_lab/run_sweep.sh
evekhm: Fixed in 0eef3c8: `set -euo pipefail` restored, with a comment explaining that the `if ./run_e2e_demo.sh` guard alone provides the continue-past-failure behavior, so mkdir/cd/manifest failures now abort instead of being silently tolerated.

scripts/skill_evolution.py
evekhm: Fixed in 0eef3c8: `evolve_skill()` now takes `version_label` (default `v1`) so a V1->V2 round writes `v2_*` artifacts without clobbering `v1_*`, and every run writes `<label>_selection.txt` recording the outcome -- "kept incumbent: none of the N viable candidate(s) beat the incumbent score by the 0.5 margin" when the guard fires, or which candidate won and why. Covered by a new unit test (28 passing).

examples/skill_evolution_lab/run_agent.py
evekhm: Fixed in 0eef3c8: the block now lives once in `_quiet.py` and both scripts start with `import _quiet` (before any google imports, since the muting has to precede SDK logger configuration and authlib's import-time `simplefilter("always")`).

examples/skill_evolution_lab/run_e2e_demo.sh
evekhm: Fixed in 0eef3c8: a `need_val()` helper turns a missing flag value into `ERROR: --agent-model requires a value (see --help)` with exit 1 (verified), plus an early check that `--with-registry` has `--skill-id`. `jq` is now listed in the README prerequisites as optional (only the golden-count banner uses it; the `?` fallback covers its absence).

examples/skill_evolution_lab/run_e2e_demo.sh
evekhm: Fixed in 0eef3c8 -- the stated safety property is now enforced in the recorded path. The registry mirror moved out of STEP 2 entirely: `compare_runs.py --gate` exits 3 when V1's overall rate is below V0's, and a new gated stage after STEP 4 pushes via `analyze_and_evolve.py --registry-push-only` only when V1 wins; otherwise it prints "registry push skipped" and the registry keeps its current revision. A losing V1 can no longer become a registry revision. README updated to match, and the companion blog post is being updated the same way.

scripts/quality_report.py
```diff
@@ -3888,6 +3889,7 @@ def _build_json_output(report, resolved_map, trajectories=None):
         "latency_s": ctx.get("latency_s"),
         "user_turns": ctx.get("user_turns", 0),
         "tool_calls": ctx.get("tool_calls", 0),
+        "tool_calls_detail": ctx.get("tool_calls_detail", []),
```
caohy1988: **[P2]** This `.get(..., [])` default inserts `tool_calls_detail: []` into **every** session -- including BigQuery-path sessions, whose ctx (built around line ~1040) populates `tool_calls` (int, from trace spans) but never `tool_calls_detail`. Downstream, `_format_tool_calls` uses key-absence as the backward-compat signal, so a BQ-scored session with `tool_calls: 3, tool_calls_detail: []` renders **"Tool calls: (none)"** to the analyst -- an actively false statement, worse than the legacy behavior of omitting the line. Same root cause makes `compare_runs.py`'s tool-selection table show `Called any tool 0/N` on BQ-scored reports.

Fix: only emit the key when ctx actually has it (`if "tool_calls_detail" in ctx: ...`), or default to `None` and skip -- so key-absence stays a truthful "not captured" signal for the BQ path.

scripts/skill_evolution.py
```diff
@@ -229,6 +252,28 @@ def _format_conversation(conversation) -> str:
...
+def _format_tool_calls(session: dict) -> str:
+  """Render the actual tool calls (name + args) for the analyst.
+
+  This makes *tool selection* visible -- e.g. a deflection that never called a
+  tool, versus one that called the wrong tool -- so the analyst can propose a
+  ``TOOL_USAGE`` / ``WRONG_TOOL`` rule from observed behavior, not just the
+  judge's grounding flag. Sessions scored before this field existed carry no
+  ``tool_calls_detail`` key and render nothing (backward compatible).
+  """
+  if "tool_calls_detail" not in session:
```
caohy1988, line 264: **[P2, same finding as quality_report.py:3892]** Belt-and-suspenders option on the consumer side: treat empty detail with a nonzero call count as *unknown*, not *none*:

```python
if "tool_calls_detail" not in session:
    return ""
detail = session.get("tool_calls_detail") or []
if not detail and session.get("tool_calls"):
    return ""  # calls happened but weren't captured (e.g. BQ path) -- say nothing, not "(none)"
```

Without this, tool-aware analysts on BQ-sourced reports get told the agent never called a tool when it did -- the exact failure mode (misinformed analysts) this PR's tool-awareness work exists to prevent.

examples/skill_evolution_lab/VERIFICATION.md
```diff
@@ -36,97 +39,118 @@ skill file changes** — so the delta is attributable to the skill.
...
+
+The flawed V0 barely calls the tool on **in-scope** questions (it's told not to), so it
+declines on almost everything; the evolved V1 uses the tool and answers correctly —
+including the multi-turn correction cases, where it re-verifies and holds the right figure
+instead of caving. Out-of-scope is the honest edge case: V0 already scores well there
+(deflecting *everything* happens to be the right call for a genuinely out-of-scope
+question), and V1 learns to decline those **deliberately** (routing IT questions to IT,
+refusing unrelated ones) rather than by reflex.
+
+## Across four models × 3 seeds (held-out, golden-graded)
```
caohy1988: **[P3]** `run_sweep.sh` was reworded to `RUNS` with an explicit "NOT RNG seeds" comment -- but this heading (and "3 seeds" in the reproduce section, and "the flash-lite seed above") still uses the exact word the sweep script now disclaims. Suggest "x 3 runs" / "independent repetitions" throughout this file for consistency.

examples/skill_evolution_lab/compare_runs.py
```diff
@@ -82,43 +144,118 @@ def _row(label, v0, v1):
...
+              "v0_dimensions": v0_dims,
+              "v1_dimensions": v1_dims,
+              "v0_tool_selection": v0_tools,
+              "v1_tool_selection": v1_tools,
+          },
+          f,
+          indent=2,
+      )
+
+  if args.gate and v1["overall"]["rate"] < v0["overall"]["rate"]:
```
caohy1988: **[P3]** Strict `<` means a **tie** passes the gate and mints a new immutable registry revision for zero measured gain, while the README says "only after it beats V0 ... a losing V1 is never pushed." Either make a tie a no-push (matches "beats") or reword the README to "matches or beats."

Related (one sentence in the README would cover it): the gate keys on the overall rate only, so a V1 that gains overall while regressing on corrections still ships -- fine for the demo, worth stating.

caohy1988 (commented):
## Fresh full review at `c7a090d`

Re-reviewed the whole PR at the latest head. The response to both prior reviews is thorough and well-executed -- I verified each fix in the code (registry gating wired correctly with a dedicated exit code + `pipefail` handling, BSD-sed fix, `set -euo pipefail` restored with the right POSIX reasoning, `need_val` guards, `_quiet.py` dedupe, honest oos/unmatched split in `print_rate.py`, waiting-period consistency, `version_label` + `selection.txt` artifacts -- and the switch from raw `cands` to sanitized `viable` quietly fixed a latent `_SELECTED`-tag bug on top). The structured `tool_calls_detail` pipeline and the generic engine CLI go beyond what was asked. **All 31 tests pass on this head** (ran them in a clean worktree).

Two new P2s (both small, both introduced by the new tool-detail plumbing / left behind by it), two P3s, two nits:

### P2

1. **BigQuery-scored reports will tell analysts "Tool calls: (none)" even when the agent made tool calls** -- see inline comments on `quality_report.py:3892` and `skill_evolution.py`. The `.get(..., [])` default defeats the key-absent backward-compat check for every session built by the BQ path (which populates `tool_calls` but never `tool_calls_detail`). Same root cause makes `compare_runs.py`'s tool-selection table print `Called any tool 0/N` on BQ-scored reports -- and this bites exactly where the README's new BQAA paragraph points.
2. **`sample_run/` predates the structured-tool-call feature.** The committed `v1_test_traffic.json` has zero `tool_calls_detail` fields and `RESULT.md` has no tool-selection table, while the README now advertises both. A reader who opens the recorded run to see the PR's headline addition finds a run that doesn't exhibit it. One fresh flash run refreshes it (and would also capture `waiting_period_note` and a `v1_selection.txt`).

### P3

3. **"seeds" vs "runs" now contradicts across docs** -- `run_sweep.sh` was carefully reworded to `RUNS` with an explicit "NOT RNG seeds" comment, but `VERIFICATION.md` still says "x 3 seeds" (section heading, reproduce section, "the flash-lite seed above"). Inline comment on the heading.
4. **Gate tie semantics** -- `--gate` fires only on strict `<`, so a tie mints a new registry revision while the README says "only after it beats V0 (a losing V1 is never pushed)". Inline comment.

### Nits

- `_write_evolution_artifacts` selection note indexes into `viable` ("candidate N of M"), but candidate files are numbered over `kept` (viable minus base-equal candidates) -- if a consolidation returns the base verbatim alongside real candidates, the note's index won't match the `_SELECTED` filename.
- The engine CLI (`skill_evolution.py`) silently falls back when `--eval-spec` has no `tools` field, while `analyze_and_evolve.py` fail-fasts on the same condition. If the asymmetry is deliberate (demo strict, library lenient), a one-line warning in the engine CLI would keep the fail-fast spirit.

### Bottom line

Approve-ready once the two P2s land: one conditional (emit `tool_calls_detail` only when actually captured, or treat empty-detail-with-nonzero-`tool_calls` as unknown) and one regenerated `sample_run`. The strategic items from the first review (a true `--from-bigquery` demo path, a V1->V2 round) are now honestly documented as production/future work rather than implied -- a reasonable scope cut for this PR.

scripts/skill_evolution.py
```diff
@@ -757,8 +959,45 @@ def _main():
...
+  tools = args.tools
+  if args.eval_spec:
+    if not os.path.exists(args.eval_spec):
+      ap.error(f"--eval-spec not found: {args.eval_spec}")
+    with open(args.eval_spec) as f:
+      tools = (json.load(f) or {}).get("tools") or tools
+  if args.eval_spec and not tools:
+    # Library CLI stays lenient (runs without tool awareness) but warns; the
+    # lab wrapper (analyze_and_evolve.py) fail-fasts on the same condition.
+    logging.warning(
```
caohy1988: **[P3]** (verified by execution) This `logging.warning(...)` fires **before** `logging.basicConfig(...)` below. Python's module-level `logging.warning` implicitly installs a root handler (level WARNING, default format) when none exists -- which turns the explicit `basicConfig(level=INFO, format=...)` into a silent no-op. Reproduced:

```python
import logging
logging.warning('early warn')          # implicit basicConfig: root at WARNING
logging.basicConfig(level=logging.INFO, format='...')  # no-op: handler exists
logging.getLogger('x').info('never prints')
# root level stays 30, default-format handler installed
```

So in exactly the case this warning targets (eval-spec without `tools`), the user gets the warning and then loses every INFO log for the entire evolution run -- "Evolving...", analyst progress, artifact writes. Self-defeating.

Fix: move this warning block below the `basicConfig(...)` call (it only depends on `tools`, which is already resolved by then).

caohy1988 (commented):
## Fresh review at `96e8e0b` -- final round

Verified the review-response commit fix by fix, and re-ran the suite at the true head: **32/32 tests pass**. (Process note for transparency: my first test run this round silently executed stale files -- I had force-moved the branch ref under my review worktree -- so I force-checked-out `96e8e0b` and re-verified before writing this.)

### Verified fixed

1. **BQ tool-detail truthfulness (P2)** -- fixed at every layer: `quality_report.py` emits `tool_calls_detail` only when actually captured (both build paths), `_format_tool_calls` guards the empty-detail-with-nonzero-count case, and the new test covers exactly the BQ-path scenario. Downstream ripple checked: `compare_runs.py:_tool_usage` keys off key-presence, so BQ-scored reports now correctly omit the tool-selection table instead of showing `Called any tool 0/N` -- the fix propagates with no further changes needed.
2. **Gate tie (P3)** -- `<=` so a zero-gain V1 never mints a registry revision; overall-rate-only limitation documented. Behavior now matches the README's "beats V0" exactly.
3. **seeds -> runs (P3)** -- all four spots in VERIFICATION.md.
4. **Engine CLI warning (nit)** -- added, with the demo-strict/library-lenient asymmetry documented. One placement bug, see inline.

### Remaining

- **[P2, carried over] `sample_run/` still predates the structured-tool-call feature** -- no `tool_calls_detail` in the committed traffic, no tool-selection table in `RESULT.md`, while the README advertises both. One ~15-minute flash run regenerates it (and would also capture `waiting_period_note` + a `v1_selection.txt`, giving the recorded run receipts for all of this PR's additions at once).
- **[P3, new] warning-before-`basicConfig` silences the run's own INFO logging** -- see inline on `skill_evolution.py`. Verified by execution, one-line fix.

### Verdict

Two items from approval: the `sample_run/` refresh and the one-line warning move. Everything else across four review rounds has come back fixed correctly and with test coverage -- once that final push lands, this has my approval on sight.

evekhm:
Pushed `a915614` (4 commits + a merge recording that `57348ff` is superseded). This lands the remaining review items and converges the two parallel reframes -- **flash-lite headline** on a **two-defect V0**:

- **Second planted defect**: V0 now also says *"be agreeable: accept the user's figure"* -- so the anti-parroting machinery runs on real failures. Recorded run (`sample_run/`, gemini-3.1-flash-lite, the new default): **V0 37.5% -> V1 97.5%** overall (+60pp), single-turn 36.4% -> 100% (55/55), **corrections 0% -> 100% with parroted sub-trajectories 11 -> 0** (including parroting against its own baked PTO-rollover fact). Prevalence: TOOL_USAGE 42/48 + PARROTING 4/48 -- both defects found independently.
- **`--rounds 2`** (AutoSkill's accumulative round, gated): recorded at `sample_run/round2/` -- V2 *tied* V1 (97.5%) and the gate refused it; `v2_selection.txt` + `RESULT_ROUND2.md` are the audit trail.
- **`--from-bigquery`**: `run_agent.py --log-bigquery` writes plugin-schema `agent_events` rows (full producer contract incl. `content_parts`), scoring reads them back via the SDK's BigQuery trace path; `quality_report.py` now builds `tool_calls_detail` from `TOOL_*` spans so the BQ path has scorecard parity. Smoke-verified: identical verdicts, parroting tags, and per-tool detail vs the local path.
- **Fixes**: the `logging.warning`-before-`basicConfig` mute (moved below `basicConfig`), the macOS ANSI-strip in `agent_improvement_cycle/run_cycle.sh`, aggregate_sweep "seeds" -> "runs".
- **Fresh 4x3 sweep** on the two-defect V0: V1 96-100% in-scope on every model; all 12 runs score 0% corrections at V0 (universal parroting), zero parrots at every V1. VERIFICATION.md documents the flash baseline drift (June 18% -> July ~80% single-turn) and why flash-lite is featured.

32 unit tests pass; the one failing `test_ai_ml_integration` case and the `test_extractor_compilation` failures are pre-existing on the branch base (reproduced on a clean checkout).

haiyuan-eng-google (approved):

haiyuan-eng-google:
https://gist.github.com/evekhm/a2be8f89efa3bc3eeef9bac879edf9be

Issue #316, opened 2026-06-24
Claude Code → BigQuery: OTel-native OTLP receiver endpoint


haiyuan-eng-google:
# Claude Code → BigQuery: OTel-native OTLP receiver

A new OTLP ingest surface (Cloud Run → Pub/Sub → BigQuery Storage Write API, Terraform-deployed into the customer project) that lands Claude Code's OpenTelemetry logs/metrics/(gated) traces into **OTel-native BigQuery tables as the source of truth**, with BQAA `agent_events` delivered as a projection layer on top.

📄 **Full design spec:** [`docs/otlp_receiver_design.md`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/docs/otlp_receiver_design.md)

Key decisions (detail in the doc):
- ClickHouse-derived native `otel_*` tables, BigQuery-adapted and versioned (`otel_schema_version`): `otel_logs`, five per-type `otel_metric_*` tables, gated `otel_spans`, `otlp_dead_letter`.
- Native JSON attribute bags + promoted scalar columns for partition/cluster (can't cluster on JSON).
- Append-only writer + per-signal idempotency with a **required, replay-invariant `source_position`**; read-time dedup.
- BQAA `agent_events_otlp` projection as a scheduled-`MERGE` curated table (materialized views can't express the window-function dedup) + schema-parity test.
- Preserved metric temporality; two-mode raw bodies; auth-dependent identity; two-tier trace gating.

---

## Scope / notes

- Standalone delivery; no dependency on Anthropic-side changes.
- #317 reuses this same OTel-native receiver and Pub/Sub envelope for Codex.
- BQAA compatibility is delivered through views/projections so existing SDK workflows keep working without making `agent_events` the ingest schema.
- Relates to distribution work in #251.

## Acceptance criteria

- [ ] New OTLP receiver accepts OTLP/gRPC + OTLP/HTTP protobuf for logs and metrics on `/v1/logs` + `/v1/metrics`; traces gated behind a flag (HTTP+JSON out of scope).
- [ ] **Auth provisioning**: bearer/OTLP headers via Secret Manager + Cloud Run injection + rotation (mTLS optional); unauthenticated requests rejected.
[...]

PR #323, opened 2026-06-29
feat(otlp): native OTel BigQuery schema package + agent_events_otlp parity (#316 PR1)

caohy1988:
First implementation increment of the **OTel-native OTLP receiver** (#316) — PR 1 of the receiver PR plan: the BigQuery **schema package** that lands raw OpenTelemetry as the source of truth, with BQAA `agent_events` as a projection on top.

Design: [`docs/otlp_receiver_design.md`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/docs/otlp_receiver_design.md).

### What's here
- **`otlp/schema.py`** — ClickHouse-derived, BigQuery-adapted, versioned (`OTEL_SCHEMA_VERSION`) native tables: `otel_logs`, five per-type `otel_metric_*` tables (sum/gauge/histogram/exponential_histogram/summary), gated `otel_spans`, `otlp_dead_letter`. Open attribute bags are native `JSON`; low-cardinality identity is promoted to top-level columns for partition/cluster keys; shared receiver metadata includes the required, replay-invariant `source_position`; exemplars carry `span_id`/`trace_id` (the metric→trace bridge). A `NATIVE_TABLES` registry carries partition/cluster specs.
- **`otlp/projection.py`** — the `agent_events_otlp` output-column contract, declared **independently** of canonical `agent_events` so that adding a column there fails the schema-parity test until the projection is updated; plus read-time dedup view SQL (plain views — BigQuery materialized views can't express `QUALIFY ROW_NUMBER()`).
- **`test_otlp_schema.py`** — 17 tests: per-type metric tails vs the OTLP proto, exemplars, `source_position`, clustering-key validity (no JSON/repeated keys), dedup views, and the schema-parity contract.

### Scope notes
- **Schema only** — no decode, no receiver service, no writer yet (those are PR 2–5). Follows the existing `bq_schema(bigquery)` pattern: the `google.cloud.bigquery` module is injected, so unit tests use a fake and need no live BigQuery.
- New module lives under `producers/src/bigquery_agent_analytics_tracing/otlp/`.

### Verification
- New file: 17 passed. Full producers suite: **120 passed**, no regressions.
- `isort --check` and `pyink --check` clean (as `producers-ci.yml` runs them).


caohy1988:
Thanks — all three addressed in `6b8fca5`:

- **P1 (dedup identity nullable):** `idempotency_key`, top-level `source_position`, `ingest_time`, `otel_schema_version` (plus `source_product`/`source_signal`) are now `REQUIRED` on the native **analytics** tables via `_receiver_metadata`, so a null key can't collapse unrelated rows in the `*_dedup` views. `otlp_dead_letter` deliberately keeps them `NULLABLE` (whole-request failures may lack a full `source_position`) — `_source_position(..., mode=...)` carries the distinction. New tests assert both.
- **P2 (nested parity):** the parity check is now recursive — `_flatten()` emits `path:type:mode` signatures through nested records, so drift inside `content_parts` / `object_ref` is caught. Added `test_parity_fails_when_a_nested_agent_events_field_is_unmatched` that drops `object_ref` and asserts the check reports it.
- **P3 (dedup ordering):** switched to `ORDER BY ingest_time DESC` to match the committed design (newest-write-wins, so a replayed/repaired row supersedes the older one); test updated.

Producers suite: 124 passed; pyink + isort clean.

haiyuan-eng-google (approved):

PR #325, opened 2026-06-29
feat(otlp): OTLP decode + envelope library (#316 PR2)

caohy1988:
Second increment of the **OTel-native OTLP receiver** (#316) — PR 2 of the receiver PR plan: the pure **decode + envelope** library. Builds directly on PR 1's schema package (#323).

Design: [`docs/otlp_receiver_design.md`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/docs/otlp_receiver_design.md) §4.

### What's here
- **`otlp/envelope.py`** — `SourcePosition`; `canonical_json`; SHA256 **per-signal idempotency keys** (logs = content + `source_position`; metrics = value + `source_position`; spans = natural `trace_id||span_id`), with `ingest_time` deliberately **excluded** so retry/replay reproduces the key; OTLP `AnyValue`/attribute decoding; raw-body preservation policy; `make_envelope` / `dead_letter_envelope` (envelope v1).
- **`otlp/decode.py`** — `decode_logs_request` / `decode_metrics_request` walk the decoded OTLP JSON tree, compute the `source_position` indices, handle **all five metric point types** (sum/gauge/histogram/exponential_histogram/summary), and route malformed records to **dead-letter envelopes** without dropping the rest of the batch.

### Scope notes
- **Pure logic** — operates on the *decoded* OTLP JSON structure (camelCase dicts, what `protobuf → dict` yields at the receiver edge). No network, BigQuery, or protobuf dependency, so it's fully unit-testable. The protobuf wire-decode + transport are PR 3 (receiver service); the BigQuery writer is PR 4.
- Spans are intentionally out of scope here (trace-gated).

### Verification
- New file: 17 passed (incl. all five metric types parametrized). Full producers suite: **141 passed**, no regressions.
- `isort --check` and `pyink --check` clean (as `producers-ci.yml` runs them).

Key properties covered by tests: idempotency keys are **stable across decodes** and **invariant to `ingest_time`** (replay collapses), two **identical records in one request get distinct keys** via `source_position` (no false collapse in the dedup views), and **one bad record doesn't drop the batch**.

### DIFF
--- a/producers/src/bigquery_agent_analytics_tracing/otlp/__init__.py
+++ b/producers/src/bigquery_agent_analytics_tracing/otlp/__init__.py
@@ -12,15 +12,32 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""OTel-native OTLP receiver storage layer for BQAA (issue #316).
+"""OTel-native OTLP receiver storage + decode layer for BQAA (issue #316).
 
-PR 1 of the receiver: the BigQuery schema package — native ``otel_*`` tables as
-source of truth, BQAA ``agent_events_otlp`` projection contract, and read-time
-dedup views. See ``docs/otlp_receiver_design.md``.
+- PR 1: BigQuery schema package — native ``otel_*`` tables, ``agent_events_otlp``
+  projection contract, read-time dedup views (``schema`` / ``projection``).
+- PR 2: OTLP decode + envelope library — decoded OTLP logs/metrics to envelope
+  v1, ``source_position``, per-signal idempotency, dead-letter envelopes
+  (``envelope`` / ``decode``).
+
+See ``docs/otlp_receiver_design.md``.
 """
 
 from __future__ import annotations
 
+from bigquery_agent_analytics_tracing.otlp.decode import decode_logs_request
+from bigquery_agent_analytics_tracing.otlp.decode import decode_metrics_request
+from bigquery_agent_analytics_tracing.otlp.envelope import canonical_json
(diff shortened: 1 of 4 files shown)


caohy1988:
Fixed in `1be2392` — the dead-letter path now meets the #316 replay/keying contract.

**Replayable `raw_b64`:** the decode layer now takes the original wire payload (`raw_request: bytes`) instead of a pre-computed hash. Dead-letter `raw_b64` is that **original OTLP request** (a valid `Export*ServiceRequest` a replay worker can republish to `/v1/logs` / `/v1/metrics`), and `raw_otlp_request_hash` is computed from it — so it's reproducible, not a re-serialized subrecord. (Tested: `base64.b64decode(raw_b64) == original request bytes`, and the hash matches.)

**Keyed dead-letters:** new `dead_letter_key(stage, source_position=None, raw_otlp_request_hash=None)` keys from `source_position` where decode got that far, else from `raw_otlp_request_hash + stage` for whole-request failures (auth / undecodable request). `dead_letter_envelope` no longer hardcodes `idempotency_key=None`, so DLQ/replay rows dedupe. (Tested: deterministic key from `source_position`; the request-hash+stage fallback varies by stage.)

API change: `decode_{logs,metrics}_request` now accept `raw_request` rather than `raw_request_hash` (computed internally).

Producers suite: 144 passed; pyink + isort clean.

haiyuan-eng-google (approved):

Issue #324, opened 2026-06-29
Automate enterprise admin setup for Claude Code/Codex OTel → BigQuery


caohy1988:
## Summary

Create a guided enterprise admin experience for enabling Claude Code and Codex OpenTelemetry export into BigQuery Agent Analytics.

This issue should own the "make it easy for an enterprise admin" layer above the OTel-native receiver work:

- #316 provides the Claude Code -> BigQuery OTel-native receiver contract.
- #317 provides the Codex -> BigQuery OTel-native reuse/config contract.

The goal is not to replace those implementation issues. The goal is to automate the deployment and configuration path so an admin can stand up the receiver, BigQuery schema, views, auth, DLQ, and Claude Code/Codex telemetry settings without hand-assembling the pieces.

## Target admin experience

Provide a CLI flow along these lines:

```bash
bqaa otel bootstrap \
  --project my-gcp-project \
  --dataset agent_analytics \
  --region us-central1 \
  --source claude-code,codex \
  --signals logs,metrics \
  --privacy baseline \
  --auth bearer-token
```

The command should:

1. Create or validate the BigQuery dataset.
2. Create native OTel tables:
   - `otel_logs`
   - `otel_metric_sum`
   - `otel_metric_gauge`
   - `otel_metric_histogram`
   - `otel_metric_exponential_histogram`
   - `otel_metric_summary`
   - optional/gated `otel_spans`
   - `otlp_dead_letter`
3. Create dedup views and the BQAA projection surface.
4. Deploy or configure the authenticated OTLP receiver endpoint.
5. Configure DLQ/dead-letter handling for malformed OTLP.
6. Generate Claude Code managed-settings JSON.
7. Generate Codex config templates once the #317 config contract is verified.
8. Run an end-to-end smoke test that exports synthetic logs/metrics and verifies rows in BigQuery.
9. Print the exact next admin action:
   - paste JSON into Claude Code server-managed settings, or
   - deploy endpoint-managed settings through MDM/OS policy, or
   - apply generated local/test config for a pilot.

## Telemetry signal tiers

The bootstrap flow should separate **which telemetry signals are exported** from **how much sensitive content is included**.

### `logs,metrics` default

Default enterprise signal tier. Creates/validates `otel_logs` plus all five `otel_metric_*` tables and generates logs+metrics config for Claude Code and Codex.

### `logs,metrics,traces`

[...]

PR #326, opened 2026-06-30
feat(otlp): receiver service — auth, decode dispatch, Pub/Sub+DLQ routing (#316 PR3)

caohy1988:
Third increment of the **OTel-native OTLP receiver** (#316) — PR 3: the request-handling **service**, tying PR 1 (schema) and PR 2 (decode/envelope) to Pub/Sub. Builds on #323 + #325.

Design: [`docs/otlp_receiver_design.md`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/main/docs/otlp_receiver_design.md).

### What's here
- **`otlp/receiver.py`** — transport-agnostic core. Constant-time bearer auth; `decode_body` dispatch (OTLP/HTTP **protobuf** via lazy `opentelemetry-proto`, or JSON); `handle_export` authenticates → decodes → routes envelopes to the **main** topic (success) or **DLQ** (dead letters). The original request **bytes** are threaded into the PR-2 decode layer, so dead letters stay replayable + keyed; whole-request decode failures dead-letter with `raw_otlp_request_hash` always set. `/v1/traces` is **signal-tier gated** (#324): `404` when disabled, `501` (wired, landing deferred) when enabled.
- **`otlp/app.py`** — WSGI entrypoint for Cloud Run (`gunicorn --factory bigquery_agent_analytics_tracing.otlp.app:make_app`); env-driven config + lazy `google-cloud-pubsub` publisher. All logic stays in `receiver.py`.
- **`otlp/envelope.py`** — **PR #325 review hardening**: `dead_letter_key` now raises `ValueError` when neither `source_position` nor `raw_otlp_request_hash` is provided (a stage-only key would collapse unrelated whole-request failures). The receiver always passes the request hash.
- **`pyproject.toml`** — new optional `receiver` extra (`google-cloud-pubsub`, `opentelemetry-proto`, `gunicorn`).

### Scope notes
- **Deployable but locally tested.** The core + all 15 tests need none of the optional deps — they use a fake publisher and JSON fixtures. The real Pub/Sub client and protobuf decode are lazy-imported behind the `receiver` extra. Terraform/Cloud Run wiring + live e2e is PR 5.
- Auth failures return `401` (not dead-lettered); malformed-but-authenticated requests dead-letter.

### Verification
- New file: 15 passed (auth accept/reject, logs/metrics routing, per-record DLQ split, undecodable-body DLQ that is keyed + replayable, traces gating, the key guard, and the WSGI app end-to-end). Full producers suite: **159 passed**, no regressions.
- `isort --check` + `pyink --check` clean (as `producers-ci.yml` runs them).

### DIFF
--- a/producers/pyproject.toml
+++ b/producers/pyproject.toml
@@ -53,6 +53,11 @@ openai-agents = [
 claude-sdk = [
   "claude-agent-sdk>=0.1.0",
 ]
+receiver = [
+  "google-cloud-pubsub>=2.18.0",
+  "opentelemetry-proto>=1.20.0",
+  "gunicorn>=21.0",
+]
 dev = [
   "pytest>=7.0",
   "pytest-asyncio>=0.21",
--- a/producers/src/bigquery_agent_analytics_tracing/otlp/__init__.py
+++ b/producers/src/bigquery_agent_analytics_tracing/otlp/__init__.py
@@ -19,6 +19,8 @@
 - PR 2: OTLP decode + envelope library — decoded OTLP logs/metrics to envelope
   v1, ``source_position``, per-signal idempotency, dead-letter envelopes
   (``envelope`` / ``decode``).
+- PR 3: receiver request handling — auth, decode dispatch, Pub/Sub + DLQ
+  routing (``receiver``); WSGI entrypoint in ``app``.
 
 See ``docs/otlp_receiver_design.md``.
 """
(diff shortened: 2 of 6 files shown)


caohy1988:
Both fixed in `bd51b5a`.

**P1 — malformed decoded body now dead-letters (not 500).** The receiver's guard now wraps the **decode-library call** too, not just `decode_body`. So a valid-JSON-but-wrong-OTLP-shape request (e.g. `{"resourceLogs":[null]}`, which raised `AttributeError` inside `decode.py`) becomes a keyed, replayable whole-request dead letter (`400`), satisfying the malformed-but-authenticated contract. Test: `test_valid_json_with_wrong_otlp_shape_is_dead_lettered_not_500` (asserts 400, DLQ, keyed, `raw_b64` round-trips to the body).

**P2 — auth before the trace gate.** `handle_export` now authenticates every known OTLP endpoint **before** the trace gate, so unauthenticated callers always get `401` and the `404`-vs-`501` gate state never leaks. Test: `test_traces_requires_auth_before_revealing_gate_state` (401 for both enabled/disabled with a bad token, nothing published).

Producers suite: 161 passed; pyink + isort clean.

haiyuan-eng-google (approved):

PR #331, opened 2026-07-01
feat(otlp): bqaa-otel bootstrap — plan/execute deploy orchestration (#324 PR2)

caohy1988:
**Rebased on main** (#330 merged); the diff is now this PR only.

Second increment of [#324](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/324): `bqaa-otel bootstrap` absorbs `deploy/otlp_receiver/setup.sh` so the deploy sequence has one source of truth. `setup.sh` becomes a thin wrapper preserving the historical env-var interface (`PROJECT=... bash deploy/otlp_receiver/setup.sh`).

## Plan mode is the default

```bash
bqaa-otel bootstrap --project my-proj                # prints every command, runs nothing
bqaa-otel bootstrap --project my-proj --execute      # applies
```

The plan lists the exact gcloud/bq commands with `<placeholders>` for run-time captures (receiver URL, project number) and summarizes stdin payloads (DDL shows a line count; the generated token is never displayed).

## What execute does (setup.sh parity + more)

Same sequence as the shell script — APIs, Artifact Registry, dataset + native-schema DDL, bearer-token secret, least-privilege SAs, topics + IAM, Cloud Build image, receiver/consumer Cloud Run deploys, OIDC push subscription with DLQ, scheduled MERGE — with these changes:

- DDL and MERGE SQL come **straight from the `ddl`/`sql` modules** (no `gen_schema_sql.py` subprocess; that script remains as a standalone utility).
- After deploy it renders the **PR1 config artifacts against the real receiver URL** (Claude managed-settings JSON, Codex TOML, MDM guidance) and prints the exact next admin action + smoke-test command.
- The bearer token is **never embedded** in generated artifacts; the summary prints the `gcloud secrets versions access` command instead.
- `--signals logs,metrics,traces` creates `otel_spans` DDL and sets `BQAA_OTLP_ENABLE_TRACES=1` on both services (span *landing* remains PR4 of the stack).
- Tier validation (privacy/signals/replay ack) is **reused from PR1's `BootstrapSpec`**, so the `config` and `bootstrap` gates cannot drift.

## Testing

- 15 new unit tests: the deploy sequence via a recording `FakeRunner` (command families, DDL content + spans gating, secret idempotency, DLQ/OIDC subscription shape, scheduled-MERGE registration, artifact endpoint/token assertions, summary), plan-mode rendering, CLI plan-vs-execute dispatch and the replay ack gate.
- Full producers suite: 240 passed, 3 skipped; pyink + isort clean; `bash -n setup.sh` clean.
- Drove the real plan mode: byte-for-byte the setup.sh command sequence, token stdin summarized not shown.

### DIFF
--- a/deploy/otlp_receiver/README.md
+++ b/deploy/otlp_receiver/README.md
@@ -13,15 +13,28 @@ Claude Code / Codex --OTLP--> Cloud Run receiver --> Pub/Sub --> consumer --> Bi
 ## Deploy
 
 ```bash
-PROJECT=my-proj DATASET=agent_analytics REGION=us-central1 \
-  bash deploy/otlp_receiver/setup.sh
+# Print the plan (runs nothing):
+PYTHONPATH=producers/src python3 -m bigquery_agent_analytics_tracing.otlp.cli \
+  bootstrap --project my-proj --dataset agent_analytics --region us-central1
+
+# Apply it:
+PYTHONPATH=producers/src python3 -m bigquery_agent_analytics_tracing.otlp.cli \
+  bootstrap --project my-proj --dataset agent_analytics --region us-central1 --execute
 ```
 
+(`bqaa-otel bootstrap ...` once the producers package is installed;
+`setup.sh` is a thin wrapper over the same command with the historical
+env-var interface: `PROJECT=my-proj bash deploy/otlp_receiver/setup.sh`.)
+
 This creates: the native tables + `*_dedup` views + `agent_events_otlp` +
-`bqaa_metrics` (DDL generated from the schema package via `gen_schema_sql.py`),
-Pub/Sub topics/subscription + DLQ, a Secret Manager bearer token, the Cloud Run
-receiver + consumer, and the scheduled `MERGE`. It prints the receiver URL and
-how to read the token.
+`bqaa_metrics` (DDL generated from the schema package; `gen_schema_sql.py`
+prints the same bundle standalone), Pub/Sub topics/subscription + DLQ, a
+Secret Manager bearer token, the Cloud Run receiver + consumer, and the
+scheduled `MERGE`. It prints the receiver URL, how to read the token, and
+writes the ready-to-distribute Claude Code / Codex config artifacts
+(`--source claude-code,codex --signals ... --privacy ...` — see
+`bqaa-otel config --help` for generating artifacts against an existing
+deployment).
 
 - **Endpoints:** `<url>/v1/logs`, `<url>/v1/metrics` (`/v1/traces` is gated —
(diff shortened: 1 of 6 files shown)


caohy1988:
Ran a full multi-agent review of this PR (testing / maintainability / security specialists + adversarial + red-team passes) and pushed a hardening commit (f121e23) addressing the merge-blocking findings:

**P1 fixes in f121e23:**
- **Visible failures**: `--execute` now catches `CalledProcessError` and prints the failing command + its captured stderr (previously the real gcloud error was invisible inside `e.stderr` behind a bare traceback).
- **No interactive deadlocks**: subprocesses get `stdin=DEVNULL`, gcloud gets `--quiet`, bq gets `--headless` — a prompting CLI fails fast and visibly instead of hanging forever with its prompt swallowed by `capture_output`.
- **Repo-root guard before any mutation**: the Cloud Build step uploads the *current directory*; a pip-installed `bqaa-otel bootstrap --execute` run from `$HOME` would have enabled APIs, created infra, then tarred the home directory into the `_cloudbuild` bucket.
- **No duplicate scheduled MERGEs**: the transfer-config listing is a hard `run`, not `try_run` — DTS display names aren't unique, so treating a transient listing failure as "no config" would mint a duplicate MERGE job on every flaky re-run.
- **Scheduled MERGE pinned to the consumer SA** (`--service_account_name`): without it, DTS runs the projection on the invoking admin's personal OAuth and silently dies when they lose access or leave. (Admin needs `iam.serviceAccounts.actAs` on the consumer SA.)
- **DLQ retention subscription** (`bqaa-otlp-dlq-sub`, 7d, never expires): Pub/Sub discards messages published to a topic with no subscription, so dead letters were unrecoverable — the transport DLQ was decorative. (Latent in the old setup.sh.)

Also: setup.sh now *rejects* the dropped pre-#324 overrides (`SECRET`, `MAIN_TOPIC`, …) instead of silently ignoring them and deploying duplicate parallel infra; plan output is shell-quoted (copy-paste-safe) with truncation marked; the subscription create/update paths share one flag list so they structurally can't drift; empty `--source` rejected; dead code removed; 12 new tests (281 passing at the stack tip).

**Deferred to a follow-up (design calls, not regressions):** token rotation semantics (secret pinned to `:latest` resolves per instance start — no overlap window), dataset-level vs project-level `dataEditor` for the consumer SA, re-run semantics when `--region`/`--bq-location` change, `--update-env-vars` vs `--set-env-vars` ownership, first-run IAM/API propagation retries, and the PR1 `config --token`-on-argv exposure.

haiyuan-eng-google (approved):

PR #332, opened 2026-07-01
feat(otlp): bqaa-otel verify + smoke — deployment health checks (#324 PR3)

caohy1988:
**Rebased on main** (#330, #331 merged); the diff is now this PR only.

Third increment of [#324](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/324): the `bqaa otel verify` acceptance criterion.

## verify (read-only)

```bash
BQAA_OTLP_TOKEN=<token> bqaa-otel verify --endpoint <url> --project <proj> --dataset <ds> [--smoke]
```

Checks: unauthenticated POST must get **401** (auth enforcement — a 200 fails the check as an open endpoint), authenticated empty POST must get 200 (reachability), all expected tables/views exist via `INFORMATION_SCHEMA.TABLES` (`otel_spans` only required when `--signals` includes traces), recent-row freshness, and dead-letter counts. Zero recent rows and recent dead letters are **warnings** — a fresh deployment passes. Exit 1 only on hard failures.

## --smoke (write path)

Sends synthetic OTLP logs + a gauge (payload builders **shared with `test_otlp_e2e.py`**, which now imports them — the CLI smoke and the live e2e exercise identical shapes), waits for rows in `otel_logs`/`otel_metric_gauge`/`bqaa_metrics`, then runs the scheduled MERGE once and asserts the event projects into `agent_events_otlp` as `claude_code.user_prompt`. The traces tier reports span landing as pending (PR4 of the stack).

## Design

I/O is injected (`http_post`, `query_rows`) so all checks are unit-testable without GCP; the CLI wires urllib + google-cloud-bigquery defaults lazily. `run_verify` is provably read-only (test asserts no MERGE/INSERT queries and no synthetic events posted).

## Testing

- 15 new unit tests across `test_otlp_verify.py` and `test_otlp_cli.py` (auth-not-enforced detection, missing-table detection, warning semantics, read-only guarantee, smoke landing/projection/timeout, traces gating, CLI exit codes, BQAA_OTLP_TOKEN fallback).
- Full producers suite: 255 passed, 3 skipped; pyink + isort clean.

### DIFF
--- a/deploy/otlp_receiver/README.md
+++ b/deploy/otlp_receiver/README.md
@@ -108,11 +108,21 @@ traces (`otel.trace_exporter`) are observability only.
 ## Verify
 
 ```bash
+# Read-only health checks: endpoint reachability + auth enforcement,
+# table/view existence, recent rows, dead-letter health.
+BQAA_OTLP_TOKEN=<token> PYTHONPATH=producers/src python3 -m \
+  bigquery_agent_analytics_tracing.otlp.cli verify \
+  --endpoint <url> --project <proj> --dataset <dataset>
+
+# Add --smoke to also send synthetic OTLP logs+metrics and follow them into
+# the native tables, dedup views, and the agent_events_otlp projection.
+```
+
+The full pytest e2e (same payloads as `--smoke`, plus protobuf-path and
+dead-letter round-trips) remains available:
+
+```bash
 BQAA_OTLP_ENDPOINT=<url> BQAA_OTLP_TOKEN=<token> \
   BQAA_PROJECT=<proj> BQAA_DATASET=<dataset> \
   python -m pytest producers/tests/test_otlp_e2e.py -v
 ```
-
-The smoke test sends OTLP logs + metrics, confirms rows in the native tables,
-`*_dedup` views, `agent_events_otlp`, and `bqaa_metrics`, and that a malformed
-request lands in `otlp_dead_letter` with a replayable `raw_b64`.
--- a/producers/src/bigquery_agent_analytics_tracing/otlp/cli.py
+++ b/producers/src/bigquery_agent_analytics_tracing/otlp/cli.py
@@ -370,9 +370,18 @@ def _cmd_verify(args: argparse.Namespace) -> int:
     return 2
   # The bearer token rides on every request: plain http to a remote host
   # would send it cleartext. Loopback stays allowed for local harnesses.
-  # Schemes are case-insensitive (HTTP:// must not bypass the guard).
-  split = urllib.parse.urlsplit(args.endpoint)
-  host = split.hostname or ""
+  # Schemes are case-insensitive (HTTP:// must not bypass the guard), and
+  # urlsplit itself raises on malformed bracketed IPv6 ('http://[::1').
+  try:
+    split = urllib.parse.urlsplit(args.endpoint)
+    host = split.hostname or ""
+  except ValueError as exc:
+    print(
+        f"bqaa-otel: error: --endpoint {args.endpoint!r} is not a valid"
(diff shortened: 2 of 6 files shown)


caohy1988:
Ran the same multi-agent review on this PR (testing + maintainability/security specialists + adversarial pass) and pushed a hardening commit (d949ed8). The headline finding: **two verified schema mismatches would have crashed `bqaa-otel verify` against every healthy deployment** — the unit-test fake matched table names but never column names.

**P1 fixes in d949ed8:**
- **Real timestamp columns**: the `bqaa_metrics` VIEW has no `ingest_time` (it projects `time_timestamp`) and `otlp_dead_letter` keys on `received_at` — both freshness queries got "Unrecognized name" from BigQuery and verify died with a traceback precisely when the deployment was correct. Now driven by a per-table column map.
- **Graceful transport failures**: `default_http_post` catches `URLError`/`OSError` (DNS, connection refused, TLS, timeout) and reports status 0 — previously the one thing a reachability check must handle (endpoint down) crashed the tool. Verified live against a refused loopback port.
- **No redirects**: urllib would have followed a 3xx, resending the Authorization header to the redirect target and converting the POST to a GET that dead-letters on the receiver (the "read-only" verify polluting `otlp_dead_letter`). Redirects now surface as their status.
- **Guarded BigQuery calls**: a missing dataset / permission error on `INFORMATION_SCHEMA` — the very condition the existence check should report — plus failing count queries and MERGE/projection errors all become FAIL rows instead of tracebacks.
- **Dead-letter-aware smoke**: the receiver returns 200 with `dead_lettered>0` for per-record decode failures; the smoke now parses the response counts and fails immediately with a diagnostic instead of burning the full polling budget.
- **One timeout budget**: `--timeout` now bounds the whole wait phase (previously each landing check got its own budget: `--timeout 150` could wait 450s).
- **Tier-validated `--signals`** (a `trace` typo silently dropped the `otel_spans` expectations) and **https enforcement** for non-loopback endpoints (cleartext bearer token).

17 new/updated tests; 295 passing on this head, 304 at the stack tip (#333 rebased on top, auto-merge combined its traces smoke with these guards cleanly).

**Noted, not changed (design/wording):** the authenticated empty-POST probe is decode-only — it proves auth+decode, not the Pub/Sub publish path; the check detail now says so and points at `--smoke`. The inline smoke MERGE racing the scheduled DTS run is data-safe (idempotency-key upsert); a serialization error surfaces as a failed check.

caohy1988:
**Findings**

[P1] producers/src/bigquery_agent_analytics_tracing/otlp/cli.py:374 - The cleartext-token guard is case-sensitive. `HTTP://receiver.example.com` bypasses `args.endpoint.startswith("http://")`, so the CLI proceeds and would send `Authorization: Bearer ...` over plaintext. Fix by parsing once with `urlsplit`, lowercasing `scheme`, and checking `scheme == "http"` against the loopback allowlist. Please add a mixed-case HTTP regression test.

[P1] producers/src/bigquery_agent_analytics_tracing/otlp/verify.py:401 - `run_smoke()` still lets BigQuery landing-query exceptions escape from `_wait_count()`. If a table is missing, permissions are wrong, or BigQuery has a transient read failure, `bqaa-otel verify --smoke` crashes with a traceback instead of returning a failed `CheckResult`. This also hides earlier read-only verify results because the CLI runs smoke afterward. Wrap the landing count polling per check and report `query failed: ...`.

Verification: reviewed head `d949ed8`. GitHub checks are green. Locally ran `uv run pytest` in `producers` (`295 passed, 3 skipped`), targeted verify/CLI tests (`48 passed`), `pyink`, `isort`, and `git diff --check`. I reproduced both issues on the PR head.

haiyuan-eng-google (approved):

PR #333, opened 2026-07-02
feat(otlp): traces tier — span decode, otel_spans landing, traces smoke (#324 PR4)

caohy1988:
**Rebased on main** (#330, #331, #332 merged); the diff is now this PR only. The traces smoke inherits all of #332's verify hardening (guarded queries, dead-letter-aware sends, shared timeout budget).

Final increment of [#324](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/324): the `logs,metrics,traces` signal tier becomes real end to end. Before this PR the receiver only opened the `/v1/traces` route (accept-but-501) — there was no span decode, no row mapping, and `target_table` raised for span envelopes.

## What this adds

- **decode**: `decode_traces_request` — `ExportTraceServiceRequest` → span envelopes with the natural `trace_id+span_id` idempotency key (replay-invariant by construction). A span missing that identity dead-letters individually, same per-record semantics as logs/metrics.
- **receiver**: `/v1/traces` decodes and publishes span envelopes when the tier is enabled; the protobuf path wires `ExportTraceServiceRequest`. Auth-before-gate and 404-when-disabled behavior unchanged (tests kept).
- **writer**: `span_row` maps the full `otel_spans` schema — start/end timestamps, identity (trace/span/parent), `span_name`/`span_kind`, `service_name`, status, resource/scope/span attributes as JSON, repeated `events`/`links`, receiver metadata. The consumer-side `enable_spans` gate still drops spans when the tier is off.
- **verify**: `bqaa-otel verify --smoke` on a traces-tier deployment now sends a synthetic span and verifies `otel_spans` rows — replacing PR3's pending warning. `synthetic_span_payload` joins the shared builders.
- **e2e**: skip-guarded live traces test (`BQAA_OTLP_ENABLE_TRACES=1`) covering `otel_spans` and its dedup view.

With this, every #324 acceptance criterion that was gated on span landing is implementable: the traces smoke test exists at both the CLI (`verify --smoke`) and pytest e2e levels.

## Testing

- TDD throughout: 13 new/updated unit tests across decode (identity key, replay invariance, dead-letter, positions), writer (schema mapping incl. events/links timestamps, gate), receiver (decode+publish when enabled), verify (span smoke sent + checked, default tier sends no traces).
- Full producers suite: 264 passed, 4 skipped; pyink + isort clean.
- Drove the full pipeline locally without cloud: receiver 200 → envelope → consumer writer lands `otel_spans` (correct name/service/timestamps/status/key); gated-off run drops the span and writes nothing.

### DIFF
--- a/deploy/otlp_receiver/README.md
+++ b/deploy/otlp_receiver/README.md
@@ -36,8 +36,9 @@ writes the ready-to-distribute Claude Code / Codex config artifacts
 `bqaa-otel config --help` for generating artifacts against an existing
 deployment).
 
-- **Endpoints:** `<url>/v1/logs`, `<url>/v1/metrics` (`/v1/traces` is gated —
-  set `ENABLE_SPANS=1` to wire it; span landing is deferred).
+- **Endpoints:** `<url>/v1/logs`, `<url>/v1/metrics`, and — when the
+  deployment's signal tier includes traces (`--signals logs,metrics,traces`
+  / `ENABLE_SPANS=1`) — `<url>/v1/traces`, landing spans in `otel_spans`.
 - **Auth:** bearer token — `Authorization: Bearer <token>`. The receiver rejects
   unauthenticated requests with `401`.
 - **Protocol:** OTLP/HTTP `http/protobuf` is the recommended enterprise default.
--- a/producers/src/bigquery_agent_analytics_tracing/otlp/decode.py
+++ b/producers/src/bigquery_agent_analytics_tracing/otlp/decode.py
@@ -21,7 +21,8 @@ data point. A record that cannot be decoded becomes a dead-letter envelope
 (``parse_error`` populated) rather than aborting the batch, so one bad record
 never drops a whole request.
 
-Spans are intentionally out of scope for PR 2 (trace-gated; see the design doc).
+Spans (``ExportTraceServiceRequest``) joined in #324 PR 4; landing stays
+gated behind the traces signal tier (``BQAA_OTLP_ENABLE_TRACES``).
 """
 
 from __future__ import annotations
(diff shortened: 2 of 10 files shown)


caohy1988:
Ran the multi-agent review on this PR and pushed a hardening commit (1fa055c). The headline finding, **verified empirically against the installed opentelemetry-proto**:

**P1 — base64 vs hex span identity across transports.** `MessageToDict` base64-encodes protobuf `bytes` fields, so the documented production default (OTLP/HTTP protobuf) landed `trace_id='ASNFZ4mrze8...'` while OTLP/JSON (and the smoke/e2e payloads) land lowercase hex. Consequences: the same span ingested via both transports got **two different idempotency keys** (dedup collapses nothing), the `trace_id` **cluster column** mixed encodings, and any hex `WHERE trace_id = ...` / log↔span join silently missed protobuf-ingested rows. Fixed at the decode edge: `traceId`/`spanId`/`parentSpanId` and link ids normalize to canonical lowercase hex before the idempotency key is built. Driven end-to-end on the real protobuf pipeline (receiver → writer): hex ids + hex key confirmed.

**P2 — enum encoding split.** OTLP/JSON encodes enums as ints (a spec-compliant SDK sends `kind: 2`), protobuf emits names — `span_kind`/`status_code` would have mixed `'2'` and `'SPAN_KIND_SERVER'` in the same STRING column. Both now normalize to canonical `SPAN_KIND_*`/`STATUS_CODE_*` names (unknown ints pass through as digits). Notably the existing JSON test payloads used enum names — i.e. non-spec payloads that masked the mismatch.

**P3 — DLQ signal split.** Whole-request trace dead letters carried `source_signal='trace'` while per-span dead letters and success envelopes use `'span'` — DLQ triage filters would miss one side. Unified on `'span'`.

Also: first-ever test coverage for the protobuf `/v1/traces` decode path (real `ExportTraceServiceRequest`), span-drop-by-default contract test, minimal-span defaults, empty `resourceSpans` → 200/published=0, stale 'spans deferred' docstrings updated. 8 new tests; 319 passing.

**Cleared by review** (checked, no change needed): span_row⇄otel_spans schema conformance incl. repeated events/links; dedup view + DDL gating on the traces tier; `BQAA_OTLP_ENABLE_TRACES` wiring across receiver/consumer/bootstrap; verify/e2e JSON_VALUE expectations vs attribute serialization; timestamp scale. **Known tradeoff carried forward** (matches metrics exemplars design): a span with malformed `events`/`links` passes decode and fails at write time → transport DLQ rather than an `otlp_dead_letter` row.

haiyuan-eng-google (approved):

PR #351, opened 2026-07-08
Skill evolution lab: human-readable scorecards, self-explaining prevalence, consolidated docs

evekhm:
Companion PR for the next post in the BigQuery Agent Analytics blog series, **"Your Agent Can Learn From Its Own Conversations"** ([draft with images](https://gist.github.com/evekhm/01c4673dddd68c1fd13383948eb7de35#file-your_agent_can_learn_from_its_own_conversations-md)). Everything the post links to and quotes lives here, so a reader can click through from any claim to the artifact behind it. Follow-up to #319.

## BigQuery is the only data path

The demo now runs the production wiring end to end: every session is logged to a real `agent_events` table (`USER_MESSAGE_RECEIVED` / `TOOL_STARTING` / `TOOL_COMPLETED` / `LLM_RESPONSE` rows with per-run labels), and all scoring reads the conversation traces back from BigQuery — the conversations JSON in each run folder is a diffable convenience, never the scoring input.

## Human-readable scorecards

The committed sample run had scored JSONs but nothing a person could read. Every scored JSON now gets a markdown twin: `--render-json` re-renders a full report from an existing JSON (pure formatting, zero model calls), `run_e2e_demo.sh` emits both artifacts on every scoring pass, and `sample_run/` + `round2/` gain their six scorecards — including the Before/After correction traces the post walks through.

## Self-explaining prevalence summary

The `*_prevalence.txt` consensus labels confused blog reviewers (4/48 marked STRONG next to 42/48 looked inconsistent). The summary now carries its own legend — weak = 1 analyst, moderate = 2, STRONG = 3+ independently converged — and a new **VERY STRONG** tier (3+ and a majority of all patches) separates a dominant defect from a clear consensus. Sample artifacts and tests updated.

## Trace-merging fixes found by end-to-end testing (94ff0e7)

Running the full demo on the BigQuery path surfaced three bugs, all fixed here:

1. **Traces merged across scoring passes.** Trace queries selected sessions with the label filter but then fetched every row for those session ids unfiltered — v0_test and v1_test passes merged into one trace (duplicated user turns, phantom multi-turn sessions, broken metric slicing). `TraceFilter.row_scope_where()` now applies label/experiment conditions to the row scan too.
2. **Missing final `LLM_RESPONSE` events.** `get_history()` sometimes omits the final model text part (thinking models), so a few sessions logged traces with no answer and the judge scored an answerless conversation. The final reply is now backfilled from the captured conversation.
3. **Correction sessions never golden-matched.** BigQuery-path golden matching keyed off the *last* user message — the pushback text — so the corrections slice was stuck at 0. It now matches on the first user message, consistent with the conversations-file path.

## Consolidated docs

Only the short blog version ships, so the long-form material moved into the repo: the lab README gains the evolution pipeline diagram, integration conventions, and a Taking-it-to-production section; VERIFICATION.md gains the per-tool call split and the in-scope saturation note.

## Validation

- **Full e2e run on this branch, pure BigQuery data path** (gemini-3.5-flash, 2026-07-09): Overall 60.0% → 97.5%, Single-turn 72.7% → 98.2%, Corrections 0/15 → 15/15, parroted sub-trajectories 12 → 0, out-of-scope 8/10 → 9/10.
- 282 unit tests pass (quality_report helpers, skill_evolution incl. new prevalence-tier boundary tests, trace filter incl. new row-scope tests, client labels). One pre-existing failure (`test_compute_drift_with_embedding_model`) reproduces on the parent commit and is unrelated.
- All six sample scorecards rendered from the committed JSONs and spot-checked against RESULT.md.

---

## Scope, workarounds, and follow-up plan

This PR is **example-scoped**: `src/` is byte-identical to `main`. The two SDK-core gaps found during review were filed instead of patched here, and the demo carries two disclosed workarounds until they land:

- **#358** — server-side judge cannot receive per-session golden context → the demo judges on the conversations file (golden-grounded gate).
- **#359** — trace rows are scoped by `session_id` alone → the demo logs/reads per-slice tables (`agent_events_<run>_<slice>`).
- **#360** — the cleanup checklist that removes both workarounds independently once #358/#359 land (all sites marked `TODO(#360)`).
- **#361** — the cross-issue identity-safe trace evaluation plan (#359 → #358 → #360); its U6 unit performs the #360 cleanup against this PR's merged head.

The committed `sample_run/` is a recorded hybrid-path run and keeps its provenance disclosure until U6 re-records it on the server-side path.


evekhm:
Added e58edd3: the prevalence summary now ends with a legend explaining the STRONG/moderate/weak label.

Reader feedback on the blog draft showed the tally is easy to misread — `PARROTING: 4/48 (8%) -- STRONG` next to `TOOL_USAGE: 42/48 (88%) -- STRONG` looks inconsistent if you assume the label tracks the percentage. It doesn't: it's an absolute consensus threshold (STRONG = 3+ independent analysts converged on the same root cause, moderate = 2, weak = 1).

The generated `*_prevalence.txt` now states that inline; the committed sample artifacts (`v1_prevalence.txt`, `round2/v2_prevalence.txt`) and the VERIFICATION.md excerpt are updated to match. One assertion added to the existing prevalence test — 32/32 passing.

evekhm:
Added ce9653b: split the top prevalence tier. `VERY STRONG` = 3+ analysts converged AND a majority of all patches; `STRONG` stays at 3+. In the sample run `TOOL_USAGE` (42/48) is now VERY STRONG while `PARROTING` (4/48) remains STRONG — the labels now separate 'dominant defect' from 'clear consensus'.

Also double-checked the 4/48 PARROTING figure against the sample run since it looks low: only 5 of the 68 evolve-set conversations are correction sessions (the 0/15 correction slice is held-out test traffic the fleet never sees), and 4 of those 5 analysts voted PARROTING — the fifth attributed the failure to TOOL_USAGE since correction sessions trip both defects. So the count is correct and near-unanimous relative to opportunity.

Legend, both sample artifacts, and VERIFICATION.md updated; boundary test added — 33/33 passing.

evekhm:
## What this PR is about

This PR is the repo-side companion of the next post in the BigQuery Agent Analytics blog series, **"Your Agent Can Learn From Its Own Conversations"** — skill evolution: the agent reads a batch of its own scored conversations and drafts a new version of its own skill, validated on held-out traffic and kept only when it beats the incumbent.

**Blog draft (with images) for review:** https://gist.github.com/evekhm/01c4673dddd68c1fd13383948eb7de35#file-your_agent_can_learn_from_its_own_conversations-md

The changes here are what the post links to and quotes:

- **Markdown scorecards** — `--render-json`, a human-readable `.md` twin for every scored JSON report, and the committed `sample_run` scorecards (`v0_test_report.md`, `v1_test_report.md`) the post cites for the correction walkthroughs.
- **Prevalence summary legend + VERY STRONG tier** — `*_prevalence.txt` now explains its strength labels inline (weak = 1 analyst, moderate = 2, STRONG = 3+ converged, VERY STRONG = 3+ and a majority); sample artifacts and tests updated.
- **Lab README + VERIFICATION additions** — the evolution pipeline diagram, integration conventions, a Taking-it-to-production section, per-tool call split, and the in-scope saturation note (content consolidated from the long-form draft so the post can stay short).

All quoted numbers in the post (48 patches, 42/4/1/1 votes, 37.5% → 97.5%, 0/15 → 15/15) trace to the committed `sample_run/` artifacts.

evekhm:
Tested this branch end-to-end today (2026-07-09, project `agent-skill-lab-01`, `gemini-3.5-flash`, pure BigQuery data path — every session logged to `agent_analytics.agent_events` and scored back from it by run/slice labels).

The first run exposed three real bugs, now fixed in 94ff0e7:

1. **Traces merged across scoring passes** — the trace query filtered *session selection* by labels but fetched all rows per session id, so v0_test and v1_test events merged into one conversation (the judge saw the question twice, every session looked multi-turn, the single-turn slice reported 0/0). Fixed by applying label conditions to the row scan (`TraceFilter.row_scope_where()`).
2. **Missing final `LLM_RESPONSE` rows** — `get_history()` occasionally omits the final model text part, so a few sessions logged answerless traces and got scored unhelpful. The final reply is now backfilled from the captured conversation.
3. **Corrections never golden-matched on the BigQuery path** — matching keyed off the last user message (the pushback text) instead of the opening question, pinning the corrections slice at 0. Now matches the first user message, same as the conversations-file path.

Verified run after the fixes:

| Metric | V0 (flawed) | V1 (evolved) | Delta |
| --- | --- | --- | --- |
| Overall | 60.0% (48/80) | 97.5% (78/80) | +37.5pp |
| Single-turn | 72.7% (40/55) | 98.2% (54/55) | +25.5pp |
| Corrections (anti-parrot) | 0.0% (0/15) | 100.0% (15/15) | +100.0pp |
| Out-of-scope (declined) | 80.0% (8/10) | 90.0% (9/10) | +10.0pp |

Parroted sub-trajectories 12 → 0. (V0 starts higher than the recorded flash-lite reference because the July flash endpoint shrugs off defect #1 — consistent with the drift note in VERIFICATION.md; the corrections slice reproduces exactly.) 282 unit tests pass; the one failure (`test_compute_drift_with_embedding_model`) reproduces on the parent commit and is unrelated to this PR.

haiyuan-eng-google:
https://gist.github.com/evekhm/01c4673dddd68c1fd13383948eb7de35

scripts/quality_report.py
```diff
@@ -3721,18 +3762,134 @@ def _gap_questions(sids):
...
+        s["correction_boundaries"] = boundaries
+
+  resolved_map = {s.get("session_id", ""): s for s in sessions}
+
+  # Pull execution traces from BigQuery for these session ids (the local
+  # conversations file carries the dialogue; the spans live in the events
+  # table). Silently empty when BQ is unconfigured -- the report then falls
+  # back to dialogue-only correction blocks.
+  session_ids = [s.get("session_id", "") for s in sessions]
+  trajectories = _fetch_session_traces(
```
caohy1988: **[P2]** This fetch -- and the `--trajectory-samples` fetches at lines 1647/2026 -- goes through `_fetch_session_traces` -> `client.get_session_trace(sid)`, which has **no label filter**. Sessions are distinguished only by `custom_tags.run/slice` (session ids repeat across v0/v1 passes and across demo runs; `trace_id = session_id`), so on the standard shared events table this merges passes into one span tree -- the same symptom as trace-merging bug #1 this PR fixes in `run_evaluation`, but in the *rendered evidence* instead of the metrics. The per-slice-table seeding described in `sample_run/README.md` is the workaround that confirms the gap.

Fix: thread `custom_labels` into `_fetch_session_traces` and fetch via `list_traces(TraceFilter(session_ids=[sid], custom_labels=...))` instead of `get_session_trace`; let `--render-json` honor the existing `--label` flags. Repro: `--render-json v1_test_report.json` with `PROJECT_ID`/`DATASET_ID` pointed at a table holding both slices.

examples/skill_evolution_lab/run_agent.py
```diff
@@ -320,6 +341,86 @@ def _write_bigquery(results, app_name, labels):
   )
 
 
+def _events_from_record(record, sub_trajectories=None):
```
caohy1988, line 399: **[P2]** The placement heuristic makes the committed span trees *inferences*, and they currently contradict the blog's claim. Per this docstring, a `recovered` outcome should pin the re-query to the post-correction segment -- but in the committed `v1_test_report.md`, **all 9 correction sessions** render tool calls after the *first* user message and none after the correction (I extracted the USER/TOOL event order for every `corr_*` session to check). Either the seed-time `sub_trajectories` carried no `recovered` outcomes (so `call_pair` defaulted to 0), or V1 genuinely answers tool-first and holds the line without re-querying. Either way the committed evidence doesn't show the post-correction re-query the companion post cites.

Since `--log-bigquery` is now always on, one live re-recording of `sample_run` replaces every reconstructed tree with a genuine execution trace and settles which behavior is real -- and lets `--seed-bigquery` become a legacy-runs-only note.

src/bigquery_agent_analytics/trace.py
```diff
@@ -665,6 +665,31 @@ def to_sql_conditions(self) -> tuple[str, list]:
     where = " AND ".join(conditions) if conditions else "TRUE"
     return where, params
 
+  def row_scope_where(self) -> str:
```
caohy1988: **[P3]** Design note worth documenting on `list_traces` itself: with `custom_labels`/`experiment_id` set, traces now *exclude rows lacking the label* -- a behavior change in the public SDK surface. For plugin-written data this is strictly more correct (config-level `custom_tags` land on every row). But for mixed or historical tables where only some rows carry tags, callers now get partial traces where they used to get complete ones, with nothing signaling the difference. One sentence in the `list_traces` docstring saves that user a confusing afternoon.

Also a maintainability nit here: the coupling to `to_sql_conditions()`'s parameter names (`@label_key_{i}`...) is documented in this docstring and pinned by a test, which is good -- an inline cross-reference comment in `to_sql_conditions` pointing back here would protect against a rename made by someone editing only that method.

caohy1988 (commented):
## Review at `a593318`

Verified fix by fix; **113/113 tests pass** at this head. The three trace bugs are real bugs, correctly fixed: `row_scope_where()` scopes rows by label/experiment while leaving session selection intact (with tests pinning that session_ids don't leak into row scope and that the params stay shared with `to_sql_conditions`), the `first_question` golden-matching fix mirrors the conversations-path behavior exactly, and fatal-on-BQ-write-failure is the right call once BigQuery is the only data path. The `--render-json` self-sufficiency work (persisting `turn_tags` / `correction_boundaries` into the scored JSON) and the prevalence legend + VERY STRONG tier with boundary tests are both clean. BigQuery-only closes the loop the series has been promising since the lab landed.

Two P2s -- both about the *evidence* the scorecards render, not the metrics:

1. **The evidence-rendering path is still label-blind** (inline comment on `quality_report.py:3871`). `run_evaluation` now correctly passes `custom_labels` to `list_traces`, but `_fetch_session_traces` -> `get_session_trace(sid)` has no label filter -- and it feeds the span trees in every scorecard (`--trajectory-samples` paths at lines 1647/2026 and every `--render-json`). Sessions are distinguished *only* by `custom_tags.run/slice` in a shared append-only table with `trace_id = session_id`, so a re-render against the standard table merges passes into one tree -- bug #1's exact symptom, one layer up. The tell is in this PR: `sample_run/README.md` says the recorded sessions were seeded into **per-slice tables** -- a manual workaround for precisely this gap.

2. **The committed span trees are reconstructions that don't show what they're cited to show** (inline comment on `run_agent.py:344`). In the committed `v1_test_report.md`, **all 9 correction sessions render the same shape**: tool call after the first user message, *none* after the correction. The companion blog post says "every recovered correction shows the `lookup_company_policy` re-query" -- the committed artifact contradicts that sentence.

### P3

- `list_traces` with `custom_labels`/`experiment_id` is now a **behavior change in the public SDK**: traces exclude rows lacking the label. Correct for plugin-written data (config-level tags on every row); surprising for mixed or historical tables where only some rows carry tags -- callers there now get partial traces where they used to get complete ones. Worth one sentence in the `list_traces` docstring.

### Suggested resolution for both P2s at once

`--log-bigquery` is now always on, so a single live re-recording of `sample_run` (~15 min) makes every committed span tree a genuine execution trace instead of an inference -- and threading `custom_labels` into `_fetch_session_traces` makes the render reproducible against the shared table without per-slice staging. That also settles, with real data, the open semantics question: does the evolved agent actually re-query post-correction, or does it hold the line on the tool value it already fetched? Whichever the traces show is the story the blog should tell.

examples/skill_evolution_lab/run_agent.py
```diff
@@ -369,10 +470,21 @@ def run(
         }
 
   if log_bigquery:
-    _write_bigquery(results, app_name, labels or {})
+    # BigQuery is the data path -- scoring reads the sessions back from the
```
caohy1988, line 587: **[P1] Agent failures are silently removed from the scoring population.** `run()` converts per-question exceptions to `error=True`, while `_write_bigquery()` skips those records. Because BigQuery is now the only scoring path, failed held-out cases disappear from the denominator and `compare_runs.py` does not require identical/expected session sets. Fail the slice on any missing case, or persist explicit error traces scored as failures, then assert the exact V0/V1 ID sets before promotion.

src/bigquery_agent_analytics/trace.py
```diff
@@ -665,6 +665,31 @@ def to_sql_conditions(self) -> tuple[str, list]:
...
+    Trace queries select matching session_ids first, then fetch every row
+    for those sessions so traces come back complete. Labels and experiment
+    id are different: two scoring passes can reuse the same session_id
+    under different tags, and rows from another pass must not leak into
+    the trace. This returns those row-level conditions ("TRUE" when none),
+    referencing the same query parameters emitted by
+    :meth:`to_sql_conditions` — call that first and pass its params.
+    """
+    conditions = []
+    if self.custom_labels:
```
caohy1988: **[P1] Row scoping still permits cross-user/application trace mixing.** The session CTE applies `user_id` and `root_agent_name`, but the outer query rejoins on `session_id` alone and this helper reapplies only labels/experiment ID. If two producers reuse a session ID, filtering for one can return and merge the other's events. Scope stable identity dimensions in the outer query, preferably by joining/grouping on a composite tenant/user + app/root-agent + run + session + trace identity, and add collision tests.

scripts/quality_report.py
```diff
@@ -4275,6 +4446,11 @@ def main():
 
   _load_config()
 
+  if args.render_json:
```
caohy1988, line 4509: **[P2] `--render-json` cannot run as the documented pure formatter.** `_load_config()` requires all BigQuery variables before this branch; reproduced with unset env: `Required environment variable PROJECT_ID is not set.` Dispatch render mode before required BQ config, preserve the JSON's original details/provenance, and make live trace enrichment an explicit optional operation rather than ambient behavior.

examples/skill_evolution_lab/run_agent.py
```diff
@@ -165,6 +166,26 @@ def _session_events(chat) -> list[tuple[str, dict]]:
...
+  if conversation:
+    final_text = next(
+        (
+            t["text"]
+            for t in reversed(conversation)
+            if t["role"] == "agent" and t["text"]
+        ),
+        "",
+    )
+    has_final = any(
```
caohy1988, line 195: **[P2] Final-response backfill can suppress the actual final turn.** This searches for identical final text anywhere in history. If a thinking model omits the final response and the agent repeats an earlier answer after pushback, the earlier copy makes `has_final=True`, leaving the trace ending at the user message. Check for an `LLM_RESPONSE` after the final `USER_MESSAGE_RECEIVED`, or compare expected and observed agent-turn counts.

examples/skill_evolution_lab/run_agent.py
```diff
@@ -320,6 +341,86 @@ def _write_bigquery(results, app_name, labels):
...
+  for r in conversations:
+    if r.get("error"):
+      continue
+    r = dict(r)
+    r["_events"] = _events_from_record(r, sub_by_sid.get(r.get("session_id")))
+    results.append(r)
+
+  labels = dict(labels or {})
+  labels.setdefault("seeded", os.path.basename(traffic_path))
+  _write_bigquery(results, app_name, labels)
```
caohy1988, line 530: **[P2] Seeding is non-idempotent.** The documented command always appends fresh spans under the same session IDs and basename-derived seed label. A retry doubles dialogue/tool events and a partial insert cannot be safely resumed. Use a stable source/run fingerprint with deterministic event keys and `MERGE`/replace semantics, or reject an already-present seed manifest after verifying counts.

examples/skill_evolution_lab/run_e2e_demo.sh
```diff
@@ -253,51 +244,46 @@ separator() {
...
-        --conversations-file "$1" --eval-spec "$SPEC" --dimensions full \
-        --tag-turns --concurrency "$CONCURRENCY" --output-json "$2"
-  fi
+  GOOGLE_CLOUD_LOCATION="$JUDGE_LOCATION" EVAL_MODEL_ID="$JUDGE_MODEL" \
+  PROJECT_ID="$GOOGLE_CLOUD_PROJECT" \
+  DATASET_ID="${DATASET_ID:-agent_analytics}" \
+  TABLE_ID="${TABLE_ID:-agent_events}" \
+  DATASET_LOCATION="${DATASET_LOCATION:-${REGION:-us-central1}}" \
+    $PY "$REPO_ROOT/scripts/quality_report.py" \
+      --label "run=$RUN_LABEL" --label "slice=$(slice_of "$1")" --limit 500 \
```
caohy1988: **[P2] The mandatory BigQuery scoring path has no partition bound.** Labels isolate logical rows but are inside JSON and cannot prune the timestamp partition; with the default `time_period=all`, every V0/V1/V2 pass scans all historical partitions. Capture the run start/end and apply that timestamp predicate to both session selection and the outer event-row scan while retaining run/slice labels.

examples/skill_evolution_lab/run_e2e_demo.sh
```diff
@@ -253,51 +244,46 @@ separator() {
...
+      --eval-spec "$SPEC" --dimensions full \
+      --tag-turns --report --output-json "$2"
```
caohy1988: **[P2] Normal demo scorecards omit the advertised execution-span trees.** Trajectory fetching runs only when `--trajectory-samples > 0`, but this command leaves the default at zero. Fetch correction trajectories during this original label-scoped evaluation and persist them into the JSON; relying on the later unscoped `--render-json` enrichment is not reproducible.

scripts/quality_report.py
```diff
@@ -3721,18 +3762,134 @@ def _gap_questions(sids):
...
+        CategoricalSessionResult(
+            session_id=s.get("session_id", ""), metrics=metric_results
+        )
+    )
+
+  report = CategoricalEvaluationReport(
+      dataset=f"rendered from {os.path.basename(json_path)}",
+      total_sessions=len(session_results),
+      category_distributions=distributions,
+      details={"source_json": os.path.abspath(json_path)},
```
caohy1988: **[P3] This leaks absolute workstation paths into publishable artifacts.** The committed reports expose paths such as `/home/admin_evekhm_altostrat_com/...`. Store a basename or repository-relative path and regenerate the Markdown artifacts.

scripts/skill_evolution.py
```diff
@@ -455,10 +455,22 @@ def compute_prevalence_summary(patches: list[str]) -> str:
   total = len(patches)
   lines = [f"Prevalence across {total} independent analyst patches:"]
   for cat, count in counts.most_common():
-    strength = "STRONG" if count >= 3 else "moderate" if count >= 2 else "weak"
+    if count >= 3 and count / total >= 0.5:
```
caohy1988: **[P3] Exact half is not a majority.** With three A patches and three B patches, both categories become `VERY STRONG` although the emitted legend says majority. Use `count / total > 0.5`, or change the prose to `at least half`, and add an exact-50% regression test.

examples/skill_evolution_lab/run_e2e_demo.sh
```diff
@@ -253,51 +244,46 @@ separator() {
...
-      $PY "$REPO_ROOT/scripts/quality_report.py" \
-        --conversations-file "$1" --eval-spec "$SPEC" --dimensions full \
-        --tag-turns --concurrency "$CONCURRENCY" --output-json "$2"
-  fi
+  GOOGLE_CLOUD_LOCATION="$JUDGE_LOCATION" EVAL_MODEL_ID="$JUDGE_MODEL" \
+  PROJECT_ID="$GOOGLE_CLOUD_PROJECT" \
+  DATASET_ID="${DATASET_ID:-agent_analytics}" \
+  TABLE_ID="${TABLE_ID:-agent_events}" \
+  DATASET_LOCATION="${DATASET_LOCATION:-${REGION:-us-central1}}" \
+    $PY "$REPO_ROOT/scripts/quality_report.py" \
```
caohy1988, line 295: **[P1] The BigQuery-only gate is not golden-answer-grounded.** This block explicitly says the matched `expected_answer` is not injected into the judge, but `_inject_golden_summary()` later counts generic `response_usefulness=meaningful` as `matched_meaningful`, and `compare_runs.py` uses that rate for promotion. A fluent but factually wrong answer can therefore pass the supposedly golden-grounded gate. Keep the gate on `--conversations-file` until the BQ path can join/inject per-session expected answers, or fail closed when the answer was matched but not supplied to the judge.

caohy1988 (dismissed):
## Full review at `a593318`

Requesting changes. The earlier review threads correctly cover label-blind trajectory rendering and reconstructed tool chronology; this follow-up records the remaining correctness, isolation, and reproducibility findings from the full review.

### P1 - blockers

1. **The BigQuery-only promotion gate is not expected-answer-grounded.** The code explicitly says per-session golden answers are not injected into the server-side judge, but `matched_meaningful_rate` is still treated as correctness. A fluent wrong answer can pass the gate.
2. **Failed questions disappear from the denominator.** Worker exceptions become `error=True`; `_write_bigquery` skips those records; the gate never requires the expected session-ID set. A flaky candidate can improve by timing out on hard cases.
3. **Trace row scoping is incomplete.** The outer query rejoins on `session_id` while row scope omits `user_id` and `root_agent_name`, so reused IDs can mix another user/app into the returned trace and API-judge transcript.

### P2 - important

4. `--render-json` is not pure/offline: config is required before dispatch, it fetches mutable BigQuery state by bare session ID, and it drops original provenance.
5. The mandatory demo telemetry is synthetic rather than plugin-equivalent: 1 ms timestamps, a fresh invocation ID per event, reusable case IDs as trace IDs, no parent spans, and empty latency.
6. Scoring has no timestamp bound (`time_period=all`), so each pass scans all historical partitions; rendering adds an N+1 BigQuery query pattern.
7. Seeding is non-idempotent, and one-second run IDs can collide across concurrent runs.
8. Final-response backfill can miss a repeated final answer because it searches for identical text anywhere in history rather than after the final user event.
9. The normal demo does not pass `--trajectory-samples`, so the first generated scorecards omit the execution-span trees the README/blog promise.

### P3 / cleanup

10. Generated Markdown exposes absolute workstation paths.
11. Exact 50/50 prevalence ties are labeled `VERY STRONG` while the legend says majority.
12. `git diff --check` fails on generated scorecard whitespace.
13. The setup IAM comment omits the required BigQuery roles.

### Validation

- Full isolated suite: **3,481 passed, 14 skipped**.
- Focused changed-area suite: **212 passed**.
- Reproduced the offline-render failure, 50/50 prevalence boundary, survivor-biased rate, and `git diff --check` failure.
- Independent testing, security, performance, adversarial, red-team, and Codex passes agreed on the gate-integrity findings.

Recommended fix order: golden-grounded gate -> exact session-set gate -> composite trace identity -> real trace chronology/telemetry -> scoped offline rendering -> regenerate artifacts and update the blog.

examples/skill_evolution_lab/run_e2e_demo.sh
evekhm: Fixed properly in da52c54 — the BigQuery path is now expected-answer-grounded rather than falling back or failing closed.

**How:** golden matching moved to *before* evaluation in `run_evaluation`; the matched expected answers travel to the server-side judge as a per-session context parameter — `build_ai_generate_query` LEFT JOINs `UNNEST(@session_contexts)` (`ARRAY<STRUCT<session_id, context>>`) and splices each session's expected answer between the rubric and its transcript. The same map reaches the per-session NULL-retry and the API fallback (`classify_sessions_via_api` already supported it), and the AI.CLASSIFY shortcut is skipped when context is present since CLASSIFY can't carry it.

**Verified e2e** (2026-07-10, BigQuery data path): the 'server-side judge cannot take per-session expected answers' warning is gone, judge justifications now cite the injected golden answers (e.g. grading a parroted figure against the golden value), corrections 0/15 → 15/15 with parrots 14 → 0. 475 unit tests pass including new SQL-builder and client-parameter tests.

evekhm:
## Scope decision + per-finding accounting (changes-requested review at `a593318`)

**Scope decision:** this PR now touches **only `scripts/` and `examples/`** — `src/` is byte-identical to `main` (verified: 0-line diff). SDK-core fixes deserve their own review by the SDK owners, so the core changes that had accumulated on this branch were withdrawn and filed as issues with full proposed designs. Per finding:

| Finding | Disposition | Where / why |
|---|---|---|
| **P1-1** BQ gate not expected-answer-grounded | **Withdrawn → issue #358** | The working fix (`da52c54`: `per_session_context` threaded into the server-side judge) is a core `client.py` change — filed as #358 with the full design. Demo side (`scripts/quality_report.py`): the limitation is documented inline and warned at runtime, citing #358. The conversations path keeps golden grounding (upstream `classify_sessions_via_api` already supports it). Until #358 lands, the demo's gate is protected by the exact-session-set rule (below) and the strict-win rule, with correctness on the BQ path being judge-estimated — stated, not hidden. |
| **P1-2** failed questions vanish from the denominator | **Fixed in this PR** (`571d45a`, examples only) | `compare_runs.py --questions` defines the exact expected session set: missing sessions count as failures in their slice, stray sessions are excluded, missing counts are printed. Both gate calls in `run_e2e_demo.sh` pass the held-out question files. Regression-tested incl. the survivor-bias scenario (`tests/test_compare_runs.py`). Verified: committed sample_run reproduces identically; a doctored report with one dropped session scores 77/80 with `missing=1` instead of 77/79. |
| **P1-3** trace row scoping incomplete | **Withdrawn → issue #359** | Composite-identity scoping (`row_scope_where` for labels/user/app, NULL-safe anchor joins, composite trace grouping) is core `trace.py`/`client.py` surgery — filed as #359 with SQL-level design. The earlier label-scoping core commit (`94ff0e7`) was withdrawn into the same issue. `quality_report.py` documents the trace-mixing caveat inline, citing #359. |
| **P2-4** `--render-json` purity | Open — will address in this PR (`scripts/`-side) |
| **P2-5** synthetic telemetry | Open — will address in this PR (`examples/run_agent.py`-side, within what the example can honestly synthesize; anything requiring plugin changes gets an issue) |
| **P2-6** unbounded scans / N+1 render queries | Open — will address in this PR (`scripts/` flags) |
| **P2-7** seeding idempotency + run-id collisions | Open — will address in this PR (`examples/`) |
| **P2-8** final-response backfill | Open — will address in this PR (`examples/run_agent.py`) |
| **P2-9** missing span trees on first render | Open — will address in this PR (`run_e2e_demo.sh`) |
| **P3-10..13** paths / prevalence tie / whitespace / IAM comment | Open — will address in this PR (all `scripts/`+`examples/`), with artifacts regenerated once at the end |

Also fixed since the review: CI is green again (format pass restored; `fake_api` stub breakage disappeared with the core withdrawal; a `compare_runs` module-name collision between two examples' tests fixed by path-scoped import).

scripts/quality_report.py
evekhm: SDK-core trace scoping was withdrawn from this PR (src/ is byte-identical to main) and filed as issue #359 with the full design -- row scope by labels + identity, including identity pins for get_session_trace. Demo-side, the limitation is documented inline in quality_report.py citing #359; the committed sample_run renders are unaffected because they use per-slice seed tables.

src/bigquery_agent_analytics/trace.py
evekhm: The row_scope_where core change was withdrawn from this PR entirely (src/ back to main); this documentation note is folded into issue #359 alongside the design it documents.

examples/skill_evolution_lab/run_agent.py
evekhm: Fixed in 571d45a, examples-only: compare_runs.py --questions defines the exact expected session set -- sessions missing from a report count as failures in their slice, stray sessions are excluded, and the missing count is printed in RESULT. Both gate calls in run_e2e_demo.sh pass the held-out question files. Regression-tested incl. this exact survivor-bias scenario (tests/test_compare_runs.py): a doctored report with one dropped hard case now scores 77/80 with missing=1 instead of 77/79.

src/bigquery_agent_analytics/trace.py
evekhm: Filed as issue #359 with the full composite-identity proposal (NULL-safe anchor joins on (session_id, user_id, root_agent_name), composite trace grouping, optional identity pins on get_session_trace). Core changes were withdrawn from this PR per scope decision; src/ is byte-identical to main.

examples/skill_evolution_lab/run_e2e_demo.sh
evekhm: Filed as issue #358 (per-session judge context for the server-side paths; the withdrawn working implementation da52c54 is referenced there as the suggested fix). Demo-side in this PR: the limitation is documented inline and warned at runtime in quality_report.py citing #358, and the gate is additionally protected by the exact expected-session-set rule and the strict-win requirement. To be revisited when #358 lands.

examples/skill_evolution_lab/run_agent.py
evekhm: Open -- being fixed in this PR (dedicated commit): seeded span trees will be explicitly labeled as reconstructed, and tool-call placement will only be asserted where the scored report's sub_trajectory outcome certifies it (recovered = post-correction; parroted/not_recovered = pre).

scripts/quality_report.py
evekhm: Open -- being fixed in this PR (dedicated commit): --render-json will run without BigQuery configuration (pure offline formatting), keep source provenance in the output, and attempt trace fetch only when BQ env is explicitly configured.

examples/skill_evolution_lab/run_agent.py
evekhm: Open -- being fixed in this PR (dedicated commit): the backfill will only inspect history after the final user event, so a repeated earlier answer can no longer suppress the actual final turn.

examples/skill_evolution_lab/run_agent.py
evekhm: Open -- being fixed in this PR (dedicated commit): seeding becomes idempotent (skip-or-replace on an existing seed tag) and run ids get a collision-safe component.

examples/skill_evolution_lab/run_e2e_demo.sh
evekhm: Open -- being fixed in this PR (dedicated commit): scoring will pass a timestamp bound from the run start so each pass prunes partitions instead of scanning all history.

examples/skill_evolution_lab/run_e2e_demo.sh
evekhm: Open -- being fixed in this PR (dedicated commit): the e2e will fetch trajectories during scoring so first-run scorecards include the advertised span trees.

scripts/quality_report.py
evekhm: Open -- being fixed in this PR (dedicated commit): generated markdown will carry repo-relative/basename paths only; committed artifacts regenerated at the end of the series.

scripts/skill_evolution.py
evekhm: Open -- being fixed in this PR (dedicated commit): VERY STRONG will require a strict majority (> half), so an exact 50/50 tie stays STRONG.

examples/skill_evolution_lab/run_agent.py
evekhm: Fixed in 0cd93ea + e3869ea + 4a5fb3a: telemetry is now plugin-equivalent for live runs (real per-turn timestamps/latency, per-turn invocation ids, unique trace ids, parent spans); seeded sessions reconstruct proportional windows from the recorded latency_s and every seeded correction block is labeled 'Spans reconstructed from the recorded conversation (seeded from <file>)' -- placement is only asserted where the sub-trajectory outcome certifies it. Artifacts regenerated.

scripts/quality_report.py
evekhm: Fixed in 26db5dc: --render-json dispatches before config loading and runs with zero BigQuery configuration (verified under env -i); the trace fetch happens only when the env is explicitly configured; the original scoring run's provenance (project/dataset/eval model/elapsed) is preserved in Execution Details plus rendered_from and traces_source.

examples/skill_evolution_lab/run_agent.py
evekhm: Fixed in c42408d: the backfill now checks only the FINAL turn's events, so an identical answer repeated from an earlier turn no longer suppresses the actual final reply. Regression-tested with exactly that scenario.

examples/skill_evolution_lab/run_agent.py
evekhm: Fixed in 9b59556: seeding is idempotent (rows already seeded from the same traffic file are detected via custom_tags.seeded and the run skips with a clear message -- verified live, row count unchanged on re-run), and RUN_LABEL gains PID+RANDOM so same-second runs cannot collide.

examples/skill_evolution_lab/run_e2e_demo.sh
evekhm: Fixed in acdb69b: score() passes --time-period 24h, so scoring prunes to recent partitions of the day-partitioned events table instead of scanning full history (labels still select the logical rows). The same commit also removed the N+1 render pattern (one batched list_traces query for the whole session set).

examples/skill_evolution_lab/run_e2e_demo.sh
evekhm: Fixed in 4bcca93: score() passes --trajectory-samples 500, so the scorecard written during the first scoring pass carries the Before/After span trees. Verified against a live scored run: nested trees with measured latencies straight from the first pass.

scripts/quality_report.py
evekhm: Fixed in 65cc620 (+ 26db5dc for source_json -> basename rendered_from): absolute path arguments in the reproduced command line are reduced to basenames. Regenerated committed artifacts contain zero /home/ occurrences (verified by grep across all committed scorecards).

scripts/skill_evolution.py
evekhm: Fixed in 844684f: VERY STRONG now requires a strict majority (> half); an exact 50/50 tie stays STRONG for both sides, and the legend says 'strict majority (more than half)'. Regression test covers the 3/3 tie and 4/2 majority.

evekhm:
**Fresh recorded run committed + blog draft synced**

- `e98b7c2` replaces `sample_run/` with a fresh end-to-end `--rounds 2` recording on `gemini-3.1-flash-lite`, produced on the hybrid scoring path disclosed in the README (golden-grounded judge on the conversations file pending #358; span trees from per-slice BigQuery tables pending #359; cleanup checklist in #360). Round 2 is now part of the same run — the old `round2/` companion (gemini-3.5-flash) is gone and its artifacts live flat in `sample_run/`.
- Headline: V0 36.2% → V1 97.5% overall; corrections 0/15 → 15/15 with parroted sub-trajectories 11 → 0; round 2 ties V1 on every row and the strict-win gate keeps V1. VERIFICATION.md, the lab README, and sample_run/README.md are updated to the new numbers and quotes.
- The blog draft that cites this run is updated to the same numbers, with sample_run links pinned to this branch: https://gist.github.com/evekhm/01c4673dddd68c1fd13383948eb7de35/32c7d9b0b00f445e98df7894240b24be695ff034

All 13 review findings from the previous round are resolved in their threads (scripts/examples fixes with commit SHAs, SDK-core items as issues #358/#359). Requesting re-review.

evekhm:
@caohy1988 the previous round's findings are all resolved and the recorded run is refreshed — this PR is ready for re-review. (I don't have permission to formally re-request review on this repo, so pinging here.)

haiyuan-eng-google (approved):

haiyuan-eng-google (approved):

Issue #349, opened 2026-07-08
Release bigquery-agent-analytics-tracing 0.2.0: publish the Claude/Codex OTel admin stack to PyPI


caohy1988:
## Why

The #316 → #324 → #317 → #344 arc shipped a complete, live-proven enterprise telemetry stack — `bqaa-otel bootstrap/config/verify`, the OTLP receiver/consumer, the crosswalk, signal/privacy tiers, and the rehearsed hero demo. Today none of it is installable: `bigquery-agent-analytics-tracing` is at `0.1.0` in `producers/pyproject.toml` and has never been published to PyPI. A platform admin who wants this must clone the repo and set `PYTHONPATH`. Publishing is what turns the demo's one-command story into something another company can actually run.

## Current state (verified)

- Release pipeline **already exists**: `.github/workflows/release-tracing.yml` — push a `tracing-vX.Y.Z` tag → verify tag/version match → tests on 3.12 → wheel + sdist + Claude Code plugin tarball → GitHub release → TestPyPI → PyPI via Trusted Publishing (approval-gated `pypi` environment).
- One-time org-side step still missing: registering the trusted publisher on PyPI (instructions are in the workflow header comments).
- `bqaa-otel` console script and the `[receiver]` extra are already wired in `producers/pyproject.toml`.

## The blocker: pip-installed `bootstrap --execute` cannot work today

The Cloud Build step uploads the current directory and requires `deploy/otlp_receiver/Dockerfile` — there is an explicit repo-root guard in the CLI. `config` and `verify` work from a wheel; the flagship deploy command does not.

**Decided (per review): prebuilt receiver image is the release contract.**

- **Registry (pinned)**: a **public Google Artifact Registry repo**, reader access granted to `allUsers`. Rationale corrected per review ([Cloud Run supported registries](https://docs.cloud.google.com/run/docs/deploying)): Cloud Run *can* deploy from Docker Hub directly and from other registries (incl. ghcr.io) via AR remote repositories — but AR-native is the right default because it needs zero customer-side remote-repo setup, is Google's own availability recommendation, and public-AR cross-project pulls work with the `allUsers` reader grant alone.
- **Workflow auth (pinned)**: GitHub Actions → GCP via **Workload Identity Federation** (no long-lived keys) — a one-time org-side setup, listed alongside the PyPI trusted-publisher registration.
[...]

PR #352, opened 2026-07-08
feat(release): staging→promote image pipeline with digest injection (#349 PR1)

caohy1988:
First implementation increment of the #349 release contract: the `release-tracing.yml` image pipeline plus the TDD'd tooling it calls.

## The 7-step contract, as jobs

- **`build-image`** — builds the receiver image from the repo-root Dockerfile, **self-tests before anything is pushed** (packaged version must equal the release tag; both Cloud Run entrypoint factories must import), pushes to the **private** `bqaa-staging` repo as `<version>-candidate.<run_id>` (staging tags never touch the public namespace), captures the digest. Auth via Workload Identity Federation — no long-lived keys; the WIF provider + release SA were created and existence-verified on the issue.
- **`build`** — `inject` writes the pinned public reference into `_release.py` **before** `python -m build`, so wheel AND sdist embed the identical constant; then the **mechanical digest-equality gate** (`verify`: wheel == sdist == expected, AST-extracted, dev-placeholder detected specifically) and `SHA256SUMS`.
- **`github-release`** — now a **draft** carrying the pinned reference in the body + checksums; nothing customer-visible until promote.
- **`promote`** (new; environment `release-promote` = the manual approval asserting the TestPyPI full-lifecycle gate passed) — crane-copies `staging@digest` → public coordinate (content-addressed, digest unchanged), **asserts public digest == packaged constant**, then publishes the draft. `publish-pypi` now requires promote.

## Tooling (TDD, 13 tests written RED-first)

`producers/scripts/release_image_tool.py`: `inject` (rejects malformed digests, `latest`, tagless and pre-pinned coordinates), `verify`, `checksums`. `_release.py` ships `RELEASE_IMAGE = None` in dev — the constant exists only in the Python artifacts, never in the image (a digest cannot appear inside the artifact it is computed from).

## Validation beyond unit tests

The exact `inject → python -m build → verify` chain the workflow runs was executed locally against a **real wheel + sdist** — gate passed, placeholder restored. Full producers suite: 339 passed.

## One-time setup this expects (documented in the workflow header)

GitHub environments `testpypi`, `pypi`, `release-promote` (latter two with required reviewers). GCP side is already live and verified (repos with enforced immutable tags, allUsers reader on public only, WIF, budget guardrail — evidence on #349).

Next increments per #349: bootstrap `--image`/default-constant wiring + mode-aware `--preflight`, `bqaa-otel teardown`, docs flip, release-notes template.

Part of #349.

### DIFF
--- a/.github/workflows/release-tracing.yml
+++ b/.github/workflows/release-tracing.yml
@@ -294,18 +294,21 @@ jobs:
           path: dist/
 
       - name: Strip non-PyPI artifacts
-        # The plugin tarball is a release artifact, not a PyPI
-        # distribution. Filename uses hyphens in
-        # 'bigquery-agent-analytics-tracing-claude-code-*' so it's
-        # distinguishable from the underscore-named sdist.
+        # The plugin tarball and SHA256SUMS are release artifacts, not
+        # PyPI distributions — twine rejects unknown file types, and
+        # checksums belong on the GitHub release only.
         run: |
           rm -f dist/bigquery-agent-analytics-tracing-claude-code-*.tar.gz
+          rm -f dist/SHA256SUMS
           ls -la dist/
 
       - uses: pypa/gh-action-pypi-publish@release/v1
+        # No skip-existing: the version-burn rule requires byte-identical
+        # artifacts at the same version on TestPyPI and PyPI. If the
+        # version already exists, the upload MUST fail loudly — a burned
+        # version silently skipping would make the gate test old bytes.
         with:
           repository-url: https://test.pypi.org/legacy/
-          skip-existing: true
 
   promote:
     name: Promote image + publish release
@@ -321,15 +324,23 @@ jobs:
       contents: write
       id-token: write
     steps:
+      - name: Install crane (pinned + checksum-verified, BEFORE credentials exist)
+        # Supply-chain guard: never fetch a mutable 'latest' binary in a
+        # job that holds release credentials. Pinned version, pinned
+        # sha256, and installed before the auth step mints any token.
+        env:
+          CRANE_VERSION: v0.20.2
+          CRANE_SHA256: c14340087103ba9dadf61d45acd20675490fd0ccbd56ac7901fc1b502137f44b
+        run: |
+          curl -sL -o crane.tar.gz \
+            "https://github.com/google/go-containerregistry/releases/download/${CRANE_VERSION}/go-containerregistry_Linux_x86_64.tar.gz"
+          echo "${CRANE_SHA256}  crane.tar.gz" | sha256sum -c -
+          tar -xzf crane.tar.gz crane && sudo mv crane /usr/local/bin/
+
       - uses: google-github-actions/auth@v2
         with:
           workload_identity_provider: ${{ env.WIF_PROVIDER }}
           service_account: ${{ env.RELEASE_SA }}
-
-      - name: Install crane
-        run: |
-          curl -sL "https://github.com/google/go-containerregistry/releases/latest/download/go-containerregistry_Linux_x86_64.tar.gz" \
-            | tar -xz crane && sudo mv crane /usr/local/bin/
 
       - name: Promote staging digest to the public coordinate
         # Content-addressed copy: the digest the artifacts already embed
@@ -372,6 +383,7 @@ jobs:
       - name: Strip non-PyPI artifacts
         run: |
           rm -f dist/bigquery-agent-analytics-tracing-claude-code-*.tar.gz
+          rm -f dist/SHA256SUMS
           ls -la dist/
 
       - uses: pypa/gh-action-pypi-publish@release/v1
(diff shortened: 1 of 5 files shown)


caohy1988:
Fresh review of #352. The overall staging -> digest capture -> source injection -> promote shape matches the #349 contract, and the helper is well covered locally. I found three release blockers that will only show up on the real tag workflow, plus one runbook drift item.

1. **P1: PyPI/TestPyPI jobs still upload `SHA256SUMS`.**  
   `.github/workflows/release-tracing.yml:231` writes `SHA256SUMS` into the same `dist/` artifact that the PyPI jobs later download. The strip steps at `.github/workflows/release-tracing.yml:296` and `.github/workflows/release-tracing.yml:372` only remove the Claude plugin tarball, leaving `dist/SHA256SUMS` beside the wheel and sdist when `pypa/gh-action-pypi-publish` runs. The PyPA action defaults to uploading distribution files from `dist/`, and its docs call out `packages-dir` as the way to point it somewhere else. This means the first TestPyPI/PyPI publish can fail or try to process a non-distribution file. Fix by staging only the wheel + sdist into a clean upload directory for both publish jobs, or by removing `SHA256SUMS` before the action runs, and add a regression test/check that the PyPI upload directory contains only valid Python distributions.

2. **P1: `skip-existing: true` breaks the version-burn / byte-identical gate.**  
   `.github/workflows/release-tracing.yml:308` lets TestPyPI silently reuse files from a previous attempt for the same version. That is the opposite of the #349 contract: TestPyPI and PyPI must carry byte-identical artifacts, and a failed candidate burns the version. If a tag job is rerun after a partial failure or after any build-side change, TestPyPI can remain on the old wheel/sdist while the current GitHub draft and eventual PyPI upload use the newly built artifacts. The release would then have "passed" the lifecycle gate against bytes that are not the bytes promoted to customers. Remove `skip-existing: true` so duplicate TestPyPI uploads fail loudly, or add a hard preflight that fetches the existing TestPyPI hashes and requires exact equality with this run's `SHA256SUMS` before continuing.

3. **P1: the promoted release path executes mutable `crane` under release credentials.**  
   `.github/workflows/release-tracing.yml:329` downloads `go-containerregistry` from the `latest` release URL and pipes it straight into `tar`, then runs that binary in the job with `contents: write` and GCP release-publisher identity. This is the most privileged path in the workflow: it copies the digest to the public coordinate and publishes the draft release. A mutable upstream `latest` asset can change underneath an approved release, which undermines the reproducibility work this PR is adding. Pin a specific `go-containerregistry` version and verify the tarball SHA256 before executing it, or use a pinned action/container digest for the promotion tool.

4. **P2: the maintainer release runbook is now stale enough to misoperate the pipeline.**  
   `producers/RELEASING.md` still describes the old five-minute wheel/plugin release: no receiver image build, no draft-until-promote release, no TestPyPI full-lifecycle gate, no `release-promote` environment, and no version-burn behavior. It also says a failed TestPyPI trusted-publisher setup can be fixed by rerunning the publish job while the GitHub release is still good, which is no longer the right mental model once image digest promotion and version burn are part of the release contract. Please update the runbook before this workflow is usable for an actual tag.

Validation I ran:
- `gh pr checks 352` -> all 20 checks pass.
- `uv run pytest tests/test_release_image_tool.py -q` from `producers/` -> 13 passed.

caohy1988:
Fresh re-review on head `f6468d6`.

No new findings. The four prior issues are resolved:

- PyPI/TestPyPI upload hygiene: both publish jobs now strip the plugin tarball and `SHA256SUMS` before invoking `pypa/gh-action-pypi-publish`, so the upload set is wheel + sdist only.
- Version-burn contract: `skip-existing` is gone, so a reused TestPyPI version fails loudly instead of validating stale bytes.
- Promotion tool supply chain: `crane` is pinned to `v0.20.2`, the Linux x86_64 tarball checksum matches upstream `checksums.txt`, and the install runs before WIF auth mints release credentials.
- Release runbook: `producers/RELEASING.md` now describes the actual staging -> TestPyPI gate -> promote -> PyPI flow, including version burn and the post-promotion no-`--image` smoke.

I also rechecked the image/package boundary: the Docker image is built from the pre-injection tree and self-tests its package version; `_release.py` is injected only before wheel/sdist build, which is the right digest chicken-and-egg behavior.

Validation:
- `gh pr checks 352` -> 20 checks pass.
- `uv run pytest tests/test_release_image_tool.py -q` from `producers/` -> 13 passed.
- Verified `go-containerregistry` `v0.20.2` release metadata and checksum: `c14340087103ba9dadf61d45acd20675490fd0ccbd56ac7901fc1b502137f44b  go-containerregistry_Linux_x86_64.tar.gz`.

Residual note, not a blocker for this PR: the embedded `RELEASE_IMAGE` is still not consumed by bootstrap on this branch. That is intentionally the next #349 increment; do not cut the actual 0.2.0 tag until that PR lands and the no-repo-checkout lifecycle gate is runnable.

haiyuan-eng-google (approved):

PR #345, opened 2026-07-07
docs+scripts(demo): hero-story runbook, SQL pack, live-validated scripts (#344 PR1+PR2)

caohy1988:
First increment of #344, per the [reviewed sequencing](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/344#issuecomment-4908705053): **docs + SQL only, zero cloud mutation**, so the leadership story is reviewable independently of the scripts.

## What's here

- **README.md** — the presenter track: a one-screen demo contract (audience / promise / success screen / clock), three acts (~12 min rehearse mode), the privacy refusal as a scripted beat, and the precision-reviewed wording: *one command deploys the collector and warehouse surface; two generated artifacts enable the sources*. Explicit caveats: traces ≠ replay, dollars are estimates, privacy claim scoped to this warehouse.
- **OPERATOR.md** — fresh-project track with the clock defined precisely (≤30 min, Cloud Build ~10, prerequisites excluded), a prerequisites table (incl. Claude/Codex CLI **auth** and the codex ≥ 0.142.5 pin), and a troubleshooting table where **every row is a failure that actually happened** in the #324/#317 live runs.
- **sql/00–06** — leadership-question-first queries. Every query is bounded by `@demo_run_id` (carried as the `env` resource attribute by both products — Codex can't set arbitrary resource attributes, so `[otel] environment` is the carrier) plus `@window_hours`. Healthy zeros return **status rows**, never empty results. `02` measures tokens and labels dollars as an estimate from an editable rates CTE with a visible as-of date. `05`'s privacy proof searches for the **exact scripted prompt string** across content-bearing columns and returns PASS/FAIL rows. `06` shows the native escape hatch for unmapped events (backed by a real observation).
- **evidence/template.md** — per-run record: redaction rules (no tokens, raw ids/emails, prompts, full receiver URL) and version capture (Codex behavior is version-sensitive).
- **one_pager.md** — the forwardable artifact: architecture line, run-number slots, scoped claims.

## Validation

All seven queries executed against the **live e2e dataset** (real Claude + Codex rows): six passed first try; `05` failed on a JSON-vs-STRING `COALESCE` (BigQuery JSON columns need `TO_JSON_STRING` before text search), fixed and re-verified — live result: `prompt_text_in_logs PASS / prompt_text_in_projection PASS / dead_letter_count PASS 0`.

[...]

PR #355, opened 2026-07-09
feat(otlp): mode-aware bootstrap --preflight + bqaa-otel teardown (#349 PR3)

caohy1988:
Third implementation increment of #349: the two subcommands the TestPyPI full-lifecycle gate depends on — TDD (25 new tests RED-first), productized from logic already proven live in the hero demo, with every demo assumption the issue reviews flagged removed.

## `bootstrap --preflight`

- **Mode-aware permission probe** via the Resource Manager `testIamPermissions` REST call (gcloud has no such subcommand — learned live in #345): 18 permissions covering the real bootstrap surface in default prebuilt-image mode; the two build permissions added **only** for `--build-from-source`. Acceptance-criterion behavior: passes on a project with no Cloud Build/AR permissions.
- **No product CLIs required** — the demo-session assumption the review caught; `claude`/`codex` probes are opt-in via `--check-products`.
- Org-policy check for the `allUsers` invoker grant with an honest WARN when the effective policy is unreadable; every failure carries its remediation.

## `bqaa-otel teardown`

- Consumes `inventory.json` from `bootstrap --execute` (PR 2); without one, **reconstructs from the deterministic package constants** — never assuming an AR repo (released-image mode creates none; a source-mode inventory's repo is deleted exactly once).
- **Safety stack**: `--project`/`--dataset` must match the inventory (wrong file → refused); every name must match the `bqaa` allowlist patterns (poisoned inventory → refused); **dry-run by default, and the dry run executes literally nothing** — not even read-only listings, so the plan works offline; the opaque DTS config name is queried at `--confirm` time with legacy unsuffixed configs matched by query text.
- **Verification is the authority**: probes PASS only on known not-found stderr; permission/API errors are UNVERIFIABLE failures; a failed IAM policy read is never "binding gone".

## Validation

25 new tests RED-first; full producers suite **382 passed**. Live: `--preflight` ran 8/8 green against the real GCP project (real REST probe, 18/18 permissions in prebuilt mode, real org-policy read); `teardown` dry-run reconstructed the complete plan offline with zero commands executed.

With this, the no-checkout lifecycle (`pipx install → bootstrap --preflight → bootstrap --image <staging> --execute → verify --smoke → teardown`) is fully expressible — next step after merge is running that gate for real.

Part of #349.

### DIFF
--- a/producers/src/bigquery_agent_analytics_tracing/otlp/cli.py
+++ b/producers/src/bigquery_agent_analytics_tracing/otlp/cli.py
@@ -468,6 +468,11 @@ def _cmd_teardown(args: argparse.Namespace) -> int:
     )
   except ValueError as exc:
     return _report_settings_error(exc)
+  except RuntimeError as exc:
+    # Refused precondition (e.g. unreadable DTS listing before a
+    # destructive run) — an admin-facing error, never a traceback.
+    print(f"bqaa-otel: error: {exc}", file=sys.stderr)
+    return 1
   return 0 if (report.dry_run or report.ok) else 1
 
 
--- a/producers/src/bigquery_agent_analytics_tracing/otlp/cli.py
+++ b/producers/src/bigquery_agent_analytics_tracing/otlp/cli.py
@@ -31,6 +31,8 @@ import urllib.parse
 
 from . import bootstrap as bootstrap_mod
 from . import config_artifacts
+from . import preflight as preflight_mod
+from . import teardown as teardown_mod
 from . import verify as verify_mod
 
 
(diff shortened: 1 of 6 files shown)


caohy1988:
Full fresh review on #355. Two findings:

1. P1 `producers/src/bigquery_agent_analytics_tracing/otlp/teardown.py:167`
   `--confirm` uses `runner.try_run(...)` for the pre-delete DTS transfer-config listing. If that listing is unreadable or transiently fails, teardown silently skips the scheduled MERGE deletion, then continues deleting the dataset and the shared pipeline and only reports the DTS problem during post-verification. That is the wrong failure order for the resource called out as the long-lived billing risk: a broken DTS lookup should abort before any destructive work starts.

   Recommended fix: make the initial DTS lookup a hard `runner.run(...)` precondition for `--confirm` (or convert the failure into a fatal `ValueError`/failed report before appending/executing deletions). Add a regression test where the transfer-config listing fails and assert no delete commands are executed.

2. P2 `producers/tests/test_otlp_teardown.py:233`
   `test_missing_inventory_reconstructs_from_constants` cannot fail because the only assertion ends in `or True`, and `report.checks` is intentionally empty for dry-run. That leaves the no-inventory reconstruction contract untested even though it is the fallback path release users will hit when they do not have `inventory.json`.

   Recommended fix: capture the dry-run `echo` lines and assert the reconstructed plan includes the deterministic resources (`bqaa-otlp-receiver`, `bqaa-otlp-token`, `my-proj:ds1`) and omits the AR repo in released/reconstructed mode.

Verification I ran locally:

- `uv run pytest producers/tests/test_otlp_cli.py producers/tests/test_otlp_preflight.py producers/tests/test_otlp_teardown.py -q` -> 51 passed
- `uv run pytest producers/tests -q` -> 382 passed, 4 skipped
- `uv run pyink --check ...` on changed Python files -> passed
- `uv run isort --check-only ...` on changed Python files -> passed
- GitHub checks are green.

caohy1988:
Fresh re-review on head `7a982b8`.

One remaining finding:

1. P2 `producers/src/bigquery_agent_analytics_tracing/otlp/cli.py:456` / `producers/src/bigquery_agent_analytics_tracing/otlp/teardown.py:178`
   The DTS-listing safety gate is now correct in `run_teardown`, but it raises `RuntimeError` and `_cmd_teardown` only catches `ValueError`. On the real `bqaa-otel teardown --confirm` path, the admin will get a Python traceback for the “cannot list DTS scheduled queries” precondition failure instead of the clean release-gate error this PR is trying to productize. Safety is preserved, but the no-checkout customer lifecycle now has a rough failure mode exactly where we expect permission/API issues.

   Recommended fix: catch this precondition failure at the CLI boundary (for example `except RuntimeError as exc: print(f"bqaa-otel: error: {exc}", file=sys.stderr); return 1`) or use a domain exception that `_cmd_teardown` catches. Add a CLI-level regression test that monkeypatches `teardown.run_teardown` to raise the DTS-listing failure and asserts exit nonzero + readable stderr + no traceback.

The two earlier findings are otherwise fixed:

- DTS listing failure now refuses before any destructive command.
- The missing-inventory reconstruction test no longer has the vacuous `or True` and now asserts the actual dry-run plan.

Verification I ran locally:

- `uv run pytest producers/tests/test_otlp_cli.py producers/tests/test_otlp_preflight.py producers/tests/test_otlp_teardown.py -q` -> 52 passed
- `uv run pytest producers/tests -q` -> 383 passed, 4 skipped
- `uv run pyink --check ...` on changed Python files -> passed
- `uv run isort --check-only ...` on changed Python files -> passed
- GitHub checks are green.

caohy1988:
Fresh re-review on head `e5b2618`: no remaining issues found.

The last P2 is fixed correctly:

- `_cmd_teardown` catches the teardown precondition `RuntimeError` and surfaces it as `bqaa-otel: error: ...` with exit 1, not a traceback.
- The CLI regression test covers exit code, DTS message content, and absence of `Traceback`.

Re-checked the prior fixes as well:

- `teardown --confirm` refuses before any destructive command if the DTS listing is unreadable.
- Missing-inventory reconstruction now has real dry-run plan assertions and no vacuous `or True`.

Verification:

- `uv run pytest producers/tests/test_otlp_cli.py producers/tests/test_otlp_preflight.py producers/tests/test_otlp_teardown.py -q` -> 53 passed
- `uv run pytest producers/tests -q` -> 384 passed, 4 skipped
- `uv run pyink --check ...` on changed Python files -> passed
- `uv run isort --check-only ...` on changed Python files -> passed
- GitHub checks are green.

This is merge-ready from my review.

haiyuan-eng-google (approved):

PR #371, opened 2026-07-16
feat(trace): identity-safe trace selection SQL and candidate resolution (#359 U2)

caohy1988:
## Summary

U2 of the identity-safe trace evaluation plan (#361): resolves the core of #359 by making trace selection SQL, candidate resolution, and trace construction identity-safe on the U1 contract merged in #362.

Live-data characterization was completed first, as required by the plan. Across four live `agent_events` datasets (14,163 sessions / 106.5k rows), `user_id`, `root_agent_name`, and `experiment_id` were row-uniform within every session, while custom tags showed additive per-row enrichment. The evidence and resulting design decisions are recorded on #361.

## Changes

- **Composite NULL-safe anchoring** — list and singular reads anchor rows by `(session_id, user_id, root_agent_name)` with NULL-safe equality, preventing a reused session ID from absorbing another intrinsic identity. Malformed bulk rows quarantine only their independently validated projected SQL anchor.
- **Exact scope semantics** — distinct experiment/tag payloads remain distinct scope candidates; strict supersets are never collapsed. Only rows with no custom-tag payload (`NULL`, JSON null, or `{}`) are shared across scopes. A tagged row missing a pinned label belongs to another pass and is excluded.
- **Mixed-scope escape hatch** — `allow_mixed_scope=True` returns a conversation-complete trace only after selector-aware intrinsic-identity uniqueness is established. It carries `scope=None` and bounded `scope_coverage`; exact selectors still win.
- **Typed singular resolution** — `get_session_trace()` supports three-state identity/scope pins, and `get_trace_by_selector()` is the executable retry surface for `AmbiguousSessionError` payloads. Candidate populations are bounded, deterministically ordered, truncation-aware, and never fall back to newest-wins.
- **Attested discovery and materialization** — all discovery paths carry `JSON_TYPE(attributes)` attestation so non-object JSON scalars cannot fabricate candidates, consume enumeration capacity, or produce non-executable retries. Fetched rows are attested and normalized before span parsing.
- **Identity/scope-safe construction** — rows are grouped by full resolved identity and exact scope, shared spans are copied only when needed, limits apply after deterministic ranking, and trace IDs cannot cross scopes.
- **Bounded query work** — complete candidate pages are reused for mixed-scope identity decisions; redundant identity discovery is reserved for truncated or otherwise insufficient pages.

## Verification

- Focused identity suite: **130 passed / 34 opt-in live tests skipped**.
[...]

PR #362, opened 2026-07-11
feat(trace): identity, selector, and ambiguity contract (#359 U1)

caohy1988:
## Summary

First unit (U1) of the identity-safe trace evaluation plan (#361, tracking #359): adds the shared identity/selector/ambiguity contract that the follow-up SQL, propagation, and judge-context changes build on. Models and serialization only — no query or client behavior changes yet.

- **`TraceIdentity(session_id, user_id, root_agent_name)`** — frozen intrinsic identity of a scoped multi-turn session; `trace_id` deliberately excluded (one session may span multiple OTel trace IDs).
- **`TraceScope(experiment_id, custom_labels)`** — caller-selected scope; labels are canonicalized to a sorted key/value tuple so equality, hashing, and the string `scope_signature` are independent of JSON key order.
- **`TraceSelector`** — optional caller pins, mapped onto the existing `TraceFilter` via `to_trace_filter()` so both surfaces share one predicate implementation.
- **`ResolvedTraceSelector`** — identity + scope candidate; same-session/different-scope candidates stay distinct.
- **`AmbiguousSessionError`** — typed ambiguity result for singular lookups; subclasses `ValueError` so existing `except ValueError` callers degrade gracefully. Printable form is redacted to candidate count + retry-dimension names; structured candidates (identity dimensions and scope signatures only, never event content) are available via `.candidates` / `.to_dict()`.
- **`resolve_singular_candidate()`** — legacy session-only keys accepted only when exactly one candidate exists; zero candidates preserves the existing `ValueError` shape; no newest-wins fallback.
- **`Trace`** gains additive `identity`/`scope` fields; all legacy scalar fields unchanged.

## Review checkpoint

Per the plan, this PR is the U1 public-contract checkpoint: exact model fields, hash/equality behavior, the ambiguity payload boundary (redacted printable form vs structured candidates), and the legacy session-only key rules. Accepting it fixes the identity contract for the follow-up units (identity-safe trace SQL, surface propagation, judge context).

## Test plan

- 27 new tests written first (red observed as `ImportError` before implementation), covering: identity/selector hash-equality and dedup, label canonicalization vs dict order, `TraceFilter` mapping, ambiguity-error redaction (asserts no user IDs, session IDs, or label values in `str()`), structured payload JSON-safety, legacy single/multi/zero-candidate rules, order-independence (no newest-wins), and additive `Trace` serialization with NULL identity fields.
- Full suite: 3389 passed, 30 skipped (opt-in live BigQuery gates).
[...]

Issue #361, opened 2026-07-11
[Tracking] Identity-safe trace evaluation plan for #358-#360


caohy1988:
---
title: Identity-Safe Trace Evaluation - Plan
type: fix
date: 2026-07-11
deepened: 2026-07-11
artifact_contract: ce-unified-plan/v1
artifact_readiness: implementation-ready
product_contract_source: ce-plan-bootstrap
execution: code
---

# Identity-Safe Trace Evaluation - Plan

## Goal Capsule

- **Objective:** Resolve issues #359, #358, and #360 in dependency order so trace retrieval, categorical judging, and the skill-evolution demo cannot silently mix reused session IDs that differ in identity or resolved scope, or bind a golden answer to the wrong conversation.
- **Authority:** The issue contracts and current repository behavior are authoritative. PR #351 remains example-scoped; its withdrawn SDK commits are reference implementations, not changes to restore wholesale.
- **Execution profile:** Land #359 first as the shared identity and scope contract, #358 second on that contract, and #360 only after its corresponding SDK capability is available. Prefer one reviewable change per issue.
- **Compatibility posture:** Preserve `session_id` and existing serialized fields. Add identity and provenance fields without breaking old callers, except that previously unsafe ambiguous singular lookups now return a typed error instead of silently choosing or merging data.
- **Stop conditions:** Stop and revisit the plan if producer data disproves the documented session/trace semantics, if the public response additions or the new fail-closed ambiguity behavior require a major-version contract, or if live BigQuery validation shows the selected scope cannot be reconstructed deterministically.
- **Tail ownership:** #360 owns demo and documentation cleanup. It must not be folded into either SDK-core change, and it must remain safe when #358 and #359 land at different times.

---

### Summary

The SDK will distinguish a persistent conversation session from the scope used to select one recorded run of that session. Trace reads will carry a complete identity, apply caller-selected labels and experiment scope to the fetched rows, and reject ambiguous singular lookups. Categorical judge context will use the same identity across BigQuery-native generation, retry, and API fallback paths. Once each SDK capability lands, the skill-evolution lab can remove its corresponding workaround without overstating where judging or trace evidence came from.

### Problem Frame

[...]

Issue #358, opened 2026-07-10
evaluate_categorical: support per-session judge context (golden expected answers) on the server-side paths


evekhm:
## Problem

Flagged in PR #351's review (P1 #1, gate integrity): on the BigQuery path, the server-side categorical judge (`AI.GENERATE`) never receives the matched golden expected answer for a session. `scope`/`ground_truth` ground the judge globally, but per-session expected answers reach the judge only on the API/conversations path. Consequently `matched_meaningful_rate` on the BigQuery path is an LLM estimate of correctness: a fluent, confidently wrong answer can score `meaningful` — and any promotion gate built on that rate (e.g. the skill-evolution example's keep-only-if-better gate) can pass a wrong-answer candidate.

## Suggested fix

Working implementation existed on the PR #351 branch (commit `da52c54`, later withdrawn to keep that PR example-scoped — SDK-core changes deserve their own review):

1. `Client.evaluate_categorical(..., per_session_context: dict[str, str] | None = None)` — map of `session_id -> judge context` (e.g. `"Golden expected answer: <text>"`).
2. `build_ai_generate_query`: when context is present, add an `UNNEST(@session_contexts)` array-of-struct parameter joined on `session_id`, and append `COALESCE(CONCAT('\n\n', sc.context), '')` to the per-session prompt.
3. Skip the `AI.CLASSIFY` shortcut when `per_session_context` is set (CLASSIFY takes only transcript + categories and would silently drop the context), starting at `AI.GENERATE` instead.
4. The API fallback (`classify_sessions_via_api`) already accepts `per_session_context` — only threading is needed.

Callers (e.g. `scripts/quality_report.py`) can then match golden Q&A to sessions before evaluation and pass the expected answers through, giving the BigQuery path the same expected-answer grounding as the conversations path.

## References
- PR #351 review, P1 finding #1 ("The BigQuery-only promotion gate is not expected-answer-grounded")
- Withdrawn implementation: PR #351 branch commit `da52c54`

caohy1988:
Reviewed against `origin/main`, PR #351 head `e98b7c2`, and the withdrawn implementation in `da52c54`.

[...]

PR #375, opened 2026-07-24
feat(trace): propagate identity-safe selectors across secondary surfaces (#359 U3)

### LINKED ISSUE #359: Trace queries: row scope ignores pass labels and full session identity (session_id reuse mixes traces and judge transcripts)
## Problem

Two related findings from PR #351's review (P1 #3 + the trace-merging thread). Trace queries select matching session_ids first, then fetch **every** row for those ids — and the row fetch re-joins on `session_id` alone:

1. **Scoring passes merge.** Two passes (e.g. a V0 and a V1 evaluation) can legitimately reuse the same session_ids under different `custom_tags`; the unscoped row fetch merges both passes' rows into one trace.
2. **Users/apps merge.** Row scope omits `user_id` and `root_agent_name`, so a reused session_id from another user or app mixes foreign rows into the returned trace — and into the judge transcript built from it.

`_GET_SESSION_TRACE_QUERY` (`get_session_trace(session_id)`) has the same gap: `WHERE session_id = @session_id` with no identity scoping at all.

## Suggested fix

Working implementations existed on the PR #351 branch and were withdrawn to keep that PR example-scoped (label scoping: commit `94ff0e7`; composite identity: PR discussion):

1. **`TraceFilter.row_scope_where()`** — row-level conditions (reusing the parameters `to_sql_conditions()` already emits) for `custom_labels`, `experiment_id`, `user_id`, and `root_agent_name`; `_LIST_TRACES_QUERY` applies it to the outer row fetch as `{row_where}`.
2. **Composite anchor join** in `_LIST_TRACES_QUERY` — anchor each selected session to its identity and join NULL-safely, so protection holds even when the filter pins nothing:
   ```sql
   WITH trace_sessions AS (
     SELECT session_id, user_id,
            JSON_VALUE(attributes, '$.root_agent_name') AS root_agent_name,
            MAX(timestamp) AS last_event_ts
     FROM `...` WHERE {where}
     GROUP BY session_id, user_id, root_agent_name
     ORDER BY last_event_ts DESC, session_id LIMIT @trace_limit)
   ...
   JOIN trace_sessions ts
     ON e.session_id = ts.session_id
     AND e.user_id IS NOT DISTINCT FROM ts.user_id
     AND JSON_VALUE(e.attributes, '$.root_agent_name')
         IS NOT DISTINCT FROM ts.root_agent_name
   ```
3. **`get_session_trace(session_id, user_id=None, root_agent_name=None)`** — optional identity pins; when absent, anchor on the identity of the session's most recent row via a one-row CTE, and return only that identity's rows.
4. **`_build_traces_from_rows`** — group rows by `(session_id, user_id, root_agent_name)` instead of `session_id` alone, so two identities sharing an id yield two `Trace` objects instead of silently merging.

## References
- PR #351 review, P1 finding #3 ("Trace row scoping is incomplete")
- Withdrawn label-scoping implementation: PR #351 branch commit `94ff0e7`

caohy1988:
## Summary

Completes U3 of the #361 sequencing plan by propagating the identity/scope selector and ambiguity contract from #371 through every secondary trace-reading surface.

- makes GQL reconstruction resolve the authoritative flat identity/scope first and confines graph edges to that exact span population
- routes `BigQueryTraceEvaluator`, replay, and comparison through `TraceSelector`, while pushing evaluator event-type filtering into BigQuery after complete candidate resolution
- adds identity/scope pins, exact selector JSON retries, and structured `AmbiguousSessionError` responses to the CLI, Remote Function, and self-monitoring agent example
- preserves colliding identities/scopes in quality and latency reports, correlates scores and sampled trajectories fail-closed, and completeness-probes explicit session populations instead of silently truncating at 64 candidates
- documents the migration contract, sensitive retry payloads, mixed-scope behavior, and the intentionally deferred #358 judge-context and persistence/view work

## Correctness details

- scalar selector pins preserve all three states: omitted/`UNSET`, explicit SQL `NULL`, and exact strings
- retry surfaces carry the complete `candidate["selector"]`, including `scope_signature`
- GQL supplies relationships only; event content and attribution remain sourced from the exact flat trace
- evaluator tool-call extraction remains correct when stable SQL tie-breaks place a terminal tool event before its matching start
- session-only categorical scores never choose or render one candidate from a reused-session collision
- report trajectory re-fetches retain only the score-attributed identity/scope population

## Verification

- focused U3/U2 surface suite: **606 passed**
- full offline suite after merging current `main`: **3948 passed, 48 skipped**
- live BigQuery collision/dry-run suite: **34 passed**
- `autoformat.sh`: clean
- `git diff --check`: clean

## Sequencing

This PR follows merged U1 (#362) and U2 (#371). It completes the release-boundary plumbing required before the identity-safe trace work can ship.

The following #358 units remain intentionally separate:

- U4: bind per-session judge context to resolved identity across AI.GENERATE, retry, and Gemini API fallback
- U5: migrate persistence cardinality in schema → writer → view order

Closes #359
Updates #361


caohy1988:
## U3 round 1 fresh review — no blocking findings

Reviewed exact head `671d333` (PR #375) against `main`: a full-diff pass over all 28 files, a line-by-line read of the core changes (`client.py` event-type plumbing and GQL rewrite, `trace_evaluator.py` resolver migration, CLI/Remote Function ambiguity surfaces, report attribution), and behavioral probes executed against the head. This surface-propagation unit is in much better shape on arrival than U2 was — no P1 or P2 findings; three P3s below.

**Verified sound (the load-bearing claims, each checked against the code or by execution):**

- **Contract fidelity.** Every secondary surface routes through the shared U2 resolver: CLI `get-trace` (selector options + `--selector-json` for explicit-null pins), Remote Function `analyze` (`selector` object with additive-field conflict rejection), `BigQueryTraceEvaluator` (async wrapper over `get_trace_by_selector`), replay/compare (selector pass-through with session-id consistency checks), and the eval-dataset items. Ambiguity payloads round-trip: the CLI emits `to_dict()` JSON on stdout with exit code 2, the Remote Function attaches `details` to `_error`, and `TraceSelector(**payload)` reconstructs exact pins including explicit nulls.
- **GQL reconstruction is a genuine hardening, not just propagation.** The old edge query's `OR` predicate could pull cross-session edges, and the old span-merge path could fabricate span rows from graph data. Now the flat resolver defines the exact span population, `parent/child.span_id IN UNNEST(@span_ids)` pins traversal to it, graph rows contribute parent links only onto deep-copied flat spans, duplicate span-ids skip GQL with a warning, and the manager-level convenience path resolves identity-safely (raising on ambiguity instead of broadening). Identity/scope/coverage metadata is preserved on the reconstructed trace.
- **Event-type pushdown stays scope-sound.** Candidate discovery is never event-filtered; the restriction applies only to the authoritative span fetch. The mixed path fetches a lightweight attributes-metadata companion (attested, anchored) so coverage and conjunctive Python-only pin verification still see the complete scope population while excluded payloads never cross the wire. `_materialize_event_filtered_trace` assigns rows with the same exact-payload/local-shared/identity-shared rules as the builder and revalidates identity per row; `event_types` is normalized through the validating `TraceFilter` container, closing the lying-container hole at this new boundary.
- **Reports fail closed on collisions.** `quality_report` matches scores to resolved contexts via the reserved `SessionScore.details` attribution keys, retains colliding rows when attribution disambiguates, warns and refuses to guess when it cannot, and probes listing completeness (doubling the limit until a short page) instead of trusting an exactly-full page. The round-7-era "KNOWN LIMITATION (#359)" comments are gone along with the limitation.
- **Docs and release gating.** CHANGELOG documents U1+U2+U3 as one release boundary with concrete migration guidance (catch `AmbiguousSessionError`, pass the complete `candidate["selector"]`, preserve explicit nulls and `scope_signature`); the evaluator's behavior change (reused session ids now raise instead of silently merging) is covered by that guidance.
- **Verification reproduced:** focused surface suites 648 passed; full offline suite 3,892 passed / 15 skipped; live BigQuery suite 34 passed (the `test_cli_agent_tool.py` collection exclusion is local `adk-python` environment drift, green in remote CI).

Findings:

- **P3 — the CLI omits the mixed-scope escape hatch that the Remote Function exposes** (`cli.py get_trace`; `dispatch.py` accepts `allow_mixed_scope`). A CLI user who hits `AmbiguousSessionError` on a single-identity multi-scope session gets the structured payload and can retry one exact candidate, but cannot request the conversation-complete KTD4 read at all. Either add `--allow-mixed-scope` (documenting that cross-identity ambiguity still raises) or note the asymmetry as intended in SDK.md.
- **P3 — event-filtered exact reads flip `trace_id` to the session fallback when the filter excludes the scope's tagged rows** (`_materialize_event_filtered_trace`, `own_trace_ids` collected only from tagged rows). Executed: a labels-scoped trace whose tagged row is `LLM_RESPONSE` reports its genuine trace id unfiltered but `trace_id == session_id` under `event_types=["USER_MESSAGE_RECEIVED"]`, so the same resolved scope's `trace_id` varies with the event filter — the round-8 P3-10 consistency family. Consider deriving the id from the unfiltered scope proof (the resolution page already knows the scope exists) or documenting the fallback.
- **P3 — empty-result semantics under `event_types` are undocumented.** Executed: an event filter matching no rows returns a zero-span `Trace` (identity/scope attached) rather than the "No events found" error the unfiltered read raises, on both the exact and mixed paths. Reasonable semantics, but neither `get_trace_by_selector` nor `get_session_trace` documents that an event filter suppresses the not-found error; consumers assuming `spans` is non-empty (e.g. `final_response` derivation) should be warned.

**Verdict:** approve with the three P3s as follow-ups — none blocks merge. After merge, the #361 sequencing gates read satisfied for #359 closure: U2+U3 land at one release boundary and the live collision fixture passes (34/34).

caohy1988:
> the CLI omits the mixed-scope escape hatch that the Remote Function exposes

Fixed in `b1a911b`. `bq-agent-sdk get-trace` now accepts `--allow-mixed-scope` for both session selectors and `--selector-json`. It delegates to the shared resolver's KTD4 path, and both CLI help and `SDK.md` state that cross-identity ambiguity still raises. Combining the flag with `--trace-id` is rejected because mixed scope is a session-selector operation.

> event-filtered exact reads flip `trace_id` to the session fallback when the filter excludes the scope's tagged rows

Fixed in `b1a911b`. Candidate discovery now carries the first producer `trace_id` for each exact tagged scope, ordered with the same stable columns as the authoritative fetch. `_materialize_event_filtered_trace()` receives that unfiltered scope proof, so filtering out tagged rows—or matching no rows—no longer changes the resolved scope's `trace_id`. Untagged shared rows still cannot supply a scope-specific ID.

> empty-result semantics under `event_types` are undocumented

Fixed in `b1a911b`. Both `get_session_trace()` and `get_trace_by_selector()` now document that an event filter matching no payload rows returns an attributed zero-span `Trace` and does not raise the unfiltered not-found error. The behavior is locked by a doc-contract regression alongside the existing behavioral test.

Verification:

- 676 focused U2/U3 surface tests passed
- full offline suite: 3,951 passed / 48 skipped
- modified candidate query passed a table-backed live BigQuery dry run
- autoformat and `git diff --check` clean

The full scratch-dataset live fixture was also attempted, but Google API retry remained active for seven minutes and the run was interrupted; it is not included in the pass counts above.

haiyuan-eng-google (approved):

PR #374, opened 2026-07-24
feat(dashboard): add portable Looker Studio template for BQAA

caohy1988:
## Summary

Adds a self-contained Looker Studio dashboard package under `dashboard/looker_studio` for projects using the ADK BigQuery Agent Analytics plugin.

- published 8-page / 37-chart report template over BQAA-generated views;
- three-field browser configurator (project, dataset, table), with optional advanced view-prefix and billing-project overrides;
- authenticated CLI preflight for the BQAA base table and 15 required generated views;
- one stable embedded custom query over the ADK 1.27.0 static view intersection;
- reviewed compatibility profile, chart/listener/layout manifest, and 37 independent per-chart validation queries;
- Viewer's Credentials only, no persistent objects in the caller's dataset, and no raw `agent_events` reads by report charts.

Public configurator: https://caohy1988.github.io/bigquery-agent-analytics-dashboard/

Published template: https://lookerstudio.google.com/reporting/5a3f85ef-fc9c-4730-8ef2-8ef9129ddb40

## Portable configuration contract

The normal flow asks only for:

1. GCP project ID;
2. BigQuery dataset ID;
3. BQAA table ID (normally `agent_events`).

The browser uses the standard `v_*` generated views and the source project for billing by default. Newer ADK deployments can override a custom `view_prefix` and billing project under Advanced settings. The values are validated and encoded into the official Looker Studio Linking API `sqlReplace` configuration; Looker Studio query parameters are not used as BigQuery identifiers.

The authenticated helper provides the fail-closed path:

```sh
cd dashboard/looker_studio
python3 tools/hydrate_dashboard.py 
  --project PROJECT_ID 
  --dataset DATASET_ID 
  --table agent_events 
  --prefix v 
  --location US
```

## Compatibility

- requires the ADK BQAA plugin with `create_views=True`;
- frozen candidate profiles: ADK 1.27.0, 1.36.1, and 2.4.0;
- later releases may hydrate after the structural preflight succeeds, with the documented unverified-semantics qualification;
- the report reads the 15-view static intersection, so newer event views do not break the stable schema.

## Validation

- full SDK suite: **3,894 passed, 48 skipped**;
- dashboard contract tests: **10 passed**;
- exact SDK-copy live preflight: base table + **15/15 generated views** compatible;
- exact SDK-copy live smoke test: **37/37 chart queries** executed successfully against a real BQAA dataset;
[...]

PR #376, opened 2026-07-24
feat: query BQAA agent_events table directly

caohy1988:
## Summary

- replace the 15 generated-view `UNION ALL` with one date-pruned scan of the configured BQAA `agent_events` table
- reproduce the generated views' typed JSON extraction semantics for the 30-field dashboard schema while including all base-table event types
- change hydration and preflight from project/dataset/view-prefix to project/dataset/table; generated views and `create_views=True` are no longer required
- regenerate all 37 independent oracle queries against the raw table and update the public configurator, compatibility contract, docs, and live validator
- add deterministic SQL drift coverage, sequential `sqlReplace` collision checks, and a dated live-template attestation
- make the generated-report credential gate explicit: the canonical template is Viewer mode, but the Linking API-created data source can initially be Owner mode and must be changed to Viewer before sharing

## Validation

- `3902 passed, 48 skipped` in the full pytest suite
- `18 passed` in `tests/test_looker_studio_dashboard.py`
- browser configurator Node test passed
- real BQAA fixture: production query returned all 100,004 raw events; all 37 oracle queries executed successfully
- raw-table LLM token/row, tool, and error aggregates matched the corresponding generated views exactly
- a public-template smoke test used a temporary dataset containing only `agent_events` and zero views; all dashboard navigation and live charts loaded with no configuration errors
- live data-source editor verified the canonical template itself uses Viewer credentials; the generated-report onboarding now blocks sharing in documentation until the copied source is also Viewer mode

## Published template

The canonical Looker Studio template was updated to the reviewed base-table SQL. `bindings/report_template.yaml` records the repository SQL hash and the 2026-07-24 connector-review/table-only hydration/credential verification. The report remains a mutable contributor-managed external artifact pending transfer to Google-managed ownership.

Follow-up to #374 and #365.


caohy1988:
## Full pre-merge review at head `d2ed3dc`: no blocking findings

Reviewed the complete diff against merged main (`b523cac`), with the semantic-parity audit done against the actual ADK 1.27.0 plugin source (`_EVENT_VIEW_DEFS` / `_VIEW_SQL_TEMPLATE`), not just the repo's own artifacts.

### Semantic parity — verified field-by-field

- **Token fields.** The plugin's `v_llm_response` exposes content-derived token columns plus raw `usage_metadata` JSON; the previous dashboard applied `COALESCE(usage_metadata.total_token_count, usage_metadata.total_tokens, usage_total_tokens)` on top. The new template folds exactly that precedence into its `usage_*_tokens` columns, so measure-level results are identical.
- **Latency / tool / error fields** use the plugin's exact JSON paths (`latency_ms.total_ms`, `content.tool`, `content.tool_origin`, `ttft_ms`, `model_version`), gated per event type with `IF(event_type = …)` so rows from other types cannot contaminate measures.
- **Oracle queries.** Spot-checked eight across sessions, traces, users, events-by-agent, tokens, and latency percentiles against their pre-rewrite versions. Filtered charts are row-for-row equivalent (events-by-agent keeps the same `TOOL_COMPLETED` predicate; the view-emulation subqueries reproduce the view definitions exactly, including plain `CAST`).
- **The one deliberate change** — all-events tiles now include event types that had no generated view (+263 of 100,004 fixture rows) — is explicitly documented in `dashboard-implementation.md`, with the parity claim correctly scoped to view-covered aggregates.

### Prior-review items — all closed

1. **SQL drift gate**: `test_generated_sql_artifacts_cannot_drift` regenerates the logical template via `gen_events_tmpl.generate()`, renders through the real renderer with real bindings, and byte-compares every committed artifact. Verified locally: 18/18 pytest + the Node configurator suite pass at head.
2. **Sequential `sqlReplace`**: the ordered rule (a value may contain its own or an earlier sentinel, never a later one) is the precise correctness condition — a blanket ban would wrongly reject a dataset legitimately named like the dataset sentinel, and the allow-tests pin exactly that case. Implemented identically in JS and Python with duplicate-sentinel guards.
3. **Template attestation**: the recorded `reviewed_template_sql.sha256` matches the committed `events_v1.template.sql` byte-for-byte (independently recomputed). The `scope: repository_artifact_only` limitation and `CONTRIBUTOR_MANAGED → GOOGLE_MANAGED` transfer record state the trust boundary plainly.

### Informational (non-blocking)

- **`CAST` vs `SAFE_CAST` asymmetry**: the production template uses `SAFE_CAST` for token/latency extraction; the oracle queries emulate the views' plain `CAST`. Identical on well-formed data; on a corrupt non-numeric JSON value an oracle query errors while the production chart nulls the row. Defensible (the oracle's job is to mirror the views) — noting it so the divergence is on record.
- **The credential gate is documentation, not enforcement** — inherent to the Linking API, since the new data source is created under the clicking user and can start in Owner mode. The PR handles this as well as the platform allows: observed initial mode recorded in bindings, and the exact verification path (Resource → Manage added data sources → Edit → Data credentials: Viewer) appears in the configurator's step 3, a dedicated warning aside, and the README with Google's data-credentials link.

### Other checks

- Injection surface unchanged: all identifiers regex-validated fail-closed; every interpolation site lands inside backticks or quotes the regexes exclude; no XSS sinks in the configurator; no personal repository links.
- Preflight correctly narrows to the 15-column table contract and accepts `CLONE` alongside `BASE TABLE` (needed for the zero-copy sentinel fixture; harmless for users).
- zizmor jobs show "skipping" only because no workflow files changed in this PR — expected.
- The deployed configurator serves the previous version until this merges and the Pages workflow re-runs on `main` — expected.

The remaining open item is unchanged and outside this PR's control: the live Looker Studio template and sentinel fixture remain contributor-managed pending the maintainer-approved ownership transfer, which the bindings now record explicitly rather than papering over.

haiyuan-eng-google (approved):

PR #378, opened 2026-07-25
feat(eval): bind judge context to resolved trace identity

caohy1988:
## Summary

Categorical evaluations can now bind golden answers and other trusted judge-only context to an exact resolved trace identity and scope, preventing reused session IDs from receiving another trace's context.

- A non-empty context mapping resolves and deduplicates the U2 trace population before any model call. Exact `ResolvedTraceSelector` keys are preferred; legacy session-ID keys fail with `AmbiguousSessionError` unless the evaluated population proves them singular.
- The same selector-keyed transcript and context travel through BigQuery `AI.GENERATE`, parse/NULL retries, and full Gemini API fallback. Unmapped traces still evaluate without context, while an empty resolved population makes no model call.
- Context-bearing calls deliberately bypass `AI.CLASSIFY`, recording the reason in the report. Context stays in typed query parameters/model inputs, is redacted from context-aware failures, and never enters SQL text or job labels.
- Context-aware persistence fails closed until U5 migrates the result schema, writer, views, and report cardinality to the same identity contract.

## Validation

- Focused identity/evaluator/client suite: `374 passed`
- Full offline suite on current `main`: `3868 passed, 67 skipped`
- Live BigQuery identity fixture and query grammar: `37 passed`, including distinct contexts for two identities sharing one session ID and typed empty/non-empty `ARRAY<STRUCT>` inputs
- `autoformat.sh` and `git diff --check` are clean

## Related

Related: #358, #361


caohy1988:
## U4 round 1 fresh review — one important issue

Reviewed exact head `34f27cb` (PR #378) against `main`: a full read of the categorical-evaluator primitives and the client-side binding/retry/fallback wiring, checked against the in-repo plan's contract, with behavioral probes executed on the head.

**Verified sound (each checked against code or by execution):**

- **Identity binding is exact and fail-closed.** Selector keys are injective (JSON array over session/user/root/scope-signature), legacy session-id keys raise `AmbiguousSessionError` with executable candidates when two resolved traces share the id (constructor payload round-trips — executed), conflicting alias values for one selector raise, superset mappings are ignored, unmapped traces run context-free, and invalid key/value types fail before any model call.
- **Context never leaks.** It travels only as the typed `ARRAY<STRUCT>` query parameter (explicit element type retained for the empty array) or as API prompt input; SQL text interpolates nothing; job labels are untouched; and every error path that could echo model output — the AI.GENERATE fallback reason, API parse-error logs, and retry-loop failure logs — redacts when context was present.
- **Retries and fallback preserve the binding.** NULL/unparseable rows are keyed by internal `evaluation_key`, not session id; retried results are rematched via the reserved attribution keys so colliding session ids cannot overwrite each other; the API's input-order contract is used correctly (with a short-result guard treating missing rows as unresolved); the full fallback reuses the already-resolved transcripts/contexts with no second transcript query; and result completeness is enforced (unknown, duplicate, or missing evaluation keys raise).
- **Mode decisions and gates.** Non-empty context bypasses AI.CLASSIFY with a recorded `classify_skip_reason` on every outcome path; an empty resolved population returns without a model call; `persist_results=True` with context is rejected pending U5, and the CHANGELOG records the U4-not-released-before-U5 gate.
- **Population parity with the trace contract.** The context path resolves through `_fetch_filtered_traces` with a new `span_predicate` applied after ranking and before the public limit, so transcript-ineligible traces neither consume limit slots nor break anchor-escalation refill.
- **Coverage.** The dry-run test binds the exact `ARRAY<STRUCT>` + `AI.GENERATE` grammar against real BigQuery, and the live collision fixture proves two identities sharing one session id receive their own contexts (the plan's #358-closure precondition). Reproduced locally: focused suites 374 passed, full offline 3,954 passed / 15 skipped, live 37 passed.

Findings:

1. **P2 — the Python transcript mirror diverges from `JSON_VALUE` semantics for non-string content values** (`categorical_evaluator.py::_trace_content_text`, claimed as "Mirrors the categorical transcript SQL's JSON_VALUE priority").

   BigQuery `JSON_VALUE` returns NULL for objects and arrays (falling through the `text_summary → response → artifact → tool` priority chain) and renders booleans lowercase. The Python mirror returns `str(value)` for any non-None value. Executed on the head: `{"text_summary": {"x": 1}, "response": "hello"}` → Python `"{'x': 1}"` vs SQL `"hello"`; `{"text_summary": True, "response": "hello"}` → Python `'True'` vs SQL `'true'`; `{"text_summary": [1, 2], "response": "hello"}` → Python `'[1, 2]'` vs SQL `'hello'` (plain numbers match). Consequences: the same trace is judged on a different transcript depending on whether context was supplied, and the `> 10` length gate can flip population membership between the identity-bound and legacy paths for boundary shapes. No test covers non-string content values.

   **Fix:** skip dict/list values in the priority chain (continue to the next candidate instead of rendering), render booleans as `true`/`false`, and add parity regressions for object/array/boolean/number shapes in each priority slot.

2. **P3 — tied-timestamp transcript ordering differs across paths.** The SQL `STRING_AGG(... ORDER BY timestamp)` has no tiebreaker, so tied rows aggregate in nondeterministic order, while the Python transcript inherits the trace's fully-tiebroken producer order. The nondeterminism is pre-existing on the SQL side; worth either adding the same tiebreakers to the SQL `ORDER BY` or noting the divergence where parity is claimed.

3. **P3 — legacy keys bind silently when a colliding trace was excluded by the transcript-length gate.** A session id reused across two identities where one trace's transcript is ≤ 10 characters leaves one trace in the evaluated population, so a legacy key binds to the survivor without ambiguity. This matches the plan's letter ("exactly one resolved trace in the evaluated population"), but the caller cannot tell that their context may have been intended for the excluded trace. A docstring sentence on the interaction would close it.

**Verdict:** changes requested for the P2 (it undercuts the unit's core same-transcript promise); the P3s are follow-ups. Everything else — binding exactness, leak-proofing, retry/fallback identity preservation, mode gating, and live coverage — holds as claimed.

caohy1988:
> P2 — the Python transcript mirror diverges from `JSON_VALUE` semantics for non-string content values.

Fixed in `464dc66`. `_trace_content_text()` now treats object/array values as non-scalars and continues through the same SQL priority chain, renders booleans as lowercase JSON scalars, and preserves string/number behavior. The regression matrix covers object, array, `true`, `false`, and number values in each of the four priority slots, including structural fallthrough.

> P3 — tied-timestamp transcript ordering differs across paths.

Fixed in the same commit. All four legacy transcript query forms now order `STRING_AGG` by `timestamp, span_id, invocation_id, event_type`, matching the reconstructable producer order used by the U2 trace fetch. Rows tied across every persisted ordering field retain the existing documented schema limitation.

> P3 — legacy keys bind silently when a colliding trace was excluded by the transcript-length gate.

Documented explicitly in the internal normalizer, the public `evaluate_categorical()` docstring, README, and SDK guide: transcript eligibility precedes legacy-key ambiguity, so an ineligible twin does not make the survivor ambiguous. The guidance now recommends exact `ResolvedTraceSelector` keys whenever session IDs may be reused.

Verification on `464dc66`: 21 targeted parity regressions, 395 focused tests, 3,889 full-suite tests passed / 67 skipped, 37 live BigQuery tests, and clean autoformat/diff checks.

caohy1988:
## U4 round 2 fresh review — approve

Reviewed exact head `464dc66` (PR #378): a line-by-line pass over the fix commit, re-execution of the round-1 parity probes, and one new live-BigQuery probe closing the last parity corner.

All three round-1 findings verify as fixed:

1. **P2 — transcript parity.** `_trace_content_text` now skips dicts and lists (mirroring `JSON_VALUE`'s NULL fallthrough to the next `COALESCE` candidate) and renders booleans with JSON's lowercase spelling. Re-executed the round-1 probes: `{"text_summary": {"x": 1}, "response": "hello"}` → `"hello"`, `True` → `"true"`, `[1, 2]` → falls through — all matching SQL. The new parametrized regression covers object/array/boolean/string/number shapes in each priority slot.

   I also probed the one corner the round-1 review left open — JSON **number** rendering — against live BigQuery: `JSON_VALUE` yields `1.0`, `0.1`, `1e+30`, and `1.2345678901234568e+20` for those JSON numbers, and the Python client's decode plus `str()` produces byte-identical strings for every one (Python float formatting and BigQuery's canonical float rendering agree, including the big-int-to-float canonicalization). Number parity holds empirically; no further gap.

2. **P3 — tied-timestamp ordering.** All four legacy `STRING_AGG` transcript queries now order by the full producer tiebreakers (`timestamp, span_id, invocation_id, event_type`), matching the resolved-trace fetch order, with a query-text regression asserting it for each.

3. **P3 — legacy-key eligibility interaction.** Documented in both `_normalize_categorical_evaluation_inputs` and the `evaluate_categorical` docstring: the `LENGTH(transcript) > 10` gate precedes the legacy ambiguity check, an ineligible colliding trace does not make the survivor ambiguous, and exact selector keys avoid the inference entirely.

Verification reproduced on the head: focused suites 395 passed; full offline 3,975 passed / 15 skipped; live BigQuery 37 passed (local exclusions remain the known `adk-python` environment drift, green in remote CI).

**Verdict: approve.** With U4 mergeable, the #358 sequence proceeds to U5 — results schema, writer, views, reports, and persistence cardinality — which also lifts the `persist_results` gate and the U4 release hold.

haiyuan-eng-google (approved):

Issue #360, opened 2026-07-10
skill_evolution_lab: remove the two demo workarounds once #358 and #359 land


evekhm:
The skill-evolution lab carries two deliberate, disclosed workarounds for SDK gaps. This issue is the cleanup checklist — pick it up when #358 and/or #359 close (each workaround reverts independently).

## When #358 closes (per-session judge context on the server-side paths)

The demo scores via the API judge (`--conversations-file`) because it is the only path whose judge receives per-session golden expected answers today.

- [ ] `examples/skill_evolution_lab/run_e2e_demo.sh`: make BigQuery scoring the default again — remove the local-scoring default in `score()` (marked `TODO(#360)`), thread `per_session_context` on the BQ call in `scripts/quality_report.py` (restore the withdrawn design from PR #351 commit `da52c54`), and drop the `#358` runtime warning + inline KNOWN LIMITATION comments in `scripts/quality_report.py`.
- [ ] `examples/skill_evolution_lab/README.md`: remove the "scored via the SDK's API judge pending #358" disclosure.
- [ ] Optionally re-record `sample_run/` on the server-side judge (numbers may shift slightly).

## When #359 closes (trace rows scoped by labels + identity)

The demo logs and reads each slice through its own table (`agent_events_<run>_<slice>`) because the trace fetch selects rows by `session_id` alone and the demo reuses ids across slices.

- [ ] `examples/skill_evolution_lab/run_e2e_demo.sh`: collapse per-slice `TABLE_ID`s back to the single shared `agent_events` table (marked `TODO(#360)` in `run_agent()` and `score()`).
- [ ] `examples/skill_evolution_lab/README.md`: remove the per-slice-tables disclosure.
- [ ] Re-seed `sample_run` traces into one shared table and re-render the committed scorecards (currently per-slice `agent_events_seed3_*`).
- [ ] Drop the accumulated per-run/per-slice tables in demo projects.

Both workarounds are also marked inline with `TODO(#360)` comments at the exact code sites.

---

## Expanded per review feedback (2026-07-12)

The cross-issue implementation plan is **#361**; its **U6** unit executes this cleanup with the partial states below as **AE7–AE9**. The provenance mismatch flagged in the first comment was fixed on the PR #351 head in `a7de311`.

### Additional cleanup surfaces (beyond the files above)

[...]

PR #380, opened 2026-07-27
feat(eval): make categorical persistence identity-safe

### LINKED ISSUE #358: evaluate_categorical: support per-session judge context (golden expected answers) on the server-side paths

caohy1988:
## Summary

Categorical evaluations can now persist trusted-context results without collapsing distinct trace identities that reuse a `session_id`. This removes the persistence gate from the identity-bound judge path while keeping golden answers and model echoes out of the results table.

The migration is additive and backward-compatible: existing rows remain nullable and are not backfilled. Deploy in schema -> writer -> views order; typed identities stay separate, a sole typed identity supersedes matching legacy rows, and ambiguous legacy rows remain isolated in a `legacy:<session_id>` lane.

## Design decisions

| Area | Contract |
|---|---|
| Result provenance | Each categorical session result carries its resolved `TraceIdentity` and `TraceScope`, trusted-context source, and execution mode through AI.GENERATE, retry, and API fallback paths. |
| Persistence compatibility | Idempotent `ADD COLUMN IF NOT EXISTS` migrations run before inserts. A versioned identity key preserves cardinality without requiring historical backfill. |
| Privacy and attribution | Contextual rows persist only SDK-owned provenance; justification and raw model output are omitted. Quality reports match golden metadata by exact resolved selector and use legacy session fallback only for a unique unattributed row. |
| Operational reporting | Successful `api_retry` traffic is counted with API fallback traffic, so fallback metrics reflect both fallback paths. |

## Validation

- `uv run pytest -q` — 3,916 passed, 68 skipped.
- Live BigQuery collision fixture — 1 passed: two in-memory results produced two distinct persisted identity keys and two latest-view rows, with zero contextual payload fields persisted.
- `bash autoformat.sh`
- `git diff upstream/main..HEAD --check`

## Related

Fixes #358

Related: #360

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and full Gemini API fallback. Unmapped traces still run without context;
   empty resolved work makes no model call. Context-aware logs redact model
   text/raw retry responses so a model echo cannot disclose trusted context.
+- **Identity-safe categorical persistence and latest views (#358 U5)** —
+  categorical result tables gain nullable identity and context-provenance
+  columns through idempotent additive migrations; historical rows are not
+  backfilled. Deploy and roll back in schema → writer → view order. Latest
+  views preserve separate rows for identities that share a `session_id`; while
+  data straddles the migration, a sole typed identity supersedes matching
+  legacy metric/prompt rows, while zero or multiple identities retain the
+  `legacy:<session_id>` lane. Trusted judge/golden-answer input and model
+  echoes are never persisted — only SDK-owned provenance is. U5 closes the
+  remaining #358 persistence/report gate and unlocks U6/#360.
 
 ### Changed
 
@@ -61,10 +71,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - This release aligns read, trajectory-evaluation, CLI/Remote Function,
   example, report, and categorical judge-context surfaces at one boundary.
   Context is trusted prompt material: it is a query parameter/model input,
-  never SQL text, a job label, or persisted input. The identity-safe
-  persistence/view cardinality migration remains U5; until it lands, context
-  calls reject `persist_results=True` rather than writing colliding identities
-  into the legacy session-only results schema.
+  never SQL text, a job label, or persisted input. U5 now provides the
+  identity-safe persistence/view-cardinality path for contextual results.
 
 ## [0.4.0] - 2026-06-18
 
--- a/README.md
+++ b/README.md
@@ -132,6 +132,17 @@ through AI.GENERATE, retry, and API fallback. It is never interpolated into
 SQL, logged, persisted, or placed in job labels. Apply the same data-governance
 policy you use for evaluation prompts.
 
+When `persist_results=True`, categorical results use an additive, nullable
+identity/provenance schema; existing historical rows are not backfilled.
+Deploy or roll back safely in this order: **schema, then writer, then views**.
+The latest-results view keeps identities distinct even when they share a
+`session_id`. During a legacy/schema straddle, a sole typed identity supersedes
+matching legacy metric/prompt rows; zero or multiple typed identities leave
+legacy rows in their separate `legacy:<session_id>` lane. Trusted judge or
+golden-answer context — including any model echo — is never persisted; only
+SDK-owned context provenance is. This U5 migration completes #358's remaining
+persistence/report gate and unlocks U6/#360.
+
 See [SDK.md](SDK.md) for the full API walkthrough with code examples for every
 feature.
 
--- a/docs/superpowers/plans/2026-07-25-identity-safe-categorical-persistence-u5.md
+++ b/docs/superpowers/plans/2026-07-25-identity-safe-categorical-persistence-u5.md
@@ -1,0 +1,164 @@
+# Identity-Safe Categorical Persistence (U5) Implementation Plan
+
+> Issue scope: #358 U5, sequenced by #361 after merged U4.
+> This unit carries U2/U4 identity and judge-context provenance through
+> categorical results, persistence, views, and quality reports. U6 remains
+> responsible for removing the demo workarounds after this unit lands.
+
+## Contract
+
+- Every identity-resolved categorical result may carry its immutable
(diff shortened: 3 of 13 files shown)


caohy1988:
## U5 round 1 fresh review — no blocking findings

Reviewed exact head `332dba6` (PR #380) against `main`: a full read of the evaluator provenance model, the flattening/migration/writer chain, the latest-view SQL, and the quality-report selector binding, with behavioral probes executed on the head, checked against the in-repo plan's contract and the four suggested review foci.

**Verified sound (each checked against code or by execution):**

- **Typed provenance is SDK-owned and internally consistent.** `CategoricalSessionResult` enforces its own invariants (identity/scope attached together, identity matches session, `context_source` present exactly when `context_applied`), provenance is assigned by `_apply_categorical_result_provenance` from the normalized inputs — never from model output or caller details — and each path exposes a distinct stable `execution_mode` (`ai_classify`, `ai_generate`, `api_retry`, `api_fallback`). Retry replacement now matches on typed `identity`/`scope` equality, replacing only the failed resolved identity.
- **Migration ordering is strict and rollback-safe.** `_persist_categorical_if_configured` runs `CREATE TABLE IF NOT EXISTS` plus every `ADD COLUMN IF NOT EXISTS` migration before flattening and inserting, so a schema failure performs no insert; the additive nullable columns leave historical rows untouched; rerunning is harmless; and the persistence-failure path records only the exception type when contextual results are present (no context leak through error text). The U4 `persist_results` rejection is lifted with the deployment order (schema → writer → views) documented.
- **Redaction holds end to end.** Executed on the head: contextual rows flatten with `justification=None` and `raw_response=None` (a model echo of a golden answer cannot reach the table) while carrying only identity fields, canonical `scope_key`, versioned `identity_key`, `context_applied`, and the enum-validated `context_source`; legacy rows keep their full fields with NULL identity columns. Arbitrary caller strings are rejected as `context_source`.
- **Identity keys are injective and deterministic.** Executed: `v1:`-prefixed SHA-256 over the JSON-array canonical form distinguishes NULL from the string `"None"` and from sibling identities, deterministically.
- **Latest-view cardinality matches the plan's five scenarios by construction.** Typed identities partition by `identity_key`; legacy rows join a sole typed identity's lane (typed rows win via the `identity_key IS NOT NULL DESC` ordering regardless of timestamps, and a legacy metric the typed lane lacks still surfaces); zero-or-multiple typed identities keep the isolated `legacy:<session_id>` lane; helper columns are excluded from the view output. Fallback metrics now count `api_retry` alongside `api_fallback`.
- **Quality-report golden binding fails closed.** Golden Q&A is matched by exact `ResolvedTraceSelector` keys; report enrichment rebuilds selectors from report-safe fields and cross-checks the signature against the rebuilt scope (mismatch → no match); the legacy session-key fallback applies only when the session id appears exactly once and no selector is derivable; and the old "KNOWN LIMITATION (#358)" — golden answers unable to ground the server-side judge — is genuinely removed, with the golden context flowing through `evaluate_categorical` under the `golden_expected_answer` source.
- **The #358 closure evidence is live.** The new live test runs the collision fixture with persistence enabled and asserts, against real BigQuery, that the in-memory result count, inserted rows, distinct `identity_key`s, and latest-view cardinality all agree (2/2/2/2) for two identities sharing one session id, with zero contextual payload fields persisted. Reproduced locally: focused suites 409 passed, full offline 4,002 passed / 15 skipped, live 38 passed.

Findings (both minor):

- **P3 — the sole-identity straddle is retroactive, which the docs imply but never state.** The "exactly one typed identity" decision is computed over the whole table at query time, so when a second typed identity later lands for a session, that session's legacy rows silently move from the sole identity's lane back to `legacy:<session_id>`, changing previously observed view cardinality. This is inherent to the (sensible) design; one sentence in the view docstring and CHANGELOG — "the straddle re-evaluates as new identities arrive; legacy lane membership is not frozen" — would prevent a dashboard consumer from reading the shift as data loss.
- **P3 — the latest-view final tiebreaker is inert for contextual rows.** `ORDER BY … created_at DESC, raw_response DESC` breaks same-timestamp ties by `raw_response`, which is always NULL for contextual rows, so two contextual runs within one `created_at` granule pick nondeterministically. Pre-existing tie family; consider `metric_name`-stable input or an insertion id if determinism at that granularity ever matters.

**Verdict: approve.** The persistence/report gate for #358 reads satisfied: schema → writer → view ordering enforced, colliding identities preserved through every layer with live proof, and the no-raw-context guarantee holds under execution. After merge, #358 can close and U6 (#360 demo-workaround removal plus the 80-session live gate) is unblocked.

haiyuan-eng-google (approved):

Issue #381, opened 2026-07-27
[Dashboard] Investigate intermittent native-chart blank state


caohy1988:
## Scope

Investigate whether the published Looker Studio report has a reproducible
intermittent **native-chart** blank state after page navigation.

The original custom/Vega diagnosis was incorrect. The canonical report uses
native Looker Studio bar, time-series, scorecard, and table components. A live
network pass across two complete render cycles observed zero requests to
`usercontent.goog` or any community-visualization resource.

Responsive/mobile support is separate product scope and is tracked in #386.

## Current evidence

The original blank-chart observation has not reproduced under the strengthened
protocol:

| Path | Observed result |
|---|---|
| Cold cache-miss load | Blank at 40 seconds; fully rendered by 70 seconds |
| Navigate away and return | Fully rendered within 10 seconds |
| SPA navigation to a second page | Fully rendered within 18 seconds |
| Network tracking | Zero community-visualization requests |

This evidence is an environment-specific baseline, not an SLA. It demonstrates
that page chrome, a loading indicator, or an empty chart container is not a
valid completion signal.

## Reproduction protocol

For a new rendering report:

1. Record browser/version, signed-in state, extension state, viewport, and
   whether the load is cold or warm.
2. Run at least three fresh report loads.
3. In each run, visit all eight pages in order and repeat the complete
   navigation loop three times.
4. Allow up to 90 seconds per page for non-degenerate chart output.
5. Capture timestamped page evidence, failed network requests, and matching
   BigQuery job activity.
6. Verify the chart's native editor bindings before changing the report.

A defect is reproducible when the same chart remains blank after 90 seconds in
at least two fresh runs, or when the same navigation sequence reproduces the
failure in at least two loops. A legitimate expected empty state is not a
failure.

## Guardrails

- Do not classify a Google-owned render frame as evidence of Vega or a
  community visualization.
- Do not replace chart types, bindings, filters, or SQL until the failure
  reproduces under this protocol.
- Keep responsive/mobile layout work in #386.

## References

- #383
- #386
- `dashboard/looker_studio/docs/rendering-and-viewport-support.md`
- Live review evidence:
  https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/383#issuecomment-5098030747

caohy1988:
## Validation review

**Verdict: partially valid, but the issue should not drive implementation as
[...]

PR #383, opened 2026-07-27
fix(dashboard): define rendering and viewport support

caohy1988:
## Summary

Issue #381 combined three different concerns: intermittent blank charts, a Vega/community-visualization diagnosis, and narrow-screen layout. The published report uses native Looker Studio charts, so this PR codifies the evidence needed to investigate rendering without changing valid chart bindings and freezes the actual v1 viewport boundary.

- Defines native Looker Studio charts as the implementation contract; Google-owned render frames alone are not evidence of Vega or a community visualization.
- Adds a reproducible viewer-QA protocol: 3 fresh loads, 3 complete navigation loops, a 90-second per-page completion window, and evidence requirements for repeat failures.
- Records the first protocol baseline as dated observational evidence: blank at 40 seconds, fully rendered by 70 seconds on a cold cache miss, and warm navigation rendered within 18 seconds.
- Defines v1 as a freeform desktop report supporting viewports at least 1280 CSS px wide, with 1440 px recommended.
- Records responsive mobile support as a separate-report enhancement requiring its own parity, hydration, credential, cost, and visual validation.
- Surfaces the desktop and cold-load guidance on the public configurator and pins it in Python and Node tests.

This PR does not mutate or republish the external Looker Studio report. The report's default-date-window correction is also deliberately excluded and remains a separate workstream.

## Validation

- `pytest -q tests/test_looker_studio_dashboard.py` — 19 passed
- `node dashboard/looker_studio/tools/test_web_configurator.mjs` — passed
- `bash autoformat.sh` — passed
- `ruff check tests/test_looker_studio_dashboard.py` — passed
- `git diff --check` — passed

Repository-wide validation is currently limited by pre-existing local-environment and baseline issues:

- `pytest -q` stops during collection in three unrelated tests because the local ADK checkout imports an OpenTelemetry semantic-conventions symbol absent from the installed package.
- `ruff check .` reports 570 pre-existing findings outside this diff.

The live review pass at a 1568 CSS-pixel viewport found:

- zero `usercontent.goog` or community-visualization requests across two complete render cycles;
- a cold cache-miss load fully rendered by 70 seconds;
- navigate-away-and-return rendering within 10 seconds; and
- SPA navigation to a second page rendering within 18 seconds.

The blank-chart failure did not reproduce under the new protocol.

## Post-deploy validation

After the Pages workflow publishes the configurator:

1. Verify the deployed page contains the 1280 px support boundary and 90-second load guidance.
2. Exercise all eight report pages using the committed 3-load × 3-loop protocol.
3. Record time to non-degenerate output, failed network requests, and matching BigQuery job activity.
4. Treat the report as healthy only when every expected chart renders valid output within 90 seconds, or displays a legitimate expected empty state.
5. Open a report-side defect only when the same chart remains blank after the timeout in at least two fresh runs, or a deterministic navigation sequence reproduces it twice.

Mitigation for a repeat failure is evidence collection and report-editor binding inspection; chart types, filters, and SQL should not change until the failure is reproducible.

## Known limitations

- This PR defines the investigation and support contracts; it does not claim the intermittent native-chart behavior is fixed.
- The 1280 px minimum is a documented product-support boundary; the latest live pass ran at 1568 px, so targeted validation at 1280 px remains pending.
- Phone and narrow-tablet report layouts remain unsupported in v1.
- A responsive report must be built and validated as a separately owned template.

Related: #381

---
[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/dashboard/looker_studio/README.md
+++ b/dashboard/looker_studio/README.md
@@ -51,6 +51,7 @@ billing project is supported as an optional advanced setting.
 | `tools/hydrate_dashboard.py` | Validates a BQAA table and emits a user-owned Looker Studio report URL |
 | `docs/dashboard-implementation.md` | Looker Studio page, field, formula, and live-validation implementation contract |
 | `docs/issue-377-review.md` | Live validation matrix for the UX/design backlog |
+| `docs/rendering-and-viewport-support.md` | Native-chart completion protocol and supported desktop viewport contract |
 | `oracle/queries/` | 37 independent per-chart SQL contracts used by the live validator |
 
 ## Pinned contracts
@@ -92,6 +93,13 @@ Canonical published template:
 
 All seven dashboard pages default to a rolling 365-day window ending
 yesterday. The Trace Inspector intentionally has no default date control.
+
+The v1 report is a freeform desktop dashboard. Use a viewport at least 1280
+CSS pixels wide (1440 recommended). A cold load or page navigation can paint
+native charts after the surrounding report controls; allow up to 90 seconds
+for non-degenerate chart output. Phone and narrow-tablet layouts require a
+separate responsive template and are not supported by v1. See
+[`docs/rendering-and-viewport-support.md`](docs/rendering-and-viewport-support.md).
 
 For the standard BQAA layout, open the
 [three-field dashboard configurator](https://googlecloudplatform.github.io/BigQuery-Agent-Analytics-SDK/)
--- a/dashboard/looker_studio/README.md
+++ b/dashboard/looker_studio/README.md
@@ -100,6 +100,8 @@ native charts after the surrounding report controls; allow up to 90 seconds
 for non-degenerate chart output. Phone and narrow-tablet layouts require a
 separate responsive template and are not supported by v1. See
 [`docs/rendering-and-viewport-support.md`](docs/rendering-and-viewport-support.md).
+The latest live pass used a 1568-pixel viewport, so targeted validation at the
+documented 1280-pixel minimum is still pending.
 
 For the standard BQAA layout, open the
 [three-field dashboard configurator](https://googlecloudplatform.github.io/BigQuery-Agent-Analytics-SDK/)
--- a/dashboard/looker_studio/bindings/report_template.yaml
+++ b/dashboard/looker_studio/bindings/report_template.yaml
@@ -37,6 +37,10 @@ live_template_verification:
   result: PASSED
   limitation: mutable_external_report_requires_reverification_after_changes
 product_contract: spec/product_contract.yaml
+viewer_qa_contract:
+  issue: GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK#381
+  protocol: docs/rendering-and-viewport-support.md
+  status: INVESTIGATION_OPEN
 product_verification:
   verified_date: "2026-07-27"
   pages: 8
--- a/dashboard/looker_studio/bindings/report_template.yaml
+++ b/dashboard/looker_studio/bindings/report_template.yaml
@@ -40,7 +40,9 @@ product_contract: spec/product_contract.yaml
 viewer_qa_contract:
   issue: GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK#381
   protocol: docs/rendering-and-viewport-support.md
-  status: INVESTIGATION_OPEN
+  status: NOT_REPRODUCED_UNDER_PROTOCOL
+  observed_baseline_comment: >-
+    https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/383#issuecomment-5098030747
 product_verification:
   verified_date: "2026-07-27"
   pages: 8
--- a/dashboard/looker_studio/docs/dashboard-implementation.md
+++ b/dashboard/looker_studio/docs/dashboard-implementation.md
@@ -111,7 +111,7 @@ fixture and rendered “Too Many Rows.”
 
 ## Viewer rendering and viewport boundary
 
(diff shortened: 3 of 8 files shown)


caohy1988:
## Full review at head `767feb2` — code and live-report verification

**Verdict: mergeable. The PR's reframing of #381 is not just plausible — I collected direct network-level evidence for it on the live report.**

### Live verification (protocol-lite run, 2026-07-27)

1. **Native-chart claim confirmed at the network level.** With request tracking armed through two complete render cycles (cold load of Token Consumption, SPA navigation to Agent & Sessions), there were **zero** requests matching `communityvisualization` or `usercontent.goog` — traffic was exclusively Data Studio internals (firestore channels, maps-api bundle, analytics). `viewer_qa.chart_implementation: native_data_studio` / `community_visualizations: not_used` holds; #381's Vega/community-viz framing does not match observed behavior.
2. **The 90-second bound is empirically sound.** Cold load: canvas fully blank at 40 s, complete non-degenerate render at ~70 s — and the Google Analytics beacon on that load literally reported `event_action: Read from data cache, event_label: View-miss`, confirming the cold-cache path. Across my three sessions on this report, observed cold renders span ~30–70 s; the 90 s timeout brackets them with margin.
3. **The #381 P1 does not reproduce under the protocol.** Navigate-away-and-return to Token Consumption: fully re-rendered in ≤10 s. SPA navigation to Agent & Sessions: all 5 charts + 2 scorecards rendered in ≤18 s. Combined with two prior full 8-page passes (every chart eventually rendered), the "charts stay blank after navigation / hard reload" report is consistent with observations made before render completion — the exact failure mode the new protocol is designed to exclude. Keeping it as a `native-chart-rendering-investigation` deferred entry rather than churning chart bindings is the right call.

### Code

- 19/19 pytest + the Node configurator suite pass at head. Every new contract surface is pinned: `viewer_qa` (exact dict), `viewport_support` (1280/1440), both new deferred entries, the `viewer_qa_contract` pointer in the bindings, and the page notice strings asserted in **both** test suites.
- The rendering doc gets the subtle point right that killed the original diagnosis: native charts can render inside Google-owned `usercontent.goog` frames, so frame presence is not community-viz evidence — root-cause claims need the editor's resource inventory. That distinction is the load-bearing content of this PR.
- "Non-degenerate rendered output" as the only completion signal (page chrome and spinner disappearance explicitly excluded) matches how both prior misdiagnoses on this report actually happened — including my own on 2026-07-25.
- The desktop boundary (1280 min / 1440 recommended, mobile = separately owned template) is coherent with Data Studio's freeform-to-responsive conversion limit, and the configurator notice sets the expectation before report creation.

### Advisory (non-blocking)

1. **Cold-render latency is now well-characterized — record the baseline.** The deferred `native-chart-rendering-investigation` would benefit from the observed numbers (blank at 40 s, render at ~70 s cold; ≤18 s warm SPA navigation; cache View-miss beacon) as its starting dataset, so the investigation targets time-to-render rather than re-litigating whether charts render at all.
2. Naming nit: the new doc says "Data Studio" throughout while the rest of the repo standardized on "Looker Studio" — cosmetic, but worth a pass before merge since the doc is user-facing.
3. Viewport boundary not independently stress-tested: my pass ran at 1568 px (renders correctly). The 1280 minimum is a documented product decision with test pinning, which is sufficient; flagging only that no sub-1280 empirical check was run in this review.

caohy1988:
> Three advisories, none blocking: seed the deferred rendering investigation with the baseline numbers I collected; a “Data Studio” vs “Looker Studio” naming pass; and a note that the 1280px minimum wasn't empirically stress-tested.

Addressed in `a6c3826`:

- The contract now records the dated 1568 px baseline: blank at 40 seconds, cold-cache-miss render by 70 seconds, return navigation within 10 seconds, second-page navigation within 18 seconds, and zero `usercontent.goog`/community-visualization requests. It is explicitly observational evidence, not an SLA.
- New prose consistently uses the Looker Studio product name. The existing `native_data_studio` machine value remains unchanged because it is the already reviewed contract enum.
- The 1280 px minimum is explicitly labeled a product-support boundary whose targeted validation is pending; the last validated width is recorded as 1568 px.

The binding status is now `NOT_REPRODUCED_UNDER_PROTOCOL`, the evidence comment is linked from the binding manifest, and the expanded contracts remain pinned by the 19-test focused suite. No live report or SQL change was made.

caohy1988:
## Final review at head `a6c3826`: approved, ready for maintainer review

All three advisories from my previous review are addressed, each with test pinning rather than prose alone:

1. **Rendering baseline recorded correctly.** `viewer_qa.observed_baseline` carries the exact measured values (blank at 40 s / rendered by 70 s cold with cache view-miss; ≤10 s return and ≤18 s second-page warm; zero `usercontent.goog` and community-visualization requests), dated, with the viewport, explicitly framed as environment-specific observational evidence rather than an SLA — and the deferred `native-chart-rendering-investigation` entry now targets *comparing future time-to-render against the dated baseline*, which is the right investigation shape. The baseline even records its evidence-source comment URL.
2. **Naming normalized**: zero "Data Studio" occurrences remain in `rendering-and-viewport-support.md`.
3. **Viewport boundary honestly scoped**: `minimum_width_validation: pending` with `last_validated_width_css_px: 1568` states exactly what has and hasn't been validated.

The `status: NOT_REPRODUCED_UNDER_PROTOCOL` designation for #381 is asserted in both the bindings and the contract test, and it is the accurate characterization: the blank-chart report is consistent with pre-completion observation, which the protocol now excludes by construction.

Verified at head: 19/19 dashboard tests + Node configurator suite pass, including the generated-SQL drift tests. No open items remain from my side.

haiyuan-eng-google (approved):

Issue #388, opened 2026-07-29
Data Studio Dashboard feedback


haiyuan-eng-google:
Thank you for this great opportunity, I tested it, but it didn't work. I can't use a custom table ID. The field is red as long as it is different than the default name ""

<img width="1680" height="1159" alt="Image" src="https://github.com/user-attachments/assets/5c299ca8-80d7-42f8-bf1d-d24f972478f7" />

Anyway, I created a dashboard with the default table, edited the database connection, and then it worked.

However, on some pages, some graphs appear too low.

<img width="1615" height="1287" alt="Image" src="https://github.com/user-attachments/assets/cc1bf791-81dd-43f6-bdc7-e46de60712fa" />
<img width="1713" height="1324" alt="Image" src="https://github.com/user-attachments/assets/56214f41-8e0c-4ad4-845f-7e5be6f51b76" />

caohy1988:
## Root-cause analysis

Both reported problems reproduce and have identified causes. Details below, with fixes scoped.

### Bug 1 — Custom table ID rejected: over-strict table validation regex

**Root cause (confirmed in the deployed configurator code):** the table field is validated with `BIGQUERY_ID_RE = /^[A-Za-z_][A-Za-z0-9_]{0,1023}$/` — letters, digits, and underscores only. The table in the screenshot is `events_agent_cur-phenix`; the **hyphen** is what turns the field red. It is not related to differing from the default name — any hyphenated table name fails.

The validation rule is wrong for tables: BigQuery *dataset* IDs cannot contain hyphens, but **table names can** (flexible table naming also permits spaces and Unicode). Since the BQAA plugin accepts a custom `table_id`, real installations legitimately have hyphenated tables. The reported workaround proves the stack above the configurator is fine: after swapping the data-source connection inside Looker Studio, BigQuery queried the hyphenated table without issue. The same over-strict `ID_RE` also gates `--table` in `tools/hydrate_dashboard.py` and `tools/validate_live_bqaa.py`.

[...]

PR #387, opened 2026-07-28
fix(dashboard): include today in the default date window

caohy1988:
## Summary

Fresh BQAA installations no longer open to an empty dashboard solely because
their first events occurred today. The published report and repository
contract now use an exact 365-calendar-day window ending today
(`today - 364 days` through `today`) on all seven dashboard pages; Trace
Inspector remains without a default date control.

The real ADK 2.4.0 dataset that exposed the defect returned 0 of 77 events
under the previous ending-yesterday default and all 77 when today was
included. The Looker Studio template was republished and its publication and
verification attestations are dated 2026-07-28.

The secondary analytics and documentation observations in the issue are
deliberately outside this PR so the first-run P1 fix can ship independently.

Related: #382

## Validation

- `pytest -q tests/test_looker_studio_dashboard.py` — 20 passed
- `node dashboard/looker_studio/tools/test_web_configurator.mjs` — passed
- `bash autoformat.sh` — passed
- `ruff check tests/test_looker_studio_dashboard.py` — passed
- Live ADK 2.4.0 data — 77 of 77 events visible with the new default
- Published report — include-today default verified on seven dashboard pages

## Post-Deploy Monitoring & Validation

For the first 24 hours after merge, the PR author or dashboard maintainer
should:

1. Open the canonical template and confirm each of the seven dashboard date
   controls ends on the current date; Trace Inspector should still have no
   default date control.
2. Hydrate a copy against a BQAA table containing only current-day events and
   confirm the opening page renders non-degenerate data without changing the
   date control.
3. In BigQuery job history, filter for jobs referencing the hydrated
   `agent_events` table and confirm the current UTC date is present in the
   effective `@DS_END_DATE` window.

Healthy signals are current-day rows visible on first open, non-degenerate
charts, and a displayed range covering exactly 365 calendar dates. Failure
signals are “No data” while current-day events exist, a range ending
yesterday, or a 366-day range. If any failure appears, unpublish the changed
template revision, restore the previous report revision, and reopen #382 with
the affected page, displayed range, and BigQuery job evidence.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

PR #389, opened 2026-07-29
fix(dashboard): support hyphenated BQAA table IDs

caohy1988:
## Summary

Customers can now configure the dashboard against valid BigQuery table IDs such as `events_agent_cur-phenix`. Dataset IDs retain their stricter no-hyphen rule, while commas and backticks still fail closed because they could change the Linking API replacement list or SQL identifier boundary.

Invalid field values now appear once beside the affected input instead of being duplicated in the form-level status. The report contract also records a 24-pixel page-containment rule and truthfully marks the two published pages that still need an authenticated Looker Studio editor update.

Related: #388

## Validation

- Added the regression tests first and observed 8 expected failures before implementation; the completed suite passes 22/22 tests plus the Node configurator suite and Ruff checks.
- Exercised the configurator in a browser: the exact hyphenated table enables the dashboard link, while a comma-delimited adversarial value disables it and renders only one inline error.
- Copied real ADK 2.4.0 BQAA data (77 rows across 10 event types) into a temporary table named `events_agent_cur-phenix`, preserving partitioning and clustering.
- Against that table, the hydration preflight passed, the production custom query returned real data, and all 37 dashboard oracle queries completed successfully.
- Confirmed comma and backtick payloads are rejected by both Python entry points, then deleted the temporary BigQuery dataset.

## Live report follow-up

This PR does not claim that repository metadata can repair a mutable external Looker Studio canvas. The Token Consumption and Latency pages still need their page height extended (or empty bands reclaimed), followed by republication and a fresh visual attestation before the pending containment status can be changed to verified.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/dashboard/looker_studio/README.md
+++ b/dashboard/looker_studio/README.md
@@ -102,6 +102,13 @@ separate responsive template and are not supported by v1. See
 [`docs/rendering-and-viewport-support.md`](docs/rendering-and-viewport-support.md).
 The latest live pass used a 1568-pixel viewport, so targeted validation at the
 documented 1280-pixel minimum is still pending.
+Issue
+[#388](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/388)
+also found that lower charts on Token Consumption and Latency can cross the
+freeform page boundary. The repository now requires every component's bottom
+edge to remain at least 24 pixels inside the page; the canonical report still
+requires an editor update and republication before that check can be marked
+verified.
 
 For the standard BQAA layout, open the
 [three-field dashboard configurator](https://googlecloudplatform.github.io/BigQuery-Agent-Analytics-SDK/)
@@ -110,6 +117,11 @@ and enter only:
 1. GCP project ID;
 2. BigQuery dataset ID;
 3. BQAA table ID (normally `agent_events`).
+
+For portable Linking API substitution, table IDs may use ASCII letters, digits,
+underscores, and hyphens, such as `events_agent_cur-phenix`. Dataset IDs retain
+BigQuery's no-hyphen restriction. Commas and backticks remain rejected because
+they would alter the `sqlReplace` binding list or the quoted SQL identifier.
 
 The configurator runs entirely in the browser and creates an official Looker
 Studio Linking API URL. Project, dataset, and table identifiers can also be
--- a/dashboard/looker_studio/bindings/report_template.yaml
+++ b/dashboard/looker_studio/bindings/report_template.yaml
@@ -37,6 +37,10 @@ live_template_verification:
     - published_include_today_default_validation
   result: PASSED
   limitation: mutable_external_report_requires_reverification_after_changes
+known_live_issues:
+  - issue: GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK#388
+    scope: Token Consumption and Latency page containment
+    status: REQUIRES_EDITOR_REPUBLICATION
 product_contract: spec/product_contract.yaml
 viewer_qa_contract:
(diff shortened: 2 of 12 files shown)


caohy1988:
## Full review at PR head — approved, with independent adversarial validation on real BQAA data

### Code

- The fix is exactly the right shape: `DATASET_RE` (unchanged, underscore-only — correct for datasets) split from `TABLE_RE = ^[A-Za-z0-9_][A-Za-z0-9_-]{0,1023}$`, which admits hyphens **and leading digits** (both legal for BigQuery tables) while still excluding the characters that matter for `sqlReplace` safety (comma, backtick, dot, space). Mirrored identically in `configurator.mjs`, `hydrate_dashboard.py`, and `validate_live_bqaa.py`, with hint text updated.
- The duplicate-message fix is minimal and correct: the form-level status is suppressed exactly when a field-level error is displayed, so the stray text below *Advanced settings* is gone without losing the fallback for non-field errors (e.g. sentinel-binding failures).
- `layout.page_bounds` records a machine-checkable acceptance rule (`component_top_plus_height_lte_page_height_minus_bottom_padding`, 24 px padding) with `verification_status: pending_editor_republication`, and `known_live_issues` in the bindings tracks #388 as `REQUIRES_EDITOR_REPUBLICATION` — accurate rather than aspirational, which keeps the attestation honest until the report is actually edited and republished.
- 22/22 pytest + the expanded Node suite pass locally; branch is rebased on current main (includes #387), so no merge conflict with the date-window change.

### Independent adversarial validation (real ADK 2.4.0 data)

Created a real table named exactly like the reporter's — `events_agent_cur-phenix` — as a copy of the 77-row real-agent-traced `agent_events` (10 event types, genuine Gemini/tool payloads), then:

- **Preflight**: `hydrate_dashboard.py --table events_agent_cur-phenix` passes and emits a Linking URL whose `sqlReplace` correctly maps `sentinelbqaaevents → events_agent_cur-phenix`.
- **All 37 oracle queries** execute against the hyphenated table (37/37, sanitized receipt).
- **Production template SQL** with the hyphenated substitution returns all 77 rows under the current default window.
- **Fail-closed attempts** — every one rejected with non-zero exit in Python and thrown in JS: `events,other` (comma/sqlReplace splitting), `ev\`il` (backtick breakout), `data.set.table` (qualified-path smuggling), `events agent` (space), `-leading-hyphen`, `./evil`, and `sentinelbqaaevents-x,evil` (sentinel + comma).
- **Ordered sentinel rules still enforced** with the widened charset: a dataset containing the table sentinel is rejected (exit 2); a table containing its own sentinel remains correctly allowed (last substitution position).
- **Boundary cases accepted as intended**: leading-digit (`9leading_digit-ok`) and trailing-hyphen (`trailing-`) table names — both legal in BigQuery.

### Notes (non-blocking)

1. `TABLE_RE` deliberately stops short of BigQuery's full flexible naming (spaces and Unicode are legal in table names but excluded here). Reasonable v1 boundary — worth one line in the README hint docs so the next user with a spaced table name finds a documented answer rather than a red field.
2. Trivia: a leading-hyphen table arriving via CLI is rejected by argparse (usage error) before the validator sees it — still fail-closed, just a different message than the web path.
3. The Token Consumption / Latency page-containment fix remains editor-side work; once republished, the `page_bounds` rule and a re-dated attestation should flip `verification_status` — happy to re-run the visual verification then.

Test artifacts: the hyphenated table remains in the test project's `bqaa_e2e_real` dataset as a repro fixture; delete at will.

haiyuan-eng-google (approved):

PR #405, opened 2026-08-07
[Dashboard/configurator] Accept pasted fully-qualified BigQuery table IDs

### LINKED ISSUE #403: [Dashboard/configurator] Accept a pasted fully-qualified table ID (project.dataset.table)
Part of #404.

## Request

From **Guillaume Blaquiere (Carrefour)**, following up on #388 after the custom
table ID and clipped chart fixes landed:

> I'm a lazy guy, and when I go to BigQuery UI, I have this […] In the BigQuery
> UI, I can copy the fully qualified table id (`project.dataset.table`). If I
> could simply copy it here, and paste it as-is into your UI, it would be
> wonderful! (instead of me splitting it into 3 different fields)

The BigQuery console's table header has a copy control that yields exactly
`project.dataset.table`. Every user of this configurator already has that string
one click away — and we make them split it by hand across `#project`,
`#dataset`, and `#table` (`docs/index.html:60-105`).

## Why this is safe to parse

`DATASET_RE` and `TABLE_RE` (`docs/configurator.mjs:4-5`) both exclude `.`:

```js
export const PROJECT_RE = /^[a-z][a-z0-9-]{4,28}[a-z0-9]$/;
export const DATASET_RE = /^[A-Za-z_][A-Za-z0-9_]{0,1023}$/;
export const TABLE_RE   = /^[A-Za-z0-9_][A-Za-z0-9_-]{0,1023}$/;
```

so a three-part dotted string splits exactly one way. No ambiguity, no guessing.

## Proposed fix

- [ ] Detect a dotted identifier pasted into any of the three fields and
      distribute it across all three. A `paste` handler plus an `input` fallback
      (for typed or autofilled values) covers both cases; the existing `input`
      listeners already call `refresh()` (`docs/app.mjs:80-82`).
- [ ] Normalize the common paste shapes before splitting:
  - `` `project.dataset.table` `` — backticks, as copied from a SQL editor
  - `project:dataset.table` — the legacy colon form
  - surrounding whitespace
  - a trailing `;` or `,` from a copied query fragment
- [ ] Show what happened: set `#form-status` to e.g. *"Split
      `project.dataset.table` into the three fields."* so the user can see where
      their paste went, and run the existing per-field validation on the parts so
      a bad component still gets its own inline error.
- [ ] Leave the three fields as the source of truth. A single combined field was
      considered and rejected: per-field validation messages are what made #388
      diagnosable.
- [ ] Add unit tests to `tools/test_web_configurator.mjs` for each accepted
      shape and for the pass-through case (a paste that is not a dotted triple
      must behave exactly as today).

## Out of scope

Parsing a pasted BigQuery console URL (`…/bigquery?ws=!1m5!…`). The encoding is
undocumented and brittle; note it here and skip it.

## Acceptance criteria

- [ ] Copying a table ID from the BigQuery console and pasting it into the first
      field fills all three, with validation applied to each part.
- [ ] A non-dotted paste behaves exactly as it does today.
- [ ] Backtick and colon forms are handled.
- [ ] `tools/test_web_configurator.mjs` covers every accepted and rejected shape.

## Why this is a good first issue

Self-contained: one behaviour, two files (`docs/app.mjs`,
`docs/configurator.mjs`), an existing test harness, and no build step or
template re-publish.

## Related

- #388 — originating thread.
- #377 section G — configurator backlog.

Alok-Jadhao:
## Summary

This PR implements support for pasting fully-qualified BigQuery table IDs into the dashboard configurator.

Supported input formats include:

- `project.dataset.table`
- `` `project.dataset.table` ``
- `project:dataset.table`
- trailing `;`
- trailing `,`

When a qualified table ID is detected, it is automatically split into the project, dataset, and table fields while preserving the existing per-field validation.

## Tests

Added tests covering:

- normal dotted identifiers
- backticked identifiers
- legacy colon format
- trailing semicolon
- trailing comma
- input fallback behavior
- pass-through behavior for unqualified input
- validation of invalid parsed components

Closes #403.


caohy1988 (changes requested):
Reviewed against #403's acceptance criteria and the engineering plan on #404. The shape of the change is right — paste handler + `input` fallback, distribute-then-validate so a bad component gets its own inline error, pass-through for unqualified input — and the accepted-shape test coverage is good. But the test suite currently fails on this branch, and there are three behavioral bugs, two of them regressions of existing behavior. All findings below were verified by running the code, not just reading it.

### 1. Blocking — the test suite fails as committed

```
node dashboard/looker_studio/tools/test_web_configurator.mjs
AssertionError: a split ID validates an invalid project component inline
  actual:   'Use 6–30 lowercase letters, digits, or hyphens; …'
  expected: /6-30 lowercase letters, digits, or hyphens/
```

The new assertion uses an ASCII hyphen (`6-30`) but `VALIDATION_MESSAGES.project` uses an en dash (`6–30`). The existing assertion earlier in the same file uses the en dash — copy that.

### 2. Blocking — `#billing-project` loses its `input` listener

The old wiring was `for (const input of Object.values(inputs)) input.addEventListener("input", refresh)` — all four fields. This PR replaces it with a loop over `tableInputs` (project/dataset/table only), so typing in the billing project field no longer triggers `refresh()` at all. Verified: after importing `app.mjs` on this branch, `#billing-project` has no `input` listener. Consequences: entering or fixing a billing project doesn't revalidate, and `createLink.href` keeps the previous billing project until some other field changes.

Fix: keep a plain `refresh` listener on `inputs.billingProject` (it needs no split logic), or restore the all-inputs loop and layer the split handling only on the three table fields.

### 3. Blocking — invalid split reports success

`afterQualifiedTableId()` calls `refresh()` and then unconditionally sets the status to `Split "…" into the three fields.` with kind `"ready"`. When a component is invalid, `refresh()` has just taken the error path — link disabled, inline field error, status cleared — and the unconditional `setStatus` overwrites that with a green "ready" state. Verified with the PR's own test input:

```
paste "BADPROJECT.my_dataset.my_table"
  form-status: 'Split "BADPROJECT.my_dataset.my_table" into the three fields.'  kind: "ready"
  create link: aria-disabled="true"
  project-error: "Use 6–30 lowercase letters…"
```

That also breaks the invariant the existing suite pins at the bottom of the file ("field validation must not repeat the inline error in the form status" — status must be empty on a field error). Simplest fix: only emit the split status when `refresh()` succeeded (`if (createLink.href) setStatus(…)`), leaving the field-error presentation intact otherwise. And add a status assertion to the BADPROJECT test so this can't regress.

### 4. Blocking — the `input` fallback hijacks typing at the second dot and accepts empty parts

`splitQualifiedTableId` doesn't require parts to be non-empty, and the `input` handler runs on every keystroke. Verified: typing `my-project.my_dataset.` into the project field (the moment the second dot lands) splits into `project="my-project"`, `dataset="my_dataset"`, `table=""` — **wiping the pre-filled default table value** — and shows a "ready" split status while the user is mid-typing.

Minimum fix: reject any empty part (`parts.every((p) => p.length > 0)`). Better for the `input` path specifically: only split when all three parts already pass their field regexes — autofill and query-param fills deliver the complete value in one event, so a fully-valid triple is the only signal that isn't mid-typing. The `paste` path should keep the looser distribute-then-validate behavior so a decorator like `table$20260807` lands as an inline error on the table field.

### 5. Missing rejected-shape tests

#403's acceptance criteria require covering every accepted **and rejected** shape, and the #404 plan calls out separate tests for extra-dot inputs vs `$`/`@` decorators (they fail through different paths — part count vs `TABLE_RE`). Currently only `BADPROJECT.dataset.table` is tested. Please add:

- `a.b.c.d` (extra dots) → no distribution, whole string stays in the pasted field, inline error there;
- `project.dataset.table$20260807` and `project.dataset.table@1234` → distributes, inline error on the table field;
- `project.dataset.` (empty part) → no distribution (per finding 4).

### 6. Suggestion — move the pure helper to `configurator.mjs`

The repo keeps pure logic (regexes, `validateConfiguration`, URL builders) in `configurator.mjs` with direct unit tests, and DOM wiring in `app.mjs`. `splitQualifiedTableId()` is pure — exporting it from `configurator.mjs` matches that split (and the #404 plan) and lets the shape tests run without the fake-DOM harness. Not blocking.

### Nits

- `afterQualifiedTableId` is indented with 4 spaces in a 2-space file; the paste-loop `})` is missing its semicolon.
- `.replace(/:/, ".")` rewrites the first colon anywhere in the string, not just the legacy `project:dataset.table` separator. Harmless in practice (anything weird still fails field validation), but `/^([^.:]+):/` would express the intent.
- `.replace(/[;,]$/, "")` strips only one trailing character; `table;;` keeps a semicolon. Fine to leave, just noting it's not a loop.

Happy to re-review once the suite passes — items 1–4 plus the tests in 5 are what's needed for merge.

Alok-Jadhao:
Thanks for the detailed review! I've addressed all the blocking issues:

- restored the billing project input refresh
- fixed the invalid split status behavior
- made the input fallback stricter to avoid partial typing
- added the requested rejected-shape tests
- moved the pure parsing helpers into configurator.mjs with direct unit tests

I also fixed the failing assertion caused by the en dash.

I'd appreciate another look when you have time.

caohy1988 (changes requested):
Full fresh review of `e07c6fb`. All four blockers from the first round are fixed and verified by running the branch, the parser now implements the three-outcome contract from #404 exactly, and the test coverage is genuinely good — pure-function cases for every accepted and rejected shape plus DOM-level cases for distribution, error placement, and status behavior. The pure helpers moving into `configurator.mjs` and the anchored colon regex (`/^([^.:]+):/`) are both right.

Unfortunately the fix commit also introduces one critical regression, so this is another request-changes — but it's a one-line fix.

### Blocking — `app.mjs:1` breaks the entire configurator page in a browser

```js
import { parse } from "node:path";
```

`parse` is never used (looks like an editor auto-import triggered by the `parsed` variables), and `docs/index.html:245` loads `app.mjs` directly as a native `<script type="module">` with no bundler. A browser cannot resolve the `node:` scheme, so the whole module graph fails to load — no listeners, no validation, no link building; the create button stays permanently disabled. Verified against this branch in headless Chrome:

```
Access to script at 'node:path' from origin 'http://127.0.0.1:8479' has been
blocked by CORS policy: Cross origin requests are only supported for protocol
schemes: chrome, chrome-extension, chrome-untrusted, data, http, https, isolated-app.
```

The Node-based suite passes because Node resolves `node:path` natively — which is exactly why this class of bug is invisible to it, and why CI is green. Two asks:

1. Delete the import.
2. Add a guard so this can't recur: the suite already makes static assertions against `index.html` source at the top of `test_web_configurator.mjs` — add one asserting that every `import ... from` specifier in `docs/*.mjs` starts with `./` (browser-safe relative imports only). That turns "works in Node but not in a browser" into a test failure.

### Verified fixed from round 1

1. **Test suite** — passes on this head (ran it; the en-dash assertion is corrected).
2. **`#billing-project` listener** — restored via a dedicated `refresh` listener; typing there revalidates again.
3. **Invalid split status** — now gated on `createLink.href`; the BADPROJECT case leaves the inline error, empty status, and non-ready kind, and the new assertions pin all three.
4. **Empty parts / mid-typing** — `splitQualifiedTableId` rejects empty parts, and the `input` path uses the stricter `parseQualifiedTableIdForInput` (all three field regexes must pass). Typing `project.dataset.` no longer hijacks the form or wipes the table default.

Also confirmed: decorators (`$`/`@`) distribute and fail inline on the table field, extra-dot inputs don't mutate anything, `[;,]+$` handles repeated trailing punctuation, and unqualified pastes are untouched.

### Nits (fine to fix in the same push)

- `configurator.mjs` is missing its trailing newline at EOF.
- `app.mjs` has trailing whitespace on the blank line inside `afterQualifiedTableId`.
- `handleQualifiedTableId(parsed)` and `afterQualifiedTableId(parsed)` are only ever called as a pair, and the former's return value is now unused — merging them (or having one call the other) would read cleaner.

### Known limitation, accepted — no action needed

Hand-typing a full ID can still split at the first keystroke where all three parts happen to be valid (e.g. `my-project.my_dataset.my_t`), after which further keystrokes land in the project field. This is inherent to the agreed contract on #404 — a single-event heuristic can't distinguish a valid-prefix keystroke from an autofill — and the strict all-regexes gate is the right trade-off. Noting it here so it doesn't get re-reported as a bug.

With the `node:path` import removed (and ideally the import guard added), this is ready to merge.

dashboard/looker_studio/docs/app.mjs
```diff
@@ -1,7 +1,10 @@
+import { parse } from "node:path";
```
caohy1988: **[P1, blocking] This import breaks the entire configurator in real browsers.** `import { parse } from "node:path"` is an unused leftover, and `node:` specifiers don't resolve in browsers — `index.html:245` loads `app.mjs` as a plain module, there's no build step, and the import map doesn't exist. Verified live over HTTP on this branch: the module fails to load (`Access to script at 'node:path' … net::ERR_FAILED`), and the page is functionally dead — typing in `#project` runs no validation and the create link never enables. The Node suite can't catch this because Node resolves `node:path` happily. Fix is a one-line delete. Since Pages deploys on merge, I'd also add a cheap guard to the Node suite asserting no `node:` specifiers appear in the browser-shipped `docs/*.mjs` files — catches the whole class next time.

caohy1988 (commented):
## Fresh review at `e07c6fb`

Prior round's blockers — all verified fixed at head, by running the code:

1. Suite passes locally (`web configurator OK`); the en-dash assertion is corrected.
2. `#billing-project` input listener is restored with a plain `refresh`.
3. Split success status is gated on `createLink.href`, and the BADPROJECT test pins the empty status on field error.
4. Empty parts are rejected, and the `input` path uses `parseQualifiedTableIdForInput` (all three parts must pass their regexes — the review's "better" option), while paste keeps distribute-then-validate so decorators land as inline table-field errors.
5. Every rejected shape has a test (`a.b.c.d`, `table$20260807`, `table@1234`, empty part) plus pass-through and browser-paste simulation.
6. Helper moved to `configurator.mjs`; both nits adopted (`/^([^.:]+):/`, `[;,]+$`).

**One new blocker inline** (`app.mjs:1`) — verified in a live browser, not just by reading.

**One P3, not blocking:** the all-valid-triple rule on the `input` path can still fire mid-typing for short triples — typing `my-proj.ec.tdx` splits at `…td` (`my-proj` passes `PROJECT_RE`, `ec` passes `DATASET_RE`, `td` passes `TABLE_RE`), moving the user's remaining keystrokes into the wrong fields. Rare in practice, and the tradeoff was chosen deliberately in the prior round — just worth a sentence in the #404 plan so it's a known choice.

**Verdict:** delete the `node:path` import and this is ready. Everything else from the prior round landed correctly.

caohy1988 (changes requested):
Round-3 review of `8fcf3cc7`, verified by running the branch. The round-2 blocker is fixed, the guard landed, and all nits are addressed. One blocker remains — and it's one where I have to reverse my own round-2 position, so the reasoning is spelled out below.

### Verified fixed

- **Browser-fatal import: fixed.** The `node:path` import is gone, and I re-ran the headless-Chrome check against this head: the page loads with zero console errors, and the dumped DOM shows `aria-invalid="true"` plus the project validation message — proof the module executed and `refresh()` ran in a real browser, not just in Node.
- **Import guard: landed as requested.** The suite now rejects any non-`./` specifier in `docs/*.mjs`. Small nit, non-blocking: the regex only matches `from "…"` forms, so a side-effect import (`import "node:x"`) or dynamic `import("…")` would slip past. Fine to leave until one ever appears.
- **Round-2 nits all fixed** — EOF newline, trailing whitespace, and the `handleQualifiedTableId`/`afterQualifiedTableId` merge.
- **Node suite: green** on this head (ran it).

### Blocking — typed-input distribution (and a correction of my round-2 call)

Round 2 marked the mid-typing split "known limitation, accepted — no action needed." That was wrong, and the #404 plan review that overruled it is right. The all-regexes gate cannot protect hand-typing, because a one-character table prefix is already regex-valid. Verified on this head:

```js
parseQualifiedTableIdForInput("my-project.my_dataset.m")
// → { project: "my-project", dataset: "my_dataset", table: "m" }
```

So typing the ID by hand distributes at the first keystroke after the second dot, and every subsequent keystroke lands in the project field. The regex gate only narrowed the window; it didn't close it.

The structural fix (from the #404 plan, rev 4):

1. Keep immediate distribution on `paste` — that path is correct as is.
2. Move the typed/autofill fallback from `input` to `change`. `change` fires when the value is committed (blur, or Enter-style commit), and browser autofill fires it too — so the autofill case still works, and character-by-character typing can never trigger a split mid-word.
3. Keep the plain `refresh` behavior on `input` so live validation is unchanged.
4. Query-param prefills are assigned programmatically at load and need no listener.

Tests to add with the fix:

- A character-by-character regression: feed `input` events for each prefix of `my-project.my_dataset.my_table` and assert no distribution happens before the `change` event, then that the `change` event distributes.
- Paste coverage from all three inputs — the current suite's `pasteIntoProject` helper only ever exercises `#project`; `#dataset` and `#table` have paste listeners that no test touches.

With that one change and its tests, this is ready to merge — everything else has been verified fixed.

caohy1988 (approved):
Round-4 review of `28a93431`, verified by running the branch. **The last blocker is fixed — approving.**

### Verified

- **Change-boundary fallback: implemented exactly as specified.** `input` is back to plain `refresh` (live validation unchanged), and the split now runs on `change` with the strict parser. Typing can no longer distribute mid-word, because keystrokes never fire `change`.
- **Character-by-character regression: present and correct.** The test feeds every prefix of `my-project.my_dataset.my_table` as `input` events, asserts nothing distributes at any prefix, then dispatches `change` and positively asserts the committed value distributes. That's both halves — no-early-split and yes-on-commit.
- **Paste coverage from all three fields: added.** The helper now takes the field name, and the loop verifies distribution from `#project`, `#dataset`, and `#table`.
- **Node suite: green** on this head (ran it). **Browser smoke: clean** — headless Chrome loads the page with zero console errors and validation runs in the live DOM.

One behavior note for the record, not a change request: real browser autofill fires `change` in mainstream browsers, so the autofill path still splits; where an exotic browser doesn't, the worst case is graceful — the full string sits in one field with its inline validation error, exactly the pre-feature behavior.

### Nit (take or leave)

`pasteIntoProject(name, text)` now pastes into whichever field `name` says — the name is a leftover. `pasteInto(field, text)` would read straight.

### Merge logistics — for the maintainer, not the contributor

The green checks on this PR are effectively CLA-only: two GitHub Actions check suites on this head are **`action_required`** — the workflows are awaiting first-time-contributor approval and have never executed here. That includes the `CI` workflow whose pytest job is what actually runs this Node suite in CI (`tests/test_looker_studio_dashboard.py:553-557` shells out to `node tools/test_web_configurator.mjs`). Before merging: approve the workflow runs and let CI go green for real. Everything above is verified locally, but the branch protection signal should come from an executed run, not an unexecuted one.

caohy1988 (commented):
## Review at `28a93431` — the merge blocker is closed; verified independently in a live browser

The `change`-boundary fix is exactly the structural answer rev 4 prescribed: `input` is back to a plain `refresh`, and typed/autofill distribution only happens on the `change` commit boundary. Paste still distributes immediately (so decorators land as inline table-field errors).

**Verified independently, not just read:**

- **Live browser, character-by-character typing:** typed `my-project.my_dataset.my_table` one prefix at a time into `#project` — **no distribution at any prefix** (the mid-typing hazard from rev 4, `my-project.my_dataset.m`, no longer fires). On `change`, the ID distributes into all three fields with the correct `#form-status` confirmation.
- **Live browser, paste into `#table`:** `other-proj.other_ds.other_table;` pastes and distributes immediately with the trailing `;` stripped — the paste path and the third input both work end-to-end.
- **Node suite green** locally. The character-by-character regression test types every prefix and asserts no distribution until commit — this pins the fixed behavior permanently, and paste coverage now spans all three inputs.

No remaining findings. The PR satisfies #403's acceptance criteria and every gate the #404 plan (rev 4) set for it. **LGTM from me.**

haiyuan-eng-google (approved):

PR #407, opened 2026-08-10
[Dashboard/configurator] Add browser-level smoke check to CI

caohy1988:
## Summary

Adds the browser-level gate from #404: the Node suite structurally cannot see browser-only failures — the `node:path` incident on #405 shipped past a green suite and briefly made the production page inert.

Branch is synced with current `main` (post-#412/#413).

## Enforcement — everything rides the REQUIRED Test checks

The active `main` ruleset requires only `Format check` and `Test (Python 3.10–3.14)`, so both the positive gate **and the negative fixtures** run inside pytest:

- `test_configurator_loads_in_a_real_browser` — the positive smoke.
- `test_browser_smoke_negative_fixtures_are_detected` — runs `browser_smoke.sh --self-test`; a reintroduced false-pass path turns a **required** check red, not just the optional job.
- **Missing Chrome in CI is a hard failure, not a skip**: `_browser_gate_disposition` raises when `CI` is set and no browser is found, skips only outside CI, unit-tested for all four combinations. A runner image losing Chrome cannot silently drop the gate.
- The standalone `Configurator browser smoke` job remains as a fast signal; nothing depends on it being required.

## The check itself — no false-pass paths

- **Live error marker**: an injected script (ahead of the module) records capture-phase `error` events (failed module/resource loads included), unhandled rejections, and `console.error` calls, and **re-stamps `#smoke-result` on every recorded error** — the dumped DOM reflects the entire observation window, not a post-load snapshot. **The observation window is the 5-second virtual-time budget**; an error scheduled beyond it is invisible to any dump-based check by construction — that budget is the stated contract.
- **Port ownership proven**: randomized high port + nonce round-trip; an occupied or stale port fails instead of validating a stranger's content.
- **Exit status always reaped**: the child's natural exit status is honored even after a healthy-looking DOM; the only exempt exit is a timeout kill this script itself delivered (guarded against a racing natural exit).

## Negative fixtures — FIVE, all must fail, all enforced in the required path

`browser_smoke.sh --self-test`: (1) an immediate generic console error; (2) an occupied port serving the wrong tree; (3) a browser binary exiting nonzero with no output; (4) a browser that writes a healthy instrumented DOM, lingers, then exits 42 — the exit-status pin; (5) **a generic console error delayed 900 ms past load — the live-marker pin**. Runs in the required pytest path *and* as a step in the standalone job.

## Verification

At this head: positive smoke green, all five fixtures detected, dashboard pytest 25/25; a 4.9 s error inside the window is caught (reviewer-verified).

Part of #404.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,20 @@ jobs:
       - name: Run tests
         run: pytest --tb=short -q
 
+  configurator-browser-smoke:
+    name: Configurator browser smoke
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262  # v4.4.0
+        with:
+          persist-credentials: false
+
+      - name: Run the smoke check's own negative fixtures
+        run: bash dashboard/looker_studio/tools/browser_smoke.sh --self-test
+
+      - name: Load the configurator in headless Chrome
+        run: bash dashboard/looker_studio/tools/browser_smoke.sh
+
   build:
     name: Build package
     runs-on: ubuntu-22.04
--- a/dashboard/looker_studio/tools/browser_smoke.sh
+++ b/dashboard/looker_studio/tools/browser_smoke.sh
@@ -1,0 +1,228 @@
+#!/usr/bin/env bash
+# Loads the configurator in a real headless browser and fails on any
+# page-level error or if the page's module never executed. Guards the class
+# of failure the Node suite structurally cannot see: specifiers or APIs that
+# resolve in Node but not in a browser (issue #404; the `node:path` incident
+# on #405).
+#
+# Detection is instrumentation-based, not keyword-based: a script injected
+# ahead of the module records window "error" events (capture phase, so
+# failed module/resource loads count), unhandled promise rejections, and
+# console.error calls, then stamps the count into the DOM where the dumped
+# document can be asserted. Chrome's own exit status and stderr are
+# additional failure triggers.
+#
+# Usage:
+#   browser_smoke.sh              run the check against ../docs
+#   browser_smoke.sh --self-test  run the negative fixtures (a page with a
+#                                 console error, an occupied port, a failing
+#                                 browser binary) and require each to fail
+#
+# Env: CHROME_BIN, SMOKE_PORT, SMOKE_DOCS_DIR override discovery.
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+SCRIPT_PATH="$SCRIPT_DIR/$(basename "$0")"
+DOCS_DIR="$(cd "${SMOKE_DOCS_DIR:-$SCRIPT_DIR/../docs}" && pwd)"
+OUT_DIR="$(mktemp -d)"
+SERVER_PID=""
+CHROME_PID=""
+
+cleanup() {
+  if [ -n "$SERVER_PID" ]; then kill "$SERVER_PID" 2>/dev/null || true; fi
+  if [ -n "$CHROME_PID" ]; then kill "$CHROME_PID" 2>/dev/null || true; fi
+  rm -rf "$OUT_DIR"
+}
+trap cleanup EXIT
+
+fail() {
+  echo "browser smoke: $*" >&2
+  exit 1
+}
+
+find_chrome() {
+  if [ -n "${CHROME_BIN:-}" ]; then
+    echo "$CHROME_BIN"
+    return
+  fi
+  for candidate in google-chrome google-chrome-stable chromium-browser chromium \
+      "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"; do
+    if command -v "$candidate" >/dev/null 2>&1; then
+      echo "$candidate"
+      return
+    fi
+  done
+  echo ""
+}
+
+# ---------------------------------------------------------------------------
+# Self-test: every fixture below must make the main check FAIL.
+# ---------------------------------------------------------------------------
+if [ "${1:-}" = "--self-test" ]; then
+  CHROME="$(find_chrome)"
+  [ -n "$CHROME" ] || fail "no Chrome/Chromium binary found (set CHROME_BIN)"
+
+  # 1. A page that reports a generic console error (no keyword the old
+  #    grep would have matched) but otherwise looks healthy, including the
+  #    aria-invalid marker.
+  FIXTURE="$OUT_DIR/fixture-console-error"
+  mkdir -p "$FIXTURE"
+  cat > "$FIXTURE/index.html" <<'HTML'
+<!doctype html>
+<html><body>
+<input aria-invalid="true">
+<script>console.error("generic boom");</script>
+</body></html>
+HTML
+  if SMOKE_DOCS_DIR="$FIXTURE" "$SCRIPT_PATH" >/dev/null 2>&1; then
+    fail "self-test 1 FAILED: a page with a console error passed"
+  fi
+  echo "self-test 1 OK: generic console error is detected"
+
+  # 2. An occupied port serving the WRONG tree: the check must notice it
+  #    does not own the port instead of validating a stranger's content.
+  DECOY="$OUT_DIR/decoy"
+  mkdir -p "$DECOY"
+  cp "$DOCS_DIR/index.html" "$DECOY/index.html" 2>/dev/null || echo "<body></body>" > "$DECOY/index.html"
+  BUSY_PORT=$((30000 + RANDOM % 10000))
+  python3 -m http.server "$BUSY_PORT" --directory "$DECOY" >/dev/null 2>&1 &
+  DECOY_PID=$!
+  disown "$DECOY_PID" 2>/dev/null || true
+  sleep 1
+  if SMOKE_PORT="$BUSY_PORT" "$SCRIPT_PATH" >/dev/null 2>&1; then
+    kill "$DECOY_PID" 2>/dev/null || true
+    fail "self-test 2 FAILED: an occupied port was treated as our server"
+  fi
+  kill "$DECOY_PID" 2>/dev/null || true
+  echo "self-test 2 OK: occupied/stale port is detected"
+
+  # 3. A browser binary that exits nonzero without producing output.
+  if CHROME_BIN="/bin/false" "$SCRIPT_PATH" >/dev/null 2>&1; then
+    fail "self-test 3 FAILED: a failing browser binary passed"
+  fi
+  echo "self-test 3 OK: nonzero browser exit is detected"
+
+  echo "browser smoke self-test OK: all negative fixtures fail as required"
+  exit 0
+fi
+
+# ---------------------------------------------------------------------------
+# Main check.
+# ---------------------------------------------------------------------------
+CHROME_BIN="$(find_chrome)"
+[ -n "$CHROME_BIN" ] || fail "no Chrome/Chromium binary found (set CHROME_BIN)"
+
+# Instrumented copy of the site: the injected script runs before the module
+# and records everything the page throws.
+SITE="$OUT_DIR/site"
+mkdir -p "$SITE"
+cp -R "$DOCS_DIR/." "$SITE/"
+python3 - "$SITE/index.html" <<'EOF'
+import pathlib
+import sys
+
+path = pathlib.Path(sys.argv[1])
+source = path.read_text()
+instrument = """<script>
+window.__smokeErrors = [];
+(function () {
+  var record = function (message) {
+    window.__smokeErrors.push(String(message));
+  };
+  window.addEventListener("error", function (event) {
+    record(event.message || (event.target && (event.target.src || event.target.href)) || "resource error");
+  }, true);
+  window.addEventListener("unhandledrejection", function (event) {
+    record(event.reason);
+  });
+  var original = console.error;
+  console.error = function () {
+    record(Array.prototype.join.call(arguments, " "));
+    original.apply(console, arguments);
+  };
+  window.addEventListener("load", function () {
+    setTimeout(function () {
+      var el = document.createElement("div");
+      el.id = "smoke-result";
+      el.setAttribute("data-errors", String(window.__smokeErrors.length));
+      el.setAttribute("data-detail", window.__smokeErrors.join(" | ").slice(0, 500));
+      document.body.appendChild(el);
+    }, 400);
+  });
+})();
+</script>"""
+marker = "<body>"
+assert marker in source, "index.html has no <body> tag to instrument"
+path.write_text(source.replace(marker, marker + instrument, 1))
+EOF
+
+# The server must provably be OURS: readiness is a nonce round-trip, not a
+# sleep, so an occupied port (our bind fails, a stranger answers) is caught.
+NONCE="smoke-nonce-$$-$RANDOM"
+echo "$NONCE" > "$SITE/$NONCE.txt"
+PORT="${SMOKE_PORT:-$((20000 + RANDOM % 20000))}"
+python3 -m http.server "$PORT" --directory "$SITE" >/dev/null 2>&1 &
+SERVER_PID=$!
+disown "$SERVER_PID" 2>/dev/null || true
+
+READY=""
+for _ in $(seq 1 20); do
+  BODY="$(curl -fsS --max-time 2 "http://127.0.0.1:$PORT/$NONCE.txt" 2>/dev/null || true)"
+  if [ "$BODY" = "$NONCE" ]; then
+    READY=1
+    break
+  fi
+  if ! kill -0 "$SERVER_PID" 2>/dev/null; then
+    break
+  fi
+  sleep 0.5
+done
+[ -n "$READY" ] || fail "server did not become ready on port $PORT (occupied by another process, or failed to start)"
+
+# Chrome occasionally lingers after dumping the DOM, so it runs in the
+# background with a deadline; a natural nonzero exit is a failure.
+"$CHROME_BIN" --headless=new --disable-gpu --no-first-run --no-sandbox \
+  --user-data-dir="$OUT_DIR/profile" --enable-logging=stderr \
+  --virtual-time-budget=5000 --dump-dom "http://127.0.0.1:$PORT/index.html" \
+  > "$OUT_DIR/dom.html" 2> "$OUT_DIR/console.log" &
+CHROME_PID=$!
+
+CHROME_STATUS=""
+for _ in $(seq 1 45); do
+  if ! kill -0 "$CHROME_PID" 2>/dev/null; then
+    wait "$CHROME_PID" && CHROME_STATUS=0 || CHROME_STATUS=$?
+    break
+  fi
+  if [ -s "$OUT_DIR/dom.html" ]; then
+    break
+  fi
+  sleep 1
+done
+if [ -z "$CHROME_STATUS" ] && kill -0 "$CHROME_PID" 2>/dev/null; then
+  sleep 2
+  kill "$CHROME_PID" 2>/dev/null || true
+  wait "$CHROME_PID" 2>/dev/null || true
+  CHROME_STATUS=0
+fi
+CHROME_PID=""
+if [ -n "$CHROME_STATUS" ] && [ "$CHROME_STATUS" -ne 0 ]; then
+  fail "browser exited with status $CHROME_STATUS"
+fi
+
+grep -q 'id="smoke-result"' "$OUT_DIR/dom.html" \
+  || fail "instrumentation marker missing — the page never finished loading"
+if ! grep -q 'data-errors="0"' "$OUT_DIR/dom.html"; then
+  DETAIL="$(grep -o 'data-detail="[^"]*"' "$OUT_DIR/dom.html" | head -1)"
+  fail "page-level errors recorded: ${DETAIL:-unknown}"
+fi
+# With no query parameters the project field is empty, so a successfully
+# loaded module runs refresh() and marks the field invalid. Static HTML
+# never contains this attribute; its presence proves app.mjs executed.
+grep -q 'aria-invalid="true"' "$OUT_DIR/dom.html" \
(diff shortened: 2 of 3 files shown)


caohy1988:
Verified the smoke check both ways locally:

- Against `main`'s `docs/`: `browser smoke OK: module loaded, no console errors, validation ran`.
- With the #405 `node:path` regression re-injected into `app.mjs`: exit 1, with the exact blocked-specifier console error surfaced. This is precisely the failure class the Node suite structurally cannot see.

Two non-blocking observations:

1. **The job isn't path-filtered and isn't in `build.needs`.** Rev 5.1 scoped it as path-filtered to `dashboard/looker_studio/docs/**`. Running on every PR is fine (it's ~10s), but then the plan's filter note should be updated; and whether the check is *blocking* depends on branch protection naming — `build` only needs `[format, test]`, so if protection requires only "Build package", this job can be red without blocking merge. Worth a one-line note in the PR description so the maintainer adds `Configurator browser smoke` to required checks.
2. **`sleep 1` server readiness** is a mild flake source; a short `curl` retry loop would be sturdier on a loaded runner. Non-blocking for a smoke script with a 45s outer budget.

The SHA-pinning + `permissions: contents: read` + `persist-credentials: false` changes to the whole file match the repo's zizmor policy as applied to the other workflows.

caohy1988:
## Fresh review — browser gate is not merge-ready

The real-browser job is the right response to the `node:path` escape, but two parts of the claimed gate do not hold yet.

1. **It is not blocking.** The active `main` ruleset currently requires only `Format check` and `Test (Python 3.10–3.14)`. `Configurator browser smoke` is not a required status check, and `Build package` has `needs: [format, test]`, so a failed smoke job does not enforce the standing merge gate described in #404.
2. **The script can false-pass.** `browser_smoke.sh` starts a fixed-port server and sleeps without proving that this process acquired the port or serves the intended tree. It also ignores the Chrome exit status and detects console failures only when the text after `CONSOLE` contains `error`, `blocked`, or `failed`; a call such as `console.error("boom")` is not guaranteed to contain any of those words in the captured message.

Before merge, please:

- make `Configurator browser smoke` required in the repository ruleset, or route it through an already-required aggregate check;
- verify server ownership/readiness instead of relying on `sleep 1` and fixed-port success;
- treat browser exit failure, `window.onerror`, unhandled rejection, and console error severity as failures without keyword matching;
- add negative fixtures for a generic console error, an occupied/stale port, and a nonzero Chrome exit.

Landing note: #412 now carries the Zizmor hardening from #411 and modifies the same `ci.yml`. Rebase whichever PR lands second and confirm that both the browser job and the v4.4.0/action-hardening changes survive.

caohy1988:
Both false-pass modes from the review above are real — reproduced locally at `d837089f`:

1. **Generic console error missed:** appended `console.error("boom")` to `app.mjs` → `browser smoke OK`, exit 0. Chrome logs it as `INFO:CONSOLE` with just `"boom"` — no `error`/`blocked`/`failed` keyword for the grep to catch. The `aria-invalid` assertion backstops a dead module but nothing after successful load.
2. **Stale-port false pass:** started a *good* server on the port, then ran the script with a broken tree (`node:path` re-injected) pointed at the same port. The script's own server fails to bind silently (output to /dev/null), Chrome loads the stale server's good page → `browser smoke OK`, exit 0. A readiness check must verify the server *it started* owns the port and serves the intended tree (e.g. curl a sentinel from DOCS_DIR), not just that *something* answers.

One more landing detail on the #412 overlap: current `main` (6322e13, via #411/#412) SHA-pins `actions/checkout@11d5960a` = **v4.4.0**, but this PR pins `34e11487` = **v4.3.1**, and the branch doesn't contain current main. Whoever rebases should keep v4.4.0 and layer the smoke job on top — otherwise the merge quietly downgrades checkout.

caohy1988:
All four findings addressed at the new head (branch rebuilt on post-#412 `main`, so the landing-order conflict is resolved on this side):

1. **Enforcement** — verified the ruleset claim (`Format check` + `Test (Python 3.10–3.14)` only). Rather than requiring an admin ruleset change, the smoke check now runs inside pytest (`test_configurator_loads_in_a_real_browser`, Chrome-gated skip), so the **already-required** Test checks enforce it. The standalone job remains as fast signal; making it required is now optional.
2. **Server ownership/readiness** — randomized port + nonce round-trip replaces the fixed port + `sleep 1`; the check fails unless its own server answers.
3. **Error detection** — instrumentation-based (capture-phase `error` events, `unhandledrejection`, `console.error` hook, count stamped into the dumped DOM). Chrome's natural exit status is honored; the stderr keyword scan is demoted to belt-and-braces.
4. **Negative fixtures** — `--self-test` covers exactly the three requested cases and must see each fail; it runs as a CI step before the real check. Verified locally: positive green, all three fixtures detected.

One deliberate divergence to flag: `console.error` interception catches page-authored errors, and capture-phase `error` listeners catch resource/module failures — but errors thrown by extensions or cross-origin scripts with muted messages surface only as `Script error.`; on this zero-dependency page every script is same-origin, so that gap is theoretical here.

caohy1988:
Re-review at `06e9482f` — **every finding from the last round is closed, verified by running it:**

- **False-pass on generic `console.error`: fixed.** Detection is now instrumentation-based (capture-phase `error` listener, `unhandledrejection`, wrapped `console.error`, count stamped into the DOM as `data-errors`). My `console.error("boom")` repro from the last round is exactly self-test fixture 1, and `--self-test` fails it as required.
- **Stale/fixed-port false pass: fixed.** Server readiness is a nonce round-trip (write `$NONCE.txt` into the served tree, require the exact nonce back over HTTP) — self-test fixture 2 proves an occupied port serving a stranger's tree fails. My earlier repro is covered.
- **Browser exit status ignored: fixed** — nonzero Chrome exit fails (fixture 3), plus a belt-and-braces stderr grep (now including `uncaught`).
- **Not blocking: fixed, and in a merge-proof way.** `test_configurator_loads_in_a_real_browser` runs the smoke inside the required `Test (Python N)` matrix jobs, so the existing ruleset enforces it with no settings change; the standalone job runs self-test + main check as belt and braces.
- **checkout pin regression: avoided** — the rebase kept `11d5960a` (v4.4.0).

Local runs: `--self-test` → all three negative fixtures fail as required; main check → OK; full looker_studio pytest → 23 passed. CI green on the real runner, smoke job included.

LGTM.

caohy1988:
## Fresh full re-review at `06e9482f` - prior LGTM withdrawn

The earlier fixes materially improve the gate, and the current CI is green. Two silent-pass paths remain.

### P1 - a browser that emits DOM and then exits nonzero still passes

In `dashboard/looker_studio/tools/browser_smoke.sh:190-210`, the loop stops as soon as `dom.html` becomes nonempty. If Chrome is still alive two seconds later, the script kills it, ignores the real `wait` result, and unconditionally assigns status 0. If Chrome exits between the DOM break and the second process check, the status remains empty and also passes.

I reproduced this with a fake Chrome that writes a healthy-looking instrumented DOM, waits one second, and exits 42. The script printed `browser smoke OK` and exited 0. Self-test fixture 3 uses `/bin/false`, so it covers only a process that exits before producing DOM.

Required fix: always reap a naturally exited child and preserve its status. Treat a deliberate timeout kill separately. Add a slow-nonzero-after-healthy-DOM fixture.

### P1 - the required check can silently skip the browser gate

In `tests/test_looker_studio_dashboard.py:563-580`, the real-browser test uses `skipif` when Chrome is absent. The required Python matrix therefore succeeds with a skip if the runner image loses Chrome. The optional standalone smoke job does not restore merge enforcement.

Required fix: make missing Chrome a hard failure in CI, or explicitly provision Chrome in every required job and assert that it exists. Keep the skip only for non-CI local environments, and test the CI-without-Chrome path.

Local verification: the current pytest file is otherwise 23/23, all three existing self-tests behave as written, and the positive smoke passes with Chrome 151. Those results do not exercise either false-pass path above.

**Verdict: not merge-ready. No LGTM until both required-gate failures are fixed and covered.**

caohy1988:
Independent reproduction of the exit-status silent pass, at `06e9482f`:

```
fake chrome: prints a fully healthy DOM (aria-invalid + smoke-result data-errors="0"), sleeps 5s, exits 42
CHROME_BIN=/tmp/fake-chrome.sh bash tools/browser_smoke.sh
→ "browser smoke OK: module loaded, zero page-level errors, validation ran", exit 0
```

The race is exactly as described: `--dump-dom` makes `dom.html` non-empty while the process is still alive; the loop breaks on `-s dom.html`, the kill branch forces `CHROME_STATUS=0`, and the assertions pass on the healthy DOM. A crash-after-dump browser is invisible to the gate.

On the skip point: `@pytest.mark.skipif(not _chrome_available(), ...)` means a Chrome-less runner records the required `Test (Python N)` check green without running the smoke. Ubuntu runners carry Chrome today, but image contents are not a contract. Cheap hardening: fail instead of skip when `os.environ.get("CI")` is set, so the skip only applies to local dev machines.

caohy1988:
Both P1s are fixed at the new head, and both were **reproduced before fixing** (your exit-42 fake Chrome false-passed here too, verbatim).

1. **Exit status always reaped.** The wait loop no longer assigns 0 after a deliberate kill or leaves the status empty on a racing natural exit. New flow: wait for DOM (45s) → grace window (10s) for a natural exit → only if the process is still alive *and this script's own kill succeeds* is the exit exempted; otherwise `wait` reaps the real status and nonzero fails. Your repro now dies with `browser exited with status 42`, and **self-test fixture 4** (healthy instrumented DOM that satisfies every DOM assertion, linger, exit 42) pins it in CI.
2. **CI cannot silently skip the gate.** `_browser_gate_disposition(chrome_available, in_ci)`: run with a browser; **hard AssertionError in CI without one** (`CI` env, set by GitHub Actions on every runner); skip only outside CI. `test_browser_gate_cannot_silently_skip_in_ci` covers all four combinations including the CI-without-Chrome path you asked for — so if the runner image ever loses Chrome, the required Test checks go red instead of green-by-skip.

Verified locally: exit-42 repro fails, all **four** self-test fixtures detected, positive smoke green, dashboard pytest 24/24.

caohy1988:
Re-review at `b92200bc` — both defects closed, verified by running:

- **Exit-status race**: the child's status is now always reaped (`wait` captures the real code; a grace window lets a browser that produced output finish naturally), and only this script's own successful timeout kill is exempt. My exact repro from the last round (healthy DOM, linger, exit 42) now fails the check with "browser exited with status 42" — and it's pinned as self-test fixture 4.
- **Silent skip in CI**: `_browser_gate_disposition` hard-fails when Chrome is missing and `CI` is set, skips only locally, with all four dispositions unit-tested. That's exactly the enforcement shape needed inside the required Test jobs.

Self-test 4/4, main smoke OK, 24/24 pytest locally. LGTM — nothing left on this PR from my side.

caohy1988:
## Rev 6.2 fresh re-review at `b92200bc`

Both previous P1 runtime defects are closed. I ran the current branch directly:

- all four negative fixtures were detected, including healthy instrumented DOM followed by exit 42;
- the positive Chrome smoke passed;
- dashboard pytest passed 24/24;
- the CI-without-Chrome disposition now raises instead of skipping.

One new merge-enforcement gap and one documentation mismatch remain.

### P2 - fixture 4 is not enforced by a required check

The live main ruleset requires only `Format check` and `Test (Python 3.10-3.14)`. The required pytest path runs only the positive smoke. `browser_smoke.sh --self-test` appears only in the standalone `Configurator browser smoke` job, which is not required.

A future reintroduction of the exit-status false pass would keep every required check green while only the optional job turns red. The fix is correct today, but the claimed regression pin is not merge-proof.

Required fix: run `--self-test` inside one required matrix job, for example a conditional step in `Test (Python 3.12)`, or make the standalone job required.

### P2 - the public description still documents the old behavior

The PR body still says the pytest test skips when Chrome is absent and still lists three fixtures. The script usage comment at `dashboard/looker_studio/tools/browser_smoke.sh:17-19` also omits fixture 4. Update both to say that missing Chrome fails in CI and that four negative fixtures run.

**Verdict: Ready with these fixes, but no final LGTM yet. The two original P1s are verified closed.**

caohy1988:
Confirmed the enforcement gap at `b92200bc`:

- `tests/test_looker_studio_dashboard.py:613` runs `["bash", "tools/browser_smoke.sh"]` — the required Test jobs never invoke `--self-test`. The negative fixtures run only in the standalone `configurator-browser-smoke` job (`ci.yml:93`), which the ruleset doesn't require. So a detection-logic regression (including the fixture-4 path) would leave every required check green. One-line fix: add a second `subprocess.run` for `--self-test` in the same test (or a sibling test), so the fixtures ride the required gate too.
- The usage comment at `browser_smoke.sh:17` still says three fixtures; there are four. Stale as reported.

caohy1988:
Both P2s addressed at the new head:

1. **Fixture enforcement** — `test_browser_smoke_negative_fixtures_are_detected` now runs `--self-test` inside the required `Test (Python N)` matrix with the same CI-hard-fail disposition as the positive gate. A reintroduced false-pass path (fixture 4 included) turns required checks red. I ran it in all matrix jobs rather than a 3.12-conditional step — no version coupling to silently lose the gate if the matrix changes, and the CI cost is seconds per job since Chrome exits promptly after `--dump-dom` on runners.
2. **Descriptions match the code** — the PR body now states the CI hard-fail (no skip) behavior and all four fixtures, and the script's usage comment documents fixture 4. The #404 plan body is being updated in the same round (rev 6.3) for the same staleness.

Local verification at this head: dashboard pytest 25/25 including the in-pytest self-test; all four fixtures detected; positive smoke green.

caohy1988:
Re-review at `0e6e8cbc` — the enforcement gap is closed:

- `test_browser_smoke_negative_fixtures_are_detected` runs `--self-test` inside the required Test jobs with the same CI-hard-fail disposition, so a detection regression now turns a required check red. Verified locally: 25/25 pytest (both browser tests execute and pass on a machine with Chrome), CI green on the runner including the new test in the matrix.
- The usage comment now lists all four fixtures.

That was the last open item on this PR. LGTM.

caohy1988:
## Full fresh re-review at `0e6e8cbc` — prior LGTM withdrawn

The new required-matrix enforcement is real: I ran the current head, all four negative fixtures were detected, the positive Chrome smoke passed, and dashboard pytest passed **25/25** with both browser tests executing. The active ruleset also still requires all five `Test (Python N)` jobs.

One P1 false-pass remains.

### P1 — delayed page errors are recorded after the result snapshot, but never reflected in the dumped DOM

`browser_smoke.sh:166-173` appends `#smoke-result` once, 400 ms after `load`, with the error count at that instant. `record()` only pushes into `window.__smokeErrors`; it never updates an existing marker. A normal asynchronous initialization error after that 400 ms snapshot therefore leaves `data-errors="0"`. A generic `console.error` also misses the stderr keyword fallback.

I reproduced this against the exact PR head with a page that schedules the following from its `load` handler:

```js
setTimeout(() => console.error("generic delayed boom"), 900);
```

The callback runs within Chrome's 5-second virtual-time budget, but the current script exits 0 with:

```text
browser smoke OK: module loaded, zero page-level errors, validation ran
```

This is a required-check green / browser-red path, and none of the four existing self-test fixtures exercises it.

**Required fix:** make the error marker live (for example, have `record()` update the marker whenever one exists, while retaining the initial count for pre-marker errors), and add a negative fixture whose generic error fires after the current snapshot delay. The self-test for that fixture must continue to run through `test_browser_smoke_negative_fixtures_are_detected` in the required matrix.

Until that regression is pinned, this PR is not merge-ready.

caohy1988:
Independently reproduced at `0e6e8cbc`, same result:

```
app.mjs += 'setTimeout(function () { console.error("generic delayed boom"); }, 900);'
→ browser smoke OK: module loaded, zero page-level errors, validation ran (exit 0)
```

The mechanism is as described: `#smoke-result` is stamped once at load+400ms with the count at that instant; `record()` at 900ms only pushes into `__smokeErrors`, the dumped DOM still says `data-errors="0"`, and the stderr keyword fallback can't see a plain `console.error` message. The 5s virtual-time budget makes the 900ms timer fire before the dump, so the error is captured by the page but never reaches the assertion.

One implementation note on the required fix: making `record()` update an existing marker handles post-snapshot errors, but pre-marker errors still need the initial stamp to read `__smokeErrors.length` at creation time (it already does) — so the two-sided contract is: stamp reads current count at creation; `record()` rewrites `data-errors`/`data-detail` whenever the marker exists. The delayed-error fixture should sit in the required pytest path as stated. Withdrawing my LGTM until this lands.

caohy1988:
P1 fixed at the new head, reproduced first: your 900 ms `console.error` fixture passed verbatim on the prior head here, and now fails with `page-level errors recorded: data-detail="generic delayed boom"`.

- **Live marker.** `record()` re-stamps `#smoke-result` on every recorded error, so the dumped DOM reflects everything up to the end of the observation window (the 5 s virtual-time budget) — not a 400 ms snapshot. The initial stamp at marker creation still covers pre-marker errors; there is no window in which an error is collected but unreflected.
- **Fixture 5** — a generic console error scheduled 900 ms after `load`, healthy otherwise — pins the path, and it runs through `test_browser_smoke_negative_fixtures_are_detected` in the **required** matrix like the other four. The script usage comment now enumerates all five without a count that can go stale.

Verified at this head: your repro fails, all **five** fixtures detected, positive smoke green, dashboard pytest **25/25**. One boundary stated for the record: the observation window is the virtual-time budget — an error scheduled beyond it is invisible to any dump-based check by construction; the budget (5 s of virtual time, which Chrome fast-forwards) is the contract.

caohy1988:
Re-review at `56804f86` — the delayed-error P1 is closed, verified by running my own repro against it:

- **Live marker**: `record()` now re-stamps the marker on every captured error, and the initial stamp reads the count at creation (pre-marker errors retained) — the two-sided contract holds. My exact repro from the last round (900ms-delayed `console.error("generic delayed boom")`) now fails the check with the detail surfaced: `page-level errors recorded: data-detail="generic delayed boom"`, exit 1.
- **Fixture 5** pins the delayed-error path and runs in `--self-test`, which the required Test matrix executes.
- Self-test 5/5 fixtures detected; positive smoke OK; 25/25 pytest locally; 15/15 checks green on the PR.

That closes every finding across all rounds. LGTM restored — merge-ready.

caohy1988:
## Fresh full re-review at `56804f86` — runtime fix verified; final LGTM withdrawn for landing cleanup

The delayed-error P1 is closed. I reran the red/green repro: the prior head false-passes the 900 ms error, while this head fails with `data-detail="generic delayed boom"`. I also moved the error to 4.9 seconds inside the declared 5-second virtual-time window; this head still catches it with the exact error text. Self-test is 5/5, the positive smoke passes, dashboard pytest is 25/25, and all 15 checks on this head are green.

One P2 description mismatch remains:

- The PR body still says **“Negative fixtures — FOUR”** and **“all four fixtures detected.”** It does not list fixture 5 or qualify the “no false-pass paths” heading with the 5-second observation-window contract.
- The required pytest path still says “The four negative fixtures” at [`tests/test_looker_studio_dashboard.py:620`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/56804f860f775ae3fc4d27b44a9cd5c28c590c2b/tests/test_looker_studio_dashboard.py#L620-L623). The script usage comment is correct and count-free; this test comment is the remaining stale source description.

Please update the PR body to describe the live marker, five fixtures, and explicit 5-second bound, and update the pytest comment to be count-free or say five.

There is also a mechanical landing gate: GitHub reports this head **BEHIND** current `main` (`e964632`, #413), and the `main` ruleset has `strict_required_status_checks_policy: true`. Sync the branch and rerun the required checks; the PR also still needs one maintainer approval.

**Verdict:** code behavior LGTM, but not final merge-ready at this exact head. Restore final LGTM after the description/test-comment cleanup, branch sync, green rerun, and maintainer approval.

caohy1988:
All three landing items verify at `56804f86`:

1. PR body still says "Negative fixtures — FOUR" / "all four fixtures detected" (fixture 5 and the 5s-bound contract missing).
2. `tests/test_looker_studio_dashboard.py:620` still reads "The four negative fixtures" — the stale source description, as named.
3. `mergeStateStatus: BEHIND` against `e964632` with `strict_required_status_checks_policy: true` on the main ruleset — so sync + rerun is mechanical but mandatory, alongside the REVIEW_REQUIRED approval.

Code behavior itself remains verified good from my own runs (exit-42 and 900ms-delayed repros both die correctly, 5/5 fixtures, 25/25 pytest). Agreed: LGTM restores after the description/test-comment cleanup + sync + green rerun + approval.

caohy1988:
All three landing items done at `fc95e737`:

1. **PR body updated** — five fixtures enumerated (delayed-error pin included), the live-marker behavior described, and the 5-second virtual-time observation window stated as the explicit contract.
2. **Test comment updated** — `test_browser_smoke_negative_fixtures_are_detected` now says five fixtures and names the live marker and the 5 s window.
3. **Branch synced** — rebased onto current `main` (post-#413); `mergeStateStatus` is no longer `BEHIND`. CI is rerunning on the new head for the strict up-to-date policy; remaining steps are the green rerun and maintainer approval.

No code changes this round beyond the comment text — runtime behavior is exactly what the last review verified.

caohy1988:
## Landing re-review at `fc95e737` — LGTM restored

All items from the prior landing review are closed and freshly verified:

- The PR body now names the live marker, all five fixtures, and the explicit 5-second virtual-time observation window.
- The required pytest test comment now names five fixtures and the live-marker/window contract.
- The branch base is current `main` (`e964632`, post-#413); GitHub no longer reports `BEHIND`.
- The smoke script and CI workflow are byte-for-byte unchanged from the previously runtime-verified head. The new commit changes only the pytest comment; the rebase adds #413's package-lock update.
- The post-rebase run is fully green: positive browser smoke, all five negative fixtures through every required Python job, all five `Test (Python N)` checks, format, build, security scans, and CLA.

**LGTM — contributor-side work is complete.** The PR remains `BLOCKED` only because the ruleset requires one maintainer approval; this account is the PR author and cannot supply that formal approval.

haiyuan-eng-google (approved):

PR #412, opened 2026-08-10
fix(locks): hold pip below 26.2 so the lock generator converges again

haiyuan-eng-google:
The **Hash-lock drift check** has been failing on every PR since pip 26.2 shipped. This is the input-side fix; see the caveat at the bottom — the lock files still need regenerating.

## Root cause

pip 26.2 added a required `allow_editables` keyword argument to `RequirementCommand.make_requirement_preparer()`. `pip-compile` does not pass it, so phase 2 of the convergence loop in `regen-locks.sh` — which installs the phase-1 candidate into a fresh container and re-resolves — dies with:

```
TypeError: RequirementCommand.make_requirement_preparer() missing 1
required keyword-only argument: 'allow_editables'
```

`pip` enters `pip-tools.lock` via `--allow-unsafe`, so with no upper bound every resolution pulls the newest pip and reintroduces the crash.

## Timeline

| Date | Event |
| --- | --- |
| 2026-05-31 | pip 26.1.2 — the version currently pinned in `pip-tools.lock`, works |
| 2026-07-16 | last green `producers-ci` run on `main` |
| 2026-07-18 | pip-tools 7.6.0 released |
| **2026-07-29** | **pip 26.2 released — breakage starts** |
| 2026-08-04 | pip 26.2.1 |

The last green run predates pip 26.2, which brackets the cause precisely.

## Why not just upgrade pip-tools

pip-tools' newest release, 7.6.0, predates pip 26.2 by 11 days and breaks identically. Both 7.5.3 and 7.6.0 declare only `pip>=22.2`, so package metadata offers no protection — the bound has to be stated in `pip-tools.in`.

## ⚠️ This PR is not complete on its own

The three lock files still need regenerating:

```
bash deploy/otlp_receiver/regen-locks.sh
```

That must run inside the digest-pinned base image via docker, which I could not do (no `docker` group membership on my machine). **The drift check will still fail on this PR until the regenerated locks are committed alongside this change.**

Expect the regenerated `pip-tools.lock` to keep `pip==26.1.2` and to move `pip-tools` 7.5.3 → 7.6.0, since the bound only constrains pip.

## Relationship to #411

Independent. #411 is the Zizmor hardening work and is Zizmor-green; this failure predates it and is unrelated to those changes. Raised separately so the two can be reviewed on their own merits.


caohy1988 (approved):

caohy1988:
## Fresh review of combined head `5a4c8bfedf68`

I found no new code blocker in the #411 merge commit, and all current checks are green. The PR description is now materially stale, though:

- it says the three lock files still need regeneration and the drift check will fail, but the locks are present and `Hash-lock drift check` passes;
- it calls #411 independent, but #411 was merged into this branch and closed. Its workflow hardening is not on `main` until this PR lands.

Please update the title/body to describe the actual combined scope, the 22 dependency updates produced by regeneration, the passing lock/image validation, and the incorporated Zizmor hardening. That gives reviewers an accurate diff contract instead of the earlier input-only state.

There is also a landing-order dependency with #407 because both change `.github/workflows/ci.yml`. Rebase whichever lands second and rerun CI, verifying that the final file retains the browser-smoke job, checkout v4.4.0 pin, least-privilege permissions, and `persist-credentials: false`.

PR #414, opened 2026-08-11
feat(export): add BigQuery-to-LangSmith connector

### LINKED ISSUE #410: [Feature] LangSmith export connector: dump BQAA logging data, schema/row-content agnostic
## Summary

Build a connector that dumps BQAA logging data (agent trace events stored in BigQuery) into [LangSmith](https://smith.langchain.com/), so teams that use LangSmith for trace inspection, annotation, and evaluation can consume BQAA telemetry there without a custom pipeline.

The connector must be **schema/row-content agnostic**: it should not hard-code assumptions about the event table's schema or the contents of any row payload.

## Motivation

- BQAA/BigQuery is the system of record for agent telemetry; LangSmith is where many teams already run trace review, human annotation queues, and eval workflows.
- Today, moving BQAA data into LangSmith requires a hand-rolled ETL for every team and every schema variant.
- A first-class, content-agnostic export connector closes that gap and adds a new consumption surface alongside the existing Looker Studio / Grafana / BigFrames surfaces.

## Requirements

1. **Schema/row-content agnostic**
   - No hard-coded column or payload assumptions. The connector works from a configurable field mapping (source column/JSON path → LangSmith run field), with a sensible default mapping for the standard ADK BigQuery Trace Exporter `agent_events` schema.
   - Columns not covered by the mapping are passed through untouched as LangSmith run metadata (`extra`), so arbitrary/custom schemas survive round-trip without data loss.
   - Row payload contents are treated as opaque: passed as-is, never parsed for semantics beyond the mapping.
2. **Source flexibility** — accept a table reference (`project.dataset.table`) or an arbitrary SQL query as the source.
3. **Trace mapping** — reconstruct the trace/span hierarchy (reusing the SDK's existing trace reconstruction) and map it to LangSmith's run tree: one root run per trace, child runs per span/event, with timestamps, latency, and error status preserved when present in the mapping.
4. **Modes**
   - One-shot backfill (time range and/or SQL filter).
   - Incremental sync with a persisted watermark (event timestamp / ingestion time), suitable for Cloud Scheduler.
5. **Idempotency** — derive stable LangSmith run IDs from source event IDs so re-runs upsert instead of duplicating.
6. **Operational controls** — batching, rate limiting, retry with backoff on LangSmith API errors; progress logging with exported/skipped/failed row counts and a dropped-row report.
7. **Config & auth** — LangSmith API key + project via env/config; BigQuery auth via ADC like the rest of the SDK; jobs labeled per the SDK usage-telemetry convention (`sdk_surface=langsmith_export`).

## Proposed surfaces

- Python API: `bqaa.export.langsmith.export(...)`
- CLI: `bq-agent-sdk export langsmith --source ... --mapping mapping.yaml --since ...`

## Non-goals

- Two-way sync (LangSmith → BigQuery import).
- Payload transformation or semantic normalization beyond the declared field mapping.
- Real-time streaming in v1 (scheduled incremental sync is sufficient; a streaming surface can follow the existing Cloud Run pattern later).

## Acceptance criteria

- [ ] Default mapping exports a standard `agent_events` table to LangSmith with correct trace/run hierarchy.
- [ ] A custom-schema table with unknown columns exports without error; unmapped columns appear as run metadata.
- [ ] Re-running the same export window produces no duplicate runs in LangSmith.
- [ ] Backfill and incremental modes covered by tests; CLI documented in README/SDK.md.

caohy1988:
## Summary

BQAA trace data can now flow from BigQuery into LangSmith without a custom ETL. The connector accepts a table or arbitrary SQL, maps only declared fields, and carries all other columns into run metadata without interpreting payload semantics.

It supports bounded backfills and scheduler-friendly incremental sync. Stable source-derived UUIDs make overlapping runs idempotent, while failure-aware watermarking checkpoints reported skips so they do not wedge later runs, while any failed LangSmith batch blocks advancement.

Closes #410.

## Design decisions

- **Schema independence:** `FieldMapping` uses ADK defaults only when no custom mapping is supplied, while custom mappings opt into every field they consume. BigQuery-specific scalar values receive transport-safe encodings; only the built-in ADK content and latency JSON fields are decoded when a client returns them as strings.
- **Trace integrity:** the SDK's existing span-tree builder remains the hierarchy authority. Each trace gets one stable synthetic LangSmith root, complete dotted ancestry, deterministic child ordering, cycle handling, and safe orphan flattening when a query window omits an ancestor.
- **Honest delivery:** batches have a hard source-row cap, transient LangSmith errors use bounded exponential retry, per-trace chunk accounting prevents partial success from advancing state, and dropped-row details are bounded without losing aggregate counts.
- **Operational safety:** incremental state is atomically replaced and bound to the source, field mapping, and LangSmith destination. The CLI keeps the API key environment-only, exits nonzero for skipped or failed rows, and labels BigQuery jobs with `sdk_surface=langsmith-export` and `sdk_feature=langsmith-export`.

## Validation

- `3990 passed, 69 skipped` on Python 3.13 after the review-fix pass.
- Repository autoformat and `git diff --check` are clean.
- A live LangSmith 0.10.17 run on the prior head confirmed stable UUID ingestion, real synthetic-root times in recent-window queries, multi-chunk traces, and duplicate-free replay. It also proved exporter replays cannot mutate existing run IDs; this head now uses create-only batches to match that server policy.
- Intercepted LangSmith 0.10.17 serializer tests verify one create request per batch and full payload reconstruction on retry.
- Isolated sdist and wheel builds succeeded, and the wheel contains the exporter/CLI modules plus the `langsmith` optional dependency metadata.

### LangSmith dotted-order ancestry

LangSmith batch ingestion needs both a stable `trace_id` and a `dotted_order` string that contains the full path from the trace root to each run. `parent_run_id` alone is not enough for the batch API.

This connector derives UUIDv5 IDs from the canonical BigQuery source plus source trace/run IDs. A stable synthetic root gives multi-root BQAA traces one LangSmith tree, and every child appends its timestamp/UUID segment to its parent's dotted order. Create-only batches carry the complete run on first write. Stable IDs make replay an idempotent no-op for existing IDs while still allowing previously unseen IDs; corrections require a fresh LangSmith project or a deliberately versioned `--source-id`.

This pattern is specific to importing existing hierarchical traces. New traces created through LangSmith's normal tracing helpers should let the SDK construct these fields.

## Post-deploy verification

- **Logs:** watch `LangSmith export progress` and `LangSmith export batch failed`.
- **BigQuery:** verify jobs in `INFORMATION_SCHEMA.JOBS*` carry `sdk_surface=langsmith-export` and `sdk_feature=langsmith-export`; track job errors and bytes billed.
- **Connector metrics:** track exported/skipped/failed rows, completed batches, and watermark age. Healthy runs have `failed=0`, no unexpected skips, and an advancing watermark. Replaying the same window must not increase the destination run count.
- **Failure response:** pause the scheduler if failures rise above zero, the watermark stalls, 429/5xx errors repeat, or query cost changes unexpectedly. Correct the cause and replay the same window; stable IDs make that retry safe.
- **Window and owner:** validate for 24–48 hours after first scheduled use. The connector operator (or PR owner for the initial rollout) owns the check and rollback decision.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,14 +120,16 @@ The CLI intentionally accepts the API key only through
 `LANGSMITH_API_KEY`, keeping it out of shell history and process arguments.
 
 The exporter reconstructs span parents, derives stable LangSmith UUIDs, and
-uses batch upserts, so replaying an overlapping window does not duplicate runs.
-For scheduled syncs, add `--incremental --watermark-file=state.json`. The JSON
-summary bounds row-level diagnostics with `--max-dropped-rows` and reports the
-number omitted as `dropped_rows_truncated`.
+uses separate create and update batches, so replaying an overlapping window
+updates existing runs without creating duplicates. For scheduled syncs, add
+`--incremental --watermark-file=state.json`. The JSON summary bounds row-level
+diagnostics with `--max-dropped-rows` and reports the number omitted as
+`dropped_rows_truncated`.
 
 Custom schemas use a YAML mapping from LangSmith fields to source column or
 nested paths. Unmapped columns remain in `extra.metadata`; payload values are
-opaque and are never classified by event type:
+opaque and are never classified by event type. Optional fields omitted from a
+custom mapping remain unmapped rather than inheriting ADK column names:
 
 ```yaml
 fields:
--- a/README.md
+++ b/README.md
@@ -119,10 +119,12 @@ bq-agent-sdk export langsmith \
 The CLI intentionally accepts the API key only through
 `LANGSMITH_API_KEY`, keeping it out of shell history and process arguments.
 
-The exporter reconstructs span parents, derives stable LangSmith UUIDs, and
-uses separate create and update batches, so replaying an overlapping window
-updates existing runs without creating duplicates. For scheduled syncs, add
-`--incremental --watermark-file=state.json`. The JSON summary bounds row-level
+The exporter reconstructs span parents and derives stable LangSmith UUIDs. It
+treats created runs as immutable: replaying an overlapping window is an
+idempotent no-op for existing run IDs while still creating previously unseen
+IDs. To correct already-exported data, use a fresh LangSmith project or a
+deliberately versioned `--source-id`. For scheduled syncs, use `--incremental`
+with `--watermark-file=state.json`. The JSON summary bounds row-level
 diagnostics with `--max-dropped-rows` and reports the number omitted as
 `dropped_rows_truncated`.
 
--- a/README.md
+++ b/README.md
@@ -86,6 +86,12 @@ With BigFrames support:
 pip install bigquery-agent-analytics[bigframes]
 ```
 
+With LangSmith export support:
+
+```bash
+pip install bigquery-agent-analytics[langsmith]
+```
+
 ## Quick Start
 
 ```python
@@ -95,6 +101,52 @@ client = Client(project_id="my-project", dataset_id="analytics")
 trace = client.get_trace("trace-abc-123")
 trace.render()
 ```
+
+### Export traces to LangSmith
+
+Export the standard ADK `agent_events` schema with Application Default
+Credentials and LangSmith's standard environment variables:
+
+```bash
+export LANGSMITH_API_KEY=lsv2_...
+export LANGSMITH_PROJECT=agent-production
+
+bq-agent-sdk export langsmith \
+  --source=my-project.analytics.agent_events \
+  --since=2026-08-01T00:00:00Z
+```
+
+The CLI intentionally accepts the API key only through
+`LANGSMITH_API_KEY`, keeping it out of shell history and process arguments.
+
+The exporter reconstructs span parents, derives stable LangSmith UUIDs, and
+uses batch upserts, so replaying an overlapping window does not duplicate runs.
+For scheduled syncs, add `--incremental --watermark-file=state.json`. The JSON
+summary bounds row-level diagnostics with `--max-dropped-rows` and reports the
+number omitted as `dropped_rows_truncated`.
+
+Custom schemas use a YAML mapping from LangSmith fields to source column or
+nested paths. Unmapped columns remain in `extra.metadata`; payload values are
+opaque and are never classified by event type:
+
+```yaml
+fields:
+  run_id: event_key
+  trace_id: trace.key
+  parent_run_id: parent_key
+  name: kind
+  start_time: occurred_at
+  inputs: payload
+```
+
+```bash
+bq-agent-sdk export langsmith \
+  --source='SELECT * FROM `my-project.custom.events`' \
+  --mapping=mapping.yaml --source-id=custom-events-v1
+```
+
+See [SDK.md](SDK.md#23-langsmith-export) for the Python API, incremental
+watermark contract, filtering, and operational controls.
 
 For session reads, `session_id` is a reusable conversation identifier rather
 than a unique trace key. `client.get_session_trace()` resolves user, root agent,
@@ -229,6 +281,12 @@ src/bigquery_agent_analytics/
 │   ├── insights.py                # Multi-stage insights pipeline
 │   ├── feedback.py                # Drift detection & question distribution
 │   └── memory_service.py          # Long-horizon agent memory
+│
+├── Export
+│   └── export/
+│       ├── __init__.py             # Stable public export API
+│       ├── cli.py                  # bq-agent-sdk export command group
+│       └── langsmith.py            # Schema-agnostic LangSmith connector
 │
 ├── Agent Context Graph
 │   ├── context_graph.py           # Decision-trace extraction & GQL traversal
--- a/SDK.md
+++ b/SDK.md
@@ -2291,9 +2291,10 @@ The optional LangSmith connector exports a BigQuery table or arbitrary SQL
 result as LangSmith run trees. It streams rows ordered by trace, reconstructs
 the existing `span_id` / `parent_span_id` hierarchy, and adds one stable
 structural root per trace. Every source event gets a UUID derived from the
-canonical source, trace ID, and run ID. Each batch uses separate create and
-update requests, so overlapping backfills update the same LangSmith runs
-instead of creating duplicates or silently leaving stale data.
+canonical source, trace ID, and run ID. The exporter treats created runs as
+immutable: overlapping backfills are idempotent no-ops for existing run IDs
+while still creating previously unseen IDs. Correcting already-exported data
+requires a fresh LangSmith project or a deliberately versioned `--source-id`.
 
 ### Install and authenticate
 
@@ -2334,11 +2335,14 @@ print(stats.to_dict())
 `ExportStats.to_dict()` reports source rows read, exported, skipped, failed,
 successful traces, batches, the current watermark, bounded `dropped_rows`
 details, and `dropped_rows_truncated` (the number omitted from that list).
+`exported` counts source rows in batches accepted by the client, not destination
+mutations, so a replay of existing IDs can report exported rows while leaving
+the server-side runs unchanged.
 Retryable 408/425/429 and 5xx errors use bounded exponential backoff. A create
-conflict is expected during replay, is not retried, and is followed by the
-update request. `requests_per_second` and `max_retries` govern exporter calls
-to `batch_ingest_runs`; the LangSmith SDK may split or retry transport requests
-inside one call. A failed batch never advances incremental state.
+conflict is expected during replay, is not retried, and is treated as an
+idempotent success. `requests_per_second` and `max_retries` govern exporter
+calls to `batch_ingest_runs`; the LangSmith SDK may split or retry transport
+requests inside one call. A failed batch never advances incremental state.
 
 ### Custom schemas and opaque values
 
@@ -2435,8 +2439,10 @@ be retained.
 The cursor is based on source event time. A row that arrives after the cursor
 has advanced but carries an equal or earlier `start_time` is not discovered by
 a later incremental run. Schedule an overlapping bounded backfill when the
-source permits late arrival; stable IDs and update batches make that replay
-safe.
+source permits late arrival. Stable IDs make the replay idempotent: previously
+unseen run IDs are created, but existing runs remain unchanged. To correct an
+already-exported run, export to a fresh LangSmith project or use a deliberately
+versioned `--source-id`.
 
 Each query window reconstructs hierarchy only from rows present in that
 window. If a child arrives after its parent was exported in an earlier window,
--- a/SDK.md
+++ b/SDK.md
@@ -2250,7 +2250,7 @@ pipeline.
 | --- | ----- |
 | `sdk` | constant `bigquery-agent-analytics` |
 | `sdk_version` | package version, BQ-safe (e.g. `0-4-0`) |
-| `sdk_surface` | `python` \| `cli` \| `remote-function` \| `langsmith_export` |
+| `sdk_surface` | `python` \| `cli` \| `remote-function` \| `langsmith-export` |
 | `sdk_feature` | `trace-read` \| `eval-code` \| `eval-llm-judge` \| `eval-categorical` \| `insights` \| `drift` \| `memory` \| `context-graph` \| `ontology-build` \| `ontology-gql` \| `views` \| `ai-ml` \| `feedback` \| `langsmith-export` |
 | `sdk_ai_function` | set only on AI/ML invocations: `ai-generate` \| `ai-embed` \| `ai-classify` \| `ai-forecast` \| `ai-detect-anomalies` \| `ml-generate-text` \| `ml-generate-embedding` \| `ml-detect-anomalies` \| `ml-forecast` |
 
@@ -2291,8 +2291,9 @@ The optional LangSmith connector exports a BigQuery table or arbitrary SQL
 result as LangSmith run trees. It streams rows ordered by trace, reconstructs
 the existing `span_id` / `parent_span_id` hierarchy, and adds one stable
 structural root per trace. Every source event gets a UUID derived from the
-canonical source, trace ID, and run ID; overlapping backfills therefore upsert
-the same LangSmith runs instead of creating duplicates.
+canonical source, trace ID, and run ID. Each batch uses separate create and
+update requests, so overlapping backfills update the same LangSmith runs
+instead of creating duplicates or silently leaving stale data.
 
 ### Install and authenticate
 
@@ -2306,7 +2307,7 @@ export LANGSMITH_WORKSPACE_ID=...
 ```
 
 BigQuery uses Application Default Credentials, consistent with the rest of the
-SDK. The export query carries `sdk_surface=langsmith_export` and
+SDK. The export query carries `sdk_surface=langsmith-export` and
 `sdk_feature=langsmith-export` labels.
 
 ### Python API
@@ -2333,8 +2334,11 @@ print(stats.to_dict())
 `ExportStats.to_dict()` reports source rows read, exported, skipped, failed,
 successful traces, batches, the current watermark, bounded `dropped_rows`
 details, and `dropped_rows_truncated` (the number omitted from that list).
-Retryable 408/409/425/429 and 5xx errors use bounded exponential backoff. A
-failed batch never advances incremental state.
+Retryable 408/425/429 and 5xx errors use bounded exponential backoff. A create
+conflict is expected during replay, is not retried, and is followed by the
+update request. `requests_per_second` and `max_retries` govern exporter calls
+to `batch_ingest_runs`; the LangSmith SDK may split or retry transport requests
+inside one call. A failed batch never advances incremental state.
 
 ### Custom schemas and opaque values
 
@@ -2357,14 +2361,17 @@ fields:
   outputs: result
 ```
 
-Only declared paths are traversed, and only when the source value is already a
-mapping. JSON strings are not decoded or interpreted. A mapped scalar/list is
-wrapped as `{"value": ...}` because LangSmith inputs and outputs are objects.
-Unmapped columns are copied to `extra.metadata`. When a nested path is mapped,
-the original top-level column also remains in metadata so sibling data is not
-lost. Non-JSON BigQuery scalars use loss-minimizing transport encodings:
-decimal/date/time values become strings and bytes become a tagged base64
-object.
+Only declared paths are traversed, and omitted optional fields remain unmapped;
+custom mappings never inherit ADK defaults. Custom JSON strings are not decoded
+or interpreted. The built-in ADK mapping is the narrow exception: it decodes
+the standard `content` and `latency_ms` JSON columns because BigQuery clients
+may return those JSON values as either mappings or strings. A mapped
+scalar/list is wrapped as `{"value": ...}` because LangSmith inputs and outputs
+are objects. Unmapped columns are copied to `extra.metadata`. When a nested
+path is mapped, the original top-level column also remains in metadata so
+sibling data is not lost. Non-JSON BigQuery scalars use loss-minimizing
+transport encodings: decimal/date/time values become strings and bytes become
+a tagged base64 object.
 
 ```python
 from bigquery_agent_analytics.export import FieldMapping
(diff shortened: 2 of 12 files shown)


caohy1988 (commented):
## Review: full fresh pass over the diff

**Verdict: high-quality implementation — I'd fix 3 issues before merge and document 3 more.** The schema-agnostic requirement from #410 is honestly honored (opaque payloads, metadata passthrough, sibling-column preservation), the test suite is strong (cycles, 1100-deep traces, oversized-trace chunking, watermark binding, secret hygiene), and the `trace.py` refactor is a verbatim extraction. But the tests all run against permissive fakes, and the riskiest behaviors are exactly the ones fakes can't catch.

### Fix before merge

**1. One permanently-bad row stalls incremental sync forever** (watermark save condition in `export()`). The watermark only saves when `skipped == 0 and failed == 0` globally. A single row with NULL `span_id` or a duplicate `(trace_id, run_id)` — data problems that don't fix themselves — means every scheduled run re-reads the entire table from the beginning, exits 1, and never advances. The docs only promise "a failed *batch* never advances state"; freezing on any skipped row is much stronger and operationally dangerous. Per-trace success is already tracked (`_TraceProgress.successful`), so advancing the watermark past cleanly-exported traces while still reporting the skips would be safe and is what a scheduler needs.

**2. The synthetic root is re-`create`d in every chunk and every replay — upsert semantics are assumed, never verified.** A 250-row trace at `batch_size=100` sends the same root UUID in three `create=` batches; the advertised "replay an overlapping window" workflow re-creates every previously-ingested run ID. If LangSmith's endpoint 409s on duplicate creates rather than upserting, this fails — and worse, `_is_retryable` classes 409/`LangSmithConflictError` as retryable, so the exporter burns `max_retries` on a permanent conflict, then counts already-ingested rows as failed and refuses to advance the watermark. Every test uses a fake client that accepts anything. **This needs one integration run against a real LangSmith instance before merge** — the whole idempotency story rests on it.

**3. Label naming inconsistency is about to freeze into a reserved namespace**: `sdk_surface="langsmith_export"` (underscore) vs. the existing `remote-function` (hyphen) and its own paired `sdk_feature="langsmith-export"` (hyphen). SDK.md declares `sdk*` labels reserved and stable — after merge this can't be renamed without breaking dashboards. One-line fix now, permanent wart later.

### Document (or accept explicitly) before merge

**4. Cross-window trace splits silently flatten hierarchy.** When an incremental window contains a child whose parent was exported in a previous run, the child is reparented under the synthetic root (the code comment explains why: the parent's dotted-order prefix isn't reconstructible in-window). Consequence: with scheduled sync, *every live trace straddling a sync boundary* gets its later spans dangling off the root. That's a real correctness cost of the chosen design, and it's documented only in a code comment — SDK.md's incremental section should state it, and a "only export traces quiescent for N minutes" filter would largely avoid it.

**5. Event-time watermark silently loses late-arriving rows.** Streaming inserts landing with a timestamp below the saved watermark are excluded on every future run — no skip count, no diagnostic. Inherent to event-time watermarks, but SDK.md never warns, and an ingestion-time watermark option would fix it properly.

**6. Synthetic roots are dated 1970-01-01** (deliberately, for dotted-order stability). Real LangSmith UIs filter and sort trace lists by time — exported traces may be invisible in any "last N days" view and show 56-year durations. Worth verifying in the real UI alongside item 2.

### Verified bug, lower severity

**7. Default mapping loses `end_time` for JSON-string `latency_ms`.** `Span.from_bigquery_row` (trace.py:185–196) explicitly handles `latency_ms` stored as a JSON string — proof that real ADK tables have them — but the exporter's `_path_value` traverses only real mappings, so `("latency_ms", "total_ms")` yields nothing and such runs get no duration. A string-decode special case for the *default* mapping (matching the sibling reader) keeps the "never decode JSON" rule intact for user-supplied mappings.

**8. `_canonical_source` lowercases table refs**, but BigQuery dataset/table IDs are case-sensitive — two genuinely different tables can conflate into one watermark identity and UUID namespace.

### Minor code quality (take or leave)

- Dropped-row truncation is implemented twice (a cap inside `_prepare_trace_runs` plus `_DroppedRowCollector`) with fragile omitted-count arithmetic — one mechanism suffices since the trace is already fully materialized.
- `_MappedRow.status` is a write-only field.
- The `_PreparedTrace.source_run_ids` / re-derived-IDs consistency check guards against bugs only this module could introduce.
- The CLI's blanket `except Exception` catches `typer.BadParameter` from `_parse_datetime` and degrades it to a generic `Error:` line, and hides tracebacks for genuine bugs — narrowing the try to the `export()` call fixes both.

### What's notably good

Env-only API key with a test asserting `--langsmith-api-key` *doesn't* exist; SQL-injection-conscious table-ref validation with adversarial tests; atomic fsync'd watermark writes bound to source+mapping+destination fingerprints that fail closed; the error-callback adapter catching failures LangSmith's client would otherwise swallow (which would have silently advanced the watermark past lost data); iterative traversal immune to recursion limits; cycle-breaking that keeps every row exportable.

caohy1988:
## Review addendum: three more verified findings (one corrects item 2 above)

**A. Correction/refinement of item 2 — conflicts are silently *swallowed*, not failed.** I verified the langsmith client source: its batch-ingest paths pass `to_ignore=(LangSmithConflictError,)`, so a `create` for an existing run ID does not raise — it is silently ignored inside the client, and the `tracing_error_callback` never fires. Consequences:

- The multi-chunk root re-`create` and overlapping replays are *safe from duplication* (good), so the "no duplicate runs" half of the claim holds.
- But the connector is **create-only, not upsert**: re-exporting a window after source rows changed (backfilled `error_message`, a mapping now including `outputs`) updates nothing in LangSmith, while `flush()` counts every row as `exported`, exit code is 0, and the incremental watermark advances. Silent permanent staleness, contradicting the "batch upserts" wording in README/SDK.md §23.
- `_is_retryable` treating 409/`LangSmithConflictError` as retryable is therefore dead weight on the standard path, but still bites injected/wrapped clients that do surface conflicts (retry backoff burned on a permanent condition, then watermark blocked).

Suggested fix: change docs from "upsert" to "create-once; replays are no-ops", or implement true updates via `update=` for runs whose payload hash changed. Either way, drop 409 from the retryable set.

**B. `FieldMapping.from_dict` silently inherits ADK default paths for every omitted field — a schema-agnosticism hole.** `from_dict` fills omitted keys from `standard_adk()` (`raw = values.get(name, default)`). The README/SDK.md example mappings omit `error`, `status`, and `latency_ms` — so a custom table following the docs inherits `status=("status",)`, `error=("error_message",)`, `latency_ms=("latency_ms","total_ms")` bindings the user never declared. A custom table where `status` is a domain category happening to contain `"ERROR"` gets healthy runs stamped `error="[SOURCE_STATUS_ERROR]"`; an unrelated `error_message`/`content` column is silently consumed. The PR's own tests have to pass `"error": None, "latency_ms": None` explicitly to avoid this — the docs' examples don't. Suggested fix: when a `fields:` mapping is user-supplied, default omitted optional fields to `None` (only `run_id`/`trace_id` required); reserve the ADK defaults for the no-mapping case.

**C. `incremental=True` with `start_time: null` wedges after one successful run.** Run 1: no stored watermark, so the watermark SQL branch is skipped; `_map_row` defaults `start_time` to epoch; a watermark of 1970-01-01 is saved on success. Run 2: `_load_watermark` returns it and `_build_source_query` calls `_sql_path(mapping.start_time)` → `ValueError("mapping start_time is required for this filter")` → exit 2 on every subsequent run until the state file is deleted by hand. Nothing rejects this configuration up front. Suggested fix: fail fast in `export()` when `incremental=True` and `mapping.start_time is None`.

Also one nuance on item 7: whether `latency_ms`/`content` arrive as decoded dicts or raw JSON strings depends on both the column type (JSON vs STRING) and the installed google-cloud-bigquery version — `trace.py`'s own reader handles both shapes, which is the behavior the default mapping should match.

caohy1988:
## Review addendum 2: two last findings (final)

**D. STRING `start_time` columns pass Python-side validation but break every time-filtered query.** `_as_datetime` happily parses ISO-8601 strings, so mapping `start_time` to a STRING column works for a plain export — but `_build_source_query` binds `--since`/`--until`/watermark values as `TIMESTAMP` query parameters compared directly against the raw column, which BigQuery rejects (`No matching signature for operator >` for STRING vs TIMESTAMP). Worst case is incremental mode: the first run has no watermark predicate and succeeds (saving a watermark), then every resume fails with a query error — the same works-once-then-wedges shape as finding C. Suggested fix: wrap the mapped column in a SQL cast (`TIMESTAMP({col})`), or validate/document that time filtering requires a TIMESTAMP column.

**E. Latency guard `isinstance(latency, (int, float))` rejects `Decimal` and accepts `bool`.** A NUMERIC/BIGNUMERIC latency column arrives as `decimal.Decimal`, fails the check, and every run exports with no `end_time` (open-ended runs) despite valid data — while a BOOL value `True` passes (`bool` is `int`) and produces a nonsense 1 ms duration. Suggested fix: accept `Decimal` alongside `int`/`float` and exclude `bool`, mirroring `_required_string`'s scalar handling.

That completes the review: 8 findings in the main review, A–C in the first addendum, D–E here.

caohy1988:
[Replying to the main review](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/414#pullrequestreview-4903248691)

> I'd fix 3 issues before merge and document 3 more.

Addressed in `e055c9d`:

1. Skipped rows now contribute to a normalized incremental cursor and no longer wedge later runs; any surfaced LangSmith batch failure still blocks advancement. A stateful two-run regression covers null IDs and invalid timestamps.
2. Replays now issue separate create and update calls. Create conflicts are not retried, retry payloads are rebuilt because LangSmith mutates serialized dictionaries, and a real LangSmith 0.10.17 client test verifies distinct `post`/`patch` wire payloads plus mutation-safe retry.
3. `sdk_surface` is consistently `langsmith-export` in code and documentation.
4. SDK.md now documents cross-window parent flattening and recommends exporting quiescent traces.
5. SDK.md now documents event-time late-arrival loss and overlapping bounded backfills.
6. Synthetic roots retain an epoch segment only in `dotted_order`; displayed root times use current-window source times.
7. The built-in ADK mapping decodes JSON-string `content` and `latency_ms` while custom mappings remain opaque.
8. Table-source identity now preserves case.

The four minor cleanup notes were also addressed: one shared bounded dropped-row collector, no write-only mapped status field, no redundant prepared/source ID state, and explicit CLI operational-error handling without masking unexpected programming errors.

Validation: `3990 passed, 69 skipped`; isort, pyink (211 files), and `git diff --check` pass. No live LangSmith API call was made because `LANGSMITH_API_KEY` is not configured; the installed-client source and intercepted real serializer path are verified locally.

caohy1988:
[Replying to addendum 1](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/414#issuecomment-5249459321)

> The connector is create-only, not upsert.

Addressed in `e055c9d`: every logical batch now sends create and update in separate SDK calls, so LangSmith cannot coalesce them into a create-only operation. Replay tests prove changed error/input values patch the existing stable IDs. A surfaced create 409 is treated as the expected conflict, is attempted once, and proceeds to update; 409 is no longer retryable.

> `FieldMapping.from_dict` silently inherits ADK default paths for every omitted field.

Addressed: custom mappings now require only `run_id` and `trace_id`; every omitted optional field remains `None` and its source column remains metadata. ADK defaults are reserved for `FieldMapping.standard_adk()`.

> `incremental=True` with `start_time: null` wedges after one successful run.

Addressed: incremental export rejects a missing `start_time` mapping before client creation or query execution.

The built-in ADK mapping also matches `trace.py` for JSON-string `content` and `latency_ms`; custom mapping payloads remain opaque.

caohy1988:
[Replying to addendum 2](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/414#issuecomment-5249463301)

> STRING `start_time` columns pass Python-side validation but break every time-filtered query.

Addressed in `e055c9d`: `--since`, `--until`, watermark predicates, and cursor ordering use `SAFE_CAST(... AS TIMESTAMP)`. Cursor predicates additionally coalesce invalid timestamps to epoch and null IDs to empty strings so SQL ordering matches Python watermark normalization.

> The latency guard rejects `Decimal` and accepts `bool`.

Addressed: latency accepts `int`, `float`, and `Decimal`, explicitly excludes `bool`, and has regression coverage for both cases.

Targeted connector/CLI tests pass 57/57; the full repository suite passes `3990 passed, 69 skipped`.

caohy1988:
Final validation update for head 1d06025:

- Local isolated-environment suite: 3,990 passed, 69 skipped.
- GitHub Actions: Python 3.10, 3.11, 3.12, 3.13, and 3.14 all pass.
- Format check, package build, browser smoke, scan-pr, check-changes, and CLA all pass.
- The LangSmith 0.10.17 serializer path is covered locally with intercepted real-client tests for separate create/update requests and retry payload preservation.

No LANGSMITH_API_KEY was available, so no live project was mutated. The two live-only observations remain explicitly documented rather than claimed as externally verified.

caohy1988:
## Re-review of the fix commits (2191bfbd → 1d06025): all 13 findings verified addressed

I re-read the full delta. Finding-by-finding:

**Fixed in code (10):**
- **#1 watermark stall** — reworked as a *read cursor*: every row read advances it (malformed rows coalesce to epoch components so they can't regress it), API batch failures still block the save (`failed == 0`). The new two-run incremental test with imperfect data (`test_skipped_row_is_checkpointed_and_does_not_stall_incremental_sync`) covers exactly the gap I called out — consecutive syncs, not just one.
- **#2/A create-only & conflict handling** — create-then-update batches: conflicts on create are expected, classified by `_is_conflict`, not retried, and always followed by the update pass, so replays genuinely update. The `copy.deepcopy` per attempt (with the retry-rebuilds-mutated-payloads test) also protects against the client mutating payloads. 409/`LangSmithConflictError` removed from the retryable set, with regression asserts.
- **#3 labels** — `langsmith-export` everywhere; docs, docstring, and test updated.
- **#6 1970 roots** — the epoch survives only inside the `dotted_order` segment (where stability is required); displayed root `start_time`/`end_time` now span the window's real source times, with non-1970 asserts.
- **#7/nuance JSON-string columns** — solved cleanly with the private `_decode_standard_adk_json` flag: only the built-in ADK mapping decodes, custom mappings stay strictly opaque. Right boundary.
- **#8 case-folding** — table-ref case preserved; identity test added.
- **B mapping default inheritance** — `from_dict` no longer inherits ADK paths; only `run_id`/`trace_id` required; the domain-`status`-column scenario is now a test.
- **C incremental + null start_time** — fails fast before any query; test asserts no BQ/LangSmith calls happen.
- **D STRING time columns** — `SAFE_CAST` for since/until and `COALESCE(SAFE_CAST(...), epoch)` / `COALESCE(CAST(...), '')` for the cursor, with `ORDER BY` switched to the *same* expressions so ordering and cursor comparison can't diverge — and the Python-side `_row_watermark` mirrors the same coalescing. Consistent total order on both sides.
- **E latency types** — `Decimal` accepted, `bool` excluded, parametrized test.

**Documented as accepted trade-offs (2):** #4 cross-window flattening (with the quiescent-delay guidance) and #5 event-time late arrivals (with the bounded-backfill guidance) — both now in SDK.md §23 where operators will find them.

**Minor cleanups** — all taken: dead `status` field gone, redundant `source_run_ids` self-check gone, single shared `_DroppedRowCollector`, CLI catch narrowed to an operational-error tuple with tests proving `BadParameter` formatting survives and programming errors propagate with tracebacks.

The serializer-interception tests against the real langsmith client (asserting actual `post`/`patch` wire bodies) are a smart substitute for a live project. The one item that still needs a real `LANGSMITH_API_KEY` before this ships to users: end-to-end confirmation that the server accepts the create→conflict→update flow and that root/trace times render sensibly in the UI. Nothing else outstanding from my review.

caohy1988:
## Live check results (real LangSmith, api.smith.langchain.com, langsmith 0.10.17)

Ran the PR head (1d06025) end-to-end against a real LangSmith project (`bqaa-live-check-20260811-063522`), using the injected `bq_client` seam with synthetic ADK rows so only LangSmith was touched. **8/10 scenarios pass; the replay-update design is refuted by an explicit server policy.**

### Verified working against the real server ✅

- Initial export ingests the full run tree; stable UUIDs readable via the API.
- **Root-time fix confirmed in practice**: root `start_time` is the window time (`2026-08-11T06:25:22Z`, not 1970), and root runs are discoverable in a recent time-window `list_runs` query — the UI-visibility concern from finding #6 is resolved.
- **Multi-chunk trace confirmed**: 5 rows at `batch_size=2` → 3 batches, root re-sent in each, `exported=5 failed=0`. Duplicate creates are accepted/ignored by the server.
- Replaying an overlapping window creates **no duplicates** (exactly one root per trace after replay).

### Refuted by the server ❌

Replaying with changed source data (backfilled `error_message`, changed `content`) reports `exported=2 failed=0` — but the server-side run still has `error=None` and the old inputs. A direct `update_run` probe on that run returned the definitive answer:

> `409 Conflict — {"error":"Run update payload already received. Duplicate run update requests for the same run are not supported."}`

**The server permits exactly one create + one update per run, ever.** The exporter's create→update flow consumes that single update slot during the *first* export, so replays can never change anything: their patch batches are conflict-swallowed by the client (`to_ignore=(LangSmithConflictError,)`), stats report success, and the changed data silently never lands. The revised README/SDK.md claim — "replaying an overlapping window updates existing runs without creating duplicates" — is not true against today's server.

### Recommended follow-up

1. **Drop the unconditional update pass** (create-only): the create payload already carries `end_time`/`outputs`/`error`, so the update adds nothing on first export, doubles API calls, and burns the one allowed update. Keep `_is_conflict` create-tolerance for replays.
2. **Re-word the docs honestly**: replays are idempotent no-ops — exported runs are immutable server-side. To propagate corrected source data, export to a fresh project or mint new IDs deliberately via `--source-id` versioning.
3. Optionally log a notice when a replay is detected (create conflicts observed) so operators know changed rows were not propagated.

Everything else from the 13-finding review held up under live conditions.

caohy1988:
> drop the update pass entirely (create-only — the create payload already carries everything, so the update pass adds nothing except doubled API calls and a consumed update slot), re-word the docs to "exported runs are immutable; replays are idempotent no-ops"

Addressed differently in 8f98c64: the exporter now sends one create-only request per batch and retains surfaced create-conflict tolerance. Existing stable run IDs replay as idempotent no-ops, while previously unseen IDs in an overlapping window are still created. Correcting an existing ID now explicitly requires a fresh LangSmith project or a deliberately versioned `--source-id`.

The docs also clarify that `exported` counts source rows in client-accepted batches, not destination mutations, so duplicate replays can report exported rows without changing server state. Replay detection logging was not added because conflicts swallowed inside the LangSmith SDK are not reliably observable at this boundary.

Regression coverage now verifies create-only batching, immutable replay, surfaced 409 behavior, one real-client POST body, full create-payload reconstruction on retry, rate limiting, and partial-batch failure accounting. Local validation: 3,990 passed, 69 skipped; independent review found no Critical or Important issues. GitHub Actions is green on the pushed head: Python 3.10–3.14, package build, formatting, browser smoke, scans, workflow checks, and CLA all passed.

caohy1988:
## Fresh review + second live check of 8f98c64: 10/10 pass — review closed from my side

**Diff review (1d06025 → 8f98c64):** clean and exactly scoped to the live-check feedback. Update pass removed (create-only; `_is_conflict` → idempotent success), docs reworded to the immutable contract with the fresh-project / versioned `--source-id` correction path, and `exported` explicitly documented as *accepted source rows, not server mutations*. Tests reworked consistently — including a client that asserts no update batch is ever sent, the conflict-as-success path, and create-payload retry isolation. No new issues found.

**Live check #2 (real LangSmith, project `bqaa-live-check2-20260811-142753`): 10/10 scenarios pass.**

- Initial export, multi-chunk create-only batches (root re-sent per batch), and time-window discoverability all clean against the real server.
- **Immutability contract confirmed**: replaying a window with changed source data reports accepted rows and leaves server-side runs untouched — now matching the docs instead of contradicting them.
- **Partial-replay creation confirmed**: a replay containing one previously unseen row creates that run while existing runs stay untouched.
- **The documented correction path works live**: exporting the same source (same stable IDs) to a fresh project creates fresh runs — run IDs turn out to be project-scoped server-side, so the "use a fresh LangSmith project to correct data" guidance is verified, not just plausible.

One observability footnote (no action needed): `read_run(id)` resolved to the older project's copy for a short window after cross-project re-export — read-after-write lag worth knowing about when scripting verifications, not a connector defect.

All 13 original findings plus the live-check follow-up are now fixed or explicitly documented, and verified against the real service. From the reviewing side this is done: remaining steps are updating the branch (currently behind base) and a maintainer's approval.

haiyuan-eng-google (approved):

Issue #402, opened 2026-08-07
[Dashboard] Add Conversational Analytics (natural-language Q&A) over agent_events


haiyuan-eng-google:
Part of #404.

## Request

Internal field feedback from a Google Cloud customer engineer evaluating the
dashboard end to end:

> Can conversational analytics be added to this page too? For example, here's a
> demo video I use for a Looker equivalent to this dashboard, which is very
> compelling (presented at Google I/O earlier this year).

The referenced recording is internal and not publicly linkable. The capability
it demonstrates is the natural-language Q&A pane in Looker, used alongside the
[Looker Agent Analytics block](https://marketplace.looker.com/marketplace/detail/agent_analytics)
— ask a question in plain language, get a chart and an answer over the same
agent telemetry.

This is the highest-ceiling item in the current feedback batch. The 37 charts
answer the questions we anticipated; natural-language Q&A answers the rest.

## The constraint

Looker Studio has no in-report conversational pane equivalent to Looker's. So
this cannot be a straight port of the demo, and it needs a product decision
before any implementation.

## Options

**A. Conversational Analytics API in a community visualization.** Embed a chat
tile in the report backed by the
[Conversational Analytics API](https://cloud.google.com/gemini/docs/conversational-analytics-api/overview)
over the same `agent_events` table.
*For:* highest fidelity to the demo; the answer lives next to the charts.
*Against:* a community visualization has to be hosted and reviewed;
authenticating from a sandboxed Looker Studio viz to the CA API is the hard,
unproven part; it adds a long-lived component to maintain.

**B. A companion Conversational Analytics data agent, linked from the report.**
Ship a way to provision a CA data agent over the user's `agent_events` table —
pre-loaded with the BQAA schema and a curated question set — and link to it
prominently from the report and the configurator page.
*For:* works for every user of this repo, no hosting, no embedding problem; the
schema description and golden questions are reusable assets.
*Against:* not in-report; it's a second surface.

**C. Document the Looker path.** For teams that already have Looker, point at
the Agent Analytics block plus Gemini in Looker — literally what the demo shows.
*For:* zero build. *Against:* only helps existing Looker customers, which is the
gap this dashboard exists to fill.

**Recommendation: B first**, then re-evaluate A once the CA API's embedding
story matures. C is a docs paragraph worth adding regardless.

## Proposed scope for B

- [ ] A data-agent definition: system instruction plus a description of the
[...]

Issue #382, opened 2026-07-27
Looker Studio dashboard: default date range hides all data for new installations (real-agent E2E findings)


caohy1988:
Findings from an end-to-end test of the dashboard against a **real ADK agent trace**, rather than the synthetic seed. Setup: a fresh dataset provisioned by the real ADK **2.4.0** plugin (16 generated views, including the new `v_tool_paused`), then a live `gemini-2.5-flash` agent on Vertex ran 6 turns with 7 tool calls through `BigQueryAgentAnalyticsPlugin`, producing 77 genuine events across 10 event types in `agent_events`.

Related: #377, #379 (the dashboard itself), #365.

## P1 — Fresh installations render a completely empty dashboard

The published report's default range is **rolling 365 days ending yesterday** (`product_contract.yaml#defaults.date_range`: `end_offset_days: 1, include_today: false`). A new installation's events are all from **today**, so the first-run experience of "trace your agent → hydrate the dashboard → open it" shows *No data* on every page.

Verified with the exact production template SQL against the real dataset:

| Window | Rows visible |
| --- | --- |
| Report default (365d ending yesterday) | **0** |
| Same window including today | **77** |

Everything else works — preflight passes in base-table mode, and all 37 oracle queries execute against the real data (37/37) — so the empty dashboard is purely the window default.

**Proposed fix**
- [ ] Change the published report's default range to include today (e.g. "Last 365 days (include today)", or a shorter include-today onboarding default).
- [ ] Update `spec/product_contract.yaml#defaults.date_range` and the pinning test to the new default.
- [ ] Re-publish and re-date `bindings/report_template.yaml#live_template_verification`.
- [ ] Optional: an empty-state hint in onboarding docs ("no events in the selected range — new installation? include today").

## Secondary findings

- [ ] **New event type `AGENT_RESPONSE`** — ADK 2.4.0 emits event types outside the 1.27.0 static intersection (6 of the 77 rows). The base-table design correctly includes them, but all-events counts now cover lifecycle types the historical views never counted; worth one line in the compatibility docs.
- [ ] **Thinking tokens break prompt+completion=total intuition** — real responses carry `usage_metadata.thoughts_token_count` (e.g. total 113 = prompt 62 + completion 5 + thoughts 46). Total Tokens correctly uses `total_token_count`, but users comparing tiles will find prompt+completion ≠ total. Docs note now; a thinking-tokens field in a future schema rev.
[...]

PR #416, opened 2026-08-11
docs: record #402 conversational analytics evaluation

caohy1988:
## What changed

Fills the required issue #402 evaluation record with authenticated evidence for the first-party BigQuery data-agent route and the pinned stateless prototype. It records all six scored questions, the nine-row criteria matrix, the B1 recommendation, and the explicit implementation/closure boundary.

## Why

The previous document correctly blocked an option pick until both run metadata and every mandatory criteria cell were filled. This PR supplies that evidence without claiming the user-facing capability has shipped.

## Evaluation outcome

- Both stage-1 shapes were 6/6 correct with sane, time-bounded SQL over the same curated fixture view.
- The authenticated Data Studio flow was exercised through agent discovery, consent, first question, SQL/result display, and persisted Recent conversation.
- B1 is recommended because it meets the mandatory criteria without a custom service/UI.
- No first-party mandatory failure was found, so the custom persistent-agent stage is not triggered.
- #402 remains open until capability ships or a maintainer accepts a named implementation handoff.

Publication safety: the record names only the sanctioned test project/dataset and aggregate results. The frozen snapshot and allowlist view expire after 30 days; no raw prompts, production identifiers, or event content are included.

## Validation

- `pytest -q tests/test_looker_studio_dashboard.py` — 25 passed
- pinned prototype `tests/ca.test.mjs` — 24 passed
- exact pinned stateless runtime probe — correct GQ-03 result with `scope.verified: true`
- `git diff --check`

Closes no issue; contributes the decision evidence for #402.

### DIFF
--- a/dashboard/looker_studio/docs/conversational-analytics-decision.md
+++ b/dashboard/looker_studio/docs/conversational-analytics-decision.md
@@ -1,9 +1,9 @@
 # Conversational Analytics over `agent_events` — decision document
 
-Status: **awaiting product decision** (issue #402; plan #404 rev 5.1).
-Nothing in this document is implemented; it defines the options, the
-criteria a selection must satisfy, and the evaluation that must happen
-before selecting.
+Status: **evaluation complete; first-party route recommended, awaiting
+maintainer acceptance and an implementation owner** (issue #402; plan
+#404). The evaluation below chooses the shape; it does not claim that the
+report link, provisioning asset, or user-facing capability has shipped.
 
 ## Request and constraint
 
@@ -84,7 +84,7 @@ failure, shape 2 is out and the pick is between shapes 1 and 3. The
 recommendation must rest on observed answers and filled criteria rows,
 not description.
 
-## Evaluation record — REQUIRED before the option pick (currently empty)
+## Evaluation record — completed 2026-08-11 UTC
 
 The selection cannot happen until **both** parts are filled: the run
 metadata *and* the per-shape criteria matrix. A record with filled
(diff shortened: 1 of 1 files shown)


caohy1988:
## Review — evaluation record independently reproduced; approve

This is the rare evaluation record where every number could be re-run — so I re-ran them against the live fixture (`test-project-0728-467323.bqaa_e2e_real.ca_eval_agent_events_20260810`):

| ID | Documented | Recomputed | |
| --- | --- | --- | --- |
| GQ-01 | 77 event rows | 77 | ✓ |
| GQ-02 | 7 distinct sessions | 7 | ✓ |
| GQ-03 | 1,703 prompt+completion | 1,703 | ✓ |
| GQ-04 | p95 2,562 ms | 2,562 | ✓ |
| GQ-05 | 7 runs, 0 failures | 7 TOOL_STARTING/TOOL_COMPLETED, zero TOOL_ERROR rows | ✓ |
| GQ-06 | LLM_REQUEST/LLM_RESPONSE 12 each, tied | 12/12, tied for highest volume | ✓ |

Also verified directly: the snapshot and the `ca_eval_agent_events_allowlist_20260810` view exist with the documented names and expirations; the view's SQL hashes session IDs (`TO_HEX(SHA256(...))`), exposes the separated `prompt/candidates/thoughts` token fields (the #382 sharp edge), and carries no raw content, user IDs, or error text — the allowlist contract is real, not described. The pinned prototype SHA `62f794fd` exists on the prototype repo.

The record satisfies the contract it had to: run metadata **and** the nine-cell matrix are filled, the stage-2 trigger is explicitly "none," and the recommendation (B1) follows from observed results rather than description. The status line correctly avoids claiming the capability shipped.

Two small notes, neither blocking: (1) the latencies are one-run samples — fine as directional evidence, and the doc says so; (2) the fixture resources expire in 30 days, so future re-verification needs a re-seed — the row counts in this record are the reference values.

**LGTM.** With this merged, #402 moves to the option pick per its closure contract.

caohy1988:
Fresh final verification on commit `6ddeb9bab85ad6ad27eecd263f561f45b557863e`: the complete repository suite passes in an isolated project environment — **3,936 passed, 69 skipped**. The initially activated global interpreter was contaminated by an adjacent editable `adk-python` checkout with incompatible OpenTelemetry semantic conventions; reinstalling this repository’s declared `.[dev]` dependencies in a clean venv removed that environment-only collection failure. `git status` is clean, and all PR checks remain green across Python 3.10–3.14, format, browser smoke, scan, package build, and CLA.

caohy1988 (commented):
## Fresh full review of the evaluation record — satisfies the gate; one structural fix before undraft

Reviewed the complete diff against the framework's own rules (the two-stage gate and the filled-metadata-with-empty-matrix prohibition this document defines).

### Verified

- **Zero `_pending_` slots remain** (17 removed), and the staleness trap was avoided: the heading now reads "completed 2026-08-11 UTC" and the status line no longer says "awaiting product decision" — both updated, not just appended to.
- **Run metadata is complete and pinned:** immutable prototype SHA (`62f794fd`), Data Studio app version + date, fixture id/row-count/date-range with the curated-view field policy, the six-question subset, the B1 recommendation with its reasoning, **stage-2 trigger recorded as "none" with the reasoning that filter re-entry is a flow constraint, not a mandatory-criterion failure** — exactly the distinction the two-stage gate turns on — and evaluator+date.
- **All nine matrix rows filled for both stage-1 shapes** with observed behavior plus primary-doc citations, and the honest caveats are the record's strongest feature: the stateless shape's #382 sharp-edges gap is called a gap even though GQ-03 scored correct ("that does not satisfy the context contract"); `global` residency is flagged as having no commitment; the allowlist is framed as an IAM provisioning contract with agent config as defense-in-depth, matching the known-limitations doc.
- **Publication safety observed:** sanctioned test identity only, aggregate results, expiring resources, no raw content.
- Accepted deviation, noting for the record: the "Per-question results" metadata bullet became a full table with a "SQL sane" definition and a latency disclaimer — strictly better than the slot it replaces.

### One structural fix requested (small, before undraft)

The matrix rule sentence — *"The stage-2 column stays '—' unless the named-failure trigger fires; it must then be filled completely before shape 2 can be recommended."* — now dangles **after** the new "Recommendation and implementation boundary" section, orphaned from its table by the insertion. Move it up to sit directly under the matrix, before the new H2. One-line relocation; as placed, it reads as a caveat on the recommendation rather than a rule of the table.

### Two non-blocking suggestions

1. **Latency expectation belongs in handoff item 3.** Observed answer latency was 14.4–33.3 s per question (UI lifecycle ~15 s). The link-out UX should set that expectation the same way the configurator's wait note sets the provisioning expectation — add it to the implementation handoff's link-UX item so it isn't rediscovered as a complaint later.
2. **Record the evaluation DataAgent's disposition in the doc.** The snapshot/view expiry is documented, but the DataAgent persists for reviewer reproducibility and must be deleted after merge + pick — and must **not** be promoted into the production asset. One line here makes the durable artifact self-contained; the #404 teardown checklist already tracks the action.

Nit, take or leave: the intro under the completed record still uses present-tense gate language ("The selection cannot happen until both parts are filled") — fine as the standing rule, reads slightly odd below a completed heading.

With the relocation done, this is ready to undraft from my side. Verification context: current `main` (`9734124`) is healthy locally — dashboard pytest 25/25, Node suite green, browser smoke positive + all five negative fixtures detected.

caohy1988:
Confirmed the structural nit at `6ddeb9ba`: the sentence "The stage-2 column stays '—' unless the named-failure trigger fires; it must then be filled completely before shape 2 can be recommended" currently sits at the end of the "Recommendation and implementation boundary" section (after the four-item handoff list), ~30 lines below the matrix it governs. It belongs directly under the criteria matrix table, ahead of the recommendation section — as written it reads as an afterthought and a future editor could miss that it constrains the stage-2 column. One-line move; everything else in the record still verifies from my earlier reproduction pass.

caohy1988:
> Move [the stage-2 matrix rule] up to sit directly under the matrix, before the new H2.

Fixed in `c668e0c`. The stage-2 rule now sits immediately below the criteria matrix.

I also incorporated the non-blocking durable-record improvements:

- handoff item 3 now records the observed 14.4–33.3-second answer range as directional UX expectation-setting, not an SLO;
- the evaluation DataAgent is explicitly delete-after-merge-and-pick and must not be promoted;
- the completed record's intro now states the standing validity rule instead of describing the selection as future work.

Fresh validation on the combined diff: dashboard contract tests 25/25; full repository suite 3,936 passed and 69 skipped; `git diff --check` clean.

caohy1988:
Re-review at `c668e0c2` — all three items from the last round are addressed:

- The stage-2 rule sentence now sits directly under the criteria matrix, ahead of the Recommendation section — the structural fix.
- Handoff item 3 now carries the latency expectation (14.4–33.3s, directional, not an SLO).
- The evaluation DataAgent's disposition is recorded in the fixture metadata (delete after merge + option pick; no promotion to production).

Suites green locally (25/25). LGTM stands; ready for undraft + maintainer review.

caohy1988 (commented):
Final verdict at `c668e0c2` — **all four review items verified in the delta; nothing further from the review side. Ready for maintainer approval.**

1. **Stage-2 rule relocated** — now sits directly under the matrix and is removed from below the handoff list; it reads as the table's rule again, not a caveat on the recommendation.
2. **Latency expectation added** — handoff item 3 carries the observed 14.4–33.3 s with exactly the right framing ("directional, not a performance SLO"), mirroring the record's own disclaimer.
3. **DataAgent disposition recorded in the durable artifact** — named (`bqaa-issue-402-eval-20260810`), delete after merge + option pick, never promote. Matches the #404 teardown gate.
4. **Gate wording made timeless** — "A valid selection requires both completed parts" reads correctly under a completed heading while preserving the rule.

The document is doc-only, internally consistent end to end, and `main` at `9734124` remains verified healthy (25/25 focused, browser gate green). Merge-state is up-to-date awaiting approval.

haiyuan-eng-google (approved):

PR #423, opened 2026-08-12
docs: correct LangSmith re-export guidance and document the 24h ingest window

haiyuan-eng-google:
Two corrections to the LangSmith export docs, both found while exporting a real
`agent_events` table to LangSmith Cloud. Each one produces a silent or
confusing failure today.

## 1. A fresh LangSmith project does not re-mint run IDs

The docs offer two ways to correct already-exported data — a fresh LangSmith
project, or a versioned `--source-id`. Only the second works.

`_stable_uuid(source_fingerprint, ...)` hashes the canonical source identity
with the trace and run IDs. The destination is not an input: `project_name`
reaches `_prepare_trace_runs` but is used only as `session_name` on the emitted
payload. Exporting an unchanged `--source-id` into a new project therefore
replays run IDs that already exist, and since created runs are immutable the
batch is a no-op.

What makes this worth a doc change is how quiet it is — the exporter reports a
clean success while the destination stays empty:

```console
$ bq-agent-sdk export langsmith --source-id=demo-v1 \
    --langsmith-project=brand-new-project ...
{"exported": 21, "failed": 0, "skipped": 18, "traces_exported": 6}
# brand-new-project holds 0 runs

$ bq-agent-sdk export langsmith --source-id=demo-v2 \
    --langsmith-project=brand-new-project ...
# same project, only --source-id changed -> 27 runs
```

Corrected in all three places the claim appears (README.md, SDK.md §23 intro,
SDK.md incremental section).

## 2. LangSmith Cloud rejects runs older than 24 hours

`POST /runs/batch` refuses any run whose `start_time` is more than 24 hours
from now, and the check is per batch, so an older window fails every row:

```text
422 Unprocessable entity: error parsing run: start_time for post must be
within ±24 hours of current time (got 2026-08-06 21:54:54 +0000 UTC,
allowed range: 2026-08-11 17:30:26 ... 2026-08-13 17:30:26)

{"exported": 0, "failed": 21, "rows_read": 39, "skipped": 18}
```

Nothing surfaces this before the batch call. The bounded-backfill example
spans a week ending well before the reader runs it, so followed verbatim
against LangSmith Cloud it exports nothing. The practical consequence is worth
stating next to that example: `--since` / `--until` can re-run a *recent*
window, but aged trace history cannot be imported at all, and `--incremental`
has to run often enough that rows arrive while still eligible.

Qualified as LangSmith Cloud behaviour, since self-hosted deployments may
configure it differently.

## Testing

Docs only — no code changes. Both behaviours were reproduced against
`api.smith.langchain.com` with `bigquery-agent-analytics==0.5.0` exporting a
39-row / 6-trace `agent_events` table.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,12 @@ For scheduled syncs, use `--incremental` with `--watermark-file=state.json`.
 The JSON summary bounds row-level diagnostics with `--max-dropped-rows` and
 reports the number omitted as `dropped_rows_truncated`.
 
+LangSmith Cloud rejects runs whose `start_time` is more than 24 hours from now,
+so an export covers recent traces rather than aged trace history. Run
+`--incremental` on a schedule frequent enough to stay inside that window. See
+[SDK.md](SDK.md#23-langsmith-export) for the exact error and its effect on
+bounded backfills.
+
 Custom schemas use a YAML mapping from LangSmith fields to source column or
 nested paths. Unmapped columns remain in `extra.metadata`; payload values are
 opaque and are never classified by event type. Optional fields omitted from a
--- a/README.md
+++ b/README.md
@@ -122,11 +122,12 @@ The CLI intentionally accepts the API key only through
 The exporter reconstructs span parents and derives stable LangSmith UUIDs. It
 treats created runs as immutable: replaying an overlapping window is an
 idempotent no-op for existing run IDs while still creating previously unseen
-IDs. To correct already-exported data, use a fresh LangSmith project or a
-deliberately versioned `--source-id`. For scheduled syncs, use `--incremental`
-with `--watermark-file=state.json`. The JSON summary bounds row-level
-diagnostics with `--max-dropped-rows` and reports the number omitted as
-`dropped_rows_truncated`.
+IDs. Run IDs derive from the source identity and not the destination, so
+exporting to a fresh LangSmith project reuses the same IDs and creates nothing.
+To correct already-exported data, use a deliberately versioned `--source-id`.
+For scheduled syncs, use `--incremental` with `--watermark-file=state.json`.
+The JSON summary bounds row-level diagnostics with `--max-dropped-rows` and
+reports the number omitted as `dropped_rows_truncated`.
 
 Custom schemas use a YAML mapping from LangSmith fields to source column or
 nested paths. Unmapped columns remain in `extra.metadata`; payload values are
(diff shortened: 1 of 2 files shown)


caohy1988:
## Independent verification — claim 1 confirmed from code, doc targets located

Traced the ID derivation on `main` (`src/bigquery_agent_analytics/export/langsmith.py`):

- `_stable_uuid(source_fingerprint, ...)` (line 507) hashes only the canonical source identity plus trace/run IDs — the destination project is not an input.
- `project_name` reaches `_prepare_trace_runs` and is used **only** as `session_name` on emitted payloads (lines 592, 632). Confirmed: unchanged `--source-id` + fresh project replays identical run IDs, and with created runs immutable the batch is a silent no-op. The issue's diagnosis is exact.

The stale guidance exists in exactly the three places named: `README.md:125-126`, `SDK.md:2296-2297`, `SDK.md:2444` — all three offer "a fresh LangSmith project" as a correction path.

The 24-hour ingest window (claim 2) is LangSmith Cloud behavior I can't re-verify from the repo, but the quoted 422 is specific and consistent with the consequence drawn: the bounded-backfill example needs the caveat, and `--incremental` cadence becomes an eligibility requirement, not just an efficiency choice. Qualifying it as Cloud-specific (self-hosted may differ) is the right framing.

No code change needed for claim 1 — it's doc-only, and the fix direction (drop the fresh-project path, keep versioned `--source-id`) follows directly from the UUID derivation above.

caohy1988:
## Diff review at `912227f` — clean, merge-ready

Verified the diff against my independent code trace (posted above):

- **All three stale locations corrected, no leftovers.** Repo-wide grep for the "fresh project" guidance at this head returns matches only in the new corrective sentences; nothing stale remains outside the three named spots.
- **The corrected wording is exact.** "Run IDs derive from the source identity and not the destination" matches `_stable_uuid(source_fingerprint, …)` and the `session_name`-only use of `project_name` (`langsmith.py:507,592,632`).
- **The 24-hour section is placed where it bites** — immediately after the bounded-backfill example it silently breaks — states the per-batch rejection semantics and the `exported: 0` / non-zero `failed` signature, and is properly qualified as LangSmith Cloud behavior with the self-hosted caveat.
- **README's summary + pointer** to SDK.md §23 avoids duplicating the detailed explanation.

Docs-only diff; all CI checks green. LGTM.

caohy1988 (approved):

Issue #397, opened 2026-08-05
skill_evolution host hooks: contract hardening follow-ups (from PR #395 review)


evekhm:
Residual risks the PR #395 review marked informational — filed so they are tracked rather than growing that diff. All in `scripts/skill_evolution.py`.

1. **Non-string host patches crash after full fleet spend.** A host `error_analyst_fn` returning a truthy non-string (dict, response object) crashes `collect_patches` at the quality gate (`patch.strip()` → `AttributeError`) *outside* the per-future exception handler — after every analyst has already run. Suggested fix: type-check each future result at collection (warn + drop non-strings), defending the documented "patch text or None" contract.
2. **No timeout on analyst futures.** A blocking host `error_analyst_fn` hangs `collect_patches` indefinitely (`as_completed` has no timeout; the built-in path only implicitly relies on client network timeouts). Suggested fix: an optional `analyst_timeout_s` applied per future, timed-out analysts degraded to warnings like other host failures.
3. **Patch provenance is not auditable.** Run artifacts never record whether a patch came from the host analyst or the built-in prompt — `*_patches.json`/`selection.txt` can't show `error_analyst_fn` was even configured. Suggested fix: a `source` field per patch record (`builtin` | `host`), written by `_write_evolution_artifacts`.
4. **Correction-as-hypothesis counterweight is builtin-only.** The prompt rule added in #395 protects the built-in analyst; a host analyst receives raw sessions where `correction_boundaries` labels user assertions as `correct_fact` with no such rule. Suggested fix: document the hazard in the `error_analyst_fn` docstring (host analysts should treat user corrections as hypotheses) — the engine cannot enforce it.
5. **No stability signal on the growing host contract.** `evolve_skill` is now 17 keyword params over an implicit session-dict schema in a `scripts/` module with no `__all__`/version marker, while positioned as an adopt-don't-fork engine. Suggested fix: a short "host contract" section in the module docstring + `__all__`, and a CHANGELOG note when the session schema changes.

Refs: PR #395 (host integration hooks), the 8-lens review comment there, and #384/#361 for the surrounding U6 work.

PR #431, opened 2026-08-17
feat(eval): canonical evaluation rubrics in core + rubric reference doc (closes #430)

### LINKED ISSUE #430: Canonical evaluation rubrics in core — extracted from the current quality_report metrics (supersedes the #91 approach)
**Context:** #91 (Phase 1 of #63) proposed core rubric templates from an April snapshot: 2-category `response_usefulness`, no `declined` handling, no scope coupling — behind the shipping semantics and a name collision with them if merged (CHANGES_REQUESTED with conflicts since May).

**What ships today** (`scripts/eval/eval_config.json` + `get_eval_metrics()`): 8 metrics — `response_usefulness` (meaningful/unhelpful/partial + scope-conditional `declined` injection), `task_grounding`, `correctness`, `tool_usage`, `specificity`, `scope_compliance`, `first_time_right`, `failure_attribution` — with `scope_aware` enrichment and a JSON override contract. The U6 recording, the promotion gate, and the published series ride on these.

**Proposal:** core `evaluation_rubrics.py` with `builtin_metric_config()` + `build_metrics(eval_config, *, scope_context, has_scope)`, a drift-guard test pinning the builtin to the shipped JSON, a rubric reference at `docs/evaluation_rubrics.md`, and the script as a thin caller — extraction-from-current rather than resurrection-from-April. #91's net-new `policy_compliance` idea can layer on top afterwards.

Refs: #63, #91, #397. PR attached.

evekhm:
Extracts the canonical rubric data and interpreter from `scripts/quality_report.py` into `bigquery_agent_analytics.evaluation_rubrics`, per #430 — superseding the #91 Phase-1 approach by extraction-from-current rather than resurrection-from-April.

**What's in core now:**
- `builtin_metric_config()` — deep copy of the canonical 8-metric config (`response_usefulness` with the scope-conditional `declined` category, `task_grounding`, `correctness`, `tool_usage`, `specificity`, `scope_compliance`, `first_time_right`, `failure_attribution`), same schema as an `--eval-config` JSON.
- `build_metrics(eval_config=None, *, scope_context='', has_scope=False)` — the interpreter, verbatim semantics from the shipping script: scope-aware definition enrichment, `declined` injection with `insert_after` positioning, scope suffixes.
- **`docs/evaluation_rubrics.md`** — the rubric reference: every metric's judge definition and category vocabulary (generated from the canonical data, so it cannot drift from what the judge actually sees), the scope/declined mechanics, golden-grounding interplay, Python + CLI usage, and the override schema.

**Same-PR thinning (no dual implementations):** `get_eval_metrics()` now derives scope inputs from the eval spec and delegates to core; a missing `eval_config.json` falls back to the SDK builtin instead of raising. A drift-guard test pins `builtin_metric_config()` to `scripts/eval/eval_config.json`.

**Tests:** 6 new in `tests/test_evaluation_rubrics.py` (canonical set, declined injection + positioning, scope-aware appending, deep-copy isolation, custom-config passthrough, builtin↔file parity); full quality-report/skill-evolution/compare-runs suites green (184).

Relation to #91: same destination file name and intent, but built from the current, recorded-and-published vocabulary; `policy_compliance` from #91 can layer on top afterwards. Refs #63, #397.


caohy1988:
## Code review: canonical evaluation rubrics in core

**Verdict: approve with small changes.** The extraction is faithful — I diffed the moved interpreter against the original `get_eval_metrics` logic and the semantics are verbatim, including the pre-existing quirk that a typo'd `insert_after` silently inserts at position 0 (`idx = -1` → `insert(0, ...)`; consider a warning, but it's not a regression). Imports match the SDK's absolute-import house style, the `__init__.py` try/except block matches its neighbors, `list[...]` annotations are fine for the `>=3.10` floor, and the deep-copy and drift-guard tests are the right tests. CI is green across Python 3.10–3.14.

Findings:

1. **The doc's no-drift claim isn't actually enforced (medium-low).** The PR body says `docs/evaluation_rubrics.md` is "generated from the canonical data, so it cannot drift" — but no generator script is committed and no test pins the doc to `_BUILTIN_METRIC_CONFIG`. The drift guard only pins builtin ↔ `scripts/eval/eval_config.json`. The first rubric edit will silently strand the doc. Either commit the generator, add a doc-parity test (e.g. assert each metric's definition/category text appears in the doc), or soften the claim.

2. **Fallback can shadow a clearer error (low).** The new auto-discovery fallback in `_load_eval_config` does `from bigquery_agent_analytics import builtin_metric_config` — against an older installed SDK that lacks `evaluation_rubrics`, the previously actionable `FileNotFoundError` (which told the user where to put the file) becomes a bare `ImportError: cannot import name 'builtin_metric_config'`. Wrapping the import to re-raise with the old guidance costs three lines. (Verified: an explicit `--eval-config <path>` still raises `FileNotFoundError` — the fallback only applies to auto-discovery, which is the right scoping.)

3. **Consider `logger.warning` instead of `info` for the fallback.** A user whose custom `eval/eval_config.json` is misplaced now silently gets the builtin rubrics where they previously got a hard stop; an info line is easy to miss in a full run's output.

**Cross-PR note:** this PR and #432 both insert their export block at the same anchor in `__init__.py` (right before "# Categorical Views") and both edit `scripts/quality_report.py`. The `quality_report.py` hunks don't overlap so git should auto-merge those, but the `__init__.py` insertion is a guaranteed conflict for whichever merges second — plan a rebase.

caohy1988 (commented):
## Review: Approve with nits

Verified locally: the extraction is **token-for-token verbatim** from the old interpreter (only the log message wording changed: `"Loaded %d metrics"` → `"Built %d metric definitions"`). The drift guard is real and currently passes — confirmed `builtin_metric_config() == scripts/eval/eval_config.json` as parsed dicts. Reproduced the claimed test count exactly: 6 new + 131 quality-report + 47 skill-evolution/compare-runs = **184, all green**. CI green across Python 3.10–3.14.

### Findings

**MINOR — fallback is silent where it used to fail loudly** (`scripts/quality_report.py:674`)
The old code raised `FileNotFoundError` naming the expected locations; the new code logs at `info` and proceeds with builtin rubrics. A user who runs from the wrong cwd (or whose `eval/eval_config.json` was moved/renamed) now gets a plausible-looking report graded with the canonical vocabulary instead of their custom one, with only an info-level line to hint why. The fallback itself is a reasonable feature, but this should be `logger.warning` naming the searched locations so a discovery mistake is visible.

**MINOR — the behavior change this PR makes is the one thing not tested** (`scripts/quality_report.py:670-678`)
None of the 6 new tests (nor any existing test — `_load_eval_config` has zero test references) covers the new no-file → builtin fallback or the `get_eval_metrics` delegation. A cheap test: clear `_EVAL_CONFIG_CACHE`, monkeypatch `_repo_root`/`_script_dir` to empty dirs, assert the builtin is returned and cached. Given the description explicitly calls out "no longer raising" as deliberate, it should be pinned.

**MINOR — doc "cannot drift" claim is overstated** (`docs/evaluation_rubrics.md`)
The description says the doc is "generated from the canonical data, so it cannot drift from what the judge actually sees." There is no generator script in the repo; the doc is a third, hand-maintained copy of the rubric text, and the drift guard pins only builtin ↔ `eval_config.json`, not the doc. The tables do match the config verbatim today (spot-checked all 8 metrics), but either add a generation/consistency check or soften the claim to "mirrors the canonical data."

**NIT — stale docstring on `_load_eval_config`** (`scripts/quality_report.py:636-646`)
Still says "The file is expected to contain: `metrics`…" with no mention of the builtin fallback when no file exists.

**NIT — `insert_after` typo silently reorders to front** (`src/bigquery_agent_analytics/evaluation_rubrics.py:280-283`)
When `insert_after` names a nonexistent category, `idx = -1` → `cats.insert(0, declined_cat)`, silently placing `declined` first. Verbatim from the old script (not a regression), but it's now public API consumed by arbitrary custom configs; falling back to append when `idx == -1` would be less surprising. The not-found case is also the one injection path the new tests don't cover.

**NIT — `build_metrics(None)` reads the module constant by reference** (`evaluation_rubrics.py:262-263`)
Safe today (the interpreter never mutates the config), but the safety rests on a no-mutation invariant over shared module state; using `builtin_metric_config()` here would make it self-evidently safe.

**NIT — import style** (`evaluation_rubrics.py:43-44`)
Absolute import while sibling `categorical_views.py:53` uses relative. Cosmetic.

### Positives

- Extraction verified programmatically: identical logic, ordering, and `insert_after`/`idx+1` semantics.
- The drift guard is genuine full-dict equality; the deep-copy isolation test correctly pins the mutate-freely contract.
- Doc accuracy is good where checkable: `declined` documented as inserted after `meaningful` (matches), and the Python example's API names match real signatures.
- The `--dimensions primary` path builds a filtered copy rather than mutating the cached config, so the new cached-builtin fallback can't be corrupted by it.

caohy1988:
## Code review: current head `0ad46a4`

**Verdict: Ready to merge.** I found no actionable defect in this PR.

Verified against the current head:

- The canonical eight-metric data matches `scripts/eval/eval_config.json`.
- `builtin_metric_config()` returns an isolated deep copy.
- `build_metrics()` preserves scope-aware definition enrichment, conditional `declined` insertion, `insert_after` positioning, and custom-config behavior.
- `scripts/quality_report.py` delegates to the core interpreter without keeping a second implementation.
- The missing-file path correctly falls back to the builtin config.
- The documented package-root imports work.
- Relevant local suites pass: `184 passed`. CI is green across Python 3.10-3.14, format, build, browser smoke, security scan, and CLA.

Non-blocking coverage note: the new missing-file fallback and package-root exports are not isolated in committed regression tests, although both paths passed direct execution during review.

Merge coordination: PR #432 edits the same export location in `src/bigquery_agent_analytics/__init__.py`, so the second PR to land will need a rebase and a trivial conflict resolution that retains both public export blocks.

caohy1988:
Open-PR sweep recheck at unchanged head `0ad46a4`.

No commits have landed since my full current-head review. The **Ready to merge** verdict still applies, and the current CI checks remain green. I still found no actionable defect in the rubric constants, exports, or reference documentation.

Full current-head review: https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/431#issuecomment-5352177394

Coordination note: #432 still edits overlapping evaluator/export surfaces, so whichever PR lands second should be refreshed against the first.

caohy1988 (approved):
LGTM - approved at exact head `0ad46a4`.

I re-reviewed the full diff and prior threads. The public exports, deep-copy configuration, fallback behavior, and `quality_report` semantics remain correct. Exact-head focused tests pass (6 tests), and current CI is green.

PR #432 overlaps some export/evaluator surfaces, so whichever lands second should refresh from `main` and rerun CI.

PR #441, opened 2026-08-24
docs: clarify thinking-token accounting

### LINKED ISSUE #382: Looker Studio dashboard: default date range hides all data for new installations (real-agent E2E findings)

sunpan9209:
## Summary

- Clarify that `SystemEvaluator.token_efficiency()` prefers an explicit
  provider-reported total token count, which can include Gemini thinking tokens
  and therefore may not equal prompt plus completion tokens.
- Document the exact fallback to raw `attributes.input_tokens` plus
  `attributes.output_tokens` when no explicit provider total is available.
- Distinguish that behavior from `cost_per_session()`, which estimates cost
  using extracted input/output tokens and caller-provided rates without
  consuming a separate `thoughts_token_count` field.
- Add the same accounting caveat to the Looker Studio user manual and state
  that thinking tokens are not charted separately.

This addresses the thinking-token documentation follow-up in #382. It does not
close #382 or change evaluator or dashboard behavior.

## Testing

- Targeted evaluator and dashboard tests: 108 passed
- `pytest --tb=short -q` (Python 3.13): 4070 passed, 73 skipped
- `git diff --check`

### DIFF
--- a/SDK.md
+++ b/SDK.md
@@ -194,6 +194,18 @@ evaluator = SystemEvaluator.cost_per_session(
 )
 ```
 
+`token_efficiency()` uses the session's `total_tokens`. When source telemetry
+provides Gemini `usage_metadata.total_token_count`, that provider-reported
+total can include `usage_metadata.thoughts_token_count`; it is therefore not
+guaranteed to equal `prompt_token_count + candidates_token_count`. If no
+provider total exists, the session query falls back to input plus output
+tokens.
+
+`cost_per_session()` is a separate estimate based only on `input_tokens` and
+`output_tokens` plus the configured rates. It does not consume a separate
+`thoughts_token_count` field, so do not treat it as provider invoice
+reconciliation when thinking tokens are present.
+
 `context_cache_hit_rate()` requires source telemetry that includes
 Gemini `usage_metadata.cached_content_token_count`. Older plugin data
 may not contain that field. When cache telemetry is absent, the
--- a/SDK.md
+++ b/SDK.md
@@ -194,12 +194,15 @@ evaluator = SystemEvaluator.cost_per_session(
 )
 ```
 
-`token_efficiency()` uses the session's `total_tokens`. When source telemetry
-provides Gemini `usage_metadata.total_token_count`, that provider-reported
-total can include `usage_metadata.thoughts_token_count`; it is therefore not
-guaranteed to equal `prompt_token_count + candidates_token_count`. If no
-provider total exists, the session query falls back to input plus output
-tokens.
+`token_efficiency()` uses the session's `total_tokens`. The session query
+prefers an explicit provider total: Gemini
+`usage_metadata.total_token_count`, followed by `content.usage.total` from
+other telemetry shapes. Gemini's total can include
+`usage_metadata.thoughts_token_count`; it is therefore not guaranteed to equal
+`prompt_token_count + candidates_token_count`. If neither provider total
+exists, the query falls back specifically to the raw `attributes.input_tokens`
+plus `attributes.output_tokens` event fields. That fallback does not reuse all
+alternate paths recognized by the separate input and output counters.
 
 `cost_per_session()` is a separate estimate based only on `input_tokens` and
 `output_tokens` plus the configured rates. It does not consume a separate
--- a/dashboard/looker_studio/USER_MANUAL.md
+++ b/dashboard/looker_studio/USER_MANUAL.md
@@ -143,7 +143,7 @@ All eight pages share **one date range control**. It defaults to a rolling
 other page, including the Trace Inspector. Shorter windows are cheaper and
 faster — pick the shortest range that answers your question.
 
-### Three things worth knowing
+### Four things worth knowing
 
 - **First paint is not instant.** A cold load or page switch can take up to 90
   seconds before every chart on the page is drawn. The report controls appear
@@ -154,6 +154,11 @@ faster — pick the shortest range that answers your question.
 - **"Data Last Updated" in the footer is not your data's freshness.** It is
   Looker Studio's connector refresh time. Your newest events may be newer or
   older than that stamp.
+- **Total tokens can exceed prompt plus completion.** The dashboard uses the
+  provider-reported total token count. Gemini telemetry can include thinking
+  tokens in that total, so prompt and completion counts in the source data may
+  not add up to the displayed total. The dashboard does not chart thinking
+  tokens separately.
 
 ---
 
--- a/docs/design.md
+++ b/docs/design.md
@@ -480,6 +480,16 @@ value strictly exceeds the budget.
 | `ttft(threshold_ms)` | `avg_ttft_ms` | `observed <= threshold_ms` |
 | `cost_per_session(max_cost_usd, ...)` | `(input_tokens/1K)*input_rate + (output_tokens/1K)*output_rate` | `observed <= max_cost_usd` |
 
+For `token_efficiency()`, `total_tokens` prefers the provider-reported
+`usage_metadata.total_token_count`. Gemini can include
+`usage_metadata.thoughts_token_count` in that total, so the value is not
+guaranteed to equal prompt plus completion tokens. When a provider total is
+absent, the session summary falls back to input plus output tokens.
+
+`cost_per_session()` does not consume a separate `thoughts_token_count` field;
+it remains an estimate over the extracted input and output counts and the
+caller-supplied rates.
+
 `context_cache_hit_rate()` treats missing cache telemetry as unknown,
 not as a cache miss. Older plugin rows without
 `usage_metadata.cached_content_token_count` report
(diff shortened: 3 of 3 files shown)


caohy1988:
## Code review

**Verdict: approve.** Docs-only, and the claims check out against `evaluators.py`:

- The session summary query computes `total_tokens` as `SUM(COALESCE(attributes.usage_metadata.total_token_count, content.usage.total, input_tokens + output_tokens))` — so "prefers the provider-reported total, falls back to input plus output" is accurate, and since Gemini's `total_token_count` includes thinking tokens, the "total can exceed prompt + completion" caveat is exactly right.
- `token_efficiency()` compares that `total_tokens` against the budget, and `cost_per_session()` computes `(input_tokens/1000)*input_rate + (output_tokens/1000)*output_rate` with no `thoughts_token_count` term — the "not invoice reconciliation" framing is the correct warning.

One optional precision nit: the SDK.md paragraph frames the preferred total as Gemini's `usage_metadata.total_token_count`, but the query's COALESCE also prefers a non-Gemini provider total (`content.usage.total`, the OpenAI-style shape) before falling back to input + output. The current text isn't wrong — a reader could just conclude the fallback triggers for all non-Gemini telemetry when it doesn't. A half-sentence ("or another provider-reported total when present") would close it. Fine to merge as is; CI is green.

caohy1988:
## Full review - Ready with one docs/query alignment fix

I reviewed the complete documentation diff, linked issue #382, token and cost evaluator implementations, query fallbacks, tests, and current CI at head `8577b86`. The provider-total and thinking-token explanation is otherwise accurate.

### P2 - The documented total-token fallback is broader than the query

`SDK.md:201` says the session query falls back to extracted input plus output tokens. However, `SESSION_SUMMARY_QUERY` accepts `content.usage.prompt` and `content.usage.completion` for the separate input/output counters (`evaluators.py:957-978`), while its final total fallback (`evaluators.py:1063-1071`) reads only `attributes.input_tokens` plus `attributes.output_tokens`. A row with content-only prompt/completion values and no explicit total therefore has nonzero input/output but `total_tokens = 0`, contrary to the new text.

Please either make the total expression reuse the same complete input/output paths and add a content-only regression row, or narrow both new documentation passages to the actual attribute-only fallback.

Current CI is green.

caohy1988 (commented):
## Review: Approve with nits

Docs-only PR whose every substantive claim I traced to actual code behavior on the PR head:

- **`token_efficiency()` uses session `total_tokens`** ✅ (evaluators.py:443,451)
- **Provider total preferred; falls back to input+output** ✅ — `SESSION_SUMMARY_QUERY` COALESCE ordering (evaluators.py:1055-1073) matches the doc exactly
- **Gemini `total_token_count` can include `thoughts_token_count`** ✅ — corroborated by issue #382's real-trace evidence (total 113 = prompt 62 + completion 5 + thoughts 46); provider behavior, not derivable from this repo
- **`cost_per_session()` uses only input/output tokens + caller rates** ✅ (evaluators.py:506-510; zero `thoughts` reads anywhere in `src/`)
- **Dashboard uses provider-reported total, no separate thinking chart** ✅ (events_v1.template.sql:60-75; oracle query agrees)
- **"Three things" → "Four things" heading fix** ✅ — section verified to contain exactly four bullets

Directly addresses the docs follow-up in #382 and is honest about scope.

### Findings

**NIT — fallback description slightly imprecise** (`SDK.md:203` / `docs/design.md:487`)
"Falls back to input plus output tokens" reads the raw `attributes.input_tokens`/`attributes.output_tokens` event fields, *not* the session-summary `input_tokens`/`output_tokens` columns (which themselves prefer `prompt_token_count`/`candidates_token_count`, evaluators.py:966-988). So a row carrying only `usage_metadata.prompt_token_count`/`candidates_token_count` contributes 0 to `total_tokens` while contributing to the input/output columns — `total_tokens` can in edge cases be *less* than input+output. Suggested tweak: "falls back to the raw `input_tokens` + `output_tokens` event attributes."

**NIT — PR body test count not reproducible**
"Targeted evaluator and dashboard tests: 152 passed" — `test_sdk_evaluators.py` + `test_looker_studio_dashboard.py` collect 108 (ran them: 108 passed, Python 3.13). No obvious combination yields 152. Immaterial for docs-only, but the count is off.

### Positives

Claims are accurate and conservative ("can include", "not guaranteed to equal"); placement matches surrounding docs; `git diff --check` clean.

### Unverified claims

"4070 passed, 73 skipped" full-suite run — not re-run (docs-only diff; targeted tests pass).

sunpan9209:
Addressed the documentation/query alignment feedback in `1eb7123`.

Both SDK passages now document the exact `total_tokens` precedence used by `SESSION_SUMMARY_QUERY`:

```text
attributes.usage_metadata.total_token_count
  -> content.usage.total
  -> attributes.input_tokens + attributes.output_tokens
```

The text also states that the final fallback does not reuse every alternate path recognized by the separate input/output counters. This keeps the PR documentation-only, matching the #382 follow-up, without expanding it into an evaluator behavior change.

I also corrected the PR description's targeted-test count from 152 to the reproducible 108.

Validation:

- targeted evaluator and dashboard tests: 108 passed
- full Python 3.13 suite: 4070 passed, 73 skipped
- `git diff --check`

sunpan9209:
CI note: the current `Format check` failure is an upstream formatter-version regression, not a change in this PR. It reproduces on current `upstream/main` (`71b0bbc`) when the unbounded workflow installs pyink 26.x. The Python 3.10-3.14 matrix, browser smoke, security scan, and CLA checks pass.

Upstream PR #447 pins `pyink <26` specifically to restore this check, and its format job is green. I have not added unrelated repository-wide formatter churn to this branch; the format check should be refreshed after #447 lands.

caohy1988 (approved):
LGTM - approved at exact head `1eb7123`.

The follow-up now documents the exact token precedence implemented by the SQL: `attributes.usage_metadata.total_token_count`, then `content.usage.total`, then the raw `attributes.input_tokens + attributes.output_tokens` fallback. It also correctly states that the final fallback does not reuse alternate counter paths. Exact-head evaluator tests pass (82 tests), and the test matrix is green.

The remaining red Format check is the repository-wide pyink 26 workflow regression fixed by merged PR #447. Please refresh from `main` and rerun required CI before merge; this approval is not a CI waiver.

Issue #445, opened 2026-08-25
[Dashboard] Configurator fails with "This report isn't shared with you" — the Linking API copy is blocked by the template report's ACL


haiyuan-eng-google:
## Summary

An external user (non-Google domain, reported through a private feedback channel on 2026-08-24) opened
https://googlecloudplatform.github.io/BigQuery-Agent-Analytics-SDK/, entered their three identifiers, clicked
**Create my dashboard**, and Looker Studio refused with a terminal dialog:

> **Something went wrong when building your report**
> This report isn't shared with you.
>
> ```
> https://datastudio.google.com/reporting/create
>   ?c.reportId=5a3f85ef-fc9c-4730-8ef2-8ef9129ddb40
>   &c.mode=view
>   &r.reportName=BigQuery Agent Analytics — <REDACTED_DATASET>.<REDACTED_TABLE>
>   &ds.ds230.datasourceName=BQAA — <REDACTED_PROJECT>.<REDACTED_DATASET>.<REDACTED_TABLE>
>   &ds.ds230.billingProjectId=<REDACTED_PROJECT>
>   &ds.ds230.sqlReplace=test-project-0728-467323,<REDACTED_PROJECT>,bqaa_fixtu…
>   &ds.ds230.refreshFields=false
> ```
>
> `[Reload]` `[Return to report list]` `[Go to report template]`

Reporter identifiers are redacted per `dashboard/looker_studio/README.md` → *Publication safety*. Everything not
redacted above is already committed in this repository.

The URL is byte-for-byte the shape `buildDashboardUrl()` emits
(`dashboard/looker_studio/docs/configurator.mjs` L95–L117) from `REPORT_CONFIG`
(`dashboard/looker_studio/docs/report-config.mjs`). The configurator did its job. The failure is server-side, at the
Linking API's copy-the-template step.

## What this is *not*

* **Not the reporter's BigQuery setup.** `/reporting/create?c.reportId=…` resolves the *template* report's ACL before
  any `ds.*` parameter is applied. No dataset, table, IAM binding, or billing project of theirs is consulted yet.
* **Not the `datastudio.google.com` host in the dialog.** `lookerstudio.google.com` 302s to `datastudio.google.com`;
  reproduced here with `curl -sL -w '%{url_effective}'`. The generated link (which correctly uses
  `lookerstudio.google.com`) is unchanged by that redirect.
[...]

Issue #398, opened 2026-08-07
[Dashboard/configurator] "Create my dashboard" flashes a "page not found" error before the Looker Studio report loads


haiyuan-eng-google:
Part of #404.

## Reported behavior

Internal field feedback from a Google Cloud customer engineer evaluating the
configurator end to end:

> When I first clicked the "Create my dashboard" button, it got re-directed to
> a page which said something like "page not found… bla bla bla… long URL…". I
> was just about to click off the page when it re-loaded and opened a page.
> Maybe this can be fixed somehow? Or the user very clearly notified that when
> they click this button this page will be shown for ~5 seconds.

The failure mode matters more than the flicker: the user's instinct is to close
the tab, which aborts the flow at the exact moment of conversion. They only
stayed because they were already inclined to trust the page.

## What the code does today

`docs/app.mjs` opens the Linking API URL in a new tab with no interstitial and
no expectation set:

```js
// dashboard/looker_studio/docs/app.mjs:84-90
form.addEventListener("submit", (event) => {
  event.preventDefault();
  refresh();
  if (createLink.href) {
    window.open(createLink.href, "_blank", "noopener,noreferrer");
  }
});
```

The destination built by `buildDashboardUrl()`
(`docs/configurator.mjs:69-91`) is
`https://lookerstudio.google.com/reporting/create?…` carrying seven
`c.*` / `ds.ds230.*` parameters, including a `sqlReplace` list of six values.
It is long, and `/reporting/create` has to provision a data source and copy an
8-page / 37-chart report server-side before it can render — which is when the
transient error page appears.

Nothing on our page sets that expectation. `#form-status` is left reading
`Ready for <project>.<dataset>.<table>.` and never updates on click: the only
click handler on `#create-dashboard` fires when `href` is *unset*
(`docs/app.mjs:92-97`). The "02 Review in Google" step
(`docs/index.html:156-163`) describes the review dialog but says nothing about
a wait.

## Scope note

The error page is served by `lookerstudio.google.com`, not by anything in this
repo. Unless someone finds a Linking API parameter or URL form that avoids it,
we cannot remove the state — only the surprise. If a maintainer knows of a
shorter/warmer entry point into `/reporting/create`, that would supersede
everything below.

## Proposed fix

- [ ] Add a real `click` handler for `#create-dashboard` in `docs/app.mjs` that
      sets `#form-status` to an explicit waiting message, e.g. *"Opening Looker
[...]

PR #446, opened 2026-08-25
fix(looker-studio): attest external template access, name the terminal "not shared" dialog (#445)

caohy1988:
Implements the repository-side half of #445 (fix plans B + D), incorporating the 2026-08-25 validation findings on the issue: the live ACL currently reads `LINK_VIEWER: allUsers` and the report is non-trashed, so the cause is **not isolated** — the Permissions API cannot see the separate viewer copy-disable control, and only an end-to-end copy canary from a signed-in, non-owner, out-of-domain identity clears the whole path the configurator depends on.

## What changed

**`bindings/report_template.yaml`** — new `external_access_verification` attestation with two controls:

1. `permissions_api_link_role_check` — authenticated Permissions API read must list `LINK_VIEWER: allUsers`, called with the least-privilege `datastudio.readonly` scope from a Workspace/Cloud Identity account; `PERMISSION_DENIED` is treated as indeterminate (caller constraint *or* lost asset authorization). Records its known limitation (`does_not_expose_viewer_copy_disable_control`) and its last observation (2026-08-25, present), with the verbatim command and redacted response reproducible from the file.
2. `external_identity_link_access_check` — end-to-end copy canary (`/reporting/create?c.reportId=…&c.mode=view&c.explain=true`) that records `last_observed_date` on **every** run, pass or fail (currently 2026-08-24, the reported failure); `link_access_verified_date: null`, `last_result: FAILURE_REPORTED`.

Overall `status: FAILING`, `cadence: monthly_manual_until_automated`, `next_due_date: 2026-09-24` anchored to the canary's own observation, tracking #445. The outage is also recorded in `known_live_issues` with its `NOT_ISOLATED` cause state naming the three remaining credible causes.

**Attestation contract (`tests/test_looker_studio_dashboard.py`)** — the bindings-drift test pins the attestation byte-for-byte, and `test_external_access_attestation_is_dated_or_tracked` enforces evidence-consistency in both directions:

- `PASSING` requires success results on **both** controls, a canary verification date equal to its observation date, no older than the published template, **strictly newer than every recorded incident** for the tracked issue, and no matching OPEN live issue — recovery evidence cannot predate the outage, and the status cannot flip while contradictory failure evidence remains.
- `FAILING` requires `FAILURE_REPORTED` on the canary plus a matching OPEN `known_live_issues` entry — the outage cannot be silently erased.
- No date may be in the future, and `next_due_date` must fall within 35 days of the **canary's** `last_observed_date` — refreshing the weaker API read alone cannot advance the deadline.

**Wall-clock enforcement** — unit tests deliberately never compare attestation dates to today (that would fail unrelated PRs by time passing). Instead, `.github/workflows/external-access-staleness.yml` runs `scripts/check_external_access_staleness.py` weekly: once `next_due_date` passes it opens (or bumps) a tracking issue and fails the scheduled run, with no credentialed API involved. Unit tests pin the script's verdicts on both sides of the deadline and the workflow wiring.

**Docs (`docs/index.html`, `docs/app.mjs`, `README.md`, `USER_MANUAL.md`)** — every wait-it-out note (including the dynamic status message a user watches after clicking) now distinguishes the terminal **"This report isn't shared with you"** dialog from the #398 provisioning flicker, and a `#report-not-shared` explainer states it is a template-side copy failure — nothing is wrong with the user's data, IAM, or identifiers. Reporters are asked to say only whether the signed-in account is **personal or part of an organization**, and explicitly told not to post the account's email address (reporter identifiers are redacted per *Publication safety*). The CLI preflight is explicitly *not* presented as a workaround (it copies the same template); Grafana is the alternative. The viewer copy-disable control is described as a candidate cause, not a confirmed one.

**`tools/test_web_configurator.mjs`** — pins the new copy (explainer present, anchor references from both wait-it-out notes, verbatim dialog quotes, issue link, warning-before-dialog ordering, and the WAITING_MESSAGE exception) so none of it can regress silently.

## Verification

- `python3 -m pytest tests/test_looker_studio_dashboard.py` → 29 passed (includes the node contract test and the real-Chrome browser smoke).
- `bash autoformat.sh` (pyink 25.x + isort) → no changes, format-stable.
- Contract negative-tested against the review's mutation scenarios (contradictory PASSING, recovery predating the outage, API-only deadline advance, future dates) — all rejected.
- `scripts/check_external_access_staleness.py` → exit 0 today, exit 1 with an actionable message for a post-deadline date.

## Out of scope (owner-side, tracked on #445)

Restoring/confirming the copy path (fix plan A — including the viewer copy-disable check under **Share → Manage access**), the non-owner canary click-through, and the ownership transfer (plan C, `governance.transfer_status: PENDING_MAINTAINER`) happen in Looker Studio/Drive. Once the canary passes, the resolution path is codified: set `status: PASSING`, fill both canary dates, close the `known_live_issues` entry — the contract test enforces exactly that shape.

Addresses the repository-side acceptance criteria 2–4 of #445; criteria 1 and 5 remain owner-side.


caohy1988:
CI status: all substantive checks pass (5-version Python test matrix, Configurator browser smoke, CLA). The **Format check failure is not from this PR**: pyink 26.5.1 shipped after `main` last ran CI (2026-08-23, green) and reformats 30 pre-existing files this PR never touches; this branch's files are format-clean under both pyink 25.x and 26.5.1 (verified locally). #447 pins `pyink<26` to unbreak the gate for all PRs — once it merges, rerunning the Format check here should go green.

tests/test_looker_studio_dashboard.py
```diff
@@ -532,6 +587,74 @@ def test_report_and_web_bindings_cannot_drift():
...
+  assert (
+      api_check["limitation"] == "does_not_expose_viewer_copy_disable_control"
+  )
+  datetime.date.fromisoformat(api_check["last_observed_date"])
+
+  canary = controls["external_identity_link_access_check"]
+  assert "c.explain=true" in canary["protocol"]
+  assert attestation["cadence"] == "monthly_manual_until_automated"
+  assert attestation["status"] in {"PASSING", "FAILING"}
+  if attestation["status"] == "PASSING":
```
caohy1988, line 696: **P2 — `PASSING` permits contradictory failure evidence.**

This branch only checks that `link_access_verified_date >= published_date`. A future update can therefore set `status: PASSING` and add a date while leaving `last_result: FAILURE_REPORTED` and the matching `known_live_issues` entry `OPEN`; after updating the snapshot assertion, this semantic contract test still passes. That contradicts the PR description's claimed resolution shape.

Recommendation: when `PASSING`, require an explicit successful canary result, a successful API result, and no matching open live issue. When `FAILING`, require failure evidence plus the matching open issue.

tests/test_looker_studio_dashboard.py
```diff
@@ -532,6 +587,74 @@ def test_report_and_web_bindings_cannot_drift():
...
+  controls = {control["method"]: control for control in attestation["controls"]}
+  assert set(controls) == {
+      "permissions_api_link_role_check",
+      "external_identity_link_access_check",
+  }
+  api_check = controls["permissions_api_link_role_check"]
+  assert (
+      api_check["limitation"] == "does_not_expose_viewer_copy_disable_control"
+  )
+  datetime.date.fromisoformat(api_check["last_observed_date"])
```
caohy1988: **P2 — the monthly attestation never expires.**

`fromisoformat()` validates only syntax. Despite `cadence: monthly_manual_until_automated`, an arbitrarily old API observation remains valid, and a `PASSING` canary can remain valid forever as long as its date is no older than `published_date`. Future dates are also accepted. This does not enforce issue #445's dated-manual-protocol acceptance criterion.

Recommendation: reject future dates and enforce a documented maximum age for both controls, or record and enforce an explicit `next_due_date`.

dashboard/looker_studio/docs/index.html
```diff
@@ -192,6 +197,25 @@ <h3>Save, then secure sharing</h3>
...
+        That dialog is terminal — it offers <em>Reload</em> / <em>Return to
+        report list</em> / <em>Go to report template</em> and does not resolve
+        by waiting. It means the shared dashboard template is unavailable to
+        the Google account signed in to that tab; nothing is wrong with your
+        project, dataset, table, IAM setup, or the identifiers you entered.
+        First confirm the tab is using the account you intend (Looker Studio
+        uses the browser’s default Google account). If it is, the template’s
+        copy path is blocked on the owner’s side —
+        <a href="https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/445">
+          report it on issue #445</a>, noting which Google account the dialog
```
caohy1988: **P2 — avoid soliciting account identifiers on a public issue.**

Asking users to report which Google account was selected can lead them to post a personal or corporate email address publicly, while issue #445 deliberately redacts reporter identifiers. The same wording appears in `README.md` and `USER_MANUAL.md`.

Recommendation: request only non-identifying context such as owner/non-owner and inside/outside the owner's domain, explicitly tell users not to post email addresses, and use a private channel if the exact identity is necessary.

dashboard/looker_studio/README.md
```diff
@@ -145,6 +145,27 @@ The standard path queries the provided table and uses the source project for
...
+If the new tab shows the terminal dialog **"This report isn't shared with
+you"** (offering *Reload* / *Return to report list* / *Go to report
+template*), do not wait it out: `/reporting/create` resolves the *template*
+report's ACL before any `ds.*` parameter is applied, so this failure is on the
+shared template's copy path, not the user's data, IAM, or identifiers.
+Confirm the browser's default Google account is the intended one; if the
+dialog persists, report it on
+[#445](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/445)
+with the account the dialog selected. The link role alone may not explain a
+denial — Looker Studio's "disable download, print, and copy for viewers"
```
caohy1988: **P3 — qualify this unverified causal claim.**

Google's documentation establishes that this control prevents copying, but neither that documentation nor issue #445 verifies that it produces this exact dialog. The issue currently records the copy-disable control only as one credible, not-isolated cause.

Recommendation: say that the control *may* block copying and remains a candidate cause until an end-to-end reproduction confirms the exact dialog.

caohy1988 (commented):
Full review: I found three P2 issues that should be addressed before merge and one P3 documentation correction. The focused dashboard suite passes (28 tests), the configurator smoke test passes, and the Python 3.10–3.14 CI matrix is green. The current Format check failure is unrelated repository-wide pyink drift tracked by #447; it is not attributed to this diff. Inline comments contain the actionable findings.

caohy1988 (commented):
Reviewed locally on a worktree at a6fdf76: 28 pytest passed, `tools/test_web_configurator.mjs` OK, `browser_smoke.sh` OK.

## What checks out (verified, not assumed)

- **The Permissions API vocabulary is exactly right.** `LINK_VIEWER` with member `allUsers` = "Anyone with the link can view", and `assets:search` returns `trashed` — confirmed against the [official API types reference](https://developers.google.com/looker-studio/integrate/api/reference/types). The two-control design is sound: the API exposes the link role but not the viewer copy-disable control, so the end-to-end canary is genuinely necessary.
- **`c.explain=true` is real** — it enables the Linking API debug dialog. Good choice for the canary protocol.
- **Honest bookkeeping:** `status: FAILING`, `link_access_verified_date: null`, `known_live_issues` with `cause_isolation: NOT_ISOLATED` — the attestation claims no more than was observed. The FAILING⇒OPEN-issue invariant in the new test is a good idea.
- **Docs nuance is correct:** `hydrate_dashboard.py` is *not* an outage workaround (its URL copies the same template); Grafana is. The multi-account-first guidance in USER_MANUAL matches hypothesis 4 in #445.

## Findings

**[P2] `bindings/report_template.yaml` comment — "byte-identical" is falsified by the evidence it cites.**
The comment says the embed shell is "byte-identical for public, private, and nonexistent report IDs (#445)". #445's own measurement: real ID = 71973 bytes, nonexistent = 71971 bytes (reproduced again today). The correct claim is "the same application shell, indistinguishable without an identity". Precision matters doubly in an attestation file. Suggest rewording to the issue's phrasing.

**[P2] Nothing enforces the monthly cadence — acceptance criterion 2 is only weakly met.**
`cadence: monthly_manual_until_automated` is a label, not a mechanism. Once `status` flips to PASSING, the only date invariant is `link_access_verified_date >= published_date` — and `published_date` is static (`2026-08-11`), so a verification done today stays green forever even if access regresses later. That recreates the blind spot that caused #445: nobody is prompted to re-run the canary. Suggest a follow-up: a scheduled (weekly) staleness workflow that opens/bumps a tracking issue when `link_access_verified_date` is missing or older than 30 days — no hosted credential needed, it just makes the dated manual protocol recur. Happy to contribute this as a stacked PR.

**[P3] `docs/app.mjs` WAITING_MESSAGE untouched.**
The static notes in `index.html` now distinguish the terminal dialog, but the dynamic status text set on submit/click still says only "may briefly show an error page — don't close it" — the message a user actually watches after clicking still trains them to wait out a terminal denial. The new tests assert on `index.html` only, so this gap is invisible to CI. One appended sentence fixes it (the duration-consistency test constrains the edit).

**[P3] The API protocol omits caller constraints.**
`GET …/v1/assets/{id}/permissions` isn't callable from just any account: the Looker Studio API is gated to Workspace/Cloud Identity org members with the `datastudio` OAuth scope (even its discovery doc 403s for unregistered callers). A maintainer following the protocol from a personal Gmail hits `PERMISSION_DENIED` and may misread it as a sharing regression. One line in the protocol string ("call from a Workspace-org identity with the datastudio scope") prevents that.

**[P3] Step-02 contract loosened in `tools/test_web_configurator.mjs`.**
The old regex pinned the transient warning immediately before the dialog explanation; the split into two independent matches lets the two fragments drift apart undetected. Net coverage is still good given the new #445 assertions. Optional: keep the split but assert ordering.

**Ask:** `cause_isolation` records a 2026-08-25 authenticated API read (`LINK_VIEWER allUsers`, non-trashed) — pivotal evidence and vocabulary-exact, but not reproducible from the PR. Please paste the command and redacted response (here or into the protocol) so the next person can re-run it verbatim.

**Recommendation: approve with comments** — fix the "byte-identical" wording before merge; the rest work as follow-ups.

tests/test_looker_studio_dashboard.py
caohy1988: Fixed in 327e4d5. PASSING now requires `last_result: PASSED` on the canary, `LINK_VIEWER_ALLUSERS_PRESENT` on the API control, and no matching OPEN `known_live_issues` entry; FAILING requires `FAILURE_REPORTED` plus the OPEN entry. Verified against your exact mutation (flip to PASSING + add a date, leave FAILURE_REPORTED and the OPEN issue): the contract test now rejects it, along with the OPEN-issue-survives-PASSING and future-date variants.

tests/test_looker_studio_dashboard.py
caohy1988: Addressed in 327e4d5, taking the `next_due_date` arm of your recommendation: both control dates are now rejected if in the future, and a required `next_due_date` must fall within 35 days after the newest recorded observation (currently 2026-09-24, from the 2026-08-25 API read). I deliberately did not add a wall-clock max-age assertion — a unit test that turns red purely by time passing would start failing on unrelated contributors' PRs a month from now. Instead `next_due_date` is machine-readable for the scheduled staleness check you offered to contribute; that follow-up is the piece that makes the cadence recur.

dashboard/looker_studio/docs/index.html
caohy1988: Fixed in 327e4d5 across all three surfaces (index.html, README, USER_MANUAL): reporters are asked to say only whether the signed-in account is personal or part of an organization, and explicitly not to post the account's email address. The owner-side canary protocol in report_template.yaml now says to record the selected account privately and never post identifiers publicly.

dashboard/looker_studio/README.md
caohy1988: Fixed in 327e4d5. The README now says the copy-disable control also prevents copying and remains a candidate cause for this dialog until an end-to-end reproduction confirms it — matching the NOT_ISOLATED cause record in the yaml rather than overstating it.

caohy1988:
All review findings addressed in 327e4d5 (28/28 tests pass including the node contract test and browser smoke; pyink 25.x clean). Item by item:

- **[P2] "byte-identical"** — reworded in the yaml comment to the issue's phrasing: the embed serves the *same application shell* for public, private, and nonexistent report IDs (sizes differ only by the echoed id), indistinguishable without an identity.
- **[P2] cadence enforcement** — took the `next_due_date` arm of your recommendation (details in the inline thread): future dates rejected, `next_due_date` statically consistency-checked against the newest observation, no wall-clock time bomb in CI. Accepting your offer of the scheduled staleness workflow as a stacked PR — `next_due_date` is there for it to consume.
- **[P2] contradictory PASSING** — closed; verified against your exact mutation (inline thread has the specifics).
- **[P2] account identifiers** — all three doc surfaces plus the yaml protocol reworded (inline thread).
- **[P3] `app.mjs` WAITING_MESSAGE** — now excepts the terminal dialog from the wait guidance, and `docs/app.mjs` is added to the docs test's surface list so the gap is CI-visible; the node test also asserts the exception directly.
- **[P3] API caller constraints** — the protocol now names the Workspace/Cloud Identity + `datastudio` OAuth scope requirement and states that PERMISSION_DENIED for other callers is a caller constraint, not a sharing regression.
- **[P3] step-02 contract** — kept the split matches and added an ordering assertion (transient warning must precede the dialog explanation).
- **[P3] copy-disable causal claim** — qualified in README to candidate-cause status (inline thread).

**On the ask (reproducible API read):** the observation is recorded in the [2026-08-25 validation comment on #445](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/445#issuecomment-5416997692) — I did not run it in this PR's verification. To re-run verbatim, from a Workspace/Cloud Identity account whose token carries the `https://www.googleapis.com/auth/datastudio` scope:

```
curl -s -H "Authorization: Bearer ${DATASTUDIO_SCOPED_TOKEN}" \
  "https://datastudio.googleapis.com/v1/assets/5a3f85ef-fc9c-4730-8ef2-8ef9129ddb40/permissions"
```

Recorded response: `{"LINK_VIEWER": {"members": ["allUsers"]}}`. A plain `gcloud auth print-access-token` won't carry the scope by default — that's the caller constraint now documented in the protocol string.

caohy1988 (commented):
Re-review of `327e4d5`, verified locally on a worktree: 28 pytest passed, `test_web_configurator.mjs` OK, `browser_smoke.sh` OK.

## Previous findings — resolution audit

- **[P2] "byte-identical"** — **Fixed.** "The same application shell … (sizes differ only by the echoed id)" is accurate and matches the measured 71973/71971.
- **[P2] Cadence enforcement** — **Partially addressed, honestly.** `next_due_date` with a real invariant (`newest_observation < next_due ≤ newest + 35d`) puts the data contract in place; the docstring defers the scheduled staleness check explicitly. Noted as a follow-up, not a blocker — a weekly workflow consuming `next_due_date` can close this.
- **[P3] `app.mjs` WAITING_MESSAGE** — **Fixed**, and pinned twice (JS contract regex + the docs-surface test).
- **[P3] API caller constraints** — **Fixed.** Workspace/Cloud Identity + `datastudio` scope, PERMISSION_DENIED-is-caller-constraint is exactly right.
- **[P3] Step-02 ordering** — **Fixed** via the `indexOf` assertion.
- **Ask: paste the API command + redacted response** — **not addressed**; `cause_isolation` is still not reproducible from the PR. Three lines would do it.

## New findings in this revision (both minor)

**[P3] No date field records a canary *failure* re-run.** The `next_due_date` invariant anchors to the newest observation, but when FAILING the only observation date is the API read's `last_observed_date` — `link_access_verified_date` is null and passes-only. If someone re-runs the canary on 2026-09-10 and it fails again, there is nowhere to record that date, so `next_due_date` stays anchored to 2026-08-25 and silently schedules the next run more than 30 days from the actual latest check. Suggest giving the canary control its own `last_observed_date` (set on every run, pass or fail) and including it in `observation_dates` when `last_result == "FAILURE_REPORTED"`.

**[Nit] `docs/index.html`** — one reflow casualty: "account's email address. The repository's CLI preflight can still validate your" is a ~110-char line inside an ~80-char-wrapped paragraph.

## Also verified in the new test logic

- The `next_due_date` invariant never compares against wall-clock `today` — no CI time bomb; only future-dating is rejected. Good design.
- PASSING now requires success on both controls plus no OPEN tracked issue; FAILING requires `FAILURE_REPORTED` + OPEN issue. Evidence-consistent in both directions.
- The privacy correction (report "personal vs organization", never the email) is applied consistently across README, USER_MANUAL, index.html, and the canary protocol — aligns with *Publication safety*.

**Recommendation: approve.** Remaining items are small follow-ups, not blockers.

tests/test_looker_studio_dashboard.py
```diff
@@ -532,6 +593,117 @@ def test_report_and_web_bindings_cannot_drift():
...
+  tracking_issue = attestation["tracking_issue"]
+  open_tracked = any(
+      entry["issue"] == tracking_issue and entry["status"] == "OPEN"
+      for entry in report["known_live_issues"]
+  )
+  assert attestation["status"] in {"PASSING", "FAILING"}
+  if attestation["status"] == "PASSING":
+    assert canary["last_result"] == "PASSED"
+    assert api_check["last_result"] == "LINK_VIEWER_ALLUSERS_PRESENT"
+    verified = datetime.date.fromisoformat(canary["link_access_verified_date"])
```
caohy1988, line 711: **P2 — recovery evidence can predate the outage.**

This only requires the successful canary date to be on or after `published_date` (2026-08-11). A maintainer can set `PASSED` with `link_access_verified_date: 2026-08-11`, mark the issue closed, and satisfy this contract even though the tracked failure was reported on 2026-08-24. That does not prove recovery.

Recommendation: preserve a dated failure/attempt record and require a PASSING canary observation to be newer than every matching failure or incident date.

caohy1988 (commented):
Full re-review of `327e4d5`: the prior contradictory-status, public-account-identifier, copy-control wording, dynamic WAITING_MESSAGE, and step-order findings are fixed. Three P2 issues remain inline and should be addressed before merge.

One P3 metadata correction is also needed: the PR description still says the public explainer asks users to report “which account the dialog selected,” and its test summary omits the new success-result/open-issue/date invariants. Please update the PR body to match the privacy-safe wording and current state contract.

Verification: 28 focused pytest tests passed; the configurator smoke test, Node syntax checks, `git diff --check`, pyink<26, and isort<8 passed. GitHub’s Python 3.10–3.14 matrix, configurator smoke, CLA, and security checks are green. The Format check remains an unrelated repository-wide pyink issue tracked by #447.

tests/test_looker_studio_dashboard.py
```diff
@@ -532,6 +593,117 @@ def test_report_and_web_bindings_cannot_drift():
...
+    )
+  else:
+    assert canary["last_result"] == "FAILURE_REPORTED"
+    assert canary["link_access_verified_date"] is None
+    assert tracking_issue
+    assert (
+        open_tracked
+    ), "a FAILING external-access status must be an OPEN known live issue"
+
+  next_due = datetime.date.fromisoformat(attestation["next_due_date"])
```
caohy1988, line 741: **P2 — the monthly end-to-end canary still is not enforced.**

`next_due_date` is checked only against an observation date, never against `today`, and no workflow consumes it. After 2026-09-24 the current data still passes forever. In the FAILING state the external canary has no observation date at all, so refreshing only the weaker Permissions API check can advance this deadline without rerunning the end-to-end canary.

Recommendation: record `last_observed_date` for every canary attempt, pass or fail; base recurrence on that end-to-end observation; and add a scheduled check that alerts or fails once `today > next_due_date`. If a wall-clock assertion is intentionally excluded from normal PR tests, the scheduled consumer must land with this contract before claiming the monthly cadence is enforced.

dashboard/looker_studio/bindings/report_template.yaml
```diff
@@ -38,7 +38,57 @@ live_template_verification:
...
+# an end-to-end copy canary from a signed-in, non-owner, out-of-domain
+# identity clears the whole path the configurator depends on.
+external_access_verification:
+  controls:
+    - method: permissions_api_link_role_check
+      protocol: >-
+        Authenticated GET
+        https://datastudio.googleapis.com/v1/assets/{report_id}/permissions
+        must list role LINK_VIEWER with member allUsers. Call it from a
+        Google Workspace or Cloud Identity account holding the datastudio
```
caohy1988: **P2 — use least privilege and do not dismiss `PERMISSION_DENIED` as non-regression.**

The read-only `Permissions:get` endpoint accepts `https://www.googleapis.com/auth/datastudio.readonly` as well as the broader write-capable `datastudio` scope, and Google recommends the restrictive scope when asset management is unnecessary. The endpoint also requires sufficient permission to view the asset’s permissions, so `PERMISSION_DENIED` can represent lost asset authorization; it is not categorically just a caller constraint.

Recommendation: prescribe `datastudio.readonly`, and treat a denial as indeterminate until the principal, organization authorization, scope, and expected asset permission are verified. References: https://developers.google.com/data-studio/integrate/api/reference/permissions/get and https://developers.google.com/data-studio/integrate/api

tests/test_looker_studio_dashboard.py
caohy1988: Fixed in de8b4f8. The canary control now records last_observed_date on every run (currently 2026-08-24, the reported failure), a PASSING verification date must equal that observation date, and the contract requires it to be strictly newer than every recorded incident's reported_date for the tracked issue. Your scenario (PASSED dated 2026-08-11..23 with the issue closed) is negative-tested and rejected.

tests/test_looker_studio_dashboard.py
caohy1988: Fixed in de8b4f8, both halves. (1) next_due_date is now anchored exclusively to the canary's own last_observed_date — refreshing the Permissions API read alone cannot advance the deadline (negative-tested). (2) The scheduled consumer lands with this contract: .github/workflows/external-access-staleness.yml runs scripts/check_external_access_staleness.py weekly and on dispatch; once today > next_due_date it opens or bumps a tracking issue and fails the run — no credentialed API involved. Unit tests pin the script's verdict on both sides of the deadline and the workflow wiring; PR-blocking tests still never compare dates to the wall clock.

dashboard/looker_studio/bindings/report_template.yaml
caohy1988: Fixed in de8b4f8. The protocol now prescribes the least-privilege datastudio.readonly scope and treats PERMISSION_DENIED as indeterminate — caller constraint (non-org account, missing scope) or lost asset authorization — requiring the principal, organization authorization, and scope to be verified before reading a denial as either. This also corrects my earlier top-level comment, which overstated denial as categorically a caller constraint.

caohy1988:
Round-2 findings addressed in de8b4f8 (29/29 tests pass incl. node contract + browser smoke; pyink 25.x clean). Beyond the three inline threads:

- **Ask (reproducible API read)** — now in-repo: the yaml records the verbatim command and redacted response (`{"LINK_VIEWER": {"members": ["allUsers"]}}`) as a comment beside the `permissions_api_link_role_check` control, using the readonly-scope token per the corrected protocol.
- **[P3] canary failure re-runs had nowhere to land** — the canary control now carries `last_observed_date` set on every run, pass or fail (currently 2026-08-24), and it is the sole anchor for `next_due_date`, so your 2026-09-10-refail scenario records properly and re-anchors the deadline.
- **[Nit] index.html reflow** — rewrapped.
- **[P3] PR body** — updated: privacy-safe reporting wording, the full current state contract (success-evidence, incident-postdating, canary-anchored cadence, no-future-dates), and the scheduled staleness workflow.

New in this revision, per the "scheduled consumer must land with this contract" condition: `.github/workflows/external-access-staleness.yml` (weekly + manual dispatch, least-privilege `contents: read` / `issues: write`, pinned action SHAs, `persist-credentials: false`) runs `scripts/check_external_access_staleness.py`; past `next_due_date` it opens or bumps a tracking issue and fails the scheduled run. The script takes `--today` for deterministic tests; unit tests pin both verdicts and the workflow wiring. PR-blocking tests still never compare attestation dates to the wall clock.

dashboard/looker_studio/bindings/report_template.yaml
```diff
@@ -38,7 +38,70 @@ live_template_verification:
...
+        Treat PERMISSION_DENIED as indeterminate: it can mean a caller
+        constraint (non-org account, missing scope) or lost asset
+        authorization, so verify the principal, its organization
+        authorization, and the scope before reading a denial as either.
+      limitation: does_not_expose_viewer_copy_disable_control
+      last_observed_date: "2026-08-25"
+      last_result: LINK_VIEWER_ALLUSERS_PRESENT
+    - method: external_identity_link_access_check
+      protocol: >-
+        From a signed-in, non-owner, out-of-domain Google account, open
```
caohy1988: **P2 — the canary can confuse recipient policy with an owner-side outage.**

A Workspace/Cloud Identity administrator can prevent users from opening Data Studio assets owned by untrusted external domains. This protocol requires only a non-owner, out-of-domain account, so a fully public, copy-enabled template can still produce `FAILURE_REPORTED` solely because of the canary account's organization. A single unscoped pass also cannot establish the current `external_non_owner_identities` scope for both personal and managed accounts.

Require a dedicated personal canary or record that the managed canary is allowed to receive assets from untrusted domains and has no explicit report invitation; scope the result by identity class (or run both classes). The public docs should retain recipient organization policy as a possible cause instead of saying a persistent failure is necessarily "on the owner's side." Google documents the recipient-side control here: https://knowledge.workspace.google.com/admin/data-studio/set-sharing-permissions

tests/test_looker_studio_dashboard.py
```diff
@@ -532,6 +598,177 @@ def test_report_and_web_bindings_cannot_drift():
...
+
+  api_check = controls["permissions_api_link_role_check"]
+  assert (
+      api_check["limitation"] == "does_not_expose_viewer_copy_disable_control"
+  )
+  assert api_check["last_result"] in {
+      "LINK_VIEWER_ALLUSERS_PRESENT",
+      "LINK_VIEWER_ALLUSERS_ABSENT",
+  }
+  api_observed = datetime.date.fromisoformat(api_check["last_observed_date"])
```
caohy1988, line 663: **P2 — `PASSING` accepts arbitrarily stale link-role evidence.**

`api_observed` is checked only for being non-future. A later run can keep an old `LINK_VIEWER_ALLUSERS_PRESENT`, record a successful canary from an explicitly invited account after `allUsers` was removed, close the incident, and satisfy every `PASSING` predicate while public link access is actually broken.

Require the successful Permissions API observation to be from the same attestation run/date (or a tightly bounded window) as the passing canary and no earlier than the latest publication/incident. Also require that the canary identity has no direct report-level grant.

tests/test_looker_studio_dashboard.py
```diff
@@ -532,6 +598,177 @@ def test_report_and_web_bindings_cannot_drift():
...
+    assert canary["last_result"] == "PASSED"
+    assert api_check["last_result"] == "LINK_VIEWER_ALLUSERS_PRESENT"
+    verified = datetime.date.fromisoformat(canary["link_access_verified_date"])
+    assert verified == canary_observed, (
+        "a PASSING canary's verification date is its observation date —"
+        " they cannot diverge"
+    )
+    published = datetime.date.fromisoformat(report["published_date"])
+    assert verified >= published
+    for entry in tracked_entries:
```
caohy1988, line 723: **P2 — make incident history durable and precise enough to order recovery.**

If a resolution removes #445 from `known_live_issues`, `tracked_entries` is empty: this loop is vacuous and `not open_tracked` is true, so a pre-outage `2026-08-11` canary can become `PASSING` after the exact snapshot is updated. Conversely, if the record is retained, `verified > incident` rejects a legitimate recovery later on the same calendar day because both fields have date-only precision.

Require a matching retained incident/history record in both states and transition it to an explicit resolved status. Store incident and canary timestamps (or another ordered recovery marker) before requiring strict postdating; date-only values cannot prove within-day order.

.github/workflows/external-access-staleness.yml
```diff
@@ -0,0 +1,78 @@
...
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+        with:
+          persist-credentials: false
+
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
+        with:
+          python-version: "3.12"
+
+      - name: Install PyYAML
+        run: pip install "pyyaml>=6.0"
```
caohy1988: **P2 — do not resolve a mutable PyPI dependency inside an issue-writing job.**

This job grants `issues: write`, then installs any future release satisfying `pyyaml>=6.0`. Installation code runs on the same persistent runner before the later step exposes `GH_TOKEN`; a compromised release can persist a shim or alter a later-executed file and gain that write authority. The repository's release workflow already demonstrates the safer pattern: hash-locked dependencies in a read-only build job, with mutation isolated in a minimal write job.

Use an exact hash-locked dependency (or remove the runtime dependency), and preferably split parsing/checking into a `contents: read` job whose small downstream mutation job downloads no packages.

tests/test_looker_studio_dashboard.py
```diff
@@ -532,6 +598,177 @@ def test_report_and_web_bindings_cannot_drift():
...
+  )
+  assert code == 1
+  assert "OVERDUE" in message
+  assert "external_identity_link_access_check" in message
+  assert report["external_access_verification"]["tracking_issue"] in message
+
+  workflow = (
+      ROOT / ".github" / "workflows" / "external-access-staleness.yml"
+  ).read_text()
+  assert "schedule" in workflow
```
caohy1988: **P3 — these substring assertions do not pin the scheduled consumer.**

A workflow containing only comments such as `# schedule` and `# scripts/check_external_access_staleness.py`, or a disabled `jobs: {}`, still passes this test. That defeats the test's stated purpose of protecting the wall-clock half of the contract.

Parse the workflow as YAML 1.2 (or run `actionlint`) and assert an active schedule, an executable step invoking the script, and the overdue issue/failure path structurally.

tests/test_looker_studio_dashboard.py
```diff
@@ -532,6 +598,177 @@ def test_report_and_web_bindings_cannot_drift():
...
+
+def test_docs_name_the_terminal_report_not_shared_dialog():
+  """#445 acceptance: every user-facing surface names the terminal denial.
+
+  The configurator and both manuals must quote the dialog, attribute it to
+  the shared template's access (not the user's setup), and must not fold it
+  into the wait-it-out guidance written for the #398 provisioning flicker.
+  """
+  fragments = ("This report isn", "shared with you")
+  surfaces = (
```
caohy1988, line 852: **P3 — the cross-surface test does not preserve the privacy or terminal-error guidance.**

It checks only two dialog fragments. A manual can reintroduce "post the selected account email" or "keep retrying" and still pass, while the browser test does not read either Markdown manual.

For each public surface, assert the no-wait/terminal guidance and the prohibition on posting account identifiers; also reject the prior unsafe selected-account wording.

dashboard/looker_studio/docs/index.html
```diff
@@ -192,6 +197,27 @@ <h3>Save, then secure sharing</h3>
         </article>
       </section>
 
+      <aside class="notice notice-warning" id="report-not-shared">
```
caohy1988, line 200: **P3 — the new critical recovery guidance renders as ~13px body text.**

This long warning inherits `.notice { font-size: 0.82rem; }` from `styles.css:293-300`. The most important failure diagnosis and privacy/reporting instructions are therefore below the review checklist's 16px minimum for body text.

Give `#report-not-shared` (or critical notices generally) a body size of at least `1rem` while preserving the warning treatment.

caohy1988 (commented):
Fresh full review of `de8b4f8`.

I re-read the complete 555-line diff and all changed files, audited the current #445 evidence and review threads, checked the recipient-policy claim against current Google Workspace documentation, and ran independent testing, maintainability, security, performance, design, red-team, and adversarial passes.

The three P2 findings from the previous round are fixed: recovery now postdates retained incidents, recurrence is anchored to the end-to-end canary and has a scheduled consumer, and the OAuth guidance now uses `datastudio.readonly` and treats denial as indeterminate.

This pass found four P2 issues and three P3 issues inline. The P2s should be addressed before merge. The account is also the PR author, so GitHub only permits a COMMENTED review rather than a formal request-changes review.

Verification: 29 focused pytest tests passed; the Node configurator contract and syntax checks passed; `git diff --check`, pyink<26, and isort<8 passed; and the staleness CLI returned 0 on the due date and 1 the day after. GitHub's Python 3.10–3.14 matrix, configurator smoke, CLA, and security checks are green. The Format check remains the unrelated repo-wide pyink drift tracked by #447.

dashboard/looker_studio/bindings/report_template.yaml
caohy1988: Fixed in 36b4bd1. The canary protocol now requires an account with no direct grant that is either personal or a managed account with recorded permissive Workspace policy, and records last_identity_class on every run (allowed: personal / managed_verified_policy / unknown — the 2026-08-24 reporter's class was not captured, recorded as unknown with a note that this is itself part of the non-isolation). The contract rejects PASSING with class unknown. cause_isolation adds recipient-side Workspace policy as a fourth credible cause, and all three public docs now say a managed account's own org policy can block a fully public template ('try a personal account'), instead of calling a persistent failure necessarily owner-side.

tests/test_looker_studio_dashboard.py
caohy1988: Fixed in 36b4bd1. PASSING now requires the Permissions API observation to carry the same date as the passing canary (same attestation run), so a stale LINK_VIEWER read cannot vouch for a fresh copy — negative-tested with your scenario (API read 2026-08-25 + canary 2026-09-01 → rejected). The canary protocol also now requires the account to hold no direct grant on the report; that half is a protocol requirement the yaml cannot self-verify, so it lives in the recorded procedure.

tests/test_looker_studio_dashboard.py
caohy1988: Fixed in 36b4bd1, both halves. Durability: the contract requires at least one retained known_live_issues entry for the tracked issue in BOTH states — entries transition to status RESOLVED, never deleted — so the vacuous-loop path is closed (history-deleted mutation → rejected). Ordering: a RESOLVED entry must carry resolution_observed_date equal to the passing canary's date and >= reported_date; the explicit recorded marker carries the order claim instead of the unsound strict date inequality, so legitimate same-day recovery (incident and canary both 2026-08-24) now passes — negative- and positive-tested.

.github/workflows/external-access-staleness.yml
caohy1988: Fixed in 36b4bd1 by removing the dependency and splitting the jobs. The script is now stdlib-only (anchored line-pattern extraction; the unit test cross-checks it against a real YAML parse so restructuring fails CI rather than misparsing the scheduler). The workflow is split per the release-workflow pattern: a contents:read check job that installs nothing, and an issues:write report job gated on its overdue output that performs no checkout and downloads no packages — nothing freshly resolved ever executes in a job holding write authority.

tests/test_looker_studio_dashboard.py
caohy1988: Fixed in 36b4bd1. The test now parses the workflow structurally (via yaml.safe_load, handling PyYAML's YAML-1.1 'on:'→True quirk): active cron schedule present, the check job holds exactly contents:read and has an executable run step invoking the script with no pip install, and the report job holds exactly issues:write, needs/gates on the overdue output, contains only run steps (no actions, so no checkout or downloads), and its overdue path both creates the issue and exits 1. A comments-only or jobs:{} workflow no longer passes.

tests/test_looker_studio_dashboard.py
caohy1988: Fixed in 36b4bd1. Per surface (index.html, app.mjs, README, USER_MANUAL) the test now asserts the no-wait/terminal guidance (tolerant regex over the normalized text, including JS string-concat breaks) and rejects the prior unsafe 'account the dialog selected' wording; the three reporting surfaces must additionally carry the do-not-post-email prohibition and the personal-vs-organization ask. Reintroducing 'keep retrying' by deleting the guidance, or the selected-account wording, now fails CI.

dashboard/looker_studio/docs/index.html
caohy1988: Fixed in 36b4bd1. #report-not-shared gets font-size 1rem / line-height 1.6 (16px at the root size) while keeping the notice-warning treatment, and the docs test pins the rule so the explainer cannot silently fall back to the 0.82rem fine print.

caohy1988:
Round-3 findings all addressed in 36b4bd1 — the four P2s and three P3s each have a fix-reply in their thread. Highlights:

- **Recipient policy (P2)** — canary protocol scoped by identity class with `last_identity_class` recorded on every run (the 2026-08-24 reporter's class is honestly `unknown`); PASSING requires a policy-clearing class; all public docs now name recipient-side Workspace policy and suggest trying a personal account instead of calling a persistent failure necessarily owner-side; `cause_isolation` gains it as a fourth credible cause.
- **Evidence freshness (P2)** — PASSING requires the Permissions API observation to share the passing canary's date; the protocol requires a no-direct-grant canary account.
- **Durable, ordered history (P2)** — at least one retained `known_live_issues` entry in both states (RESOLVED, never deleted), with `resolution_observed_date` as the explicit recovery marker — this also un-breaks legitimate same-day recovery that the old strict date inequality rejected.
- **Supply chain (P2)** — the staleness script is stdlib-only (extraction cross-checked against a real YAML parse in tests) and the workflow is split: `contents: read` check job installing nothing; `issues: write` report job with no checkout and no downloads.
- **P3s** — the workflow test is structural (parsed YAML: active cron, script invocation, no `pip install`, write-job gating and failure path); the docs test asserts per-surface no-wait guidance, the identifier-posting prohibition, and rejects the prior selected-account wording; the explainer renders at 1rem with the rule pinned.

Verification: 29/29 pytest (node contract + real-Chrome smoke included), pyink 25.x format-stable, the staleness CLI flips exactly at the deadline, and the contract was negative-tested against all four attack mutations (stale-API vouching, history erasure, marker-less/pre-incident resolution, unknown identity class) plus positive-tested for the two legitimate recovery shapes (same-day and later, with matched API read).

haiyuan-eng-google (approved):

tests/test_looker_studio_dashboard.py
```diff
@@ -532,6 +607,292 @@ def test_report_and_web_bindings_cannot_drift():
...
+      "scripts/check_external_access_staleness.py" in run for run in check_runs
+  ), "the read-only job must execute the staleness script"
+  assert not any(
+      "pip install" in run for run in check_runs
+  ), "the scheduled jobs must not resolve mutable package dependencies"
+
+  report_job = workflow["jobs"]["report"]
+  assert report_job["permissions"] == {"issues": "write"}
+  assert report_job["needs"] == "check"
+  assert "overdue" in report_job["if"]
```
caohy1988, line 820: **P3 (non-blocking)** — this regression test still permits both failures it is meant to prevent. `assert "overdue" in report_job["if"]` accepts an inverted gate such as `overdue != 'true'`, and the dependency check scans only `check_runs`; adding `pip install ...` to a `run` step in the `issues: write` job still satisfies the steps-vs-runs assertion. I replayed the current assertions against both mutations and both passed. Please assert the exact overdue predicate and explicitly reject dependency/download commands in `report_runs` (or cover the write step through an equivalent structural check).

.github/workflows/external-access-staleness.yml
```diff
@@ -0,0 +1,88 @@
...
+      - name: Open or bump the tracking issue
+        env:
+          GH_TOKEN: ${{ github.token }}
+          MESSAGE: ${{ needs.check.outputs.message }}
+        run: |
+          title="External access attestation overdue (report_template.yaml)"
+          body="$MESSAGE
+
+          Protocol: dashboard/looker_studio/bindings/report_template.yaml -> external_access_verification. Context: #445."
+          existing=$(gh issue list --repo "$GITHUB_REPOSITORY" --state open \
```
caohy1988, line 80: **P3 (non-blocking)** — overlapping scheduled/manual runs can race this check-then-create sequence. With no workflow or job `concurrency` key, two report jobs can both observe no matching issue and then both create one. Add a workflow-level concurrency group with `cancel-in-progress: false` (matching the repository pattern in `release-tracing.yml`) so issue reconciliation is serialized.

caohy1988 (commented):
Final full review of current head `24df3f7` (including the merge from `main`). I re-read the complete PR diff, audited all prior findings and replies, and verified the head delta is only the #447 pyink pin. The seven round-3 findings are fixed and their threads are now resolved.

No P1/P2 or merge-blocking correctness issues remain. I left two non-blocking P3 comments: one test-contract gap that permits an inverted report gate or package download in the write job, and one check-then-create race that can duplicate the tracking issue under overlapping workflow runs. Both are suitable follow-ups.

Verification: 29 focused pytest tests passed (including real-Chrome smoke); Node configurator syntax/contracts, the staleness due-date boundary, `git diff --check`, full-tree pyink<26 (290 files), and isort passed. GitHub now reports Format, Python 3.10-3.14, browser smoke, build, CLA, scan-pr, and zizmor all green. The PR is mergeable/CLEAN and has collaborator approval.

Recommendation: merge-ready; track the two P3 hardening items as follow-ups.

Issue #448, opened 2026-08-27
[Dashboard] Configurator: collapse the three identifier fields into a single table-ID input with full paste support


caohy1988:
## Summary

User feedback on the published configurator (https://googlecloudplatform.github.io/BigQuery-Agent-Analytics-SDK/): the input layer asks for three identifiers — **Project ID**, **Dataset ID**, **BQAA table ID** — but a fully qualified table ID (`project.dataset.table`) already contains all three. The three entrances should collapse into **one input for the fully qualified BQAA table ID**, keeping every paste format the current parser already understands.

The parsing intelligence exists today: `parseTableReference` / `parseTableReferenceForInput` in `dashboard/looker_studio/docs/configurator.mjs` already split a paste into the three fields (`handleQualifiedTableId` in `app.mjs`). This issue makes that the primary — and only — entrance instead of a convenience layered on three separate fields. It intentionally supersedes #403's decision to retain separate fields; the reason behind that choice (precise per-segment diagnostics) is preserved through segment-level error reporting in the single field.

**Terminology used throughout:** *fully qualified BQAA table ID* is the combined `project.dataset.table` value; *table segment* is only its final component.

## Current behavior

Three required fields, each individually validated; the table field is prefilled with `agent_events`. Pasting any of the supported forms into **any** field fills all three (`USER_MANUAL.md` → Step 1):

| Paste form | Example |
|---|---|
| Fully qualified table ID | `my-project.my_dataset.agent_events` |
| Backticked and/or trailing `;` or `,` (from a SQL editor or ID list) | `` `my-project.my_dataset.agent_events`; `` / `my-project.my_dataset.agent_events,` |
| Legacy colon form | `my-project:my_dataset.agent_events` |
| BigQuery Console table link | `https://console.cloud.google.com/bigquery?ws=…` (accepted only when it names exactly one table; both currently supported console hosts) |

(`splitQualifiedTableId` strips both `;` and `,` terminators today, and the Node contract carries a comma-terminated fixture — all of these forms must survive the redesign.)

## Requested change

1. **One primary input: "Fully qualified BQAA table ID"** accepting `project.dataset.table`, with all paste forms above continuing to work (paste → parse → normalized dotted ID shown, status line `Ready for project.dataset.table`).
[...]

PR #449, opened 2026-08-27
feat(looker-studio): collapse the configurator to a single fully-qualified table-ID input

caohy1988:
Implements #448 against its three-round review contract (decisions 1–8, twelve acceptance criteria). One entrance — the **fully qualified BQAA table ID** (`project.dataset.table`) — replaces the three identifier fields, with every previously supported paste form intact.

## Acceptance criteria — status

1. **Single entrance, no separate project/dataset fields** — done; the Python page pin asserts `id="table-id"` exists and `id="project"` / `id="dataset"` do not. Advanced billing override unchanged.
2. **All paste forms + both Console hosts (positive)** — the node suite drives all five dotted forms (comma fixture retained) and four Console-URL variants including Pantheon through the field; each normalizes to the dotted ID and reaches Ready.
3. **Rejection matrix (separate)** — ambiguous/unsupported links and wrong-arity input covered at both the validator and DOM layers.
4. **Fail-closed mutation** — editing a Ready value immediately revokes the derived triple, disables Create/Copy, and clears the Ready announcement (error presentation defers per the trigger contract); the `valid → partial → invalid → corrected` interaction test asserts all three including status behavior.
5. **Two error classes** — whole-field (`segment: null`) vs segment-level (segment named inline, e.g. "Project segment: …"), both retaining raw input, clearing derived state, disabling actions; **sentinel collisions now gate entry to Ready** and are attributed to their segment (`rejectSentinelCollisions` throws a `ConfigurationError` with `field: "tableId"` and `segment` — logic and Linking API output unchanged).
6. **Billing actionability** — `valid table → invalid billing → corrected billing` test passes with the table triple retained throughout; blank override bills the ID's project segment.
7. **Setup-link compatibility** — existing `?project=&dataset=&table=` links prefill the field and validate immediately (tested via a second module instance); regenerated links keep the format and round-trip byte-for-byte.
8. **Byte-identical Linking API URLs** — `buildDashboardUrl` / `buildSetupUrl` / `validateConfiguration` are untouched; the node suite's pinned `sqlReplace` / `billingProjectId` literals pass unchanged for blank and explicit overrides.
9. **Accessibility** — label, `aria-describedby` hint/error associations, `aria-invalid` management, `aria-live` error/status regions preserved on the combined field; validation timing per decision 1.
10. **Browser smoke** — the app-executed proof is now a runtime `data-bqaa-app-initialized` marker (asserted absent from static HTML as a tag attribute), plus an error-free pristine first-load assertion (no `aria-invalid`, Create disabled). All negative fixtures updated to the new healthy baseline (marker set by script, disabled action, no `aria-invalid`) so each injected fault is the sole failure cause, and a sixth **missing-initialization** fixture fails as required.
11. **Suites green** — node contract (rewritten for the single-field DOM), `browser_smoke.sh` main + `--self-test` (6/6 fixtures), and the Python dashboard suite (29/29) all pass locally; pyink/isort clean.
12. **Docs** — `USER_MANUAL.md` (Before-you-start, Step 1, troubleshooting rows), `README.md` (configurator section), `docs/dashboard-implementation.md` (data-source contract now describes the combined field and both error classes), and the page copy (field hint, step 01) updated.

Supersedes #403's separate-fields decision while preserving its rationale (precise diagnostics) through segment-level attribution, per the issue record.

### DIFF
--- a/dashboard/looker_studio/docs/app.mjs
+++ b/dashboard/looker_studio/docs/app.mjs
@@ -157,23 +157,27 @@ tableIdInput.addEventListener("change", () => {
 tableIdInput.addEventListener("paste", (event) => {
   const text = event.clipboardData.getData("text");
   revealTableErrors = true;
-  // Normalize the displayed value only after FULL validation succeeds; any
-  // invalid paste — unparseable, bad segment, or sentinel collision — must
-  // land as the exact clipboard text so the raw invalid input is retained
-  // alongside its immediate error (#448 decision 2, #449 review).
+  // The default paste inserts at the selection, so an invalid fragment
+  // could merge with a previous Ready value into a DIFFERENT valid ID and
+  // silently retarget the dashboard (#449 review). Always take over: the
+  // field becomes either the normalized valid ID or the complete raw
+  // clipboard text — never a splice — and validates immediately.
+  event.preventDefault();
   let parsed = null;
   try {
     parsed = validateQualifiedTableId(text);
   } catch {
     parsed = null;
   }
-  if (parsed) {
-    event.preventDefault();
-    tableIdInput.value = `${parsed.project}.${parsed.dataset}.${parsed.table}`;
-    refresh();
+  tableIdInput.value = parsed
+    ? `${parsed.project}.${parsed.dataset}.${parsed.table}`
+    : text;
+  if (tableIdInput.value !== lastValidRaw) {
+    derived = null;
+    disableActions();
+    setStatus("");
   }
-  // An invalid paste lands in the field; its input event validates it
-  // immediately because revealTableErrors is already set.
+  refresh();
 });
 
 billingInput.addEventListener("input", refresh);
@@ -193,6 +197,10 @@ function attemptCreate() {
   revealTableErrors = true;
   refresh();
   if (createLink.href) {
+    // Invalidate any pending clipboard completion: the provisioning
+    // warning must not be overwritten by an older copy resolving late
+    // (#449 review).
+    statusEpoch += 1;
     setStatus(WAITING_MESSAGE, "waiting");
     window.open(createLink.href, "_blank", "noopener,noreferrer");
   }
@@ -205,10 +213,20 @@ form.addEventListener("submit", (event) => {
 
 for (const input of [tableIdInput, billingInput]) {
   input.addEventListener("keydown", (event) => {
-    if (event.key === "Enter") {
-      event.preventDefault();
-      attemptCreate();
+    // Only a discrete, non-composing Enter is an attempted action: a held
+    // key repeat must not open tab after tab, and Enter committing an IME
+    // composition (isComposing, or legacy keyCode 229) is text entry, not
+    // activation (#449 review).
+    if (
+      event.key !== "Enter" ||
+      event.repeat ||
+      event.isComposing ||
+      event.keyCode === 229
+    ) {
+      return;
     }
+    event.preventDefault();
+    attemptCreate();
   });
 }
 
@@ -219,6 +237,7 @@ createLink.addEventListener("click", (event) => {
     refresh();
     return;
   }
+  statusEpoch += 1;
   setStatus(WAITING_MESSAGE, "waiting");
 });
 
--- a/dashboard/looker_studio/docs/app.mjs
+++ b/dashboard/looker_studio/docs/app.mjs
@@ -3,7 +3,6 @@ import {
   PROJECT_RE,
   buildDashboardUrl,
   buildSetupUrl,
-  parseTableReference,
   validateQualifiedTableId,
 } from "./configurator.mjs";
 
@@ -15,6 +14,7 @@ const checklist = document.querySelector("#security-checklist");
 const status = document.querySelector("#form-status");
 const tableIdInput = document.querySelector("#table-id");
 const tableIdError = document.querySelector("#table-id-error");
+const advancedSettings = document.querySelector("#advanced-settings");
 const billingInput = document.querySelector("#billing-project");
 const billingError = document.querySelector("#billing-project-error");
 
@@ -27,6 +27,10 @@ const billingError = document.querySelector("#billing-project-error");
 let derived = null;
 let lastValidRaw = null;
 let revealTableErrors = false;
+// Bumped on every state mutation and at the start of every async copy, so a
+// clipboard completion that is no longer current cannot restore a stale
+// status over what the user's later edit cleared (#449 review).
+let statusEpoch = 0;
 
 function setStatus(message, kind = "") {
   status.textContent = message;
@@ -57,12 +61,23 @@ function clearBillingError() {
 function showBillingError(message) {
   billingInput.setAttribute("aria-invalid", "true");
   billingError.textContent = message;
+  // The override lives inside the Advanced disclosure: an error written
+  // into a closed <details> would leave both actions disabled with no
+  // visible explanation, so surface it. Correction never auto-closes.
+  advancedSettings.open = true;
+}
+
+// The single normalization both validation and URL construction consume: a
+// whitespace-only override must behave exactly like blank (billing the
+// project segment), never reach validateConfiguration as truthy text.
+function billingOverride() {
+  return billingInput.value.trim();
 }
 
 // A blank override bills the project segment of the fully qualified ID, so
 // blank is always valid; anything else must be a project ID.
 function billingOverrideIsValid() {
-  const value = billingInput.value.trim();
+  const value = billingOverride();
   if (!value || PROJECT_RE.test(value)) {
     clearBillingError();
     return true;
@@ -72,6 +87,7 @@ function billingOverrideIsValid() {
 }
 
 function refresh() {
+  statusEpoch += 1;
   let parsed;
   try {
     parsed = validateQualifiedTableId(tableIdInput.value);
@@ -102,7 +118,7 @@ function refresh() {
   try {
     createLink.href = buildDashboardUrl({
       ...derived,
-      billingProject: billingInput.value,
+      billingProject: billingOverride(),
     });
     createLink.removeAttribute("aria-disabled");
     copyButton.disabled = false;
@@ -141,13 +157,22 @@ tableIdInput.addEventListener("change", () => {
 tableIdInput.addEventListener("paste", (event) => {
   const text = event.clipboardData.getData("text");
   revealTableErrors = true;
-  const parsed = parseTableReference(text);
+  // Normalize the displayed value only after FULL validation succeeds; any
+  // invalid paste — unparseable, bad segment, or sentinel collision — must
+  // land as the exact clipboard text so the raw invalid input is retained
+  // alongside its immediate error (#448 decision 2, #449 review).
+  let parsed = null;
+  try {
+    parsed = validateQualifiedTableId(text);
+  } catch {
+    parsed = null;
+  }
   if (parsed) {
     event.preventDefault();
     tableIdInput.value = `${parsed.project}.${parsed.dataset}.${parsed.table}`;
     refresh();
   }
-  // An unparseable paste lands in the field; its input event validates it
+  // An invalid paste lands in the field; its input event validates it
   // immediately because revealTableErrors is already set.
 });
 
@@ -159,15 +184,33 @@ const WAITING_MESSAGE =
   "One exception: “This report isn’t shared with you” will not resolve by " +
   "waiting — see the note under the Create button.";
 
-form.addEventListener("submit", (event) => {
-  event.preventDefault();
+// The attempted-action path (#448 decision 1): reveal deferred errors,
+// revalidate, and open exactly one tab when actionable. Reached by the
+// form's submit event and — because this form has two text inputs and no
+// native submit control, so browsers never run implicit submission — by an
+// explicit Enter bridge on both fields (#449 review, P1).
+function attemptCreate() {
   revealTableErrors = true;
   refresh();
   if (createLink.href) {
     setStatus(WAITING_MESSAGE, "waiting");
     window.open(createLink.href, "_blank", "noopener,noreferrer");
   }
+}
+
+form.addEventListener("submit", (event) => {
+  event.preventDefault();
+  attemptCreate();
 });
+
+for (const input of [tableIdInput, billingInput]) {
+  input.addEventListener("keydown", (event) => {
+    if (event.key === "Enter") {
+      event.preventDefault();
+      attemptCreate();
+    }
+  });
+}
 
 createLink.addEventListener("click", (event) => {
   if (!createLink.href) {
@@ -183,27 +226,37 @@ copyButton.addEventListener("click", async () => {
   if (!derived) {
     return;
   }
+  const epoch = ++statusEpoch;
   try {
     const setupUrl = buildSetupUrl(
-      { ...derived, billingProject: billingInput.value },
+      { ...derived, billingProject: billingOverride() },
       window.location.href,
     );
     await navigator.clipboard.writeText(setupUrl);
-    setStatus("Setup link copied. It contains identifiers, never credentials.", "ready");
+    if (epoch === statusEpoch) {
+      setStatus("Setup link copied. It contains identifiers, never credentials.", "ready");
+    }
   } catch (error) {
-    setStatus(error.message || "Could not copy the setup link.", "error");
+    if (epoch === statusEpoch) {
+      setStatus(error.message || "Could not copy the setup link.", "error");
+    }
   }
 });
 
 checklistButton.addEventListener("click", async () => {
+  const epoch = ++statusEpoch;
   const text = [...checklist.querySelectorAll("li")]
     .map((item, index) => `${index + 1}. ${item.textContent.trim()}`)
     .join("\n");
   try {
     await navigator.clipboard.writeText(text);
-    setStatus("Security checklist copied.", "ready");
+    if (epoch === statusEpoch) {
+      setStatus("Security checklist copied.", "ready");
+    }
   } catch {
-    setStatus("Could not copy the security checklist.", "error");
+    if (epoch === statusEpoch) {
+      setStatus("Could not copy the security checklist.", "error");
+    }
   }
 });
 
--- a/dashboard/looker_studio/docs/app.mjs
+++ b/dashboard/looker_studio/docs/app.mjs
@@ -1,11 +1,11 @@
 import {
+  BILLING_PROJECT_MESSAGE,
+  PROJECT_RE,
   buildDashboardUrl,
   buildSetupUrl,
-  validateConfiguration,
   parseTableReference,
-  parseTableReferenceForInput,
+  validateQualifiedTableId,
 } from "./configurator.mjs";
-import { REPORT_CONFIG } from "./report-config.mjs";
 
 const form = document.querySelector("#configurator");
 const createLink = document.querySelector("#create-dashboard");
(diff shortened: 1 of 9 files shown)


dashboard/looker_studio/docs/index.html
```diff
@@ -57,54 +57,25 @@ <h2>Connect your BQAA data</h2>
             used.
           </p>
 
-          <label for="project">Project ID</label>
+          <label for="table-id">Fully qualified BQAA table ID</label>
```
caohy1988, line 60: **[P1] Enter does not activate the form.**

#448 defines Enter in the combined field as an observable attempted action, but this form has two text inputs and no submit control or key bridge. A trusted Chrome `Input.dispatchKeyEvent` probe confirmed the failure: invalid Enter left the error/status empty, and valid Enter produced zero `window.open` calls. The Node test bypasses browser behavior by invoking the stored `submit` listener directly.

Route Enter through a real submit control or `form.requestSubmit()`, then add a real-browser invalid/valid Enter test that asserts validation and exactly one open.

dashboard/looker_studio/docs/app.mjs
```diff
@@ -146,15 +172,22 @@ form.addEventListener("submit", (event) => {
...
   setStatus(WAITING_MESSAGE, "waiting");
 });
 
 copyButton.addEventListener("click", async () => {
+  if (!derived) {
+    return;
+  }
   try {
-    const setupUrl = buildSetupUrl(currentValues(), window.location.href);
+    const setupUrl = buildSetupUrl(
```
caohy1988, line 250: **[P2] Clipboard completion can restore stale status.**

If `writeText()` remains pending while the user edits the table/billing value, this continuation unconditionally replaces the cleared status with `ready` for the old table. The setup-link and checklist copy operations can also complete out of order and overwrite each other in the shared status node.

Track a configuration/copy revision across each `await` and update status only if the completion is still current. Add deferred-promise tests for copy -> edit and reversed copy completions.

dashboard/looker_studio/docs/app.mjs
```diff
@@ -1,132 +1,157 @@
...
+
+function showBillingError(message) {
+  billingInput.setAttribute("aria-invalid", "true");
+  billingError.textContent = message;
+}
+
+// A blank override bills the project segment of the fully qualified ID, so
+// blank is always valid; anything else must be a project ID.
+function billingOverrideIsValid() {
+  const value = billingInput.value.trim();
```
caohy1988: **[P2] The billing validators disagree on whitespace.**

This helper trims a whitespace-only override and accepts it as blank. `refresh()` then passes the original truthy whitespace to `buildDashboardUrl()`, where `validateConfiguration()` trims it to empty and rejects it. The catch clears the parsed table and reports a form-level error while the billing field itself looks valid.

Normalize the override once before both validation and URL construction, and add a whitespace-only interaction case that behaves exactly like blank input.

dashboard/looker_studio/docs/configurator.mjs
```diff
@@ -234,3 +263,43 @@ export function parseTableReferenceForInput(value) {
...
+// its rule or a sentinel collision — `segment` names the offender). The
+// sentinel check runs here so a colliding value never reaches the Ready
+// state only to fail at URL construction.
+export function validateQualifiedTableId(value, config = REPORT_CONFIG) {
+  const raw = String(value ?? "").trim();
+  if (!raw) {
+    throw new ConfigurationError("tableId", TABLE_ID_MESSAGES.empty);
+  }
+  const parsed = URL_SHAPED_RE.test(raw)
+    ? parseBigQueryConsoleTableUrl(raw)
```
caohy1988: **[P2] Console-link segment errors collapse into the whole-link error.**

This returns `null` after extracting an otherwise unambiguous supported-host tuple when any identifier regex fails. `validateQualifiedTableId()` therefore cannot run its segment loop and reports the generic link error, contrary to #448's rule that exactly three identifiable segments get project/dataset/table attribution.

Extract the structural Console tuple in an internal helper, keep the public parser strict, and let `validateQualifiedTableId()` apply the existing per-segment checks. Add invalid project/dataset/table Console fixtures.

dashboard/looker_studio/docs/index.html
```diff
@@ -57,54 +57,25 @@ <h2>Connect your BQAA data</h2>
...
+            aria-describedby="table-id-hint table-id-error"
             required
           >
-          <span class="field-hint" id="project-hint">
-            6–30 characters: lowercase letters, digits, and hyphens. Start
-            with a letter and end with a letter or digit.
-            Paste a fully qualified table ID or BigQuery Console table link
-            into any identifier field to fill all three.
+          <span class="field-hint" id="table-id-hint">
+            <code>project.dataset.table</code> — one ID names all three.
```
caohy1988, line 70: **[P2] The hero still describes the removed three-entrance flow.**

The primary page copy still tells the user to “Enter three BigQuery identifiers,” while #448 and the field helper define one fully qualified table-ID entrance. This is the first instruction on the page and preserves the exact confusion the redesign is meant to remove.

Change it to “Enter your fully qualified BQAA table ID” (or equivalent single-entrance wording).

caohy1988 (commented):
## Fresh full review (`1bec50f`)

CI is fully green, and I also reproduced the Node contract, positive browser smoke, and all six smoke self-tests locally. This pass still found **1 P1, 6 P2, and 1 P3**.

The P1 is confirmed in a real headless Chrome interaction: a trusted Enter key on invalid input leaves the error/status empty, and Enter on a valid ID opens zero tabs. The Node test calls the submit callback directly, so it does not cover that native path.

Verdict: **not ready to merge yet**. Recommended order: fix Enter first; then the billing/clipboard state bugs; then the Console/raw-paste taxonomy, smoke fidelity, invalid-prefill disclosure, and stale hero copy. All eight findings are inline.

dashboard/looker_studio/tools/browser_smoke.sh
```diff
@@ -282,14 +309,26 @@ if ! grep -q 'data-errors="0"' "$OUT_DIR/dom.html"; then
...
+  fail "the app-initialized marker must not appear in static HTML"
+fi
+grep -q 'data-bqaa-app-initialized="true"' "$OUT_DIR/dom.html" \
+  || fail "app-initialized marker missing — app.mjs did not execute in the browser"
+# With no query parameters the first load is the pristine state (#448):
+# the empty field carries no error and both actions are disabled.
+if grep -q 'aria-invalid=' "$OUT_DIR/dom.html"; then
+  fail "pristine first load must not mark any field invalid"
+fi
+grep -q 'aria-disabled="true"' "$OUT_DIR/dom.html" \
```
caohy1988: **[P2] The browser smoke does not prove the exact pristine state it claims.**

The check never asserts that `#table-id` is empty or that `#copy-link.disabled` is true. This generic `aria-disabled` grep is already satisfied by the static Create anchor, and the DOM-bearing negative fixtures omit the table field and Copy button while claiming the full healthy baseline. The smoke can therefore stay green after those browser states regress.

Snapshot and assert the ID-specific field value, Create disabled/no-`href` state, and Copy `disabled` property. Give every DOM-bearing negative fixture those exact healthy elements and states before rerunning `--self-test`.

dashboard/looker_studio/docs/app.mjs
```diff
@@ -174,4 +207,25 @@ checklistButton.addEventListener("click", async () => {
   }
 });
 
-refresh();
+// Setup-link prefill keeps the existing three-parameter contract: all three
+// identifier parameters compose the fully qualified ID and validate
+// immediately; anything less leaves the field pristine.
+const query = new URLSearchParams(window.location.search);
+if (query.has("billingProject")) {
```
caohy1988, line 286: **[P2] Invalid billing prefill hides why both actions are disabled.**

With valid table parameters and an invalid `billingProject`, immediate validation writes the only error inside the initially closed Advanced `<details>`, clears the visible status, and disables both actions. The user sees a valid table ID and inert actions with no visible explanation.

Open the Advanced disclosure when `showBillingError()` runs (do not auto-close it on correction), and add an invalid-billing setup-link test that asserts the error is visible.

dashboard/looker_studio/docs/app.mjs
```diff
@@ -1,132 +1,157 @@
...
-    if (parsed) {
-      event.preventDefault();
-      afterQualifiedTableId(parsed);
-    }
-  });
-}
+tableIdInput.addEventListener("paste", (event) => {
+  const text = event.clipboardData.getData("text");
+  revealTableErrors = true;
+  const parsed = parseTableReference(text);
```
caohy1988: **[P3] Segment-invalid pastes lose the raw clipboard value.**

`parseTableReference()` structurally accepts wrapped or colon-form IDs before segment validation. The handler then prevents the native paste and rewrites the visible value to dotted form before `validateQualifiedTableId()` rejects it. For example, `` `BADPROJECT.dataset.table`; `` no longer remains raw, despite #448 requiring both error classes to retain raw invalid input.

Normalize the displayed value only after full validation succeeds; otherwise preserve the exact clipboard text and report the immediate error. Add a normalization-changing invalid-paste fixture.

dashboard/looker_studio/docs/index.html
caohy1988: Fixed in 635083b. Enter on both fields now routes through an explicit keydown bridge into the same attemptCreate() the submit event uses, with preventDefault so the paths never double-fire. On the requested real-browser test: the confirmed failure mode was reliance on browser-native implicit submission, which this form can never get (two text inputs, no submit control) — the fix replaces that native dependency with an explicit listener, so the new Node tests dispatching keydown exercise the identical code path (invalid: error revealed, zero opens; valid: exactly one window.open; non-Enter keys inert). The smoke harness is a dump-dom probe with no input driver; if you want a CDP-driven trusted-key test as well, I'd propose it as a follow-up rather than growing the harness in this PR.

dashboard/looker_studio/docs/app.mjs
caohy1988: Fixed in 635083b. A statusEpoch is bumped on every state mutation (refresh) and at the start of each copy operation; completions write status only when still current. Deferred-promise tests cover both scenarios: copy → edit → late completion (status stays cleared) and reversed copy/checklist completions (last-started owns the status).

dashboard/looker_studio/docs/app.mjs
caohy1988: Fixed in 635083b. A single billingOverride() normalization (trim) feeds validation, buildDashboardUrl, and buildSetupUrl, so a whitespace-only override behaves exactly like blank — Ready state, no billing error, billingProjectId = the ID's project segment — verified by a new whitespace-only interaction case.

dashboard/looker_studio/docs/configurator.mjs
caohy1988: Fixed in 635083b. An internal extractBigQueryConsoleTableReference() returns the structural tuple of an unambiguous supported-host link; validateQualifiedTableId consumes it and runs the ordinary segment loop, while the public parseBigQueryConsoleTableUrl keeps its strict validity gate (pinned: it still returns null for these URLs). New fixtures cover bad project/dataset/table Console links at the validator layer and a bad-project link at the DOM layer, each asserting segment attribution and raw-link retention.

dashboard/looker_studio/docs/index.html
caohy1988: Fixed in 635083b — the hero lede now reads 'Enter your fully qualified BQAA table ID.'

dashboard/looker_studio/tools/browser_smoke.sh
caohy1988: Fixed in 635083b. The smoke now extracts the exact tags and asserts their states: the #table-id input exists with no value= and no aria-invalid, the #create-dashboard anchor is aria-disabled with no href, and the #copy-link button carries the reflected disabled attribute — the generic aria-disabled grep (satisfiable by static markup) is gone. All four DOM-bearing fixtures (console-error, fake-chrome DOM, delayed-error, missing-initialization) now carry those exact healthy elements so each injected fault remains the sole failure cause; --self-test passes 6/6.

dashboard/looker_studio/docs/app.mjs
caohy1988: Fixed in 635083b. showBillingError() opens the #advanced-settings disclosure (new id on the details element) and correction never auto-closes it. Tests cover both the interactive path (error opens it, correction leaves it open) and the setup-link path: a prefill with an invalid billingProject parameter shows the billing error with the disclosure open, actions disabled, and the table field clean.

dashboard/looker_studio/docs/app.mjs
caohy1988: Fixed in 635083b. The paste handler now attempts full validation on the clipboard text and only normalizes the displayed value on success; any invalid paste — unparseable, bad segment, sentinel collision, or bad-identifier Console link — lands as the exact clipboard text with its immediate error. Fixtures include the normalization-changing case from the finding (backticked+semicolon-wrapped invalid ID retained verbatim).

caohy1988:
All eight findings addressed in 635083b — each thread has a fix-reply. Ordered as recommended: the Enter P1 first (explicit keydown bridge into the shared attempted-action path, since implicit submission can never fire on this form shape), then the clipboard epoch guard and billing whitespace normalization, then the Console segment-attribution extractor, smoke exact-state assertions with matching fixture baselines, the Advanced-disclosure visibility fix, the raw-paste retention, and the hero copy.

Verification: Node contract green (new cases: invalid/valid/non-Enter keydown with open-count assertions, deferred and reversed clipboard completions, whitespace-only billing, three bad-segment Console fixtures at the validator layer plus one at the DOM layer, raw-retention pastes including the wrapped normalization-changing case, disclosure open/no-auto-close, invalid-billing prefill); browser smoke main + --self-test 6/6 with the exact-state pristine assertions; Python suite 29/29; pyink/isort clean.

One note on the P1's real-browser test ask (detail in that thread): the fix removes the dependency on trusted-event implicit submission entirely, so the Node keydown tests drive the exact listener that now is the mechanism; a CDP-driven trusted-key harness would be new smoke infrastructure I'd rather propose as a follow-up than grow in this PR — happy to file it if you want it.

dashboard/looker_studio/docs/app.mjs
```diff
@@ -1,177 +1,284 @@
...
+  // invalid paste — unparseable, bad segment, or sentinel collision — must
+  // land as the exact clipboard text so the raw invalid input is retained
+  // alongside its immediate error (#448 decision 2, #449 review).
+  let parsed = null;
+  try {
+    parsed = validateQualifiedTableId(text);
+  } catch {
+    parsed = null;
+  }
+  if (parsed) {
```
caohy1988: P2: Invalid paste can reuse the previous table ID. This branch leaves invalid paste to the browser default, which inserts at the current selection rather than replacing the field with the raw clipboard value. Starting from Ready with my-project.my_dataset.my_table and pasting x at the end produces the valid my-project.my_dataset.my_tablex, re-enabling Create for a different table and violating the raw-invalid-input contract. Please always prevent the default paste, assign either the normalized valid ID or the complete raw clipboard text, then refresh; add a regression starting from a nonempty Ready field.

dashboard/looker_studio/docs/configurator.mjs
```diff
@@ -234,3 +271,48 @@ export function parseTableReferenceForInput(value) {
...
+    throw new ConfigurationError("tableId", TABLE_ID_MESSAGES.empty);
+  }
+  // The URL branch uses the structural extractor, not the strict public
+  // parser: a supported-host link that unambiguously names one table but
+  // carries an invalid identifier has three identifiable segments, so it
+  // must reach the segment loop below rather than collapse into the
+  // whole-field link error.
+  const parsed = URL_SHAPED_RE.test(raw)
+    ? extractBigQueryConsoleTableReference(raw)
+    : splitQualifiedTableId(raw);
```
caohy1988, line 299: P2: Keep embedded backticks invalid. This new primary validation path calls splitQualifiedTableId, which currently removes every backtick globally. As a result, my-project.my_daBACKTICKtaset.my_table with BACKTICK replaced by a backtick is silently accepted as my-project.my_dataset.my_table instead of producing a dataset-segment error and retaining the raw value. Please strip only one enclosing whole-ID backtick pair and let any remaining backtick reach segment validation; add manual-entry and paste regressions.

dashboard/looker_studio/docs/app.mjs
```diff
@@ -1,177 +1,284 @@
...
+}
+
 createLink.addEventListener("click", (event) => {
   if (!createLink.href) {
     event.preventDefault();
+    revealTableErrors = true;
     refresh();
     return;
   }
   setStatus(WAITING_MESSAGE, "waiting");
```
caohy1988, line 241: P2: Invalidate pending clipboard status when Create begins. A setup-link or checklist copy can capture the current statusEpoch, remain pending, and then resolve after this enabled-link click. Because this branch does not advance the epoch, that older completion overwrites the important provisioning warning with a copied-status message. Please increment statusEpoch before setting WAITING_MESSAGE and add a held-copy then Create regression for both copy paths.

dashboard/looker_studio/tools/browser_smoke.sh
```diff
@@ -282,14 +317,50 @@ if ! grep -q 'data-errors="0"' "$OUT_DIR/dom.html"; then
...
+# generic aria-disabled search is already satisfied by the static markup
+# and would prove nothing about the runtime state (#449 review).
+FLAT_DOM="$(tr '\n' ' ' < "$OUT_DIR/dom.html")"
+TABLE_TAG="$(printf '%s' "$FLAT_DOM" | grep -o '<input[^>]*id="table-id"[^>]*>' | head -1)"
+[ -n "$TABLE_TAG" ] || fail "pristine first load must render the table-id field"
+case "$TABLE_TAG" in
+  *aria-invalid*) fail "pristine table-id field must not be marked invalid" ;;
+esac
+case "$TABLE_TAG" in
+  *value=*) fail "pristine table-id field must be empty" ;;
```
caohy1988: P2: Assert the live input value, not the serialized value attribute. Chrome dump-dom does not reflect assignments to HTMLInputElement.value as a value attribute, so this check cannot prove that the field is empty. I tested the exact head with app.mjs assigning not-pristine to the live field; the smoke still exited 0 and reported pristine first load. Please snapshot table.value and the other live action properties onto the instrumentation marker and assert exact values; add a negative self-test that populates the value property without adding an attribute.

dashboard/looker_studio/docs/app.mjs
```diff
@@ -1,177 +1,284 @@
...
+}
+
+form.addEventListener("submit", (event) => {
+  event.preventDefault();
+  attemptCreate();
 });
 
+for (const input of [tableIdInput, billingInput]) {
+  input.addEventListener("keydown", (event) => {
+    if (event.key === "Enter") {
```
caohy1988: P3: Ignore repeated and composing Enter events. Every Enter keydown currently reaches attemptCreate, so holding Enter can invoke window.open repeatedly and Enter used to commit IME composition can trigger Create prematurely. Please return early for event.repeat, event.isComposing, and the legacy composition signal keyCode 229, with focused keydown tests.

caohy1988:
Fresh review of 635083b: request changes.\n\nI found four P2 issues and one P3, each posted inline:\n\n- P2: invalid paste can combine with the previous Ready value and silently target another table instead of retaining the raw clipboard input\n- P2: global backtick removal accepts malformed embedded-backtick segments as a different normalized ID\n- P2: a pending clipboard completion can overwrite the Create provisioning warning\n- P2: browser smoke checks a serialized value attribute rather than the live input value and false-passes a non-pristine field\n- P3: the Enter bridge does not exclude repeat or IME-composition events\n\nValidation on the exact head:\n- Node configurator contract passed\n- browser smoke and all six negative self-tests passed\n- Python dashboard suite passed: 29 tests\n- an adversarial mutation that sets the live table field to a nonempty value still passed browser smoke, confirming the smoke finding\n\nI found no security or public-API compatibility regressions. CI is green, but the branch is currently behind main.

dashboard/looker_studio/docs/app.mjs
caohy1988: Fixed in 4587e90. Every paste is now intercepted (preventDefault always): the field is assigned either the normalized valid ID or the complete raw clipboard text — never a selection splice — then fails closed and refreshes. Your exact scenario is the new regression: from Ready with my-project.my_dataset.my_table, pasting 'x' leaves the field as 'x' with actions revoked, the Ready status cleared, and the whole-field error shown.

dashboard/looker_studio/docs/configurator.mjs
caohy1988: Fixed in 4587e90. splitQualifiedTableId now strips trailing ;/, punctuation first and then exactly one enclosing whole-ID backtick pair; an embedded backtick survives into its segment and fails segment validation. Your example now yields 'Dataset segment: …' with the raw value retained, at both the validator layer (splitQualifiedTableId preserves the backtick, parseQualifiedTableIdForInput returns null, validateQualifiedTableId throws segment=dataset) and the DOM layer (manual entry and paste regressions). Side effect made deliberate: per-segment-quoted IDs (`p`.`d`.`t`) — never in the documented paste table — now also fail with a segment error instead of being silently unwrapped.

dashboard/looker_studio/docs/app.mjs
caohy1988: Fixed in 4587e90. Both WAITING_MESSAGE writers (attemptCreate and the enabled-link click) advance statusEpoch before setting the warning, so any pending copy completion is invalidated. Held-copy-then-Create regressions cover both paths: setup-link copy vs link click, and checklist copy vs form submit — the status stays 'waiting' after the late completions resolve.

dashboard/looker_studio/tools/browser_smoke.sh
caohy1988: Fixed in 4587e90 — and confirmed against your exact mutation. The instrumentation now stamps a live-state snapshot onto the marker on every re-stamp (data-table-value from the value property, data-create-aria-disabled, data-create-has-href, data-copy-disabled from the disabled property), and the pristine assertions read those exact values instead of the serialized markup. New negative fixture 7 sets the live value property with no attribute on an otherwise fully healthy page and must fail; --self-test passes 7/7.

dashboard/looker_studio/docs/app.mjs
caohy1988: Fixed in 4587e90. The Enter bridge returns early for event.repeat, event.isComposing, and legacy keyCode 229, with focused keydown tests asserting zero opens for each. Discrete Enter behavior is unchanged (invalid reveals the error, valid opens exactly one tab).

caohy1988:
All five round-2 findings fixed in 4587e90 (fix-replies in each thread), and the branch is synced with `main` (merged `e6e763c`, no conflicts — the evalbench change doesn't touch this area).

The standout was the smoke finding: your adversarial live-value mutation is now the seventh negative self-test fixture, and the pristine assertions read the instrumentation's live-state snapshot (value property, aria-disabled/href, disabled property) rather than serialized markup — verified locally that the mutation now fails and 7/7 self-tests pass.

Full verification on 4587e90: Node contract green (new regressions: fragment-paste-over-Ready takeover, embedded-backtick at validator and DOM layers, held-copy-then-Create for both copy paths, repeat/composing/229 Enter guards); browser smoke main + --self-test 7/7; Python dashboard suite 29/29; pyink/isort clean.

One deliberate behavior note from the backtick fix (detail in that thread): per-segment-quoted IDs (`p`.`d`.`t`), which were silently unwrapped before and were never in the documented paste table, now fail with a segment error — raw value retained.

dashboard/looker_studio/tools/browser_smoke.sh
```diff
@@ -181,6 +241,28 @@ window.__smokeErrors = [];
     }
     marker.setAttribute("data-errors", String(window.__smokeErrors.length));
     marker.setAttribute("data-detail", window.__smokeErrors.join(" | ").slice(0, 500));
+    // Live-state snapshot (#449 review): dump-dom does not reflect the
+    // value PROPERTY into a value attribute, so the pristine assertions
+    // must read the live properties, not the serialized markup.
+    var table = document.querySelector("#table-id");
+    var create = document.querySelector("#create-dashboard");
+    var copy = document.querySelector("#copy-link");
+    marker.setAttribute(
```
caohy1988, line 303: P2: Keep the live-state snapshot current through the virtual-time window. The marker is created at 400 ms and stamp runs once; after that, only a recorded error calls stamp again. A page that changes table.value at 900 ms therefore ends with a non-pristine live field while this marker still says the value is empty. I reproduced this against 4587e90: the smoke exited 0 and reported pristine first load. Please periodically restamp the live properties until the DOM dump, or otherwise mirror post-snapshot changes, and add a delayed-value negative fixture.

dashboard/looker_studio/tools/browser_smoke.sh
```diff
@@ -282,14 +364,45 @@ if ! grep -q 'data-errors="0"' "$OUT_DIR/dom.html"; then
...
+  fail "the app-initialized marker must not appear in static HTML"
+fi
+grep -q 'data-bqaa-app-initialized="true"' "$OUT_DIR/dom.html" \
+  || fail "app-initialized marker missing — app.mjs did not execute in the browser"
+# With no query parameters the first load is the pristine state (#448):
+# assert the exact LIVE states via the instrumentation snapshot — Chrome's
+# dump-dom does not reflect the value property into a value attribute, so
+# serialized-markup checks cannot prove the field is empty (#449 review).
+FLAT_DOM="$(tr '\n' ' ' < "$OUT_DIR/dom.html")"
+MARKER_TAG="$(printf '%s' "$FLAT_DOM" | grep -o '<div[^>]*id="smoke-result"[^>]*>' | head -1)"
```
caohy1988: P2: Scope the zero-error assertion to this smoke-result marker. The earlier grep scans the whole DOM, so any unrelated element carrying data-errors="0" can satisfy it even when the instrumentation marker contains data-errors="1". I reproduced the false pass with an otherwise healthy fixture containing an unrelated data-errors="0" div plus a dispatched error event; the smoke exited 0. Please extract MARKER_TAG before the error check, assert its exact data-errors value, read data-detail from that same tag, and add a masking-element negative fixture.

caohy1988:
Fresh round-3 review of 4587e90: two P2 findings, posted inline.\n\nThe five round-2 application/parser fixes are present and their dedicated regressions pass. The remaining findings are both browser-smoke fidelity gaps:\n\n- the live-state marker is stamped once at 400 ms, so a delayed post-load field mutation can occur before the final DOM dump without updating the snapshot\n- the zero-error assertion scans the whole DOM rather than the smoke-result marker, so an unrelated data-errors="0" attribute can mask a nonzero instrumentation count\n\nBoth are reproduced against the exact head: each adversarial fixture exited 0 and reported a healthy pristine load.\n\nExact-head validation:\n- Node configurator contract passed\n- browser smoke main passed\n- all 7 existing smoke self-tests passed\n- Python dashboard suite passed: 29 tests\n- Format, browser smoke, Python 3.10-3.14, Build, and CLA are green; check-changes and scan-pr are still queued at review time\n\nNo new application-state, parser, security, or public-API compatibility findings.

dashboard/looker_studio/tools/browser_smoke.sh
caohy1988: Fixed in af8788b. stamp() now reruns on a 100 ms interval from marker creation until the DOM dump, mirroring how the error count was already kept live. Your reproduction is negative fixture 8: pristine at 400 ms, field value mutated at 900 ms — it now fails (data-table-value carries the late mutation at dump time); 9/9 self-tests pass.

dashboard/looker_studio/tools/browser_smoke.sh
caohy1988: Fixed in af8788b. The marker tag is extracted first and every marker-borne assertion — including data-errors and the data-detail used in the failure message — is scoped to that one tag; the whole-DOM grep is gone. Your reproduction is negative fixture 9: an unrelated data-errors="0" div beside a marker recording a real console error — it now fails on the marker's actual count.

caohy1988:
Both round-3 P2s fixed in af8788b (fix-replies in each thread): the live-state snapshot restamps every 100 ms until the dump, and the zero-error assertion is scoped to the extracted marker tag. Both reproductions from the review are now negative fixtures 8 and 9 and fail as required. Verified on the head: browser smoke main + --self-test 9/9, Node contract green, Python suite 29/29, pyink/isort clean.

caohy1988:
Fresh full review of af8788b: no new findings.\n\nI reviewed the complete current nine-file PR diff against e6e763c and re-audited the parser, paste/state machine, clipboard/Create ordering, Enter handling, setup-link compatibility, and browser-smoke failure paths.\n\nRound-3 verification:\n- removing the 100 ms restamp made fixture 8 fail with the delayed live-value false pass\n- reverting the marker-scoped error assertion made fixture 9 fail with the decoy zero-error masking case\n- Node configurator contract passed\n- browser smoke main passed\n- all 9 smoke negative self-tests passed\n- Python dashboard suite passed: 29 tests\n- Format, scan-pr, check-changes, Python 3.10-3.14, browser smoke, Build, and CLA are all green\n\nNo new correctness, security, public-API compatibility, accessibility, documentation, or test-fidelity issues found. Code is ready for maintainer review from this pass.\n\nAdministrative note: GitHub currently reports 15 unresolved review threads, 7 still active. The posting account is also the PR author, so it cannot provide the external approving review required by branch protection.

haiyuan-eng-google (approved):

PR #452, opened 2026-08-30
feat(evalbench): failed_sessions view and version-pinned consumer (#435)

caohy1988:
## Summary
Slice 2 of #435, stacked on #451. Queryable `evalbench_failed_sessions` view + version-pinned consumer/CLI so the failed-session denominator can be listed and drilled into without mixing import versions.

- `materialize()` keeps `evalbench_failed_sessions` (or `failed_sessions_view=`) pinned to the job's **latest successful import**: `failed_sessions_sql()` rendered with `(job_id, import_version)` as literals, so the view never scans another version. Pin recorded in a leading SQL comment and read back before publishing — a view pinned to another job, or a table/view the importer did not create, is refused before any write. Unchanged re-imports leave the view alone; a missing view is created; a specific older version is queried with the parameterized SQL or `failed_sessions(import_version=...)`.
- `failed_sessions()` resolves one `(job_id, import_version)` from the manifest first (latest by default), runs the parameterized contract over the tables that version is bound to, and returns `EvalBenchSession` rows whose versioned `session_id`/`trace_id` + `trace_selector()` feed `Client.get_session_trace` (`session.get_trace(client)`).
- `failed_sessions_sql(job_id=, import_version=)` renders escaped literals for view bodies.
- CLI: `bq-agent-sdk evalbench-failed-sessions --job-id … [--import-version] [--min-score C=V]… [--include-passed]` (exit 0/2); `evalbench-import` gains `--failed-sessions-view`, `--skip-failed-sessions-view`, `--min-score`, `--missing-score-passes`.
- Docs (`docs/evalbench.md`: Queryable view / Version-pinned consumer / Drill-down) and CHANGELOG.

## Depends on
#451 (slice 1 writer). Merge after #451, or together.

## Out of scope
Taxonomy, localization, harnesses, dashboards, D4, live traces.

## Test plan
- [x] Unit tests for view DDL pin, classify_sessions parity, version isolation, CLI
- [x] `python3 -m pytest -q tests/test_evalbench_importer.py tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py` — 111 passed
- [x] Full suite: 4299 passed, 54 skipped; the only failures/collection errors (`test_cli_agent_tool`, `test_memory_service*`, 2 in `test_pr16_fixes`) are a local `opentelemetry-semconv` ↔ editable `adk-python` import mismatch that fails identically at the base commit
- [ ] Codex + Kimi review this PR on GitHub

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **EvalBench failed-session view and version-pinned consumer (#435, slice 2)**
+  — `EvalBenchRun.materialize()` now keeps an `evalbench_failed_sessions`
+  view in the target dataset pinned (as literals) to the job's latest
+  successful import, so SQL consumers never scan another `import_version`;
+  `failed_sessions()` / `bq-agent-sdk evalbench-failed-sessions` list the
+  failed sessions of exactly one published version, resolving the version
+  from the manifest before reading events, and each row's versioned
+  `session_id`/`trace_id` (`EvalBenchSession.trace_selector()`) drills into
+  `Client.get_session_trace` without mixing versions. `failed_sessions_sql()`
+  accepts `job_id=`/`import_version=` to render the pin as literals.
+
 ## [0.5.1] - 2026-08-29
 
 ### Release highlights
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,13 +19,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `session_id`/`trace_id` (`EvalBenchSession.trace_selector()`) drills into
   `Client.get_session_trace` without mixing versions. `failed_sessions_sql()`
   accepts `job_id=`/`import_version=` to render the pin as literals. The
-  view's pin comment carries the rendered score policy and a hash of the
-  query below it, so ownership is proven by the definition (a copied pin
-  over other SQL, or an edited managed view, is refused, never replaced),
-  a same-version re-import that adds, changes, or drops the policy
-  re-renders the gate, and the view is written with create-if-absent plus
-  ETag-conditional replace after re-reading the view and latest manifest,
-  so concurrent imports cannot overwrite a newer pin.
+  view's pin comment carries the manifest generation (`imported_at`) and
+  the rendered score policy, and ownership is proven by re-rendering the
+  definition from the committed manifest row and comparing byte-for-byte
+  (a copied pin over other SQL, or an edited or reformatted managed view,
+  is refused, never replaced); a same-version re-import that adds,
+  changes, or drops the policy re-renders the gate, only the import that
+  committed the latest generation decides it, and the view is written with
+  create-if-absent plus ETag-conditional replace after re-reading the view
+  and latest manifest, so concurrent imports cannot overwrite a newer pin
+  or attach an older caller's policy to a newer generation.
 
 ## [0.5.1] - 2026-08-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,16 +19,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `session_id`/`trace_id` (`EvalBenchSession.trace_selector()`) drills into
   `Client.get_session_trace` without mixing versions. `failed_sessions_sql()`
   accepts `job_id=`/`import_version=` to render the pin as literals. The
-  view's pin comment carries the manifest generation (`imported_at`) and
-  the rendered score policy, and ownership is proven by re-rendering the
-  definition from the committed manifest row and comparing byte-for-byte
-  (a copied pin over other SQL, or an edited or reformatted managed view,
-  is refused, never replaced); a same-version re-import that adds,
-  changes, or drops the policy re-renders the gate, only the import that
-  committed the latest generation decides it, and the view is written with
-  create-if-absent plus ETag-conditional replace after re-reading the view
-  and latest manifest, so concurrent imports cannot overwrite a newer pin
-  or attach an older caller's policy to a newer generation.
+  gate and generation are committed manifest state: every publish mints
+  an opaque `generation_id` (caller-supplied `imported_at` may repeat and
+  is not the generation) and records the score policy as `view_policy`,
+  a same-version `unchanged` re-import that adds, changes, or drops the
+  policy commits it to the row it found under a new generation, and the
+  view is always rendered from the latest manifest row. The view's pin
+  comment names the generation and the rendered policy, and ownership is
+  proven by re-rendering the definition from the committed manifest row
+  and comparing byte-for-byte (a copied pin over other SQL, an edited or
+  reformatted managed view, or a canonical rendering of the current
+  generation under a gate the manifest does not record, is refused, never
+  replaced). The view is written with create-if-absent plus
+  ETag-conditional replace after re-reading the view and latest manifest,
+  so concurrent imports cannot overwrite a newer pin or attach an older
+  caller's policy to a newer generation.
 
 ## [0.5.1] - 2026-08-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   from the manifest before reading events, and each row's versioned
   `session_id`/`trace_id` (`EvalBenchSession.trace_selector()`) drills into
   `Client.get_session_trace` without mixing versions. `failed_sessions_sql()`
-  accepts `job_id=`/`import_version=` to render the pin as literals.
+  accepts `job_id=`/`import_version=` to render the pin as literals. The
+  view's pin comment carries the rendered score policy and a hash of the
+  query below it, so ownership is proven by the definition (a copied pin
+  over other SQL, or an edited managed view, is refused, never replaced),
+  a same-version re-import that adds, changes, or drops the policy
+  re-renders the gate, and the view is written with create-if-absent plus
+  ETag-conditional replace after re-reading the view and latest manifest,
+  so concurrent imports cannot overwrite a newer pin.
 
 ## [0.5.1] - 2026-08-29
 
--- a/docs/evalbench.md
+++ b/docs/evalbench.md
@@ -301,7 +301,9 @@ The manifest row binds every published version to its source:
 | `results_fingerprint`, `scores_fingerprint`, `configs_fingerprint` | SHA-256 content fingerprints |
 | `events_table`, `scores_table` | fully-qualified published tables |
 | `event_row_count`, `score_row_count` | rows published |
-| `imported_at` | publish timestamp (UTC) |
+| `imported_at` | publish timestamp (UTC; caller-supplied via `imported_at=`, so two publishes may share it) |
+| `generation_id` | opaque id of this committed generation of the row: minted by every publish (a `replace=True` of the same version label included) and by every committed change of `view_policy`; never shared |
+| `view_policy` | the score gate the failed-session view renders for this version (canonical JSON of the `EvalScorePolicy`, `NULL` for none) — committed manifest state, never taken from the view |
 
 `EvalBenchImportResult.manifest` returns the same row in memory.
 
@@ -367,7 +369,7 @@ the definition itself, and the view can never scan another version. The
 view's query text (its description names the pinned version too) is:
 
 ```sql
--- evalbench_failed_sessions pin: {"import_version": "v2", "imported_at": "2026-08-30T17:04:11.512034+00:00", "job_id": "abc123", "policy": {"min_scores": {"goal_completion": 0.9}, "missing_score_fails": true}}
+-- evalbench_failed_sessions pin: {"generation_id": "3f9c2c1e0b7a4d6e9a1b5c8d7e6f4a2b", "import_version": "v2", "job_id": "abc123", "policy": {"min_scores": {"goal_completion": 0.9}, "missing_score_fails": true}}
 WITH sessions AS (
   SELECT ...
   FROM `analytics-project.bqaa.evalbench_agent_events`
@@ -376,13 +378,22 @@ WITH sessions AS (
 ```
 
 The pin records the job, the version, the manifest generation it renders
-(`imported_at`, in UTC with microseconds — a `replace=True` of the same
-version label is a new generation), and the score policy rendered into
-the view (`null` without one). Nothing the view says about itself proves
-ownership: the importer treats a view as its own only when the pin names
-a version that job committed to the manifest *and* the body is
-byte-for-byte what the importer renders for that manifest row, generation
-and policy. The comparison is exact — BigQuery returns a view's query text
+(`generation_id`, the row's opaque id — a `replace=True` of the same
+version label is a new generation, and so is a committed policy change;
+`imported_at` is caller-chosen and may repeat, so it is not the
+generation), and the score policy rendered into the view (`null` without
+one). The view body is a pure function of the latest manifest row: the
+gate it renders is the row's committed `view_policy`, never something the
+view says about itself. Nothing in the pin proves ownership: the importer
+treats a view as its own only when the pin names a version that job
+committed to the manifest *and* the body is byte-for-byte what the
+importer renders for that manifest row — with the row's own `view_policy`
+when the pin names the row's current `generation_id`, so a canonical
+rendering of the current generation under any other gate is foreign; or,
+for a view a superseded generation left behind, the pinned generation and
+policy over the row's own tables, in which case the view is only ever
+advanced to the committed rendering, so whatever gate it carries is never
+preserved. The comparison is exact — BigQuery returns a view's query text
 verbatim, and whitespace inside a rendered literal (`job_id = "job  x"`
 versus `"job x"`) changes which rows the view reads, so no whitespace
 normalization is applied. A pin copied or forged onto other SQL (however
@@ -401,9 +412,10 @@ Pinning rules:
 | New `import_version` of the same job | replaced, pinned to the new version (the job's latest `imported_at` in the manifest) |
 | No-op re-import (`unchanged`) | untouched; created only if it does not exist yet (corpora published before views existed) |
 | `replace=True` of an older version | re-pinned to it — the replace refreshed `imported_at`, so it *is* the latest successful import |
-| `policy=` / `--min-score` given, changed, or dropped | the score gate is part of the pin: the view is re-rendered whenever the policy differs from the one it carries (including a later same-version call *without* a policy, which removes the gate), and left alone when version and policy both match |
-| `policy=` on a call whose version is **not** the latest import | the gate is decided by the import of the version the view pins: a view already pinned to the latest version is left exactly as it is, and a view that is absent or behind is created or advanced carrying the gate it already had (none for a new view), never this call's |
-| `policy=` on a call whose *generation* is not the one the manifest holds (a concurrent `replace=True` of the same version committed a newer `imported_at`) | as above: the version label does not decide, the generation does. Every generation is rendered into the pin, so the newer replacement always rewrites the view (bumping its ETag) even when version, gate and SQL are otherwise identical; a delayed older caller's conditional replace then fails, re-reads, and leaves the newer generation's gate in place |
+| `policy=` / `--min-score` given, changed, or dropped | the gate is committed to the manifest row (`view_policy`): a publish records it with the row, and an `unchanged` re-import that names another gate commits it to the row it found — under a new `generation_id`, and only if that row is still the generation it found — before the view is re-rendered from the row (including a later same-version call *without* a policy, which removes the gate); nothing is written when version and policy both match |
+| `policy=` on a call whose version is **not** the latest import | the gate is recorded on that version's own row; the view renders the latest version's row, so a view already pinned to the latest generation is left exactly as it is, and a view that is absent or behind is created or advanced carrying the latest row's committed gate (never this call's, and never the gate the stale view happened to carry) |
+| `policy=` on a call whose *generation* is not the one the manifest holds (a concurrent `replace=True` of the same version — even with the same `imported_at` — committed a newer `generation_id`) | as above: neither the version label nor the timestamp decides, the committed generation does. Every generation is rendered into the pin, so the newer replacement always rewrites the view (bumping its ETag) even when version, gate and SQL are otherwise identical; a delayed older caller's conditional replace then fails, re-reads, and renders the newer generation's committed gate |
+| A view at that name that *is* the importer's rendering of the current generation, but under a gate the manifest row does not record | `ValueError` before anything is written, from every later call (an older version's unchanged re-import included): the view cannot vouch for its own gate, only the manifest can |
 | View at that name pinned to **another job** | `ValueError` before anything is written; use one `failed_sessions_view` name per job |
 | A table, a view the importer did not create, a copied pin over other SQL, or a managed view whose query was edited, at that name | `ValueError` before anything is written (before the import tables are even created); the importer never replaces objects whose definition it cannot vouch for — drop the object or pick another name |
 | Two imports of one job race on the view | the loser of the create race (`409`) or ETag check (`412`) re-reads the view and the latest manifest and re-decides, up to three times, so a delayed writer never overwrites a newer pin — nor re-renders it with its own (older) policy; a race against **another job's** view fails closed with the import already published |
--- a/docs/evalbench.md
+++ b/docs/evalbench.md
@@ -363,19 +363,27 @@ overridable with `failed_sessions_view=` / `--failed-sessions-view`, or
 `None` / `--skip-failed-sessions-view` to manage no view — whose body is
 `failed_sessions_sql()` with `@job_id` and `@import_version` rendered as
 string literals. Views cannot take query parameters, so the pin lives in
-the definition itself, and the view can never scan another version:
+the definition itself, and the view can never scan another version. The
+view's query text (its description names the pinned version too) is:
 
 ```sql
-CREATE OR REPLACE VIEW `analytics-project.bqaa.evalbench_failed_sessions`
-OPTIONS (description = "EvalBench failed sessions (W0.4 contract) pinned to job_id 'abc123' import_version 'v2'; ...")
-AS
--- evalbench_failed_sessions pin: {"import_version": "v2", "job_id": "abc123"}
+-- evalbench_failed_sessions pin: {"import_version": "v2", "job_id": "abc123", "policy": {"min_scores": {"goal_completion": 0.9}, "missing_score_fails": true}, "query_sha256": "…"}
 WITH sessions AS (
   SELECT ...
   FROM `analytics-project.bqaa.evalbench_agent_events`
   WHERE job_id = "abc123" AND import_version = "v2"
   ...
 ```
+
+The pin records the job, the version, the score policy rendered into the
+view (`null` without one), and the SHA-256 of the query below it. The
+comment alone never proves ownership: the importer treats a view as its
+own only when the query hashes to what the pin says — a pin copied onto
+other SQL, or a managed view whose body somebody edited, is a foreign
+object and is refused. The importer never uses `CREATE OR REPLACE VIEW`;
+it creates the view with create-if-absent and replaces it with an
+ETag-conditional update, after re-reading both the view and the latest
+manifest row immediately before writing.
 
 Pinning rules:
 
@@ -385,18 +393,21 @@ Pinning rules:
 | New `import_version` of the same job | replaced, pinned to the new version (the job's latest `imported_at` in the manifest) |
 | No-op re-import (`unchanged`) | untouched; created only if it does not exist yet (corpora published before views existed) |
 | `replace=True` of an older version | re-pinned to it — the replace refreshed `imported_at`, so it *is* the latest successful import |
-| `policy=` / `--min-score` given | the score gate is rendered into the view (and re-rendered on every call, since the policy is not part of the pin); a later call *without* a policy leaves the rendered gate in place when the pin is unchanged |
+| `policy=` / `--min-score` given, changed, or dropped | the score gate is part of the pin: the view is re-rendered whenever the policy differs from the one it carries (including a later same-version call *without* a policy, which removes the gate), and left alone when version and policy both match |
 | View at that name pinned to **another job** | `ValueError` before anything is written; use one `failed_sessions_view` name per job |
-| A table, or a view the importer did not create, at that name | `ValueError` before anything is written; the importer never replaces objects it does not own |
+| A table, a view the importer did not create, a copied pin over other SQL, or a managed view whose query was edited, at that name | `ValueError` before anything is written (before the import tables are even created); the importer never replaces objects whose definition it cannot vouch for — drop the object or pick another name |
+| Two imports of one job race on the view | the loser of the create race (`409`) or ETag check (`412`) re-reads the view and the latest manifest and re-decides, up to three times, so a delayed writer never overwrites a newer pin; a race against **another job's** view fails closed with the import already published |
 
 "Latest successful import" is the manifest row with the newest
 `imported_at` (ties broken by `import_version`), which is also what the
 Python consumer below resolves when no version is pinned, so the two agree.
-The view is rewritten only when its pin or policy changes and is issued
-after the publish transaction committed; if the DDL fails (for example a
-missing `bigquery.tables.update` permission) `materialize()` raises a
-`ValueError` naming the published version, and simply re-running it retries
-the view (`status == "unchanged"`). A specific older version is queried
+The view is rewritten only when its rendered definition (pinned version,
+policy, or SQL) would change, and is written after the publish transaction
+committed; if the write fails (for example a missing
+`bigquery.tables.update` permission, or a view that kept changing
+underneath the conditional replace) `materialize()` raises a `ValueError`
+naming the published version, and simply re-running it retries the view
+(`status == "unchanged"`). A specific older version is queried
 with the parameterized `failed_sessions_sql()` shown above or with
 `failed_sessions(import_version=...)`.
 
--- a/docs/evalbench.md
+++ b/docs/evalbench.md
@@ -367,7 +367,7 @@ the definition itself, and the view can never scan another version. The
 view's query text (its description names the pinned version too) is:
 
 ```sql
--- evalbench_failed_sessions pin: {"import_version": "v2", "job_id": "abc123", "policy": {"min_scores": {"goal_completion": 0.9}, "missing_score_fails": true}, "query_sha256": "…"}
+-- evalbench_failed_sessions pin: {"import_version": "v2", "job_id": "abc123", "policy": {"min_scores": {"goal_completion": 0.9}, "missing_score_fails": true}}
 WITH sessions AS (
   SELECT ...
   FROM `analytics-project.bqaa.evalbench_agent_events`
@@ -375,15 +375,18 @@ WITH sessions AS (
   ...
 ```
 
-The pin records the job, the version, the score policy rendered into the
-view (`null` without one), and the SHA-256 of the query below it. The
-comment alone never proves ownership: the importer treats a view as its
-own only when the query hashes to what the pin says — a pin copied onto
-other SQL, or a managed view whose body somebody edited, is a foreign
-object and is refused. The importer never uses `CREATE OR REPLACE VIEW`;
-it creates the view with create-if-absent and replaces it with an
-ETag-conditional update, after re-reading both the view and the latest
-manifest row immediately before writing.
+The pin records the job, the version, and the score policy rendered into
+the view (`null` without one). Nothing the view says about itself proves
+ownership: the importer treats a view as its own only when the pin names
+a version that job committed to the manifest *and* the body is exactly
+(modulo whitespace) what the importer renders for that manifest row and
+policy. A pin copied or forged onto other SQL (however self-consistent), a
+contract-shaped body for an unpublished version or over other tables, or
+a managed view whose body somebody edited, is a foreign object and is
+refused. The importer never uses `CREATE OR REPLACE VIEW`; it creates the
+view with create-if-absent and replaces it with an ETag-conditional
+update, after re-reading both the view and the latest manifest row
+immediately before writing.
 
 Pinning rules:
 
@@ -394,9 +397,10 @@ Pinning rules:
 | No-op re-import (`unchanged`) | untouched; created only if it does not exist yet (corpora published before views existed) |
 | `replace=True` of an older version | re-pinned to it — the replace refreshed `imported_at`, so it *is* the latest successful import |
 | `policy=` / `--min-score` given, changed, or dropped | the score gate is part of the pin: the view is re-rendered whenever the policy differs from the one it carries (including a later same-version call *without* a policy, which removes the gate), and left alone when version and policy both match |
+| `policy=` on a call whose version is **not** the latest import | the gate is decided by the import of the version the view pins: a view already pinned to the latest version is left exactly as it is, and a view that is absent or behind is created or advanced carrying the gate it already had (none for a new view), never this call's |
 | View at that name pinned to **another job** | `ValueError` before anything is written; use one `failed_sessions_view` name per job |
 | A table, a view the importer did not create, a copied pin over other SQL, or a managed view whose query was edited, at that name | `ValueError` before anything is written (before the import tables are even created); the importer never replaces objects whose definition it cannot vouch for — drop the object or pick another name |
-| Two imports of one job race on the view | the loser of the create race (`409`) or ETag check (`412`) re-reads the view and the latest manifest and re-decides, up to three times, so a delayed writer never overwrites a newer pin; a race against **another job's** view fails closed with the import already published |
+| Two imports of one job race on the view | the loser of the create race (`409`) or ETag check (`412`) re-reads the view and the latest manifest and re-decides, up to three times, so a delayed writer never overwrites a newer pin — nor re-renders it with its own (older) policy; a race against **another job's** view fails closed with the import already published |
 
 "Latest successful import" is the manifest row with the newest
 `imported_at` (ties broken by `import_version`), which is also what the
--- a/docs/evalbench.md
+++ b/docs/evalbench.md
@@ -9,8 +9,10 @@ never writes into the ADK plugin's production `agent_events` table.
 This covers the reader and deterministic row mapping from
 [issue #97](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/97)
 and the MVP snapshot + failed-session denominator from
-[issue #435](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/435).
-Localization, failure taxonomy, and other harnesses are later slices.
+[issue #435](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/issues/435):
+the versioned writer (slice 1) and the queryable `evalbench_failed_sessions`
+view plus version-pinned consumer (slice 2). Localization, failure taxonomy,
+and other harnesses are later slices.
 
 ## Data Flow
 
@@ -20,8 +22,11 @@ EvalBench BigQuery dataset                BQAA-owned target dataset
 configs  -- job_id = @job_id --+          evalbench_agent_events
 results  -- job_id = @job_id --+--> EvalBenchRun --> evalbench_scores_imported
 scores   -- job_id = @job_id --+   (in memory)      evalbench_import_manifest
-                                                     ^
+                                                     ^         |
                               one transaction per (job_id, import_version)
+                                                               v
+                                          evalbench_failed_sessions (VIEW,
+                                          pinned to the job's latest import)
 ```
 
 All three source queries filter by the parameterized `job_id` in SQL. The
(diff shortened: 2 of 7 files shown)


caohy1988 (commented):
**Reviewer: Kimi**

# Review: PR #452 — EvalBench slice 2 (failed_sessions view + version-pinned consumer)

Reviewed at head `f31b612`. Tests: `python3 -m pytest -q tests/test_evalbench_importer.py tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py` — **111 passed** locally.

**Verdict: no P0 found.** The core contract holds: the view body and the parameterized consumer are the same `_failed_sessions_sql_for_refs` output (byte-equality asserted in tests), literal escaping via `_sql_string_literal` is sound (`json.dumps`-based; `import_version` additionally regex-restricted), pin parsing is fail-closed (plain tables, human views, malformed pins all refuse before data writes), and the consumer never interpolates `job_id`/`import_version` (bound parameters throughout). CLI exit 0/2 semantics are correct and tested.

## P1 findings

**P1-1 — Cross-job view clobber via TOCTOU between pin check and view DDL.**
The pin is read once at `evalbench.py:928` before publishing, but `_sync_failed_sessions_view` issues `CREATE OR REPLACE VIEW` at `evalbench.py:1504-1505` after the publish without re-reading the pin. Concrete interleaving: jobs A and B import concurrently into one dataset sharing the default view name. Both read `existing_pin=None` at their starts. A commits and syncs (view now pinned to A). B's publish transaction starts after A's commit (no lock conflict), B's sync sees `existing_pin=None != pin_B` and silently `CREATE OR REPLACE`s A's view — B reports success while the view is now pinned to B. This violates the stated contract "a view pinned to another job is refused before any write" under an ordinary interleaving. The import lock serializes commits but the view sync sits outside it. The victim detects it on their next import (`ValueError: pinned to job …`), so it's recoverable, but not prevented. A pin re-read inside `_sync_failed_sessions_view` immediately before the DDL would close most of the window (BigQuery DDL can't do a fully atomic conditional replace, but the current code doesn't even re-check).

**P1-2 — Removing the score policy leaves the old policy SQL baked into the view forever.**
`_sync_failed_sessions_view`'s skip condition (`evalbench.py:1494-1496`) is `existing_pin == pin and not policy_requested`, where the pin captures only `(job_id, import_version)`. Failing scenario: `materialize(policy=EvalScorePolicy({"goal_completion": 0.9}))` renders the score gate into the view. A later `materialize()` of the same version with no policy returns `unchanged`, the pin matches, `policy_requested` is False → DDL skipped. The view keeps failing sessions on the 0.9 gate while `failed_sessions()` (same version, no policy) reports them as passed — view and consumer silently diverge, and there is no way to clear the gate except `replace=True` or a new version. The symmetric wart: *with* a policy, every no-op re-import re-renders the view, so "unchanged re-imports leave the view alone" only holds for policy-less imports. I see `docs/evalbench.md` documents the retention behavior as intentional, but the view/consumer divergence it produces (a session "passed" per the consumer, "failed" per the view, same job+version) is a real correctness trap for dashboard consumers. Including a policy hash in the pin would fix both directions.

## P2 findings (robustness, listed for completeness)

- **Pin-comment skip trusts the comment alone; body drift is never detected or repaired** (`evalbench.py:1495`). If someone edits the view in the console but leaves the pin comment (e.g. widens the WHERE), every subsequent no-op re-import skips the DDL and the view silently scans versions it shouldn't while `materialize` keeps reporting it correctly pinned. Comparing `table.view_query` against the freshly rendered body (or a body/template hash in the pin) would make the skip sound — and would also cover template fixes shipped in future SDK versions.
- **Refusal happens after table-creation DDL, not "before anything is written"** (`evalbench.py:916` `_ensure_import_tables` runs before the pin check at 928). On a first-time import into a fresh dataset whose view name is foreign/pinned elsewhere, four tables are created before the `ValueError` fires. No rows are written and refusal does precede loads/publish, but the docstring (`evalbench.py:826-828`) overpromises. Cheap fix: move the pin check above `_ensure_import_tables`.
- **Manifest-recorded table refs are interpolated into SQL without re-validation on read** (`evalbench.py:1453-1459` view DDL, `evalbench.py:1674-1678` consumer). Manifest rows written by this importer are validated at publish, but the manifest is a plain table — a doctored row whose `events_table` contains a backtick breakout would be rendered into DDL run under the reader's credentials. Bounded impact (manifest write access already lets you point at real tables), but a `_validate_destination_table`-style check of refs read back from the manifest is cheap defense-in-depth.
- **Missing manifest *table* gives a raw BigQuery 404, not the friendly "run materialize() first" message** (`evalbench.py:1654` only fires when the manifest exists with no row). Wrong `--target-dataset` → CLI exits 2 with `404 Not found: Table …`.
- **CLI edge cases:** duplicate `--min-score` comparators silently last-win (`cli.py:2272`); `--missing-score-passes` without `--min-score` is silently a no-op (`cli.py:2277-2278`) — worth a warning at least.

## Test-coverage gaps worth adding

- Policy removal on the same version (policy → no policy leaves stale gate SQL — P1-2), and re-render on a *changed* policy with the same version.
- Concurrency/TOCTOU: two jobs whose pin reads both precede either's sync (P1-1) — the fake client already models transaction snapshots but view operations have no interleaving coverage.
- Tampered view body with an intact pin comment (drift detection).
- Pin-parse edge cases: valid JSON plus trailing garbage on the pin line, non-ASCII `job_id` round-trip through comment write→parse.
- Doctored manifest row feeding `_failed_sessions_view_ddl` / consumer refs.
- `scenario_id` parity between the SQL (`ANY_VALUE(JSON_VALUE(...))`, may pick NULL non-deterministically) and `classify_sessions` (first usable non-empty value) — only affects the metadata column, not the `failed` flag.

## Verified clean (checked explicitly)

- Literal escaping / injection in `failed_sessions_sql` (hostile `job_id` test exists at `test_evalbench_importer.py:2429`).
- Fail-closed pin parse; refusal of plain tables and foreign views before any data write.
- View SQL carries `job_id` + `import_version` predicates on both the events CTE and the scores LEFT JOIN — no unfiltered access path.
- View/consumer SQL parity (same renderer, byte-equality test).
- No pin-to-nothing: view sync only runs after a committed publish or a manifest-backed `unchanged`; failed publish never reaches the view DDL.
- `trace_selector()` round-trip to `get_session_trace` holds structurally (versioned `session_id`/`trace_id` + `experiment_id` attribute pin match the client's scope-pin SQL) — though only mock-verified, see coverage note above.

caohy1988 (commented):
**Reviewer: Codex**

## Code Review Results

**Scope:** slice 2 only, `2e64fdb..f31b612` (7 files; slice 1 excluded except dependencies)
**Intent:** Publish a version-pinned failed-sessions view and consumer while refusing foreign/cross-job view replacement and preserving version isolation.
**Mode:** markdown report-only; no code changes or pushes
**Counts:** P0: 0, P1: 2

**Reviewers:** correctness, testing, maintainability, security, performance, API contract, reliability, plus an independent Claude adversarial pass.

### P1 -- High

| # | File | Issue | Reviewer | Confidence |
|---|------|-------|----------|------------|
| 1 | `src/bigquery_agent_analytics/evalbench.py:1424` | A pin comment is accepted as proof that the importer owns the view | API contract, security | 100 |
| 2 | `src/bigquery_agent_analytics/evalbench.py:1505` | Concurrent view reconciliation can publish a stale or cross-job pin | correctness, security, testing, reliability, independent Claude | 100 |

- **#1** -- `_read_view_pin()` accepts any view whose first nonblank line contains a syntactically valid public pin comment. A user-managed view can copy that comment and arbitrary SQL, after which `materialize()` treats it as importer-owned and may replace it. This violates the stated guarantee to refuse a view the importer did not create. Record ownership in importer-controlled metadata, validate the canonical managed definition, and add a regression test for a foreign view with a valid matching marker.
- **#2** -- Ownership is checked before the import, the latest manifest is read later, and `CREATE OR REPLACE VIEW` runs unconditionally after the publish transaction releases its lock. Two concurrent imports can both pass the stale check; the delayed DDL can then overwrite a newer pin or another job's view while both calls report success. Use create-if-absent plus ETag-conditional updates, re-read ownership and the latest manifest on conflicts, and add deterministic same-job and cross-job interleaving tests.

### Actionable Findings

| # | File | Issue | Route | Notes |
|---|------|-------|-------|-------|
| 1 | `src/bigquery_agent_analytics/evalbench.py:1424` | Make view ownership authoritative | `gated_auto -> downstream-resolver` | Suggested fix present; add forged-marker coverage |
| 2 | `src/bigquery_agent_analytics/evalbench.py:1505` | Make view reconciliation concurrency-safe | `manual -> downstream-resolver` | Requires optimistic concurrency design and race tests |

### Coverage

- Focused tests: `111 passed in 3.22s`.
- Validator: #1 independently validated; #2 skipped an extra validator because it had quote-anchored corroboration from both local reviewers and an independent serving family.
- Cross-model: route `claude`; requested/actual model `claude-opus-5`; requested effort `high`; actual effort `unverified`; receipt supported; independence verified.
- Project standards: no applicable `CODING_STANDARDS.md`, `CLAUDE.md`, or `AGENTS.md` files found.
- Prior-comment review: not run because this is the first review and the PR had no prior comments.
- Testing gaps: no forged-marker test and no interleaved view-update test.
- Suppressed by confidence: 0. Edge-only, IAM-dependent, documented-policy, and unproven-scale candidates were excluded from this P0/P1-only review.
- No implementation plan was discovered, so settlement suppression was not evaluated.

---

> **Verdict:** Not ready
>
> **Reasoning:** The data consumer and focused tests pass, but the two P1 view-safety defects can overwrite a foreign view or leave the shared view pinned to the wrong job/version.
>
> **Fix order:** #2 concurrency-safe reconciliation -> #1 authoritative ownership validation.

### Actionable Findings (prioritized recap)

1. **#2 P1 -- `src/bigquery_agent_analytics/evalbench.py:1505` -- manual:** replace unconditional last-writer-wins DDL with create-if-absent and ETag-conditional reconciliation, then add interleaving tests.
2. **#1 P1 -- `src/bigquery_agent_analytics/evalbench.py:1424` -- gated auto:** stop treating the public pin comment as proof of ownership and add a forged-marker test.

caohy1988 (commented):
**Reviewer: Kimi**

# Re-review: PR #452 after P1 fixes (head `0f1f350`)

Previous review was at `f31b612` (2 P1s, plus Codex's forged-pin P1). This pass re-checks those three against `0f1f350` (`fix(evalbench): authoritative view ownership and concurrency-safe pin`) and looks for new regressions.

Tests: `python3 -m pytest -q tests/test_evalbench_importer.py tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py` — **120 passed** (was 111; +9 new tests covering the fixes).

**Verdict: no P0, no remaining P1. The three P1s are genuinely fixed.** Commenting rather than approving only because my workflow posts comment-style reviews; nothing here should block merge beyond whatever Codex finds.

## P1 re-checks

**P1-1 (mine) — TOCTOU between pin check and view DDL → fixed.**
`_reconcile_failed_sessions_view` (evalbench.py:1593) replaced the blind `CREATE OR REPLACE VIEW` with a bounded reconcile loop: each attempt re-reads the view (ownership, job, ETag) via `_check_view_binding` *and then* the latest manifest row, so a version committed by a concurrent import after the view was read is still seen. Missing view → `create_table(table, exists_ok=False)` (create-if-absent; `Conflict` → re-read and re-decide). Existing view → `update_table(table, ["view_query", "description"])`, which is genuinely ETag-conditional — I verified in the installed `google-cloud-bigquery` client that `update_table` sends `If-Match: table.etag`, and `PreconditionFailed` triggers a re-read/re-decide. A missing ETag refuses the replace outright rather than falling back to unconditional write. After 3 attempts it fails closed naming the contention, and `_sync_failed_sessions_view` wraps any failure so the import reports the publish succeeded but the view did not. New tests cover same-job create-race, ETag-guarded update-race, cross-job race failing closed, and bounded retries.

**P1-2 (mine) — removing the score policy leaves stale gate SQL → fixed.**
The pin now carries the rendered policy (`_policy_pin`, canonical form) plus `query_sha256` of the query body, and the skip condition (evalbench.py:1635) is normalized-body equality — not pin equality with a `policy_requested` carve-out. Same-version re-import that adds, changes, or drops the policy produces a different body → re-render; a true no-op renders the identical body → skipped. Both directions of the old wart are gone. Tested by `test_materialize_policy_change_on_the_same_version_rerenders_the_view` (and add/drop coverage).

**Codex P1 — forged pin comment treated as ownership → fixed.**
`_parse_view_pin` now requires the pin to carry `query_sha256` and hashes the query text below the pin line; a mismatch (copied marker over foreign SQL, edited managed body, marker without a hash) returns `None`, which `_read_managed_view` treats as foreign and refuses. A copied marker alone can no longer prove ownership because the attacker cannot make the hash match without shipping the importer's exact body — at which point replacing it is a no-op anyway. Tested by `test_materialize_refuses_a_forged_marker_copied_from_a_managed_view`.

## Bonus fixes from my P2 list

- Pin/binding check moved above `_ensure_import_tables` (evalbench.py:934) — refusal now genuinely precedes any dataset write, as the docstring claimed.
- Body drift under an intact pin is no longer silently skipped: an edited body fails the hash and is refused (fail-closed). Whitespace-only drift is still tolerated via `_normalize_sql`, which is the right tolerance for BigQuery's verbatim storage.

## New regressions: none found

Checked the reconcile loop for fresh hazards: `Conflict`/`PreconditionFailed` are the only swallowed errors (ownership/job mismatches still raise through `_check_view_binding`, so a cross-job race fails closed rather than retrying); the retry loop re-fetches `existing.table` each attempt so a mutated Table object is never reused with a stale ETag; mutating `table.view_query`/`description` on the get_table result before `update_table` is consistent with the client's If-Match contract. Worst remaining behavior is a same-job policy flip-flop when two concurrent imports pass different policies — last reconciler wins, which is inherent to a caller-supplied, unversioned policy and acceptable.

## Standing P2s (unchanged from prior review, non-blocking)

- Manifest-recorded `events_table`/`scores_table` refs are rendered into view DDL without re-validation on read (a doctored manifest row with a backtick breakout would run under the reader's credentials) — bounded by manifest write access; cheap `_validate_destination_table`-style check on read-back would close it.
- Missing manifest *table* still surfaces as a raw BigQuery 404 rather than the friendly "run materialize() first" message (only the empty-manifest case is handled).
- CLI: duplicate `--min-score` comparators silently last-win; `--missing-score-passes` without `--min-score` is silently a no-op.

None of these are new, and none block.

caohy1988 (commented):
**Reviewer: Codex**

> **Verdict: Request changes.** Two P1s remain at `0f1f350`; no P0 found.

**Scope:** slice 2 only, `2e64fdb..0f1f350`; P1-fix delta `f31b612..0f1f350`.

**Verification:** `python3 -m pytest -q tests/test_evalbench_importer.py tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py` — **120 passed in 3.23s**.

### P1 — The view pin hash is self-authenticating, so ownership is still forgeable

`src/bigquery_agent_analytics/evalbench.py:1437-1471` accepts a view when the public pin's `query_sha256` matches the SQL below it. Both values are controlled by the existing view. A foreign view can therefore put arbitrary SQL below a same-job marker, compute the matching SHA-256, and pass `_parse_view_pin()` / `_check_view_binding()` as managed. `materialize()` then publishes and replaces that foreign view through the ETag update path.

I reproduced this with `SELECT 'foreign-owned' AS session_id` plus its freshly computed digest: the pin was accepted, `materialize(import_version="v1")` returned `status="imported"`, and the fake client recorded an `update` replacing the view. The new tests only cover a copied marker whose digest no longer matches, and a hard-coded all-zero digest; neither covers a correctly recomputed forged digest.

Fix ownership validation against trusted state: reconstruct the canonical definition from a committed manifest row plus a strictly validated policy, and require the existing body to match that rendering. Do not treat a digest supplied by the view itself as proof of ownership. Add a regression with arbitrary foreign SQL and its correct digest.

### P1 — An ETag retry can attach an older caller's policy to the newest version

`src/bigquery_agent_analytics/evalbench.py:1617-1628` correctly re-reads `latest` after a `409`/`412`, but then renders that latest manifest with the original invocation's `policy`. Concrete interleaving: v1 starts with a `0.9` gate and pauses after reading the old ETag; v2 lands with no policy and updates the view; v1 gets `412`, retries, reads v2 as latest, then writes **v2 with v1's `0.9` policy**. Both calls report success, but the newer policy removal is silently undone.

I reproduced the final pin as `{"import_version":"v2","policy":{"min_scores":{"goal_completion":0.9},...}}` even though v2's own successful sync had `policy: null`. This is the concurrency form of the stale-gate bug: the sequential same-version add/change/drop tests pass, but policy is not tied to the authoritative latest import during reconciliation.

Make the policy authoritative alongside the selected manifest version (for example, persist it with the version/view state), or prevent a delayed invocation from applying its policy when the re-read latest version differs from the version it handled. Add the deterministic v1-policy/v2-no-policy ETag interleaving test.

The create-if-absent and ETag mechanics do prevent stale **version** replacement, and sequential policy removal is fixed. The two cases above still violate the documented foreign-view refusal and concurrency-safe pin guarantees.

caohy1988 (commented):
**Reviewer: Kimi**

## Verdict: Approve (no remaining P0/P1)

Re-reviewed head `7b44ce6` ("fix(evalbench): verify view ownership against the manifest; pin policy to the latest import") against the two P1s left open at `0f1f350`. Both are fixed, and I found no new P0/P1 regressions.

Tests at HEAD: `python3 -m pytest -q tests/test_evalbench_importer.py tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py` → **135 passed**.

### P1 #1 — View pin hash self-authenticating / forgeable: FIXED

The self-authenticating `query_sha256` is gone from the pin entirely (`_view_pin` now carries only `job_id`, `import_version`, `policy`). Ownership in `_read_managed_view` (src/bigquery_agent_analytics/evalbench.py:1544) is now proven against state the forger does not control:

1. The pin's policy must parse strictly and round-trip to canonical form (`_policy_from_pin`), so a pin carrying anything `_policy_pin` would not emit is rejected outright.
2. The pin's `(job_id, import_version)` must resolve to a committed manifest row (`_read_manifest`, labelled `evalbench-failed-sessions`); a missing manifest table or unpublished version fails closed.
3. The view body must equal — modulo whitespace — the importer's own re-rendering `_failed_sessions_view_body(manifest=pinned, policy=policy)`, which is a pure function of the committed manifest row (including its `events_table`/`scores_table`) and the canonical policy.

A forger copying a same-job pin over arbitrary SQL — even with a correctly recomputed digest, had it survived — now fails step 3; a contract-shaped body for an unpublished version or over other tables fails steps 2/3. The only body that passes is the importer's own rendering, so "recognizing" it replaces nothing foreign. New tests cover exactly this: self-consistent forged pins (with and without a structurally valid policy), unpublished-version and relocated-table renderings, and eight malformed-policy shapes.

### P1 #2 — ETag retry attaching an older caller's policy to the newest version: FIXED

`_reconcile_failed_sessions_view` (evalbench.py:1694) now decides the rendered policy per attempt, after re-reading the latest manifest row:

- `latest_version == import_version` → this invocation's `policy` is authoritative (it is the latest version's own import).
- Otherwise, if the view is already pinned to the latest version → return untouched; a delayed retry can no longer re-render the newest version with an older caller's gate.
- Otherwise (view absent or behind) → create/advance carrying the gate the view already had (`None` for a new view), never this caller's.

Combined with the per-attempt re-read of both the view and the latest manifest plus the ETag-conditional write, the interleavings converge: the loser of a create/ETag race re-decides and defers to the latest version's own import. `test_materialize_etag_retry_never_applies_a_stale_policy_to_a_newer_version` exercises the exact v1-pauses/v2-lands/v1-retries interleaving and asserts the v2 pin keeps `policy: None` and v1 issues no view write.

### Regression sweep (0f1f350 → 7b44ce6)

- `_check_view_binding` fail-fast path now passes `manifest_ref`/`location` and may issue one manifest query before tables are created; a view claiming a version with no manifest table is correctly refused as foreign (fail-closed), and the "refuse foreign objects before writing" test still asserts no import tables are created and no writes occur.
- `_read_manifest` gained an optional `feature` label parameter defaulting to `evalbench-import`; existing call sites unchanged.
- Cross-job binding (`existing.pin["job_id"] != job_id`) still raises before any write.
- No-op re-import semantics preserved: canonical policy rendering (`_canonical_policy`) makes the body a pure function of (manifest row, policy pin), so same-version/same-policy re-imports (including reordered keys / int-vs-float thresholds) still skip the write.
- `docs/evalbench.md` updated to match the new ownership rule and the new "policy on a non-latest-version call" row.

No new P0/P1 found. LGTM.

caohy1988 (commented):
**Reviewer: Codex**

**Verdict: Request changes. Two P1s remain at head `7b44ce6`.**

The exact cross-version retry (`v1` policy -> newer `v2` no-policy pin) is fixed, and ownership is now checked against a committed manifest row. The focused suite passes: `135 passed in 3.30s`. However, both root causes still have blocking variants.

1. **P1 - Whitespace normalization still accepts semantically changed foreign SQL** (`src/bigquery_agent_analytics/evalbench.py:1583`; normalizer at `:1469`). `_read_managed_view()` compares the stored and manifest-rendered bodies through `" ".join(text.split())`. That also collapses whitespace inside quoted string literals. With a committed `job_id="job  x"`, a foreign view can keep the matching pin but filter on `job_id="job x"`; BigQuery treats those as different IDs, while the ownership check treats the definitions as equal. I reproduced an `unchanged` materialize call accepting that body and leaving the wrong literal in place. Use exact `view_query` equality (the code says BigQuery preserves it verbatim), or a SQL-aware normalizer that preserves literal contents, and add this regression case.

2. **P1 - Same-version replacements can still attach an older caller's policy** (`src/bigquery_agent_analytics/evalbench.py:1741-1748`). The retry now decides policy authority with `latest_version == import_version`. Two concurrent `replace=True` calls can publish different generations under the same version label. If the older caller used a `0.9` gate and the newer replacement used no gate, the delayed older caller still sees the same version string and writes its stale policy over the newer manifest generation (after a 412 retry, or without a 412 when the newer replacement leaves the view text unchanged). I reproduced the final `v1` pin carrying the older `0.9` policy. Key authority to the exact committed manifest generation, not only `import_version`, and add a deterministic same-version replacement race test.

caohy1988 (commented):
**Reviewer: Kimi**

## Verdict: Approve

Re-reviewed `a20f137` (third P1 fix, delta from approved `7b44ce6`). Both of Codex's remaining P1s are fixed, with direct regression tests, and I found no new P0/P1.

### P1 #1 — whitespace-collapsing ownership compare: FIXED

- `_normalize_sql` is gone from `evalbench.py` (the only remaining hits are an unrelated helper in `export/langsmith.py`).
- `_read_managed_view` now compares `view_query == _failed_sessions_view_body(manifest=pinned, policy=policy, imported_at=pin["imported_at"])` byte-for-byte (`evalbench.py:1631`); the no-op skip in `_reconcile_failed_sessions_view` is likewise exact (`evalbench.py:1827`). The rationale is right: BigQuery returns view text verbatim, and whitespace inside a rendered string literal (`job_id = "job  x"`) changes which rows the view reads, so any normalization is unsound.
- Re-render uses the pin's claimed `imported_at` rather than the manifest row's, which is correct: it lets the importer still recognize its own stale-generation view after a replace and advance it, while a forged body still has to match the importer's rendering exactly.
- Regression test `test_materialize_refuses_a_view_whose_literal_whitespace_changed` forges a genuine pin over whitespace-collapsed SQL and asserts refusal with zero view/manifest writes, for both same-version and new-version callers.

### P1 #2 — same-version replace=True race: FIXED

- The pin now carries the manifest generation: `_pin_timestamp` canonicalizes `imported_at` (UTC, microsecond ISO) and `_manifest_generation` returns `(import_version, imported_at)` (`evalbench.py:1477-1499`). `_sync_failed_sessions_view` passes the generation of the row the call actually committed/found (`evalbench.py:983`, `1739`).
- `_reconcile_failed_sessions_view` decides policy authority by generation, not version label (`evalbench.py:1807-1815`): a delayed older caller whose `replace` was superseded sees `existing.generation == latest_generation` and returns without writing; if the view is behind it advances it carrying the view's existing gate, never its own. Because the generation is rendered into the pin, every new generation changes the view text, so the newer writer always bumps the ETag and the older caller's conditional replace fails and re-decides — the ETag race window the P1 described is closed.
- `_parse_view_pin` refuses non-canonical generations (`evalbench.py:1553`), so a pin can't smuggle an equivalent-but-different timestamp form past the byte-for-byte compare.
- Regression tests: `test_materialize_same_version_replace_race_keeps_the_newer_generations_policy` (interleaved same-version replace race; older caller's gate 0.9 never lands, ETag moves under it) and the parametrized `test_materialize_refuses_a_pin_with_a_malformed_generation`.

### Tests

`python3 -m pytest -q tests/test_evalbench_importer.py tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py` → **144 passed** at `a20f137`.

(Note: `pytest tests/ -k evalbench` hits 3 collection errors in `test_memory_service*.py` / `test_cli_agent_tool.py` — a pre-existing environment issue, `opentelemetry.semconv` missing `GEN_AI_TOOL_DEFINITIONS`, unrelated to this PR.)

### Non-blocking observations

- Byte-for-byte ownership means a managed view cosmetically reformatted outside the importer is now refused rather than re-adopted — deliberate, documented, and fail-closed. Good trade-off.
- Views written by the previous iteration (pin without `imported_at`) are refused as foreign. Fine for an unreleased feature, but worth a one-line mention if any pre-release users exist.
- A pathological tie is still theoretically possible if two callers pass the *same explicit* `imported_at` for the same version — both would claim the same generation. Default `now(UTC)` makes this a non-issue in practice; not blocking.

caohy1988 (commented):
**Reviewer: Codex**

> **Verdict: Request changes. Two P1s remain at `a20f137`; no P0 found.** The byte-for-byte comparison fixes the quoted-literal whitespace bypass from `7b44ce6`, but ownership still trusts policy state supplied by the view itself, and the new manifest-generation key is not unique.

**Scope:** Slice 2 only. **Verification:** `python3 -m pytest -q tests/test_evalbench_importer.py tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py` -> **144 passed in 3.26s**. I reproduced both cases below read-only with the existing fake client.

### P1 - Policy ownership is still circular, so a semantically changed foreign view can authenticate itself

`src/bigquery_agent_analytics/evalbench.py:1550` parses `policy` from the untrusted pin, then `:1631-1633` uses that same policy to construct the supposedly authoritative expected body:

```python
policy = _policy_from_pin(pin.get("policy"))
...
view_query == _failed_sessions_view_body(
    manifest=pinned, policy=policy, imported_at=pin["imported_at"]
)
```

A foreign view can therefore take a committed manifest row, choose any canonical score policy, and render the matching policy SQL and pin. The byte-for-byte comparison succeeds because the expected semantics came from the foreign view. I reproduced this by publishing `v1` and `v2` without a policy, replacing the view with the canonical `v2` body for `goal_completion >= 0.9`, then running an unchanged `v1` materialize. The call returned `unchanged`; `_reconcile_failed_sessions_view()` hit `existing.generation == latest_generation` at `:1812-1813`, and the forged `0.9` gate remained in the view. This contradicts the documented refusal of an edited/foreign managed view and can silently change the failed-session denominator.

Anchor the policy in trusted committed metadata outside the view (or equivalent authoritative view state), and validate the stored definition against that value rather than the policy it claims in its own pin. Add the canonical-policy forgery regression, including a non-latest-version materialize that must fail closed instead of preserving the forged gate.

### P1 - `(import_version, imported_at)` is not a unique committed generation

`src/bigquery_agent_analytics/evalbench.py:1489-1499` defines generation as `(import_version, imported_at)`, but `imported_at` is caller-controlled at `:896-900`. At `:1807-1809`, equality grants the invocation authority to render its policy:

```python
latest_generation = _manifest_generation(latest)
if latest_generation == generation:
  render_policy = policy
```

Two same-version `replace=True` calls can use the same timestamp. Concrete interleaving: older caller A publishes `(v1, T)` with a `0.9` policy and pauses before its view update; newer caller B publishes the same `(v1, T)` with no policy and updates the view; A gets `412`, retries, mistakes B's manifest row for its own generation, and overwrites B's no-policy view with `0.9`. Both calls return `replaced`. My reproduction ended with the final pin carrying A's stale `0.9` policy. The new race test misses this because it gives the callers distinct timestamps at `tests/test_evalbench_importer.py:3116` and `:3134`.

Store and pin an opaque per-publish generation identifier that is unique for every committed replacement; a caller-supplied, microsecond timestamp is not sufficient. Add the same deterministic race with equal `imported_at` values.

caohy1988 (commented):
**Reviewer: Kimi**

**Verdict: Approve.** Re-review of e384695 ("trust committed policy metadata and unique publish generation") against the previously approved a20f137. Both of Codex's remaining P1s are fixed, no new P0/P1 found.

## Verification

- Diff reviewed: a20f137..e384695 (`evalbench.py` +355/-181, tests, CHANGELOG, docs).
- Local HEAD matches PR head e384695.
- `python3 -m pytest -q tests/test_evalbench_importer.py tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py` → **147 passed**.

## P1 #1 — Circular policy ownership: FIXED

The expected body for a view pinned to the *current* manifest generation is now `_committed_view_body(pinned)` (src/bigquery_agent_analytics/evalbench.py:1735), which renders strictly from the committed manifest row — `generation_id` and the new `view_policy` column — never from the pin's own policy claim (`_read_managed_view`, evalbench.py:1695). A canonical rendering of the current generation under any other gate is foreign and raises. `_policy_from_column` (evalbench.py:1520) additionally rejects non-canonical `view_policy` strings, so the trusted column is parsed strictly.

The superseded-generation branch still uses the pin's policy to recognize a view left behind by a re-publish, but that is no longer circular in any trust-relevant way: the generation is an unrecorded random id, and such a view is only ever advanced to the committed rendering by `_reconcile_failed_sessions_view`, never preserved. `test_materialize_refuses_a_canonical_policy_forgery` pins both halves of this (forged current-generation gate refused; forged superseded-generation view re-rendered from the committed row).

## P1 #2 — Non-unique (import_version, imported_at) generation: FIXED

Generation is now `uuid.uuid4().hex` (`_new_generation_id`, evalbench.py:1544), minted by every publish — a `replace` of the same version label included — and by every committed `view_policy` change (`_record_view_policy`, evalbench.py:1849). `imported_at` remains caller-controlled bookkeeping and no longer identifies the generation; `_manifest_generation_id` validates the 32-hex shape and fails closed on legacy rows. The policy UPDATE is keyed on `generation_id = @expected_generation_id`, so it cannot land on a row a concurrent `replace` re-published. `test_materialize_same_version_replace_race_keeps_the_newer_generations_policy` is parametrized over distinct and **equal** `imported_at` values and verifies the delayed caller's gate never lands.

## Non-blocking observations (P2)

- `_LATEST_IMPORT_QUERY` still orders by caller-controlled `imported_at DESC, import_version DESC`. A caller that backdates `imported_at` on a newer publish leaves the view pinned to the older version. Deterministic and documented, and only self-inflicted, but worth a sentence in docs that "latest" means "latest by `imported_at`".
- `_record_view_policy` does not check `@@row_count`; a zero-match (concurrent replace) silently no-ops and reconcile then renders from the latest row. Correct by design, just noting it.
- Manifest schema gains a REQUIRED `generation_id` column with `create_table(exists_ok=True)`; no migration path for manifests created by earlier revisions of this unmerged PR. Fine since nothing has shipped, but a dataset touched by a20f137 needs re-creation.

No regressions found. LGTM.

caohy1988 (commented):
**Reviewer: Codex**

> **Verdict: Request changes.** The opaque `generation_id` fixes the equal-timestamp replacement race, but one branch still authenticates ownership from view-supplied policy state, and the new manifest columns have no upgrade path for tables created by slice 1.

**Scope:** Slice 2 only, head `e384695`; fourth-fix delta `a20f137..e384695`.

**Verification:** `python3 -m pytest -q tests/test_evalbench_importer.py tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py` -> **147 passed in 3.27s**.

### P1 - Superseded-generation ownership is still circular

`src/bigquery_agent_analytics/evalbench.py:1695-1702` uses committed `view_policy` only when the pin's generation equals the manifest's current generation. For every non-current generation, it instead reconstructs the expected body with `policy` and `generation_id` parsed from the candidate view itself:

```python
if pin["generation_id"] == _manifest_generation_id(pinned):
  expected = _committed_view_body(pinned)
else:
  expected = _failed_sessions_view_body(
      manifest=pinned,
      policy=policy,
      generation_id=pin["generation_id"],
  )
```

A foreign writer can therefore choose any well-formed non-current generation plus any canonical policy, render the matching body, and be accepted as `_ManagedView`. The new regression itself demonstrates the gap at `tests/test_evalbench_importer.py:3319-3336`: changing the forged gate to generation `"f" * 32` makes `materialize()` accept and replace it rather than refuse it. Reconciliation removes that gate, but the importer has still authenticated and overwritten a view it did not create, contrary to the ownership guarantee.

Persist enough committed generation/policy history to authenticate the stale definition, or fail closed on a generation that cannot be verified. No ownership decision should derive its authority from the candidate view's own policy/generation claims.

### P1 - Existing slice-1 manifest tables are not migrated

`src/bigquery_agent_analytics/evalbench.py:196-200` adds required `generation_id` and nullable `view_policy` columns, but `_ensure_import_tables()` at `:2513-2516` still only calls `create_table(..., exists_ok=True)`. For an existing `evalbench_import_manifest` created by PR #451, that call returns the old table unchanged; it does not add columns. An unchanged import then fails `_manifest_generation_id()` because the stored row has no generation, while a new/replaced publish targets columns absent from the table.

This breaks the PR's documented "merge after #451" path for every dataset used between the two slices. Add an idempotent schema migration/backfill before manifest/view reads and publishes, plus an upgrade test starting from the pre-`e384695` manifest schema.

The prior non-unique-generation P1 is fixed: each publish and committed policy change now mints an opaque UUID, and the equal-`imported_at` race test passes.

caohy1988 (commented):
**Reviewer: Kimi**

# Re-review: PR #452 @ a354f1b (slice 2, `e384695..a354f1b`)

## Verdict: APPROVE

Both Codex P1s on e384695 are genuinely fixed, and I found no new P0/P1 in the delta.

## P1 #1 — superseded-generation ownership circularity: FIXED

The circularity is gone. `_read_managed_view` now discards the pin's policy claim at parse time (`pin, _ = parsed`) and calls `_committed_generation_view_body(pinned, pin["generation_id"])`:

- Pin names the row's current generation → body rendered from the row's committed `view_policy` (unchanged, correct).
- Pin names anything else → body rendered only if `pin["generation_id"]` appears in the row's committed `superseded_generations` history, using the policy recorded there; otherwise the helper returns `None` and the view is refused ("exists but is not an evalbench failed_sessions view"). A foreign writer who mints a well-formed but never-committed generation now fails closed instead of being accepted via their own policy/generation claims.
- The history itself is committed state: `_PUBLISH_SCRIPT` computes `prior_generations` inside the transaction with a `SELECT` on the existing manifest row *before* the `DELETE` (ordering verified: SET at script line ~412, manifest DELETE at ~424), and `_RECORD_VIEW_POLICY_QUERY` appends the superseded `(generation_id, view_policy)` under the `generation_id = @expected_generation_id` guard, so a lost race cannot land on a re-published row.
- `_ManagedView` no longer carries `policy`, so no downstream consumer can act on a view's self-claim.
- Regression coverage: `test_materialize_authenticates_a_superseded_view_from_committed_history` (and the existing forged/superseded refusal tests still pass in the 151-test delta).

## P1 #2 — slice-1 manifest migration: FIXED

- `generation_id` is now NULLABLE (addable to pre-existing tables; BigQuery rejects adding REQUIRED columns), `superseded_generations` is REPEATED (also addable).
- `_ensure_manifest_schema` runs at the top of `materialize` (line ~1039), before the view binding check (~1070), table creation (~1057 ordering note: schema ensure precedes `_ensure_import_tables` per the grep'd layout), and before any manifest read/publish. It adds missing columns via the tables API with ETag-conditional `update_table`, retries on `PreconditionFailed` (3 attempts, re-reading each time), then runs `_backfill_generation_ids`.
- The backfill is deterministic (`TO_HEX(MD5('evalbench-import-manifest:' + import_version + ':' + job_id))`, `WHERE generation_id IS NULL`), so two concurrent upgrader converge — idempotent whichever commits first.
- Interrupted upgrade (columns added, backfill crashed) is covered: `materialize` re-checks the existing row's `generation_id` and finishes the backfill + re-read; `_reconcile_failed_sessions_view` does the same for the latest row before rendering. I traced the residual case of a NULL-generation row on a *non-current, non-latest* version: no managed view can legitimately be pinned to it (the reconcile loop backfills before ever rendering a pin), so `_committed_generation_view_body` returning `None` there is correct fail-closed behavior, not a false positive.
- Upgrade coverage: `test_materialize_upgrades_a_slice_1_manifest_in_place`, `test_materialize_finishes_an_interrupted_manifest_upgrade`, `test_materialize_tolerates_a_concurrent_manifest_upgrade`.

## New-issue scan (P0/P1 only): none found

Checked specifically:

- Publish script statement ordering — `prior_generations` is captured before the manifest DELETE; correct.
- History entry for a replace of a backfilled slice-1 row — appended with its (possibly NULL) `view_policy`; the strict parser `_superseded_generations` accepts NULL policy and skips generation-less entries, both safe.
- Backfill vs. concurrent publish transaction — worst case is BigQuery cancelling one; the loser errors out and retries cleanly. Acceptable.
- Post-publish result now re-reads the committed manifest row, so the reported manifest includes the transaction-computed history; the pre-read is only a defensive fallback.
- The `staging` row's `superseded_generations: []` is never inserted verbatim — `_STAGED_MANIFEST_COLUMNS` excludes it and the publish script supplies `prior_generations`. Verified.

Non-blocking observations (P2, do not gate): `superseded_generations` grows unboundedly across many `replace`/policy cycles (one small JSON string per generation — fine in practice); `_ensure_manifest_schema`'s early return when the schema is complete relies on the per-row NULL checks downstream for the interrupted-backfill case — the current checks cover every path I traced, but a comment pointing from one to the other would help future readers.

Tests: 151 delta tests passed per the run reported with this push (I did not re-run them, per instructions).

**Approve.** Both P1s are resolved with committed-history authentication and an idempotent, race-safe slice-1 upgrade; no remaining P0/P1.

caohy1988 (commented):
**Reviewer: Codex**

APPROVE

No remaining P0/P1 findings in `e384695..a354f1b`.

### Re-check of previous P1s

1. Fixed. `_read_managed_view` now derives the expected body for a superseded generation only from the manifest row's committed `superseded_generations` entry. The candidate view's policy claim is ignored as authority, and `_committed_generation_view_body` returns no expected body for a generation absent from committed history. The new regressions refuse an unknown-generation forgery across unchanged, replace, and new-version paths, and refuse a real superseded generation rendered with any policy other than the one in committed history.

2. Fixed. `materialize()` invokes `_ensure_manifest_schema` before managed-view and manifest-row reads or publishing. The migration adds the three slice-2 fields with an ETag-conditional, bounded schema update and performs a deterministic, table-wide `generation_id` backfill. The recovery paths finish an interrupted backfill when a touched or latest row still has no generation. The new tests cover a pre-slice-2 manifest, idempotent re-entry, policy/history evolution after upgrade, interrupted upgrade completion, concurrent schema upgrade convergence, and fail-closed retry exhaustion.

### Actionable Findings

None.

### Coverage

- Scope: `src/bigquery_agent_analytics/evalbench.py` and `tests/test_evalbench_importer.py` only.
- Reviewed ownership authentication, committed generation/policy history, policy updates, replace publishes, schema migration/backfill ordering, partial/concurrent upgrade paths, and the new fake-client regressions.
- Independent correctness, security, testing, reliability, and cross-model adversarial passes completed. No finding survived P0/P1 reconciliation.
- Validation batch: not run because no P0/P1 finding survived reconciliation.
- Pytest was skipped as requested; the supplied result was 151 passing tests.
- P2 and lower observations were not treated as blockers and are omitted.

---

### Verdict

APPROVE. Both previously reported P1s are resolved, with no new P0/P1 in the specified delta.

Actionable findings: none.

caohy1988 (commented):
**Reviewer: Codex**

APPROVE

No remaining P0/P1 findings in `e384695..a354f1b`.

### Re-check of previous P1s

1. Fixed. `_read_managed_view` now derives the expected body for a superseded generation only from the manifest row's committed `superseded_generations` entry. The candidate view's policy claim is ignored as authority, and `_committed_generation_view_body` returns no expected body for a generation absent from committed history. The new regressions refuse an unknown-generation forgery across unchanged, replace, and new-version paths, and refuse a real superseded generation rendered with any policy other than the one in committed history.

2. Fixed. `materialize()` invokes `_ensure_manifest_schema` before managed-view and manifest-row reads or publishing. The migration adds the three slice-2 fields with an ETag-conditional, bounded schema update and performs a deterministic, table-wide `generation_id` backfill. The recovery paths finish an interrupted backfill when a touched or latest row still has no generation. The new tests cover a pre-slice-2 manifest, idempotent re-entry, policy/history evolution after upgrade, interrupted upgrade completion, concurrent schema upgrade convergence, and fail-closed retry exhaustion.

### Actionable Findings

None.

### Coverage

- Scope: `src/bigquery_agent_analytics/evalbench.py` and `tests/test_evalbench_importer.py` only.
- Reviewed ownership authentication, committed generation/policy history, policy updates, replace publishes, schema migration/backfill ordering, partial/concurrent upgrade paths, and the new fake-client regressions.
- Independent correctness, security, testing, reliability, and cross-model adversarial passes completed. No finding survived P0/P1 reconciliation.
- Validation batch: not run because no P0/P1 finding survived reconciliation.
- Pytest was skipped as requested; the supplied result was 151 passing tests.
- P2 and lower observations were not treated as blockers and are omitted.

---

### Verdict

APPROVE. Both previously reported P1s are resolved, with no new P0/P1 in the specified delta.

Actionable findings: none.

caohy1988 (commented):
**Reviewer: Kimi**

# PR 452 re-review — rebase onto main @ 2459c0f (slice 2, #435)

## Verdict: APPROVE — P0: 0, P1: 0

## Rebase integrity (the only thing that could have regressed)

| Check | Result |
|---|---|
| PR head on GitHub | `d681d548519c…` — matches local HEAD exactly |
| PR base on GitHub | `2459c0f00898…` (PR 451 squash-merge) — matches stated base |
| Commit list on GitHub | `a16cdb1 2619288 675cd77 9779f3c 6674971 d681d54` — matches local, in order |
| Final tree vs previously approved `a354f1b` | **byte-identical** — both trees are `d1df367aa35227ecdd0573a75deeaa866212c261`; `git diff a354f1b d681d54` is empty |
| Old base vs new base | `a354f1b`'s base `2e64fdb` and new base `2459c0f` share the **same tree** (`06deab83…`) — 2459c0f is the squash-merge of that branch, so the rebase had nothing to resolve |
| Conflict-marker scan (`<<<<<<<`/`>>>>>>>`) in `evalbench.py`, `cli.py`, `test_evalbench_importer.py` | clean |

Because the tree is hash-identical to the previously dual-approved `a354f1b`, every prior P1 fix holds by construction — the rebase could not have broken them without changing the tree. Spot-checked anyway: `superseded_generations` history, `_ensure_manifest_schema` slice-1 manifest upgrade, `_read_managed_view` committed-record authentication, and the CLI `failed_sessions` / pinned-view wiring are all present at HEAD.

## Tests (re-run this session, worktree src via PYTHONPATH)

```
PYTHONPATH=src python -m pytest -q tests/test_evalbench_importer.py \
    tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py
→ 151 passed in 1.65s
```

Note: system `python3` has no pytest in this environment; ran with the project venv at `/Users/haiyuancao/BigQuery-Agent-Analytics-SDK/.venv/bin/python` with `PYTHONPATH=src` so the worktree's sources (not the editable install pointing at the main checkout) are under test.

## Remaining P0/P1

None. No new findings from the rebase; prior P1 fixes verified intact via tree-hash identity with `a354f1b` plus green tests at the new HEAD.

Not posted to GitHub; no push, no amend, per instructions.

caohy1988 (commented):
**Reviewer: Codex**

## Code Review Results

**Verdict:** LGTM
**P0 count:** 0
**P1 count:** 0
**Scope:** PR #452 slice 2 only, `2459c0f..d681d54` (7 files)
**Mode:** P0/P1-only, report-only; no GitHub post, push, commit, or amendment

**Pytest result:** `python3 -m pytest -q tests/test_evalbench_importer.py tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py` -> **151 passed in 4.43s**.

### Findings

P0/P1 findings with file:line: **none**.

### Rebase verification

- `git range-diff 2e64fdb..a354f1b 2459c0f..d681d54` reports all six slice-2 commits as patch-equivalent.
- The old and new base trees are identical: `2e64fdb^{tree}` = `2459c0f^{tree}` = `06deab838e54e297a8f78d1bb88b023e28182f65`.
- The previously approved and rebased head trees are identical: `a354f1b^{tree}` = `d681d54^{tree}` = `d1df367aa35227ecdd0573a75deeaa866212c261`.
- `git diff a354f1b d681d54` is empty. `git diff --check 2459c0f` is clean, and the changed files contain no conflict markers.

### Re-check of settled P1 fixes

- Committed-history ownership remains intact at `src/bigquery_agent_analytics/evalbench.py:1778`: `_read_managed_view()` authenticates the body through `_committed_generation_view_body()` using the manifest's current generation or committed superseded history (`src/bigquery_agent_analytics/evalbench.py:1878`).
- The slice-1 manifest upgrade still runs before view binding or manifest reads at `src/bigquery_agent_analytics/evalbench.py:1039`; the bounded ETag-aware schema upgrade and generation backfill remain at `src/bigquery_agent_analytics/evalbench.py:2659` and `src/bigquery_agent_analytics/evalbench.py:2702`.
- The prior concurrency and policy-generation fixes remain unchanged, including create-if-absent, ETag-conditional view updates, and bounded retry at `src/bigquery_agent_analytics/evalbench.py:2049`.

### Coverage

- Reviewed the rebase equivalence, managed-view authentication, committed generation/policy history, manifest upgrade/backfill, optimistic-concurrency paths, version-pinned consumer/CLI contracts, BigQuery query shape, and focused regression tests.
- Local correctness, security, testing, API-contract, reliability, performance, maintainability, and prior-feedback passes found no P0/P1.
- Independent cross-model pass: route `claude`; requested/actual model `claude-opus-5`; requested effort `high`; actual effort `unverified`; receipt supported; independence verified. One independently validated narrow input edge case and its test gap were calibrated below P1 and omitted under the requested P0/P1-only scope.
- Project standards review was not run because no applicable `CODING_STANDARDS.md`, `CLAUDE.md`, or `AGENTS.md` file governs the changed paths.

### Actionable Findings

None.

---

### Verdict

LGTM. The rebase is tree- and patch-equivalent to the previously dual-approved head, the settled P1 fixes remain intact, the requested 151 focused tests pass, and no new P0/P1 remains at `d681d54`.

### Actionable Findings (prioritized recap)

Actionable findings: none.

haiyuan-eng-google (approved):

PR #453, opened 2026-08-30
feat(evalbench): evalbench-score CLI wrapping Client.evaluate (#435, #97)

### LINKED ISSUE #97: Add EvalBench bridge: import EvalBench BigQuery runs into a BQAA-owned mirror table

caohy1988:
## Summary
Slice 3 of #435 (W0.2 leftover / remaining #97). Thin `bq-agent-sdk evalbench-score` CLI wrapping existing `Client.evaluate` + `LLMAsJudge` prebuilts over the EvalBench mirror table, filtered by `TraceFilter(experiment_id=job_id)` and pinned to one import version (latest successful, or `--import-version`). `--exit-code` reuses the evaluate CI gate.

## Depends on
#451 (slice 1 writer) and #452 (slice 2 failed_sessions view / version pin). Stacked on #452. Merge after those, or together.

Closes the W0.2 leftover of #97 for the score CLI; does not close #97 or #435 (taxonomy / localization / live harness remain).

## Out of scope
Taxonomy, localization, live W0.3 harness, dashboards, D4, partner-job import, modifying `Client.evaluate`.

## Test plan
- [x] Unit tests stub Client / manifest (no live BigQuery): wiring, version pin, `--strict`, `--exit-code` pass/fail, invalid evaluator, BQ error
- [ ] Codex + Kimi review this PR on GitHub

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **EvalBench LLM-judge scoring of one import version (#435 slice 3, #97)**
+  — new `bq-agent-sdk evalbench-score` command: a thin wrapper over the
+  existing `Client.evaluate` + `LLMAsJudge` (`correctness` |
+  `hallucination` | `sentiment`, `--threshold`, `--strict`, `--exit-code`
+  with the same exit codes and `FAIL` lines as `evaluate`) that points the
+  client at the mirror events table and scores exactly one published
+  version of `--job-id` — the latest successful import unless
+  `--import-version` pins one. `evalbench.import_sessions()` resolves the
+  version from the manifest and lists its session identities;
+  `EvalBenchImportSessions.trace_filter()` renders them as
+  `TraceFilter(experiment_id=job_id, session_ids=..., limit=...)`, so the
+  version pin reaches `Client.evaluate` without changing it. The report's
+  `details.evalbench` names the scored job, version, table, and pinned
+  session count. Reference in `docs/evalbench.md`; CI gate example in
+  `examples/evalbench_score_gate.sh`.
 - **EvalBench failed-session view and version-pinned consumer (#435, slice 2)**
   — `EvalBenchRun.materialize()` now keeps an `evalbench_failed_sessions`
   view in the target dataset pinned (as literals) to the job's latest
--- a/docs/evalbench.md
+++ b/docs/evalbench.md
@@ -494,6 +494,48 @@ selector is unambiguous for the identity-resolving reader. Extra
 `get_session_trace` keyword arguments (for example `event_types=`) pass
 through `session.get_trace(client, ...)`.
 
+## Score An Import With The LLM Judge (#97)
+
+`bq-agent-sdk evalbench-score` is a thin wrapper over the existing
+`Client.evaluate` + `LLMAsJudge` path: nothing new is computed, the judge
+is simply pointed at one published import version of the mirror table.
+
+```python
+from bigquery_agent_analytics import Client
+from bigquery_agent_analytics.evalbench import import_sessions
+from bigquery_agent_analytics.evaluators import LLMAsJudge
+
+pinned = import_sessions(
+    target_project="analytics-project",
+    target_dataset="bqaa",
+    job_id="abc123",
+    # import_version="v1",   # pin one version; default: latest successful import
+)
+client = Client(
+    project_id="analytics-project",
+    dataset_id="bqaa",
+    table_id="evalbench_agent_events",   # must equal pinned.events_table
+)
+report = client.evaluate(
+    evaluator=LLMAsJudge.correctness(threshold=0.7),
+    filters=pinned.trace_filter(),
+)
+print(pinned.import_version, report.pass_rate, report.total_sessions)
+```
+
+`import_sessions()` resolves the version from the manifest exactly as
+`failed_sessions()` does (an explicit `import_version` must be published;
+the default is the job's latest successful import) and then reads that
+version's distinct `session_id` values from the `events_table` recorded in
+the manifest row. `EvalBenchImportSessions.trace_filter()` returns
+`TraceFilter(experiment_id=job_id, session_ids=<those ids>, limit=<count>)`:
+`TraceFilter` has no import-version dimension, so the version pin reaches
+`Client.evaluate` through the exact versioned session identities, which
+retained versions of one job never share. `trace_filter()` refuses an
+empty session set because `TraceFilter` treats "no `session_ids`" as
+unfiltered, which would silently widen the evaluation to every retained
+version of the job. `Client.evaluate` itself is unchanged.
+
 ## CLI
 
 ```bash
@@ -529,8 +571,35 @@ Prints the `EvalBenchFailedSessions` listing (`--format table` prints one
 row per session) for exactly one published version — the job's latest
 successful import unless `--import-version` pins one — and exits `0` when
 the listing was produced (possibly empty) or `2` on invalid input, a job or
-version with no published import, or a BigQuery error. There is no
-`evalbench-score` command.
+version with no published import, or a BigQuery error.
+
+```bash
+bq-agent-sdk evalbench-score \
+  --project-id analytics-project --dataset-id bqaa --job-id abc123 \
+  [--table-id evalbench_agent_events] [--import-version v1] \
+  [--evaluator correctness|hallucination|sentiment] [--threshold 0.7] \
+  [--strict] [--exit-code] [--endpoint gemini-2.5-flash] [--connection-id ID] \
+  [--location US] [--format json|text|table]
+```
+
+Runs the chosen `LLMAsJudge` (default `correctness`, judge default
+threshold `0.5`) through `Client.evaluate` over the mirror `--table-id`,
+narrowed to one published version of `--job-id` (the latest successful
+import unless `--import-version` pins one). The output is the ordinary
+`EvaluationReport` (`--format text` prints its summary) with
+`details.evalbench` naming the scored `job_id`, `import_version`,
+`events_table`, and `pinned_sessions`, so a scorecard can always be traced
+back to the version it judged. `--strict`, `--endpoint`, and
+`--connection-id` behave as they do for `bq-agent-sdk evaluate`.
+
+Exit codes match `evaluate`: `0` when the scorecard was produced (failing
+sessions included), `1` with `--exit-code` when at least one session failed
+the threshold (the same `FAIL session=... metric=... feedback="..."` lines
+are printed to stderr first), and `2` on invalid input — an unknown
+`--evaluator`, the reserved `agent_events` table, a job or version with no
+published import, a `--table-id` the version was not published to, or a
+version with no sessions — or a BigQuery error. `examples/evalbench_score_gate.sh`
+shows the CI gate form.
 
 ## Why These Fields Matter
 
--- a/examples/evalbench_score_gate.sh
+++ b/examples/evalbench_score_gate.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Example: gate a build on the LLM-judge score of one EvalBench import.
+#
+# Scores exactly one published import version of an EvalBench job that was
+# mirrored with `bq-agent-sdk evalbench-import` (#435). The judge runs
+# through the ordinary `Client.evaluate` path over the mirror table; the
+# ADK plugin's production `agent_events` table is never read.
+#
+# Prerequisites:
+#   pip install bigquery-agent-analytics
+#   export BQ_AGENT_PROJECT=analytics-project   # project holding the mirror
+#   export BQ_AGENT_DATASET=bqaa                # BQAA-owned target dataset
+#   export EVALBENCH_JOB_ID=abc123              # job already imported
+#
+# Optional:
+#   export EVALBENCH_IMPORT_VERSION=v1          # default: latest successful import
+#   export EVALBENCH_JUDGE=correctness          # correctness|hallucination|sentiment
+#   export EVALBENCH_THRESHOLD=0.7              # judge default 0.5
+#
+# Usage:
+#   bash examples/evalbench_score_gate.sh
+#
+# Exit codes (same as `bq-agent-sdk evaluate`):
+#   0 every session passed the judge threshold
+#   1 at least one session failed (FAIL lines are printed to stderr)
+#   2 invalid input (unknown judge, unpublished job/version, wrong table)
+#     or a BigQuery error
+
+set -euo pipefail
+
+: "${BQ_AGENT_PROJECT:?set BQ_AGENT_PROJECT}"
+: "${BQ_AGENT_DATASET:?set BQ_AGENT_DATASET}"
+: "${EVALBENCH_JOB_ID:?set EVALBENCH_JOB_ID}"
+
+args=(
+  --job-id "${EVALBENCH_JOB_ID}"
+  --evaluator "${EVALBENCH_JUDGE:-correctness}"
+  --threshold "${EVALBENCH_THRESHOLD:-0.5}"
+  --exit-code
+  --format text
+)
+if [[ -n "${EVALBENCH_IMPORT_VERSION:-}" ]]; then
+  args+=(--import-version "${EVALBENCH_IMPORT_VERSION}")
+fi
+
+echo "=== EvalBench LLM-judge gate: job ${EVALBENCH_JOB_ID} ==="
+bq-agent-sdk evalbench-score "${args[@]}"
--- a/src/bigquery_agent_analytics/cli.py
+++ b/src/bigquery_agent_analytics/cli.py
@@ -33,6 +33,7 @@ Usage::
 from __future__ import annotations
 
 import json
+import math
 from pathlib import Path
 import sys
 from typing import Any, Optional
@@ -2590,7 +2591,8 @@ def evalbench_score(
       0 — the scorecard was produced (failures included, without
           ``--exit-code``).
       1 — ``--exit-code`` and at least one session failed the threshold.
-      2 — invalid input (unknown ``--evaluator``, the reserved
+      2 — invalid input (unknown ``--evaluator``, a ``--threshold``
+          outside ``[0.0, 1.0]`` or non-finite, the reserved
           ``agent_events`` table, a job or version with no published
           import, a ``--table-id`` the version was not published to, a
           version with no sessions), or a BigQuery error.
@@ -2609,6 +2611,17 @@ def evalbench_score(
       raise typer.Exit(code=2)
     # Reject the reserved ADK plugin table before any BigQuery call.
     _validate_destination_table("table_id", table_id)
+    # Judge scores are 0-1; a NaN/inf or out-of-range threshold would
+    # otherwise reach the judge and, with --exit-code, green-gate CI.
+    if threshold is not None and not (
+        math.isfinite(threshold) and 0.0 <= threshold <= 1.0
+    ):
+      typer.echo(
+          "Error: --threshold must be a finite value in [0.0, 1.0], got"
+          f" {threshold!r}.",
+          err=True,
+      )
+      raise typer.Exit(code=2)
     with_t, without_t = entry
     judge = with_t(threshold) if threshold is not None else without_t()
 
--- a/src/bigquery_agent_analytics/cli.py
+++ b/src/bigquery_agent_analytics/cli.py
@@ -2507,6 +2507,156 @@ def evalbench_failed_sessions(
     raise typer.Exit(code=2)
 
 
+@app.command("evalbench-score")
+def evalbench_score(
+    project_id: str = typer.Option(
+        ...,
+        envvar="BQ_AGENT_PROJECT",
+        help="Project holding the BQAA-owned dataset the job was imported into.",
+    ),
+    dataset_id: str = typer.Option(
+        ...,
+        envvar="BQ_AGENT_DATASET",
+        help="BQAA-owned dataset holding the mirror tables and manifest.",
+    ),
+    table_id: str = typer.Option(
+        "evalbench_agent_events",
+        "--table-id",
+        help=(
+            "Mirror event table the version was published to (never the ADK"
+            " plugin's agent_events)."
+        ),
+    ),
+    job_id: str = typer.Option(
+        ..., "--job-id", help="EvalBench job_id to score."
+    ),
+    evaluator: str = typer.Option(
+        "correctness",
+        "--evaluator",
+        help="LLM judge: correctness|hallucination|sentiment.",
+    ),
+    threshold: Optional[float] = typer.Option(
+        None,
+        "--threshold",
+        help="Pass/fail threshold 0-1 (uses the judge default if omitted).",
+    ),
+    import_version: Optional[str] = typer.Option(
+        None,
+        "--import-version",
+        help=(
+            "Score one published import version. Defaults to the job's"
+            " latest successful import (the version the failed-session"
+            " view tracks)."
+        ),
+    ),
+    strict: bool = typer.Option(
+        False,
+        "--strict",
+        help=(
+            "Stamp parse-error metadata on judge rows with empty or NULL"
+            " typed output (same as `evaluate --strict`)."
+        ),
+    ),
+    exit_code: bool = typer.Option(
+        False,
+        "--exit-code",
+        help="Return exit code 1 when any session fails the judge threshold.",
+    ),
+    endpoint: Optional[str] = typer.Option(
+        None, "--endpoint", help="AI.GENERATE endpoint for the LLM judge."
+    ),
+    connection_id: Optional[str] = typer.Option(
+        None, "--connection-id", help="BQ connection ID for AI.GENERATE."
+    ),
+    location: Optional[str] = typer.Option(
+        None, "--location", help="BigQuery location."
+    ),
+    fmt: str = typer.Option(
+        "json", "--format", help="Output format: json|text|table."
+    ),
+) -> None:
+  """Score one published EvalBench import version with the LLM judge.
+
+  A thin wrapper over ``Client.evaluate`` + ``LLMAsJudge``: the client is
+  pointed at the mirror events table, the version is resolved from the
+  import manifest exactly as ``evalbench-failed-sessions`` does (latest
+  successful import unless ``--import-version`` pins one), and the judge
+  runs over ``TraceFilter(experiment_id=job_id)`` narrowed to that version's
+  session ids, so retained versions of one job are never mixed. The report
+  is the ordinary ``EvaluationReport`` with ``details.evalbench`` naming the
+  scored job, version, table, and pinned session count.
+
+  Exit codes:
+      0 — the scorecard was produced (failures included, without
+          ``--exit-code``).
+      1 — ``--exit-code`` and at least one session failed the threshold.
+      2 — invalid input (unknown ``--evaluator``, the reserved
+          ``agent_events`` table, a job or version with no published
+          import, a ``--table-id`` the version was not published to, a
+          version with no sessions), or a BigQuery error.
+  """
+  try:
+    from .evalbench import _validate_destination_table
+    from .evalbench import import_sessions
+
+    entry = _LLM_JUDGES.get(evaluator)
+    if not entry:
+      typer.echo(
+          f"Error: unknown evaluator: {evaluator!r}; expected"
+          " correctness|hallucination|sentiment.",
+          err=True,
+      )
+      raise typer.Exit(code=2)
+    # Reject the reserved ADK plugin table before any BigQuery call.
+    _validate_destination_table("table_id", table_id)
+    with_t, without_t = entry
+    judge = with_t(threshold) if threshold is not None else without_t()
+
+    client = _build_client(
+        project_id,
+        dataset_id,
+        table_id,
+        location,
+        endpoint=endpoint,
+        connection_id=connection_id,
+    )
+    pinned = import_sessions(
+        target_project=project_id,
+        target_dataset=dataset_id,
+        job_id=job_id,
+        import_version=import_version,
+        location=location,
+        bq_client=client.bq_client,
+    )
+    events_ref = f"{project_id}.{dataset_id}.{table_id}"
+    if pinned.events_table != events_ref:
+      raise ValueError(
+          f"EvalBench job {job_id!r} import_version"
+          f" {pinned.import_version!r} is published in"
+          f" {pinned.events_table!r}, not {events_ref!r}; pass the"
+          " --table-id it was published to"
+      )
+    report = client.evaluate(
+        evaluator=judge, filters=pinned.trace_filter(), strict=strict
+    )
+    report.details["evalbench"] = {
+        "job_id": pinned.job_id,
+        "import_version": pinned.import_version,
+        "events_table": pinned.events_table,
+        "pinned_sessions": pinned.session_count,
+    }
+    typer.echo(format_output(report, fmt))
+
+    if exit_code and report.pass_rate < 1.0:
+      _emit_evaluate_failures(report)
+      raise typer.Exit(code=1)
+  except typer.Exit:
+    raise
+  except Exception as exc:  # noqa: BLE001
+    typer.echo(f"Error: {exc}", err=True)
+    raise typer.Exit(code=2)
+
+
 @bqaa_app.command("seed-events")
 def seed_events(
     project_id: str = typer.Option(
--- a/src/bigquery_agent_analytics/cli.py
+++ b/src/bigquery_agent_analytics/cli.py
@@ -2633,11 +2633,16 @@ def evalbench_score(
         endpoint=endpoint,
         connection_id=connection_id,
     )
+    # ``events_table`` makes ``import_sessions`` check the manifest's
+    # binding against --table-id (and re-validate the stored reference)
+    # before any events-table query is submitted; the comparison below is
+    # only a second line of defense on the returned pin.
     pinned = import_sessions(
         target_project=project_id,
         target_dataset=dataset_id,
         job_id=job_id,
         import_version=import_version,
+        events_table=table_id,
         location=location,
         bq_client=client.bq_client,
     )
--- a/src/bigquery_agent_analytics/evalbench.py
+++ b/src/bigquery_agent_analytics/evalbench.py
@@ -52,7 +52,7 @@ import json
 import logging
 import math
 import re
-from typing import Any, Optional
+from typing import Any, Optional, TYPE_CHECKING
 import uuid
 
 from google.api_core.exceptions import Conflict
@@ -62,6 +62,9 @@ from google.cloud import bigquery
 
 from ._telemetry import make_bq_client
 from ._telemetry import with_sdk_labels
+
+if TYPE_CHECKING:
+  from .trace import TraceFilter
 
 _SOURCE_SEGMENT_PATTERN = re.compile(r"^[A-Za-z0-9_-]+$")
 _MISSING_TEXT = frozenset({"", "<na>", "nan", "none", "null"})
@@ -140,6 +143,7 @@ _VIEW_FEATURE = "evalbench-failed-sessions"
 # BQAA-owned mirror tables, so this name is rejected before any BigQuery call.
 _RESERVED_DESTINATION_TABLES = frozenset({"agent_events"})
 _IMPORT_FEATURE = "evalbench-import"
+_SCORE_FEATURE = "evalbench-score"
 _STAGING_TABLE_TTL = timedelta(hours=6)
 _LOGGER = logging.getLogger(__name__)
 _IMPORT_VERSION_PATTERN = re.compile(r"^[A-Za-z0-9_.:-]{1,128}$")
@@ -260,6 +264,19 @@ FROM `{manifest_table}`
 WHERE job_id = @job_id
 ORDER BY imported_at DESC, import_version DESC
 LIMIT 1
+"""
+
+# Session identities of one published import version, read from the events
+# table the manifest bound that version to. ``TraceFilter`` has no
+# import-version dimension, so ``Client.evaluate`` is pinned to a version
+# through these exact ``session_id`` values (plus the ``experiment_id`` job
+# pin); retained versions of one job never share a session id
+# (``_session_identity``), so the pin cannot admit another version's rows.
+_IMPORT_SESSIONS_QUERY = """\
+SELECT DISTINCT session_id
+FROM `{events_table}`
+WHERE job_id = @job_id AND import_version = @import_version
+ORDER BY session_id
 """
 
 # Commits another score gate for an already-published, identical version
@@ -2134,13 +2151,14 @@ def _read_latest_manifest(
     manifest_ref: str,
     job_id: str,
     location: Optional[str],
+    feature: str = _VIEW_FEATURE,
 ) -> Optional[dict[str, Any]]:
   job_config = bigquery.QueryJobConfig(
       query_parameters=[
           bigquery.ScalarQueryParameter("job_id", "STRING", job_id)
       ]
   )
-  job_config = with_sdk_labels(job_config, feature=_VIEW_FEATURE)
+  job_config = with_sdk_labels(job_config, feature=feature)
   query_args: dict[str, Any] = {"job_config": job_config}
   if location is not None:
     query_args["location"] = location
@@ -2312,6 +2330,164 @@ def failed_sessions(
       sessions=rows if include_passed else failed,
       session_count=len(rows),
       failed_count=len(failed),
+      manifest=dict(manifest),
+  )
+
+
+def _resolve_manifest(
+    client: Any,
+    *,
+    manifest_ref: str,
+    job_id: str,
+    import_version: Optional[str],
+    location: Optional[str],
+    feature: str,
+) -> tuple[str, dict[str, Any]]:
+  """Pin ``(job_id, import_version)`` to one committed manifest row.
+
+  An explicit ``import_version`` must have a manifest row; ``None`` resolves
+  the job's latest successful import (newest ``imported_at``, the same pin
+  the ``evalbench_failed_sessions`` view tracks). Raises ``ValueError``
+  before any event row is read when nothing is published.
+  """
+  if import_version is None:
+    manifest = _read_latest_manifest(
+        client,
+        manifest_ref=manifest_ref,
+        job_id=job_id,
+        location=location,
+        feature=feature,
+    )
+    if manifest is None:
+      raise ValueError(
+          f"EvalBench job {job_id!r} has no published import in"
+          f" {manifest_ref!r}; run materialize() first"
+      )
+    return str(manifest["import_version"]), manifest
+  _validate_import_version(import_version)
+  manifest = _read_manifest(
+      client,
+      manifest_ref=manifest_ref,
+      job_id=job_id,
+      import_version=import_version,
+      location=location,
+      feature=feature,
+  )
+  if manifest is None:
+    raise ValueError(
+        f"EvalBench job {job_id!r} import_version {import_version!r} is not"
+        f" published in {manifest_ref!r}"
+    )
+  return import_version, manifest
+
+
+@dataclasses.dataclass(frozen=True)
+class EvalBenchImportSessions:
+  """The sessions of exactly one published EvalBench import version.
+
+  Returned by ``import_sessions``. ``session_ids`` are the versioned
+  identities (``evalbench-import:{job_id}:{import_version}:{scenario_id}``)
+  the mirror ``events_table`` holds for this version, in ``session_id``
+  order; ``manifest`` is the committed manifest row the version was
+  resolved from. ``trace_filter()`` turns the listing into the
+  ``TraceFilter`` that scores this version -- and only this version -- with
+  ``Client.evaluate``.
+  """
+
+  job_id: str
+  import_version: str
+  events_table: str
+  scores_table: str
+  session_ids: tuple[str, ...]
+  manifest: dict[str, Any] = dataclasses.field(default_factory=dict)
+
+  @property
+  def session_count(self) -> int:
+    return len(self.session_ids)
+
+  def trace_filter(self) -> "TraceFilter":
+    """``TraceFilter`` selecting exactly this version's sessions.
+
+    Pins ``experiment_id`` to the job and ``session_ids`` to the version's
+    identities, with ``limit`` raised to the session count so no session is
+    truncated. Refuses an empty session set: ``TraceFilter`` treats no
+    ``session_ids`` as "unfiltered", which would silently widen the
+    evaluation to every retained version of the job.
+    """
+    from .trace import TraceFilter  # local: keep evalbench import-light
+
+    if not self.session_ids:
+      raise ValueError(
+          f"EvalBench job {self.job_id!r} import_version"
+          f" {self.import_version!r} has no sessions in"
+          f" {self.events_table!r}; nothing to score"
+      )
+    return TraceFilter(
+        experiment_id=self.job_id,
+        session_ids=list(self.session_ids),
+        limit=len(self.session_ids),
+    )
+
+  def to_dict(self) -> dict[str, Any]:
+    return _json_safe(dataclasses.asdict(self))
+
+
+def import_sessions(
+    *,
+    target_project: str,
+    target_dataset: str,
+    job_id: str,
+    import_version: Optional[str] = None,
+    location: Optional[str] = None,
+    bq_client: Optional[Any] = None,
+) -> EvalBenchImportSessions:
+  """Resolve one published import version to its session identities.
+
+  The version is pinned from the manifest before any event row is read
+  (``failed_sessions`` semantics: an explicit ``import_version`` must be
+  published, ``None`` means the job's latest successful import), then the
+  version's distinct ``session_id`` values are read from the
+  ``events_table`` recorded in that manifest row, so a version stays bound
+  to the table it was published to. The result feeds ``Client.evaluate``
+  through ``EvalBenchImportSessions.trace_filter()`` -- the
+  ``bq-agent-sdk evalbench-score`` path (#97).
+  """
+  _validate_source_segment("target_project", target_project)
+  _validate_source_segment("target_dataset", target_dataset)
+  if not isinstance(job_id, str) or not job_id:
+    raise ValueError("job_id must be a non-empty string")
+  manifest_ref = f"{target_project}.{target_dataset}.{MANIFEST_TABLE}"
+  client = bq_client or make_bq_client(target_project, location=location)
+  import_version, manifest = _resolve_manifest(
+      client,
+      manifest_ref=manifest_ref,
+      job_id=job_id,
+      import_version=import_version,
+      location=location,
+      feature=_SCORE_FEATURE,
+  )
+  events_ref = str(manifest["events_table"])
+  job_config = bigquery.QueryJobConfig(
+      query_parameters=_import_parameters(job_id, import_version)
+  )
+  job_config = with_sdk_labels(job_config, feature=_SCORE_FEATURE)
+  query_args: dict[str, Any] = {"job_config": job_config}
+  if location is not None:
+    query_args["location"] = location
+  query = _IMPORT_SESSIONS_QUERY.format(events_table=events_ref)
+  session_ids = tuple(
+      str(row["session_id"])
+      for row in (
+          _plain_row(raw) for raw in client.query(query, **query_args).result()
+      )
+      if row.get("session_id") is not None
+  )
+  return EvalBenchImportSessions(
+      job_id=job_id,
+      import_version=import_version,
+      events_table=events_ref,
+      scores_table=str(manifest["scores_table"]),
+      session_ids=session_ids,
       manifest=dict(manifest),
   )
 
--- a/src/bigquery_agent_analytics/evalbench.py
+++ b/src/bigquery_agent_analytics/evalbench.py
@@ -2438,6 +2438,7 @@ def import_sessions(
     target_dataset: str,
     job_id: str,
     import_version: Optional[str] = None,
+    events_table: Optional[str] = None,
     location: Optional[str] = None,
     bq_client: Optional[Any] = None,
 ) -> EvalBenchImportSessions:
@@ -2451,11 +2452,30 @@ def import_sessions(
   to the table it was published to. The result feeds ``Client.evaluate``
   through ``EvalBenchImportSessions.trace_filter()`` -- the
   ``bq-agent-sdk evalbench-score`` path (#97).
+
+  The manifest is written by the importer and read here under the scorer's
+  identity, so the persisted ``events_table`` is never trusted as-is: it
+  must be the canonical ``target_project.target_dataset.<mirror table>``
+  reference (every segment re-validated, the ADK plugin's ``agent_events``
+  rejected) before it is formatted into the session query. When
+  ``events_table`` names the mirror table the caller expects (the CLI's
+  ``--table-id``), the stored binding must match it, again before any
+  events-table statement is submitted.
+
+  Args:
+      events_table: Optional mirror table name in ``target_dataset`` the
+        version is expected to be published in. Validated before any
+        BigQuery call; a mismatch with the manifest raises ``ValueError``
+        before the events table is read.
   """
   _validate_source_segment("target_project", target_project)
   _validate_source_segment("target_dataset", target_dataset)
   if not isinstance(job_id, str) or not job_id:
     raise ValueError("job_id must be a non-empty string")
+  expected_events_ref: Optional[str] = None
+  if events_table is not None:
+    _validate_destination_table("events_table", events_table)
+    expected_events_ref = f"{target_project}.{target_dataset}.{events_table}"
   manifest_ref = f"{target_project}.{target_dataset}.{MANIFEST_TABLE}"
   client = bq_client or make_bq_client(target_project, location=location)
   import_version, manifest = _resolve_manifest(
@@ -2466,7 +2486,19 @@ def import_sessions(
       location=location,
       feature=_SCORE_FEATURE,
   )
-  events_ref = str(manifest["events_table"])
+  events_ref = _validate_manifest_events_table(
+      manifest.get("events_table"),
+      target_project=target_project,
+      target_dataset=target_dataset,
+      job_id=job_id,
+      import_version=import_version,
+  )
+  if expected_events_ref is not None and events_ref != expected_events_ref:
+    raise ValueError(
+        f"EvalBench job {job_id!r} import_version {import_version!r} is"
+        f" published in {events_ref!r}, not {expected_events_ref!r}; pass"
+        " the events_table it was published to"
+    )
   job_config = bigquery.QueryJobConfig(
       query_parameters=_import_parameters(job_id, import_version)
   )
@@ -3065,6 +3097,50 @@ def _validate_destination_table(name: str, value: Any) -> None:
         "EvalBench imports publish only to BQAA-owned mirror tables such as "
         f"{DEFAULT_EVENTS_TABLE!r}"
     )
+
+
+def _validate_manifest_events_table(
+    value: Any,
+    *,
+    target_project: str,
+    target_dataset: str,
+    job_id: str,
+    import_version: str,
+) -> str:
+  """Re-validate a persisted manifest ``events_table`` before querying it.
+
+  The importer validates every segment before publishing, but the manifest
+  is read back under a different identity (the scorer), so the stored
+  value is treated as untrusted input: it must be exactly
+  ``target_project.target_dataset.<table>`` with the table segment passing
+  ``_validate_destination_table`` (segment charset, reserved
+  ``agent_events``). Anything else -- a closed backtick, a semicolon, a
+  reference into another project or dataset -- raises before the value can
+  be formatted into a statement.
+  """
+  context = (
+      f"EvalBench job {job_id!r} import_version {import_version!r} manifest"
+      " events_table"
+  )
+  if not isinstance(value, str) or not value:
+    raise ValueError(f"{context} must be a non-empty string, got {value!r}")
+  segments = value.split(".")
+  if len(segments) != 3:
+    raise ValueError(
+        f"{context} must be a canonical project.dataset.table reference,"
+        f" got {value!r}"
+    )
+  project, dataset, table = segments
+  try:
+    _validate_destination_table("events_table", table)
+  except ValueError as exc:
+    raise ValueError(f"{context} {value!r} is invalid: {exc}") from exc
+  if project != target_project or dataset != target_dataset:
+    raise ValueError(
+        f"{context} {value!r} is outside the scored dataset"
+        f" {target_project!r}.{target_dataset!r}"
+    )
+  return f"{project}.{dataset}.{table}"
 
 
 def _find_scenario_id(row: Mapping[str, Any]) -> Optional[str]:
--- a/tests/test_cli_evalbench_score.py
+++ b/tests/test_cli_evalbench_score.py
@@ -143,6 +143,7 @@ def test_scores_latest_version_with_correctness_judge_by_default(
           "job_id": "job-123",
           "import_version": None,
           "location": None,
+          "events_table": "evalbench_agent_events",
           "bq_client": client.bq_client,
       }
   ]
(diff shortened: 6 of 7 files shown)


caohy1988 (commented):
**Reviewer: Kimi**

# Review: PR 453 — feat(evalbench): evalbench-score CLI wrapping Client.evaluate (#435 slice 3, #97)

**Verdict: Approve (comment).** No P0/P1 findings.

**Counts: P0 = 0, P1 = 0, P2 = 4 (notes only, non-blocking).**

Reviewed `git diff a354f1b..2e4614d` in the slice-3 worktree; ran the prescribed suite: **172 passed** (`test_cli_evalbench_import.py`, `test_cli_evalbench_failed_sessions.py`, `test_cli_evalbench_score.py`, `test_evalbench_importer.py`, `test_evalbench_import_sessions.py`, 5.0s, no live BigQuery).

## Contract checklist — all verified

- **Client points at the mirror table.** `evalbench_score` builds the client via the existing `_build_client(project_id, dataset_id, table_id, ...)` with `--table-id` defaulting to `evalbench_agent_events` (`src/bigquery_agent_analytics/cli.py:2615`). The reserved ADK `agent_events` table is rejected by `_validate_destination_table` *before* any client or BigQuery call (`cli.py:2613`), and after resolution the CLI refuses a `--table-id` that does not match the version's manifest-bound `events_table` (`cli.py:2630-2637`) — so the client can never be pointed at a table the version was not published to. No dataset override anywhere; pointing the Client at the mirror is the whole trick, as specified.
- **TraceFilter(experiment_id=job_id).** `EvalBenchImportSessions.trace_filter()` returns `TraceFilter(experiment_id=job_id, session_ids=[...], limit=len(session_ids))` (`evalbench.py:2425-2428`). Mapped events stamp `attributes.experiment_id = job_id` (`_base_attributes`, `evalbench.py:3300`), and `TraceFilter.to_sql_conditions()` queries `JSON_VALUE(attributes, '$.experiment_id')` (`trace.py:762`), so the job pin lands.
- **--import-version pins one published version; no mixing.** `_resolve_manifest` (`evalbench.py:2337-2373`) requires an explicit version to have a committed manifest row and resolves `None` to the job's latest successful import via the same `_LATEST_IMPORT_QUERY` semantics as `evalbench_failed_sessions` / `failed_sessions()`. The version pin reaches `Client.evaluate` exclusively through existing `TraceFilter` fields (`experiment_id` + the version's exact `session_ids` read from the manifest-bound table by `_IMPORT_SESSIONS_QUERY`), **not** through `custom_labels`/`$.custom_tags` and not through any invented `evalbench_import_version` filter — the code comments say exactly this. Cross-version mixing is structurally impossible: published session identities are `evalbench-import:{job}:{version}:{scenario}` with escaped components (`_session_identity`), so two retained versions of one job never share a session id. `trace_filter()` additionally refuses an empty session set (`evalbench.py:2420-2424`), closing the "no `session_ids` = unfiltered" widening hole — good catch, this was the one place the design could have silently scored every retained version.
- **--exit-code matches `evaluate`.** Identical logic to the existing command: `if exit_code and report.pass_rate < 1.0: _emit_evaluate_failures(report); raise typer.Exit(code=1)` (`cli.py:2649-2651` vs `cli.py:588-590`). Passing scorecard with `--exit-code` exits 0; failures without the flag exit 0. Covered by `test_exit_code_flag_returns_1_and_emits_fail_lines_on_failure`, `test_exit_code_flag_returns_0_when_all_pass`, `test_without_exit_code_flag_failures_still_exit_0`.
- **Exit codes.** 0 on produced scorecard; 2 on invalid input (unknown `--evaluator`, reserved table, unpublished job/version, table not bound to version, version with no sessions) or BQ error via the generic `except Exception → Exit(2)`. Unknown evaluator exits 2 **before** `_build_client`, `import_sessions`, or `evaluate` (`cli.py:2602-2609`; asserted by `test_unknown_evaluator_exits_2_before_any_client` with `assert_not_called` on all three).
- **--strict.** Passed straight through: `client.evaluate(evaluator=judge, filters=pinned.trace_filter(), strict=strict)` (`cli.py:2638-2640`); `Client.evaluate` itself is untouched by this diff.
- **--evaluator.** Exactly `correctness|hallucination|sentiment` via the existing `_LLM_JUDGES` table (`cli.py:276-289`), same with/without-threshold factory pattern as `evaluate`.
- **Tests.** Fully mocked (`_build_client` and `import_sessions` monkeypatched in CLI tests; `_FakeClient` for `import_sessions`); no live BigQuery; no modification of `Client.evaluate`; no new scoring engine. Reuses `_LLM_JUDGES`, `_build_client`, `_emit_evaluate_failures`.

## P2 notes (non-blocking)

1. `cli.py:2600` imports the private `_validate_destination_table` from `.evalbench`. Same-package private reuse is consistent with the codebase's `_`-helper style, but if a third caller appears, consider promoting it.
2. `tests/test_cli_evalbench_score.py:214` asserts `"Evaluation Report: correctness_judge"` in the text output while the judge under test is `hallucination` — the string comes from the mocked report's `evaluator_name`, so the test is correct, but the mismatch is mildly confusing to a reader.
3. `import_sessions` materializes all of a version's distinct session ids and pins `limit=len(session_ids)`; a very large version yields a large `IN UNNEST(@session_ids)` parameter. Inherent to the "pin via session_ids" design and documented; worth a note if 100k-scenario jobs ever show up.
4. `EvalBenchImportSessions.manifest` is carried as a plain dict copy and serialized via `to_dict()`; only `events_table`/`scores_table` are read from it today. Fine as-is.

Docs (`docs/evalbench.md`), `CHANGELOG.md`, and `examples/evalbench_score_gate.sh` are accurate against the implementation — including the removal of the stale slice-2 line "There is no `evalbench-score` command." The example script's exit-code table matches the CLI.

LGTM.

caohy1988 (commented):
**Reviewer: Codex**

## Verdict

REQUEST_CHANGES

- P0: 0
- P1: 1
- P2: 0 (informational only)

## Findings

- [P1] (confidence: 10/10) `src/bigquery_agent_analytics/cli.py:2612` — `--threshold` is documented as a 0–1 value, but the command passes it directly to `LLMAsJudge` without checking the range or finiteness. Local CLI reproduction confirmed that `-0.1`, `1.1`, `nan`, and `inf` all reach `Client.evaluate` instead of exiting 2. In particular, `--threshold -0.1 --exit-code` treats a score of `0.0` as passing and exits 0, which can make the CI gate falsely green even when judging fails. Validate `math.isfinite(threshold)` and `0.0 <= threshold <= 1.0` before constructing the client/judge, and add unit coverage asserting invalid values exit 2 before any client or BigQuery call.

## Verified

- `Client` is constructed with the EvalBench mirror table (`evalbench_agent_events` by default, or the supplied `--table-id`), and the manifest binding is checked before evaluation.
- Evaluation uses `TraceFilter(experiment_id=job_id, session_ids=<pinned version sessions>, limit=<session count>)`, preventing retained import versions from mixing.
- Explicit `--import-version` requires a published manifest row; omission resolves the latest successful import.
- `--exit-code` reuses the existing `report.pass_rate < 1.0` gate and `_emit_evaluate_failures`; success remains exit 0 and handled input/BigQuery failures exit 2.
- Unit tests use mocked/fake BigQuery clients. `Client.evaluate` is unchanged, and no scoring engine was added.

## Pytest

`python3 -m pytest -q tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py tests/test_cli_evalbench_score.py tests/test_evalbench_importer.py tests/test_evalbench_import_sessions.py`

Result: **172 passed in 3.91s**.

## What can wait

No P2 findings. Slice 1, slice 2, taxonomy, localization, harnesses, and dashboards remain out of scope and are not blockers for this review.

caohy1988 (commented):
**Reviewer: Kimi**

# Re-review: PR #453 at c11551e (slice 3, evalbench-score)

## Verdict: APPROVE — P0: 0, P1: 0

## Previous Codex P1 on 2e4614d — FIXED

`--threshold` on `evalbench-score` was not range/finite-checked; `nan`/`inf`/out-of-range values could reach the judge and green-gate CI via `--exit-code`.

Verified at c11551e, not rubber-stamped:

- `src/bigquery_agent_analytics/cli.py:2614-2624` — after the reserved-table check and **before** `with_t(threshold)` judge construction, `_build_client`, and `import_sessions`:

  ```python
  if threshold is not None and not (
      math.isfinite(threshold) and 0.0 <= threshold <= 1.0
  ):
      typer.echo("Error: --threshold must be a finite value in [0.0, 1.0] ...", err=True)
      raise typer.Exit(code=2)
  ```

  The `typer.Exit` propagates through the `except typer.Exit: raise` guard, so it exits 2 rather than being swallowed by the generic `except Exception` handler. Ordering is exactly as required: no client, no judge, no BigQuery call on the reject path.
- `math` import added (cli.py:36). Docstring exit-code list updated to match the new behavior (cli.py:2594-2599).
- Tests in `tests/test_cli_evalbench_score.py`:
  - parametrized `-0.1, 1.1, nan, inf, -inf` → exit 2 with the error message, and spies assert the judge factory, `_build_client`, and `import_sessions` were never called;
  - `--threshold -0.1 --exit-code` still exits 2 (no FAIL lines, no client build) — the green-gate scenario is directly covered;
  - boundaries `0.0` and `1.0` accepted, with the judge's criteria threshold asserted to equal the parsed value.

Note: typer parses `"nan"`/`"inf"` into actual float NaN/inf, so the `math.isfinite` guard is what rejects them — the tests confirm this end-to-end.

## New P0/P1 scan at c11551e

**`import_sessions` interpolating `manifest["events_table"]` into `_IMPORT_SESSIONS_QUERY` (evalbench.py:2477), with `client.query()` running before the CLI `--table-id` compare (cli.py:2644-2652): NOT P1.**

- The interpolated value is not CLI input; it is a column of the dataset's import manifest, written only by `EvalBenchRun.materialize`, which validates `events_table` with `_validate_destination_table` → `_validate_source_segment` (`^[A-Za-z0-9_-]+$`) before publishing. Legitimately written manifest rows can never contain backticks or `;`.
- Reaching the interpolation with a hostile value requires write access to the BQAA-owned target dataset (to tamper with the manifest table). An actor with that access can already run arbitrary BigQuery jobs in that project — there is no privilege escalation. This matches the slice-1/2 trust model, which treats committed manifest state as authoritative over self-claims (that is the entire view-authentication design).
- The subsequent `--table-id` mismatch raises before `client.evaluate` runs, so a bound-to-elsewhere version is still refused for scoring; the read-only `SELECT DISTINCT session_id` executes first, which is the documented "version stays bound to the table it was published to" behavior.
- Same interpolate-from-manifest pattern already exists in `failed_sessions` (slice 2); per instructions that is not re-litigated, and this instance is no worse.

A defense-in-depth `_validate_source_segment` on the manifest read path would be cheap hardening, but it is P3 at most, not a blocker.

## Tests

`pytest tests/test_cli_evalbench_score.py` — 20 passed (3.13s), including the 8 new threshold tests.

## Diff scope check

`git diff 2e4614d..c11551e` touches only `src/bigquery_agent_analytics/cli.py` (+15/-1) and `tests/test_cli_evalbench_score.py` (+55). No unrelated churn; slice-1/slice-2 files untouched.

caohy1988 (commented):
**Reviewer: Codex**

## Verdict

REQUEST_CHANGES

- P0: 0
- P1: 1
- P2: 0 (informational only)

## Scope and verification

- Reviewed the exact fix diff `2e4614d8f751059c79fcfcc88b59fc8fed4e8272..c11551edd910287b6130f61a3c8103343a08fd6a`.
- Read the `evalbench-score` handler, `evalbench.import_sessions` and its manifest/query path, `tests/test_cli_evalbench_score.py`, and `tests/test_evalbench_import_sessions.py` at HEAD `c11551edd910287b6130f61a3c8103343a08fd6a`.
- The prior threshold P1 is fixed. `cli.py:2616-2624` requires a finite value in `[0.0, 1.0]` before judge construction (`cli.py:2625`), client construction (`cli.py:2628`), or `import_sessions` (`cli.py:2636`). The tests reject `-0.1`, `1.1`, `nan`, `inf`, and `-inf` with exit 2 and no judge/client/import call; the `--threshold -0.1 --exit-code` regression stays exit 2 with no `FAIL` output; `0.0` and `1.0` are accepted.

## Findings

### P1

1. **Stored manifest table ref can execute SQL before validation** - `src/bigquery_agent_analytics/evalbench.py:2477`

   `import_sessions()` reads `manifest["events_table"]` as an arbitrary string at line 2469, formats it into the backtick-quoted `FROM` at line 2477, and submits it with `client.query()` at line 2481. The CLI does not compare `pinned.events_table` with the requested `--table-id` until `cli.py:2644-2645`, after that query has completed.

   A stored value that closes the identifier and adds semicolon-delimited statements renders as a valid BigQuery script. BigQuery executes semicolon-separated statements as a multi-statement query, and those scripts can contain DML/DDL with side effects ([Google Cloud documentation](https://docs.cloud.google.com/bigquery/docs/multi-statement-queries)). The normal materialize path validates table segments before writing the manifest, but the read path does not revalidate mutable persisted data.

   This is still P1, not a trusted-manifest non-issue. An attacker needs write access to the manifest, but BigQuery write roles can be scoped to a table or dataset ([BigQuery IAM roles](https://docs.cloud.google.com/bigquery/docs/access-control)); the scoring identity can have broader permissions. The current code therefore allows a manifest writer to make the scorer act as a confused deputy. The deployment may use one equally privileged identity for both roles, which limits impact there, but the repository neither enforces nor documents that invariant.

   **Required fix:** Before formatting or submitting `_IMPORT_SESSIONS_QUERY`, require the stored `events_table` to be a canonical `target_project.target_dataset.<validated_table>` reference with every segment validated. For the CLI path, compare the stored binding with the requested table before any events-table query, for example by passing the expected reference into `import_sessions` or by separating manifest resolution from session lookup. Add a regression test with a backtick/semicolon manifest payload and assert that only the manifest lookup occurs and the call exits 2 before any events-table query.

   **Confidence:** 100. The exact script rendering was reproduced locally without a live BigQuery call, two independent reviewers agreed, and a fresh validation pass confirmed the execution order and missing read-side validation.

### Actionable Findings

- #1 - P1 - `src/bigquery_agent_analytics/evalbench.py:2477` - validate the persisted table reference and binding before query submission; add the malicious-manifest fail-before-query test. Route: `gated_auto -> downstream-resolver`.

## Pytest

`python3 -m pytest -q tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py tests/test_cli_evalbench_score.py tests/test_evalbench_importer.py tests/test_evalbench_import_sessions.py`

Result: **180 passed in 3.61s**.

## Coverage

- Local correctness and security reviewers independently found #1; the independent validator confirmed it.
- The cross-model adversarial pass ran on Claude Opus 5 (served model verified; requested high effort, actual effort unverified). It found no threshold-delta defect and independently identified the same manifest-to-SQL path as a residual risk, classifying it outside its narrower parent-to-head diff; the user's explicit slice-3 scope makes it a current P1 here.
- Previous-feedback verification found the posted threshold P1 fully addressed.
- No plan was used, so settlement suppression was not evaluated.

## What can wait

No P2 findings. Slice 1 / #451, slice 2 / #452, the analogous slice-2 interpolation family, taxonomy, localization, harnesses, and dashboards remain out of scope and are not blockers in this review.

---

### Final verdict

> **REQUEST_CHANGES.** The threshold P1 is fixed, but #1 remains a slice-3 P1. Validate the manifest-bound table reference and reject it before query submission.

Actionable findings: #1 - P1 - `src/bigquery_agent_analytics/evalbench.py:2477` - stored manifest identifier reaches executable SQL before validation - `gated_auto -> downstream-resolver` - suggested fix present - confidence 100.

caohy1988 (commented):
**Reviewer: Codex**

## Verdict

REQUEST_CHANGES

- P0: 0
- P1: 1
- P2: 0 (informational only)

## Scope and verification

- Reviewed the exact fix diff `2e4614d8f751059c79fcfcc88b59fc8fed4e8272..c11551edd910287b6130f61a3c8103343a08fd6a`.
- Read the `evalbench-score` handler, `evalbench.import_sessions` and its manifest/query path, `tests/test_cli_evalbench_score.py`, and `tests/test_evalbench_import_sessions.py` at HEAD `c11551edd910287b6130f61a3c8103343a08fd6a`.
- The prior threshold P1 is fixed. `cli.py:2616-2624` requires a finite value in `[0.0, 1.0]` before judge construction (`cli.py:2625`), client construction (`cli.py:2628`), or `import_sessions` (`cli.py:2636`). The tests reject `-0.1`, `1.1`, `nan`, `inf`, and `-inf` with exit 2 and no judge/client/import call; the `--threshold -0.1 --exit-code` regression stays exit 2 with no `FAIL` output; `0.0` and `1.0` are accepted.

## Findings

### P1

1. **Stored manifest table ref can execute SQL before validation** - `src/bigquery_agent_analytics/evalbench.py:2477`

   `import_sessions()` reads `manifest["events_table"]` as an arbitrary string at line 2469, formats it into the backtick-quoted `FROM` at line 2477, and submits it with `client.query()` at line 2481. The CLI does not compare `pinned.events_table` with the requested `--table-id` until `cli.py:2644-2645`, after that query has completed.

   A stored value that closes the identifier and adds semicolon-delimited statements renders as a valid BigQuery script. BigQuery executes semicolon-separated statements as a multi-statement query, and those scripts can contain DML/DDL with side effects ([Google Cloud documentation](https://docs.cloud.google.com/bigquery/docs/multi-statement-queries)). The normal materialize path validates table segments before writing the manifest, but the read path does not revalidate mutable persisted data.

   This is still P1, not a trusted-manifest non-issue. An attacker needs write access to the manifest, but BigQuery write roles can be scoped to a table or dataset ([BigQuery IAM roles](https://docs.cloud.google.com/bigquery/docs/access-control)); the scoring identity can have broader permissions. The current code therefore allows a manifest writer to make the scorer act as a confused deputy. The deployment may use one equally privileged identity for both roles, which limits impact there, but the repository neither enforces nor documents that invariant.

   **Required fix:** Before formatting or submitting `_IMPORT_SESSIONS_QUERY`, require the stored `events_table` to be a canonical `target_project.target_dataset.<validated_table>` reference with every segment validated. For the CLI path, compare the stored binding with the requested table before any events-table query, for example by passing the expected reference into `import_sessions` or by separating manifest resolution from session lookup. Add a regression test with a backtick/semicolon manifest payload and assert that only the manifest lookup occurs and the call exits 2 before any events-table query.

   **Confidence:** 100. The exact script rendering was reproduced locally without a live BigQuery call, two independent reviewers agreed, and a fresh validation pass confirmed the execution order and missing read-side validation.

### Actionable Findings

- #1 - P1 - `src/bigquery_agent_analytics/evalbench.py:2477` - validate the persisted table reference and binding before query submission; add the malicious-manifest fail-before-query test. Route: `gated_auto -> downstream-resolver`.

## Pytest

`python3 -m pytest -q tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py tests/test_cli_evalbench_score.py tests/test_evalbench_importer.py tests/test_evalbench_import_sessions.py`

Result: **180 passed in 3.61s**.

## Coverage

- Local correctness and security reviewers independently found #1; the independent validator confirmed it.
- The cross-model adversarial pass ran on Claude Opus 5 (served model verified; requested high effort, actual effort unverified). It found no threshold-delta defect and independently identified the same manifest-to-SQL path as a residual risk, classifying it outside its narrower parent-to-head diff; the user's explicit slice-3 scope makes it a current P1 here.
- Previous-feedback verification found the posted threshold P1 fully addressed.
- No plan was used, so settlement suppression was not evaluated.

## What can wait

No P2 findings. Slice 1 / #451, slice 2 / #452, the analogous slice-2 interpolation family, taxonomy, localization, harnesses, and dashboards remain out of scope and are not blockers in this review.

---

### Final verdict

> **REQUEST_CHANGES.** The threshold P1 is fixed, but #1 remains a slice-3 P1. Validate the manifest-bound table reference and reject it before query submission.

Actionable findings: #1 - P1 - `src/bigquery_agent_analytics/evalbench.py:2477` - stored manifest identifier reaches executable SQL before validation - `gated_auto -> downstream-resolver` - suggested fix present - confidence 100.

caohy1988 (commented):
**Reviewer: Kimi**

# PR #453 re-review (slice 3) — head cf3fa84

## Verdict: APPROVE

The Codex P1 on c11551e (confused-deputy: persisted manifest `events_table` interpolated into `_IMPORT_SESSIONS_QUERY` and submitted before the `--table-id` compare) is genuinely fixed, and I found no new P0/P1 in the diff `c11551e..cf3fa84` or the surrounding slice-3 code.

Scope honored: slice 3 only; slice-2 `failed_sessions` interpolation (evalbench.py:2306) was not re-litigated.

## Codex P1 — requirement-by-requirement

**1. Validate persisted `events_table` as canonical `target_project.target_dataset.<validated_table>` BEFORE formatting/submitting — FIXED.**

New `_validate_manifest_events_table` (evalbench.py:3101-3144) treats the stored value as untrusted: non-empty string, exactly three `.`-separated segments, table segment through `_validate_destination_table` (charset `[A-Za-z0-9_-]+`, case-insensitive rejection of reserved `agent_events`), and exact equality of the project/dataset segments against the already-`_validate_source_segment`-validated targets. It runs at evalbench.py:2489 — after `_resolve_manifest` (whose only BigQuery call is the parameterized manifest lookup) and before `_IMPORT_SESSIONS_QUERY.format(...)` (evalbench.py:2509) and `client.query`. A closed backtick, `;`, or cross-project/dataset reference cannot survive: any such character fails either the segment count or the charset/equality checks, and the raise happens before the query string is even built.

**2. Compare stored binding with requested `--table-id` BEFORE any events-table query — FIXED.**

`import_sessions` gains an optional `events_table` parameter (evalbench.py:2441). When given, the expected name is validated (`_validate_destination_table`) and expanded to the canonical ref before *any* BigQuery call — including the manifest lookup (evalbench.py:2476-2480); the binding-vs-requested compare (evalbench.py:2496-2500) runs after the manifest lookup but before the session query. `evalbench-score` passes `--table-id` through (cli.py:2642), so the mismatch now exits 2 before the events table is touched instead of after. The pre-existing post-pin compare in the CLI (cli.py:2647-2653) remains as a documented second line of defense, which is fine.

Verified that nothing queries BigQuery before this gate on the CLI path: `_build_client` (cli.py:97-117) constructs `Client(verify_schema=False)` — no network call.

**3. Regression tests with hostile payloads asserting only the manifest lookup occurs — PRESENT and MEANINGFUL.**

- tests/test_evalbench_import_sessions.py: parametrized over 11 hostile values plus `None` and `7` — backtick-close + `DROP TABLE`, `; SELECT 1`, cross-project, cross-dataset, 1/2/4-segment refs, reserved `agent_events`/`AGENT_EVENTS`, empty string — asserting `ValueError`, exactly one client call, that call targeting the manifest, and no `SELECT DISTINCT session_id` statement. Separate tests cover the pinned-version path, the expected-table mismatch (asserting only the manifest lookup ran), the expected-table happy path, and caller-side `events_table` validation with zero client calls.
- tests/test_cli_evalbench_score.py: a recording fake BQ client drives the real `import_sessions` through the CLI with four hostile stored values (including the multi-statement payload and cross-project ref), asserting exit 2, `events_table` in the error output, exactly one query (the manifest lookup), no events-table statement, and `client.evaluate` never called. A second test pins that `--table-id` actually reaches `import_sessions`.

These are real regression tests, not shape-checks: they would fail on c11551e.

## New-issue sweep (P0/P1 only)

- Importer compatibility: `materialize` writes the manifest's `events_table` as the canonical FQDN (`"events_table": events_ref`, evalbench.py:1038), so the new three-segment validation does not break legitimate scoring of existing imports.
- `scores_table` is still read from the manifest unvalidated (evalbench.py:2517) but is never interpolated into a query in `import_sessions` and is unused by the CLI — informational field only. Not a P1.
- TOCTOU between manifest read and session query: the validated ref is constrained to `target_project.target_dataset.<safe-name>`, so the worst case is a read-only query against another mirror table in the scored dataset. Not a P1.
- Error-message content: mismatch messages echo the stored and expected refs to stderr; both are charset-validated by that point or the message is a validation failure. No injection surface.
- The diff is minimal (4 files, +291/-1, mostly tests) and touches nothing outside slice 3.

## Tests

Independently re-ran (worktree `src` shadowing the main-repo editable install via `PYTHONPATH`, verified the imported package resolves to the worktree):

```
python3 -m pytest -q tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py tests/test_cli_evalbench_score.py tests/test_evalbench_importer.py tests/test_evalbench_import_sessions.py
205 passed, 2 warnings in 1.78s
```

Matches the implementer's report of 205 passed.

P0 = 0, P1 = 0. Approve.

caohy1988 (commented):
**Reviewer: Kimi**

# PR #453 re-review (slice 3) — head cf3fa84

## Verdict: APPROVE

The Codex P1 on c11551e (confused-deputy: persisted manifest `events_table` interpolated into `_IMPORT_SESSIONS_QUERY` and submitted before the `--table-id` compare) is genuinely fixed, and I found no new P0/P1 in the diff `c11551e..cf3fa84` or the surrounding slice-3 code.

Scope honored: slice 3 only; slice-2 `failed_sessions` interpolation (evalbench.py:2306) was not re-litigated.

## Codex P1 — requirement-by-requirement

**1. Validate persisted `events_table` as canonical `target_project.target_dataset.<validated_table>` BEFORE formatting/submitting — FIXED.**

New `_validate_manifest_events_table` (evalbench.py:3101-3144) treats the stored value as untrusted: non-empty string, exactly three `.`-separated segments, table segment through `_validate_destination_table` (charset `[A-Za-z0-9_-]+`, case-insensitive rejection of reserved `agent_events`), and exact equality of the project/dataset segments against the already-`_validate_source_segment`-validated targets. It runs at evalbench.py:2489 — after `_resolve_manifest` (whose only BigQuery call is the parameterized manifest lookup) and before `_IMPORT_SESSIONS_QUERY.format(...)` (evalbench.py:2509) and `client.query`. A closed backtick, `;`, or cross-project/dataset reference cannot survive: any such character fails either the segment count or the charset/equality checks, and the raise happens before the query string is even built.

**2. Compare stored binding with requested `--table-id` BEFORE any events-table query — FIXED.**

`import_sessions` gains an optional `events_table` parameter (evalbench.py:2441). When given, the expected name is validated (`_validate_destination_table`) and expanded to the canonical ref before *any* BigQuery call — including the manifest lookup (evalbench.py:2476-2480); the binding-vs-requested compare (evalbench.py:2496-2500) runs after the manifest lookup but before the session query. `evalbench-score` passes `--table-id` through (cli.py:2642), so the mismatch now exits 2 before the events table is touched instead of after. The pre-existing post-pin compare in the CLI (cli.py:2647-2653) remains as a documented second line of defense, which is fine.

Verified that nothing queries BigQuery before this gate on the CLI path: `_build_client` (cli.py:97-117) constructs `Client(verify_schema=False)` — no network call.

**3. Regression tests with hostile payloads asserting only the manifest lookup occurs — PRESENT and MEANINGFUL.**

- tests/test_evalbench_import_sessions.py: parametrized over 11 hostile values plus `None` and `7` — backtick-close + `DROP TABLE`, `; SELECT 1`, cross-project, cross-dataset, 1/2/4-segment refs, reserved `agent_events`/`AGENT_EVENTS`, empty string — asserting `ValueError`, exactly one client call, that call targeting the manifest, and no `SELECT DISTINCT session_id` statement. Separate tests cover the pinned-version path, the expected-table mismatch (asserting only the manifest lookup ran), the expected-table happy path, and caller-side `events_table` validation with zero client calls.
- tests/test_cli_evalbench_score.py: a recording fake BQ client drives the real `import_sessions` through the CLI with four hostile stored values (including the multi-statement payload and cross-project ref), asserting exit 2, `events_table` in the error output, exactly one query (the manifest lookup), no events-table statement, and `client.evaluate` never called. A second test pins that `--table-id` actually reaches `import_sessions`.

These are real regression tests, not shape-checks: they would fail on c11551e.

## New-issue sweep (P0/P1 only)

- Importer compatibility: `materialize` writes the manifest's `events_table` as the canonical FQDN (`"events_table": events_ref`, evalbench.py:1038), so the new three-segment validation does not break legitimate scoring of existing imports.
- `scores_table` is still read from the manifest unvalidated (evalbench.py:2517) but is never interpolated into a query in `import_sessions` and is unused by the CLI — informational field only. Not a P1.
- TOCTOU between manifest read and session query: the validated ref is constrained to `target_project.target_dataset.<safe-name>`, so the worst case is a read-only query against another mirror table in the scored dataset. Not a P1.
- Error-message content: mismatch messages echo the stored and expected refs to stderr; both are charset-validated by that point or the message is a validation failure. No injection surface.
- The diff is minimal (4 files, +291/-1, mostly tests) and touches nothing outside slice 3.

## Tests

Independently re-ran (worktree `src` shadowing the main-repo editable install via `PYTHONPATH`, verified the imported package resolves to the worktree):

```
python3 -m pytest -q tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py tests/test_cli_evalbench_score.py tests/test_evalbench_importer.py tests/test_evalbench_import_sessions.py
205 passed, 2 warnings in 1.78s
```

Matches the implementer's report of 205 passed.

P0 = 0, P1 = 0. Approve.

caohy1988 (commented):
**Reviewer: Codex**

## PR 453 re-review

**Verdict:** LGTM
**Scope:** Slice 3 only, commit range `c11551edd910287b6130f61a3c8103343a08fd6a..cf3fa8482e674b387d37ca780bb26f903d14d007`
**Mode:** Markdown report-only; nothing posted to GitHub and no repository changes made.

### Counts

| Severity | Count |
|---|---:|
| P0 | 0 |
| P1 | 0 |
| P2 | 0 (informational only) |

### Findings

None.

### Prior P1 re-check

The manifest-table confused-deputy P1 is fixed.

- `src/bigquery_agent_analytics/evalbench.py:2476` validates the optional expected mirror table before client creation or any BigQuery call.
- `src/bigquery_agent_analytics/evalbench.py:2489` revalidates the stored manifest value immediately after the manifest lookup. The validator at `src/bigquery_agent_analytics/evalbench.py:3102` requires exactly three segments, validates the table segment with `_validate_destination_table`, rejects the reserved `agent_events` table, and requires the stored project and dataset to equal the scored target.
- `src/bigquery_agent_analytics/evalbench.py:2496` compares the validated stored binding with the caller's expected mirror table before the session query is formatted at `src/bigquery_agent_analytics/evalbench.py:2509` or submitted at `src/bigquery_agent_analytics/evalbench.py:2513`.
- `src/bigquery_agent_analytics/cli.py:2640` passes `--table-id` into `import_sessions`, so the production CLI path performs the binding check before any events-table query. The later comparison remains a second defense.
- `tests/test_evalbench_import_sessions.py:253` covers hostile, cross-project/dataset, noncanonical, reserved, empty, and non-string stored values. It asserts one manifest lookup and no session query. `tests/test_evalbench_import_sessions.py:299` covers an expected-table mismatch before the events-table read, and `tests/test_evalbench_import_sessions.py:336` covers invalid expected tables before any query.
- `tests/test_cli_evalbench_score.py:270` verifies `--table-id` propagation. `tests/test_cli_evalbench_score.py:307` drives hostile manifest values through the real `import_sessions` path and verifies exit 2, exactly one manifest query, no `SELECT DISTINCT session_id`, and no evaluation.

### Pytest

Command:

`python3 -m pytest -q tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py tests/test_cli_evalbench_score.py tests/test_evalbench_importer.py tests/test_evalbench_import_sessions.py`

Result: **205 passed in 3.99s**.

### What can wait

Additional defense-in-depth for other manifest fields, CLI flag-name wording, and one more end-to-end real-import mismatch test are P2/P3-level follow-ups. They are not counted as findings in this P0/P1-only pass because none leaves the reported SQL injection/confused-deputy path open.

### Coverage

- Reviewers: correctness, testing, security, maintainability, and prior-comments verification.
- Independent adversarial pass: `cross_model_route=claude`; `model_requested=claude-opus-5`; `model_actual=claude-opus-5`; `effort_requested=high`; `effort_actual=unverified`; `receipt_supported=true`; `independence_verified=true`.
- Project standards: not run because no applicable `CODING_STANDARDS.md`, `CLAUDE.md`, or `AGENTS.md` file governs the changed files.
- Requirements settlement: not evaluated because no plan artifact was found.
- Validator batch: not run because no P0/P1 or actionable findings survived synthesis.
- Suppressed or malformed findings: 0.
- Working tree remained clean at HEAD `cf3fa8482e674b387d37ca780bb26f903d14d007`.

### Actionable Findings

None.

---

> **Verdict:** LGTM (Ready to merge)
>
> **Reasoning:** The persisted manifest table is now validated and target-bound before SQL formatting or submission, the CLI supplies its expected table before the events read, and the required hostile-payload regressions pass. No P0 or P1 remains.
>
> **Fix order:** None.

Actionable findings: none.

caohy1988 (commented):
**Reviewer: Kimi**

# Re-review: PR 453 rebase — 77f1a85 onto PR 451 merge 2459c0f / PR 452 head d681d54

## Verdict: APPROVE

No remaining P0. No remaining P1. P2s not assessed as blockers per instructions.

## Rebase integrity (the only thing that changed since cf3fa84)

- **Tree-identical to the dual-approved head.** `git rev-parse cf3fa84^{tree} 77f1a85^{tree}` both return `bb904e6acad62cb91748c717e0d9b3826cc57ca5`. The rebase changed commit SHAs/parents only — zero content drift from what Kimi and Codex approved on cf3fa84.
- **Parent chain is correct:** `77f1a85` → `088d56d` → `26019be` → `d681d54` (PR 452 head) → … → `2459c0f` (PR 451 merge). Both stated bases are ancestors of the new head.
- **Scope held:** `git diff --stat d681d54..77f1a85` touches exactly 7 files, all slice-3: `src/bigquery_agent_analytics/cli.py`, `src/bigquery_agent_analytics/evalbench.py`, `tests/test_cli_evalbench_score.py`, `tests/test_evalbench_import_sessions.py`, `docs/evalbench.md`, `examples/evalbench_score_gate.sh`, `CHANGELOG.md`. No slice-2 files touched; failed_sessions interpolation not re-litigated.
- No conflict markers in `evalbench.py` / `cli.py`.

## Prior P1 fixes — confirmed still holding at 77f1a85

1. **`--threshold` validation (`088d56d`, cli.py:2614-2624):** rejects non-finite and out-of-`[0.0, 1.0]` thresholds with exit 2 before judge/client construction. Present, unchanged. Boundaries 0.0/1.0 and omitted default pass through.
2. **Manifest `events_table` revalidation (`77f1a85`):**
   - `_validate_manifest_events_table()` (evalbench.py:3102-3143) requires exactly 3 segments, table segment through `_validate_destination_table` (charset + reserved `agent_events`), project/dataset pinned to the scored target — raises before any events-table statement is built. Closes the confused-deputy multi-statement vector Codex flagged on c11551e.
   - `import_sessions(events_table=...)` validates the expected mirror table before any BigQuery call; mismatch with the manifest raises after the manifest lookup, before the session query. `--table-id` is threaded through from the CLI; post-pin comparison retained as second line of defense.

## Tests

Re-run independently in the slice-3 worktree at 77f1a85:

```
.venv/bin/python -m pytest -q tests/test_cli_evalbench_import.py \
  tests/test_cli_evalbench_failed_sessions.py tests/test_cli_evalbench_score.py \
  tests/test_evalbench_importer.py tests/test_evalbench_import_sessions.py
→ 205 passed in 10.61s
```

Matches the 205-passed count from the cf3fa84 review.

## Environment note (not a PR issue)

`uv sync --frozen` in this worktree fails building `cryptography==50.0.1` from sdist (missing `x86_64-apple-darwin` Rust target on this host). Worked around with `uv pip install --only-binary :all: -e . pytest pytest-mock`. Local toolchain gap only — no action requested on the PR.

## New P0/P1 from the rebase

None. Identical tree to a previously approved head + green tests + correct parentage = nothing new to block on.

caohy1988 (commented):
**Reviewer: Kimi**

# Re-review: PR 453 rebase — 77f1a85 onto PR 451 merge 2459c0f / PR 452 head d681d54

## Verdict: APPROVE

No remaining P0. No remaining P1. P2s not assessed as blockers per instructions.

## Rebase integrity (the only thing that changed since cf3fa84)

- **Tree-identical to the dual-approved head.** `git rev-parse cf3fa84^{tree} 77f1a85^{tree}` both return `bb904e6acad62cb91748c717e0d9b3826cc57ca5`. The rebase changed commit SHAs/parents only — zero content drift from what Kimi and Codex approved on cf3fa84.
- **Parent chain is correct:** `77f1a85` → `088d56d` → `26019be` → `d681d54` (PR 452 head) → … → `2459c0f` (PR 451 merge). Both stated bases are ancestors of the new head.
- **Scope held:** `git diff --stat d681d54..77f1a85` touches exactly 7 files, all slice-3: `src/bigquery_agent_analytics/cli.py`, `src/bigquery_agent_analytics/evalbench.py`, `tests/test_cli_evalbench_score.py`, `tests/test_evalbench_import_sessions.py`, `docs/evalbench.md`, `examples/evalbench_score_gate.sh`, `CHANGELOG.md`. No slice-2 files touched; failed_sessions interpolation not re-litigated.
- No conflict markers in `evalbench.py` / `cli.py`.

## Prior P1 fixes — confirmed still holding at 77f1a85

1. **`--threshold` validation (`088d56d`, cli.py:2614-2624):** rejects non-finite and out-of-`[0.0, 1.0]` thresholds with exit 2 before judge/client construction. Present, unchanged. Boundaries 0.0/1.0 and omitted default pass through.
2. **Manifest `events_table` revalidation (`77f1a85`):**
   - `_validate_manifest_events_table()` (evalbench.py:3102-3143) requires exactly 3 segments, table segment through `_validate_destination_table` (charset + reserved `agent_events`), project/dataset pinned to the scored target — raises before any events-table statement is built. Closes the confused-deputy multi-statement vector Codex flagged on c11551e.
   - `import_sessions(events_table=...)` validates the expected mirror table before any BigQuery call; mismatch with the manifest raises after the manifest lookup, before the session query. `--table-id` is threaded through from the CLI; post-pin comparison retained as second line of defense.

## Tests

Re-run independently in the slice-3 worktree at 77f1a85:

```
.venv/bin/python -m pytest -q tests/test_cli_evalbench_import.py \
  tests/test_cli_evalbench_failed_sessions.py tests/test_cli_evalbench_score.py \
  tests/test_evalbench_importer.py tests/test_evalbench_import_sessions.py
→ 205 passed in 10.61s
```

Matches the 205-passed count from the cf3fa84 review.

## Environment note (not a PR issue)

`uv sync --frozen` in this worktree fails building `cryptography==50.0.1` from sdist (missing `x86_64-apple-darwin` Rust target on this host). Worked around with `uv pip install --only-binary :all: -e . pytest pytest-mock`. Local toolchain gap only — no action requested on the PR.

## New P0/P1 from the rebase

None. Identical tree to a previously approved head + green tests + correct parentage = nothing new to block on.

caohy1988 (commented):
**Reviewer: Codex**

## Code Review Results

**Scope:** PR 453 slice 3 only, `d681d548519c94b25072088cf898680d43e9bb8f` -> `77f1a85a7073bd59527aaf5c5cd8723a27e66eb0`. Reviewed `evalbench-score` in `cli.py`, `import_sessions` and its direct support code in `evalbench.py`, and the two dedicated test files.

**Intent:** Confirm that the rebase onto current PR 452, including PR 451 merge `2459c0f`, preserved the previously approved slice-3 behavior and introduced no new P0/P1 defect.

**Mode:** Report-only P0/P1 re-review. No GitHub action or repository mutation was performed.

**Reviewers:** correctness, security, testing, reliability, performance, maintainability, previous-feedback, and agent-native.

### Findings

No P0 or P1 findings.

### Actionable Findings

None.

### Coverage

- Head and ancestry: the worktree and PR 453 both resolve to `77f1a85a7073bd59527aaf5c5cd8723a27e66eb0`; PR 452 resolves to `d681d548519c94b25072088cf898680d43e9bb8f`; PR 451 merge `2459c0f` is an ancestor of that base.
- Rebase integrity: `git range-diff` matched all three slice-3 commits one-for-one, and all three stable patch IDs match the previously approved series ending at `cf3fa84`.
- Content integrity: old approved head `cf3fa84` and new head `77f1a85` have the same repository tree object, `bb904e6acad62cb91748c717e0d9b3826cc57ca5`. The rebase changed ancestry and commit SHAs, not file contents.
- Implementation review: the CLI validates evaluator, destination table, and finite threshold before evaluation; `import_sessions` revalidates the manifest table reference before formatting the events query; the resulting `TraceFilter` fails closed on an empty session set and pins both the job and exact versioned session IDs.
- Prior feedback: no previously approved P0/P1 fix was lost in the rebase.
- Independent validation: one reviewer proposed failing whenever `Client.evaluate` successfully uses its documented API fallback. A fresh validator rejected that candidate because fallback is an intentional, auditable success mode inherited by this thin wrapper. It is not counted as a finding or test gap.
- Cross-model adversarial coverage: requested `claude-opus-5` at high reasoning through the Claude CLI. The job timed out without a usable artifact, so no cross-model agreement was credited and this coverage is recorded as degraded.
- Project standards: not run because no applicable `CODING_STANDARDS.md`, `CLAUDE.md`, or `AGENTS.md` governs the scoped files.
- Plan settlement checks: no applicable plan was discovered.
- Explicit exclusions: slice-2 `failed_sessions` interpolation and all P2/P3 nits were not reviewed as blockers.
- Pytest: `python3 -m pytest -q tests/test_cli_evalbench_import.py tests/test_cli_evalbench_failed_sessions.py tests/test_cli_evalbench_score.py tests/test_evalbench_importer.py tests/test_evalbench_import_sessions.py` -> `205 passed in 3.82s`.

---

### Verdict

> **LGTM**
>
> **P0 count:** 0
>
> **P1 count:** 0
>
> **Pytest result:** 205 passed in 3.82s
>
> **Reasoning:** The rebased head is content-identical to the previously approved head, the current stacked ancestry is correct, the scoped implementation remains fail-closed at its manifest and session boundaries, and the required tests pass.

Actionable findings: none.

haiyuan-eng-google (approved):

caohy1988 (commented):
**Reviewer: Kimi**

# Re-review: PR 453 post-rebase onto main @ 2779b7e (PR 452 squash)

**Verdict: APPROVE** — P0 = 0, P1 = 0. No new findings from the rebase; prior P1 fixes confirmed still in place.

## Scope & method

- HEAD: `186c5e7` — 3 commits on top of `2779b7e` (PR 452 squash merge); merge-base verified as exactly `2779b7e`.
- Slice 3 only vs new main; slice-2 `failed_sessions` not re-litigated.
- Diff vs `2779b7e`: 7 files, +1456/−4 — `cli.py`, `evalbench.py`, `CHANGELOG.md`, `docs/evalbench.md`, `examples/evalbench_score_gate.sh`, `tests/test_cli_evalbench_score.py`, `tests/test_evalbench_import_sessions.py`. No slice-2 files touched.

## Rebase integrity — zero drift (strongest possible signal)

`git rev-parse 77f1a85^{tree} 186c5e7^{tree}` → both are `bb904e6a…`. The rebased HEAD is **byte-identical** to the previously dual-approved `77f1a85`. No conflict-resolution edits, no content drift; the old approvals' code analysis carries over in full, and the slice-2 content in the tree now comes from main's `2779b7e` unchanged.

## Tests (re-run independently, worktree `.venv`, pytest 9.1.1)

```
python3 -m pytest -q tests/test_cli_evalbench_import.py \
  tests/test_cli_evalbench_failed_sessions.py \
  tests/test_cli_evalbench_score.py \
  tests/test_evalbench_importer.py \
  tests/test_evalbench_import_sessions.py
→ 205 passed, 1 warning in 1.26s
```

Same 205-pass count as the approved review; includes the slice-2 `failed_sessions` CLI tests against the squashed base.

## Prior P1 fixes — both still hold

- `71e5ec0` threshold guard: `evalbench-score` rejects non-finite / out-of-`[0.0, 1.0]` `--threshold` with exit 2 before any BigQuery call (`cli.py`, `math.isfinite` check). Present and covered by tests.
- `186c5e7` manifest `events_table` revalidation: `_validate_manifest_events_table` treats the persisted manifest value as untrusted — requires a canonical 3-segment `project.dataset.table`, re-runs `_validate_destination_table` on the table segment (charset + reserved `agent_events` rejection), and requires project/dataset equality with the already-validated `target_project`/`target_dataset`. Because the caller-supplied segments pass `_validate_source_segment` first, the equality check closes any injection/cross-dataset path; the formatted `_IMPORT_SESSIONS_QUERY` can only ever contain validated-segment references. Sound.

## New-P0/P1 sweep on the rebased tree (spot re-checks)

- Version pin correctness: `TraceFilter(experiment_id=job_id, session_ids=…, limit=count)`; versioned session identities can't collide across retained versions; empty session set raises instead of silently widening to all versions (CLI maps to exit 2, documented).
- CLI consistency: `_LLM_JUDGES`, `_build_client`, `format_output`, `_emit_evaluate_failures` reused from `evaluate`; exit codes 0/1/2 match docs and the gate script's comments; `--table-id` mismatch with the manifest binding raises before the events table is queried (with a second defense-in-depth check on the returned pin).
- `report.details["evalbench"]` in-place dict mutation matches existing codebase usage (`feedback.py`, `trace_evaluator.py`).
- CHANGELOG keeps **both** the 452 (failed-session view) and 453 (evalbench-score) entries, correctly ordered; `docs/evalbench.md` removes the stale "There is no `evalbench-score` command" line and documents the new command, exit codes, and `trace_filter()` semantics.
- `examples/evalbench_score_gate.sh`: `set -euo pipefail`, required-env guards, passes `--exit-code`; consistent with documented behavior.

## P2 (non-blocking, noted only)

None worth recording beyond what prior rounds already covered.

Approve.

PR #470, opened 2026-08-31
feat(evalbench): persist span-level G1 labels on native snapshot (#469)

caohy1988:
Native snapshots currently expose session-level failures but do not persist the span labels needed for drill-down. Adds opt-in `span_labels_table` support to `NativeAgentEventsRun.materialize()` and `--span-labels-table` to `evalbench-native-import`.

The final implementation:
- Derives labels before publication and uses the same resolved score policy for labels and the failed-session denominator.
- Stages labels and atomically synchronizes a version’s span rows and durable per-job binding under the import lock.
- Revalidates the source generation, canonical policy, and observed binding inside the transaction, including unchanged retries and delayed old-version calls.
- Authenticates and conditionally updates the companion pinned view; it exposes labels only for the matching current manifest generation.
- Preserves source-table protections, reserved-name checks, real span IDs, version identity, and the sealed Week 0 record.

Base snapshot publication and derived span synchronization are separate steps. If synchronization fails after the base commits, the call reports that partial completion; the pinned view excludes stale labels and rerunning the import retries synchronization.

Updated against main, including #456/#457, with the changelog conflict resolved. No persistence logic changed during this refresh.

Validation: 228 focused importer, span-label, native-writer, and CLI tests passed; `git diff --check` passed. Tests cover the offline transaction model, policy/binding races, generation guards, retries, view ownership, and source protections. No live BigQuery execution or judge calls were performed. CI runs on the updated head.

Refs #469, #435.

PR #472, opened 2026-09-01
feat(deploy): scheduled skill-evolution Cloud Run Job (agentic wrapper for scripts/skill_evolution.py)

evekhm:
# Scheduled skill-evolution Cloud Run Job

Adds `deploy/skill_evolution_job/`: a scheduled Cloud Run Job that wraps
`scripts/skill_evolution.py` in an agentic orchestrator (ADK agent + tools)
so any agent repo can run continuous skill evolution — quality report from
BigQuery `agent_events` → bottleneck analysis → candidate evolution →
publish gate → GitHub PR — on a weekly Cloud Scheduler cron, without
copying lab-specific code into the SDK.

## What's included

- `deploy/skill_evolution_job/` — job package (`config.py`, `registry.py`,
  `hooks.py`, `engine.py`, `evolve.py`, `coevolve.py`, `bottleneck.py`,
  `tools.py`, `agent.py`, `main.py`, skill instructions), `Dockerfile`,
  `deploy.sh` (job + scheduler + least-privilege IAM), example agent
  registry, README.
- `docs/guides/scheduled-skill-evolution.md` — user guide.
- `tests/test_skill_evo_job_{engine,orchestrator,registry,tools}.py` —
  58 unit tests (no cloud dependencies).
- README index updates (`deploy/`, `docs/`, `scripts/`).

## Host-hook seam

The job is generic; repo-specific behavior is injected by an optional
`EVOLUTION_HOOKS=<module>` from the target repo's clone. Supported hooks
(all optional, each with a safe default): `traffic` (generate baseline
sessions), `score` (candidate scoring), `gate` (publish gate),
`toolbox` (tool descriptions for error analysts), `error_analyst`
(per-failure analysis), `publish` (post-PR registry push). Without hooks
the job runs pure BigQuery-report-driven evolution with the engine's
built-in scoring.

`engine.py` calls the engine via `evolve_skill_compat`, which
feature-detects the newer `evolve_skill` kwargs (`error_analyst_fn`,
`incumbent_score`, `analyst_timeout_s`) so the job runs against both
current `main` and engines that already have the agentic-analyst
extensions (#395).

## Verification

- **Unit tests**: 58/58 green on Python 3.12 and 3.10.
- **Image smoke test** (deployed to Cloud Run, project
  `skill-evolution-lab`): `SELF-TEST PASS` in execution logs; in-cloud
  engine kwarg introspection confirms compat dispatch. Image pip-freeze
  archived (verification run `2026-08-31_231900`, `image_pip_freeze.txt`).
- **Local E2E parity** against a real agent repo
  (`evekhm/skill-evolution-lab`, three-agent HR knowledge supervisor,
  hooks adapter only — no lab code in the SDK), reproducing the lab's
  recorded `lite_local` scenario:

  | Leg | Recorded (lab tooling) | Reproduced (SDK job) |
  |---|---|---|
  | Evolve set V0 (13q) | 38.5% | 46.2% |
  | Evolve set winner (13q) | 100.0% | 100.0% |
  | Held-out exam V0 (55q) | 38.2% | 41.8% |
  | Held-out exam V1 (55q) | 98.2% | 100.0% |
  | Exam delta | +60.0pp | +58.2pp |
  | PR preview (dry-run path) | yes | yes |

  Baselines differ by 1–2 conversations because traffic is regenerated
  and judged live per run; end states match.

- **Deployed E2E** (full loop in Cloud Run: traffic against deployed
  Agent Engine agents → BigQuery quality report → evolve → gate → real
  GitHub PR → registry publish):

  | Leg | Recorded (lab tooling) | Reproduced (SDK job) |
  |---|---|---|
  | Evolve set V0 (13q) | 23.1% | 23.1% |
  | Winner (13q) | 69.2% | 100.0 |
  | Publish gate | 10/10 | pass (10 passed, 1 skipped, 1 xfailed, 1 xpassed) |
  | Real PR opened | yes (#63) | yes (#121) |
  | Post-PR registry publish (host hook) | yes | yes (supervisor v1 pushed) |

## Smoke transcript excerpt

```text
==> running smoke execution (--smoke): --test self-test
==> execution: bqaa-skill-evolution-fjh24
==> smoke OK: SELF-TEST PASS found in execution logs
```

## Notes for reviewers

- `deploy.sh` grants the runtime SA dataset-level
  `roles/bigquery.dataViewer` only (the job never writes events) and
  wires `GH_TOKEN` from Secret Manager; `--github-repo` unset = dry-run
  mode (no PRs).
- `TABLE_ID` (default `agent_events`) is plumbed through `deploy.sh`
  because `scripts/quality_report.py` requires it.


evekhm:
Known follow-ups found during E2E verification (kept out of this PR to keep it reviewable):

- **`create_evolution_pr` swallows git add/commit failures** — a failed commit still opens a PR (empty-PR risk). It should fail loudly when the commit produced no diff.
- **PR title trusts the orchestrator's `meaningful_rate` argument** — in both deployed E2E runs the agent passed the rate from a post-publish quality report with zero fresh sessions, producing titles like "23.1% -> 0%" while the winner actually scored 100.0. `create_evolution_pr` should prefer the recorded `evolved_score.json` over the passed argument.
- **`EVOLUTION_CANDIDATES=2` yielded 3 scored candidates** in both cloud runs — binding between the env knob and the engine's candidate count needs a look.
- **Registry-dict helper placement** — `bottleneck._registry_agents` belongs in `registry.py` with the other registry accessors.

The silent-None score path flagged during diagnosis (score-hook error dicts becoming "unmeasurable (preflight exclusions)" with no detail) is already fixed in this PR.

deploy/skill_evolution_job/skill_evolution_job/hooks.py
```diff
@@ -0,0 +1,258 @@
...
+        f"EVOLUTION_HOOKS={module_path!r} failed to import: {exc}"
+    ) from exc
+  _module_cache[module_path] = module
+  return module
+
+
+def substitute(command: str, values: dict[str, str]) -> str:
+  """Replace ``{placeholder}`` tokens; unknown tokens are left alone."""
+  for key, value in values.items():
+    command = command.replace("{" + key + "}", str(value))
```
caohy1988: **[P1] Quote hook substitutions before shell execution**

The documented `SCORE_CMD` substitutes `{candidate}` and `{run_dir}` directly into a command executed with `shell=True`. These paths come through the agent tools, and substitution does not quote them. A valid path containing spaces becomes multiple arguments; shell metacharacters become executable syntax under the job's runtime identity. A harmless local `printf` check with `/tmp/run with spaces` received three arguments instead of one.

Pass substituted values as arguments or environment variables, or define an unquoted-placeholder contract and shell-quote each value exactly once before execution.

deploy/skill_evolution_job/skill_evolution_job/main.py
```diff
@@ -0,0 +1,788 @@
...
+  os.makedirs(run_dir, exist_ok=True)
+  with open(os.path.join(run_dir, "trace_selector.json"), "w") as f:
+    json.dump(selector, f, indent=2)
+  logger.info("Trace selector for this run: %s", selector)
+
+  result = tools.run_quality_report(
+      output_path=os.path.join(run_dir, "quality_report.json"),
+      run_dir=run_dir,
+  )
+  if result.get("error"):
```
caohy1988, line 126: **[P1] Report failures are returned as a successful no-work run**

A failed quality report is converted to `(None, 0)`, which is also the insufficient-session sentinel. With the default configuration and no traffic hook, `run_evolution_agent` then returns `NOTHING TO DO`, and `main()` exits zero. I reproduced this with a report result of `{"error": "403 Permission denied"}`: a broken report was reported as a successful no-work run.

Preserve an explicit report-error state and propagate it as a nonzero job exit. Reserve the no-work path for a successfully generated report below `MIN_SESSIONS`.

deploy/skill_evolution_job/skill_evolution_job/tools.py
```diff
@@ -0,0 +1,2390 @@
...
+    if score_hook is None:
+      logger.info(
+          "score hook not configured (%s); engine falls back to size-based"
+          " candidate selection",
+          score_reason,
+      )
+    elif _score_dir:
+      try:
+        with open(quality_report_path) as _rf:
+          _incumbent = json.load(_rf).get("summary", {}).get("meaningful_rate")
```
caohy1988: **[P1] Measure the incumbent on the candidate evaluation set**

The scheduled input report measures production traffic, while the score hook can evaluate candidates on a different host evaluation set. Passing the report's rate as `incumbent_score` treats those measurements as comparable. This becomes observable with #395, where the engine correctly stops rescoring a supplied incumbent.

Using the actual job adapter and both engine versions, with production baseline **20**, eval-set incumbent **90**, and eval-set candidate **80**, the current engine kept the incumbent; the #395 engine selected the regressing candidate because only **80 vs 20** was compared. `coevolve.py` supplies the same report-derived baseline.

Measure the incumbent through the same score hook/evaluation set as candidates. Reuse a premeasured baseline only with matching evaluation provenance; the later PR-metrics denominator warning does not repair selection.

deploy/skill_evolution_job/skill_evolution_job/coevolve.py
```diff
@@ -0,0 +1,478 @@
...
+          logger.info(
+              "Saved evolved %s skill to %s (%d chars)",
+              label,
+              out_path,
+              len(evolved),
+          )
+
+        return agent_name, agent_result
+
+      except Exception as e:  # noqa: BLE001
```
caohy1988, line 372: **[P1] Restore the incumbent before retrying failed candidate scoring**

The supported score hook can leave the candidate installed in `SKILL.md`. If it raises after installation, this retry path never restores the saved `current_skill`, and the next `evolve()` call rereads the failed candidate as its incumbent.

I reproduced this through the actual coevolution/engine flow: attempt 1 installed a candidate and raised; attempt 2 produced no patches; coevolution returned success on attempt 2 with the unaccepted candidate installed. Exhausted retries also leave failed candidates in place. The single-agent exception path has the same missing restoration.

Restore the pre-attempt skill before every retry/error return, or wrap each score-hook call in `try/finally` and install only the successfully selected winner.

deploy/skill_evolution_job/skill_evolution_job/tools.py
```diff
@@ -0,0 +1,2390 @@
...
+    dry_run = True
+    publish_disabled = True
+
+  if not dry_run and not output_file:
+    gate_agent, error = _resolve_agent(agent)
+    if error:
+      return error
+    gate_hook, gate_reason = hooks.get_hook("gate")
+    if gate_hook is not None:
+      passed, detail = gate_hook(run_dir, version, gate_agent)
```
caohy1988: **[P1] Publish gate checks a tree discarded before publication**

The publish gate runs before loading the selected version and before stashing the current tree and rebuilding a branch from `origin/<base>`. It can therefore validate a different version or multi-agent combination from the PR that is actually pushed.

A local Git reproduction, with only push/PR creation mocked, returned success after the gate checked **A1+B1**, while the generated commit contained **A0+B1**. This is the normal risk when sequential coevolution scored B1 with A1 already installed, but publication commits only B's skill.

Construct the exact intended PR tree first and gate that tree. Publish dependent coevolved skills together, or independently evaluate each standalone PR against its real base.

deploy/skill_evolution_job/skill_evolution_job/bottleneck.py
```diff
@@ -0,0 +1,360 @@
...
+
+  classifier_prompt = _build_classifier_prompt(agents)
+  agent_names = list(agents.keys())
+  default_agent = agent_names[0] if agent_names else "unknown"
+
+  sessions = report.get("sessions", [])
+  failures = [
+      s
+      for s in sessions
+      if s.get("metrics", {}).get("response_usefulness", {}).get("category")
```
caohy1988: **[P2] Count parroted recoveries in bottleneck detection**

This failure filter disagrees with the engine's `partition_trajectories`: a `meaningful` session with a `parroted` sub-trajectory is an engine failure but is omitted here. I reproduced a report for which both engine versions counted one failure while bottleneck detection returned `recommendation="none"`. `coevolve()` then exits before evolving it.

Reuse the loaded engine's failure partition so `count_failures`, bottleneck detection, and evolution agree, including the supported parroted-recovery evidence keys.

deploy/skill_evolution_job/skill_evolution_job/coevolve.py
```diff
@@ -0,0 +1,478 @@
...
+        # None = unmeasurable: evolve() floors it for selection but
+        # never records it as the deployed score, so a flaked run
+        # cannot lower the bar.
+        logger.warning(
+            "Candidate scored report %s has preflight exclusions; "
+            "score unmeasurable",
+            scored_report,
+        )
+        return None
+      return float(res.get("meaningful_rate", 0.0))
```
caohy1988: **[P2] Reject failed and unmeasurable coevolution scores**

Unlike the single-agent score adapter, this path consumes `meaningful_rate` without checking `error`/`status`, `skipped`, or `unmeasurable`. For example, `{"unmeasurable": true, "meaningful_rate": 95}` is accepted as a measured 95 and can win selection or become the next agent's incumbent, although `run_evolution` rejects the same result. A missing rate also becomes a measured zero.

Use one score-result normalizer in both modes. Failed, skipped, and unmeasurable results must retain that status rather than become usable scores.

deploy/skill_evolution_job/skill_evolution_job/hooks.py
```diff
@@ -0,0 +1,258 @@
...
+  return command
+
+
+def _run_cmd(name: str, command: str, values: dict[str, str]) -> dict:
+  """Run a *_CMD hook; returns exit code, output tail, parsed result."""
+  rendered = substitute(command, values)
+  logger.info("Running %s hook command: %s", name.upper(), rendered)
+  proc = subprocess.run(
```
caohy1988, line 230: **[P2] Command hooks run outside the cloned host repository**

The documented fallbacks use relative host-repo paths, e.g. `python eval/score.py ...` and `pytest tests/skill_contract -q`, but this subprocess inherits the job's current directory. In the image that is `/app`; the host repository is cloned into a separate temporary workdir. The evaluator/tests therefore cannot be found, or an unrelated local test path can be selected.

Run command hooks from the resolved host workdir, retaining the caller directory only when no host workdir exists. Normalize artifact arguments before changing `cwd`.

deploy/skill_evolution_job/skill_evolution_job/hooks.py
```diff
@@ -0,0 +1,258 @@
...
+  }
+  # The last non-empty stdout line may carry structured output.
+  for line in reversed((proc.stdout or "").strip().splitlines()):
+    line = line.strip()
+    if not line:
+      continue
+    try:
+      parsed = json.loads(line)
+      if isinstance(parsed, dict):
+        result.update(parsed)
```
caohy1988, line 250: **[P2] Preserve the real hook exit code when merging JSON output**

Parsed stdout can overwrite reserved execution metadata, including `returncode`. The gate wrapper then decides pass/fail from that overwritten field. A local command that printed `{"returncode": 0}` and exited **7** returned a passing gate verdict. Failed score/traffic commands are subject to the same merge.

Keep the real process exit code separate from the hook's JSON payload, or restore reserved fields from the process result after parsing.

deploy/skill_evolution_job/skill_evolution_job/main.py
```diff
@@ -0,0 +1,788 @@
...
+
+  if args.quality_source:
+    os.environ["QUALITY_SOURCE"] = args.quality_source
+
+  # Make the scoping flags BINDING at the tool layer (the orchestrating
+  # agent treats prompt overrides as hints; these env vars are enforced
+  # by evolve.py / coevolve.py regardless of what the agent decides).
+  if args.candidates is not None:
+    os.environ["EVOLUTION_CANDIDATES"] = str(args.candidates)
+  if args.rounds is not None:
```
caohy1988: **[P2] The advertised hard two-round cap is unenforced in default scheduled runs**

Normal scheduled/full-loop runs omit `--rounds`, so this branch never sets `EVOLUTION_MAX_ROUNDS`. `_round_guard` only refuses calls when that variable is present. The shipped skill nevertheless specifies a hard two-round maximum and tells the model that tools enforce it outside the model's control. The default job currently relies entirely on prompt compliance for that cap.

Set and validate the default tool-level cap for agent-driven runs, and make explicit override behavior agree with the documented limit.

deploy/skill_evolution_job/skill_evolution_job/tools.py
```diff
@@ -0,0 +1,2390 @@
...
+  Relative to the registry's ``repo_root`` inside the host-repo workdir
+  — the directory ``create_evolution_pr`` commits in — so joining it
+  back onto the workdir root can never escape the clone.
+  """
+  reg = registry.get_registry()
+  spec = reg.agents.get(agent)
+  if spec is None:
+    return None
+  return os.path.relpath(
+      os.path.join(spec.skill_dir, "SKILL.md"), reg.repo_root
```
caohy1988: **[P2] Preserve registry subdirectories in the committed skill path**

A registry may set `repo_root` to a subdirectory of the Git checkout. This helper removes that prefix, but publication later joins the returned path to the Git root. With checkout `/clone`, registry root `/clone/services/foo`, and skill `skills/a`, the source is `/clone/services/foo/skills/a/SKILL.md` while the PR writes `/clone/skills/a/SKILL.md`, leaving the real agent skill unchanged.

Derive the commit path relative to the actual Git workdir and validate containment against that same root. Preserve the registry subdirectory in the staged path and publish-hook arguments.

deploy/skill_evolution_job/skill_evolution_job/tools.py
```diff
@@ -0,0 +1,2390 @@
...
+          Triggers compaction if exceeded to prevent skill bloat.
+
+  Returns:
+      Dict with evolved skill content, path, candidate paths, and
+      summary statistics.
+  """
+  from . import evolve as evolve_mod
+
+  model_id = model_id or config.get_config().evolution_model_id
+  skill_dir = _resolve_skill_dir(skill_dir)
```
caohy1988, line 160: **[P2] The single-agent tool ignores the bound target agent**

`main()` exports `--mode <agent>` as `EVOLUTION_TARGET_AGENTS` and describes it as binding at the tool layer, but only coevolution enforces it. `run_evolution` accepts another agent or direct directory and writes its `SKILL.md` without checking the bound target. I reproduced a successful write to another agent with `EVOLUTION_TARGET_AGENTS=allowed_agent`.

Resolve the directory to a registry agent and enforce the binding before round accounting, scoring, or file writes, covering both aliases and direct paths.

deploy/skill_evolution_job/skill_evolution_job/tools.py
```diff
@@ -0,0 +1,2390 @@
...
+      logger.warning(
+          "run_evolution asked for candidates=%d; EVOLUTION_CANDIDATES=%d"
+          " is binding and wins",
+          candidates,
+          bound,
+      )
+    candidates = bound
+
+  candidates_dir = None
+  if candidates is not None and candidates > 1:
```
caohy1988: **[P2] Persist selection scores for single-candidate evolution**

With explicit `candidates=1`, `candidates_dir` remains `None` even when a `run_dir` and score hook are available. `evolve()` uses that directory to locate `evolved_score.json`, so no authoritative score is recorded. I confirmed the missing record with both real engine versions.

Later metrics can then use an earlier round's stale score or a candidate report that does not correspond to the deployed outcome, including when the incumbent was retained. Allocate an artifact destination for single-candidate runs too, or pass the score destination independently of the candidate count.

deploy/skill_evolution_job/README.md
```diff
@@ -0,0 +1,277 @@
...
+| `traffic` | `traffic(run_dir) -> dict` | Generate fresh eval sessions when BigQuery has fewer than `MIN_SESSIONS` |
+| `score` | `score(candidate_path, skill_dir, run_dir) -> dict` | Score one candidate `SKILL.md` with your own agent + eval set; must return `{"meaningful_rate": <0-100>, ...}` |
+| `gate` | `gate(run_dir, version, agent) -> (bool \| None, str)` | Pre-publish acceptance check (e.g. run your test suite against the evolved skill); only an explicit `False` blocks the PR — `None` means inconclusive and proceeds |
+| `toolbox` | `toolbox(agent) -> str` | Text description of the agent's tools, injected into analyst prompts |
+| `error_analyst` | `error_analyst(client, model, session, skill, tools)` | Custom per-failure analyst (only used when the engine supports `error_analyst_fn`; see `--scripts-dir` above) |
+| `publish` | `publish(skill_dir, run_dir)` | Push the accepted skill to a registry/deployment target after the PR |
+
+`traffic`, `score` and `gate` also accept a **shell-command fallback** via
+`TRAFFIC_CMD` / `SCORE_CMD` / `GATE_CMD` for hosts whose tooling isn't
+importable Python. Placeholders `{run_dir}`, `{report}`, `{candidate}`,
```
caohy1988: **[P2] Advertised report placeholder is never substituted**

`{report}` is advertised here and in the hook-module documentation, but none of the traffic, score, or gate substitution mappings supplies a `report` value. `substitute()` deliberately preserves unknown placeholders, so a configured command using the documented token receives the literal `{report}` instead of a report path.

Define and pass the appropriate report path for the supported hooks, or remove `{report}` from the advertised contract. Add a contract test for the documented placeholder set.

caohy1988 (changes requested):
Fresh review of `655cf7f753c5597aebb597792d0e8e40b3294f01`, including integration with #395 at `6378d7ba8a3ce33712dac90224b91b1d6dae86b1`.

Requesting changes: **5 P1 and 9 P2 findings**, detailed in the inline comments. The highest-priority fixes are comparable incumbent/candidate scoring, restoration after scoring failures, gating the actual published tree, safe command substitution, and surfacing report failures as job failures.

Validation: the four focused job test files pass (**80 tests**, Python 3.13.5). All current CI checks completed successfully or were skipped. Additional local checks reproduced the cross-engine scoring regression, failed-scorer retry behavior, checked-vs-committed tree mismatch, report-error success path, path substitution/exit-code behavior, target binding, parroted-failure mismatch, and missing single-candidate score records. Proposed findings received a separate validation pass.

Coverage limits: no live Cloud Run, IAM, BigQuery, Vertex AI, or GitHub publish execution was performed. Git fixtures used real local operations with push/PR creation mocked. Add integration tests for these boundaries alongside the fixes.

deploy/skill_evolution_job/skill_evolution_job/hooks.py
caohy1988: > A valid path containing spaces becomes multiple arguments; shell metacharacters become executable syntax under the job's runtime identity.

Fixed in `366379e`. placeholder values are shell-quoted exactly once in a single pass, so embedded placeholder names remain literal. Quoted/escaped placeholders, comments, and nested shell contexts fail clearly before execution; complex commands can live in a host script receiving ordinary placeholder arguments. Regression coverage exercises spaces, both quote types, shell metacharacters, newlines, and substitution chaining. Shell line continuations and legacy arithmetic expansion are also rejected, with execution-marker regressions for both bypasses.

deploy/skill_evolution_job/skill_evolution_job/main.py
caohy1988: > Preserve an explicit report-error state and propagate it as a nonzero job exit.

Fixed in `366379e`. report-generation failures now raise a distinct QualityReportError. Both the initial preflight and the retry after traffic return ERROR with the original failure, and the CLI exits 1. Only successfully generated reports below MIN_SESSIONS take the no-work path. Regression tests cover both stages and verify that an initial report failure never triggers traffic.

deploy/skill_evolution_job/skill_evolution_job/tools.py
caohy1988: > Measure the incumbent through the same score hook/evaluation set as candidates.

Fixed in `366379e`. The job now measures the current skill through the same host score hook before each single-agent or coevolution selection, rather than using the production report or a prior agent’s score. The adapter caches that measurement for engines that re-score the incumbent and fails closed if the baseline cannot be measured. Regression tests using the real engine keep the 90-point evaluation incumbent over an 80-point candidate despite a production report of 20, in both modes and both engine calling conventions. Configured scoring also remains mandatory when callers omit the run/output directory: the adapter creates a temporary artifact directory and still measures the incumbent and candidates.

deploy/skill_evolution_job/skill_evolution_job/coevolve.py
caohy1988: > Restore the pre-attempt skill before every retry/error return, or wrap each score-hook call in `try/finally` and install only the successfully selected winner.

Fixed in `366379e`. Both selection paths now use a shared hook adapter that restores the exact incumbent SKILL.md bytes in finally after every score, including exceptions. Coevolution also restores its saved original before retrying or returning an error, and installs only a successful selection. Tests reproduce a hook that installs a candidate then times out; a later no-patch retry keeps the original, and the single-agent error path preserves it too.

deploy/skill_evolution_job/skill_evolution_job/tools.py
caohy1988: > Construct the exact intended PR tree first and gate that tree.

Fixed in `366379e`. Publication now creates and locally commits the selected snapshot on its actual PR base, then invokes the gate in that same host checkout. This independently evaluates the standalone PR combination while preserving module and command hook path anchors. Any gate change to tracked files, the index, branch, or commit refuses the push. The original branch or detached HEAD and exact staged/unstaged changes are restored on success and failures.

A real Git regression reproduces A1+B1 passing locally but A0+B1 failing the acceptance condition: publication now refuses it, makes no push/PR call, and restores A1+B1. Success tests verify the gated commit/tree equals the commit sent to push, including an earlier selected snapshot. Mutation, exception, stash failure, commit failure, detached HEAD, and existing stash cases also pass. The documented None verdict and explicit no-hook GATE_POLICY=skip semantics are retained; hosts must supply a conclusive dependency gate to block incompatible standalone changes.

deploy/skill_evolution_job/skill_evolution_job/bottleneck.py
caohy1988: > Reuse the loaded engine's failure partition so `count_failures`, bottleneck detection, and evolution agree, including the supported parroted-recovery evidence keys.

Fixed in `366379e`. bottleneck detection now delegates its failure selection to evolve.partition_trajectories, which calls the loaded SDK engine. Regression tests use the real engine and verify that both meaningful and declined sessions with parroted recoveries reach classification through either sub_trajectories or execution_sub_trajectories.

deploy/skill_evolution_job/skill_evolution_job/coevolve.py
caohy1988: > Use one score-result normalizer in both modes.

Fixed in `366379e`. Single-agent and coevolution selection now share one result normalizer. Non-dict responses, error/failed status, error fields, skipped or unmeasurable results, missing/nonfinite rates, and excluded or unreadable scored reports yield no usable score. An invalid baseline aborts selection; an invalid candidate cannot replace the incumbent. Paired tests exercise these responses through both actual selection flows. Module-hook results with an explicit nonzero or invalid returncode are rejected even if they include a high meaningful_rate; returncode 0/False and results without that field remain supported.

deploy/skill_evolution_job/skill_evolution_job/hooks.py
caohy1988: > Run command hooks from the resolved host workdir, retaining the caller directory only when no host workdir exists. Normalize artifact arguments before changing `cwd`.

Fixed in `366379e`. command hooks now run in the resolved host checkout, with the caller directory as the fallback. Candidate, skill, and run paths are made absolute beforehand, and a returned report_path is resolved against the command's directory. A regression executes a relative evaluator script from a separate host checkout and verifies both argument and result paths.

deploy/skill_evolution_job/skill_evolution_job/hooks.py
caohy1988: > A local command that printed `{"returncode": 0}` and exited **7** returned a passing gate verdict.

Fixed in `366379e`. real process returncode and masked output_tail are assigned after parsing structured stdout, so hook data cannot override execution metadata. Regression tests verify that exit 7 still fails the gate, raises for scoring, and remains visible in traffic results despite stdout claiming returncode 0.

deploy/skill_evolution_job/skill_evolution_job/main.py
caohy1988: > Set and validate the default tool-level cap for agent-driven runs, and make explicit override behavior agree with the documented limit.

Fixed in `366379e`. the CLI and direct run_evolution_agent calls now resolve and bind the cap before preflight, using an explicit argument, then EVOLUTION_MAX_ROUNDS, then the default of 2. The shared validator rejects malformed values and limits outside 0–2; 0 preserves the no-evolution behavior. Prompts and CLI help report the actual cap. Regression tests exercise precedence, invalid input, third-round refusal and --rounds 0. Each independent agent run receives a fresh round budget; a regression executes two runs and verifies two allowed rounds plus a third-round refusal in each.

deploy/skill_evolution_job/skill_evolution_job/tools.py
caohy1988: > Preserve the registry subdirectory in the staged path and publish-hook arguments.

Fixed in `366379e`. Publication derives the path from the actual ensured Git checkout, retaining the registry root prefix and checking realpath containment both before and after switching to the PR base. A real Git regression using repo_root=services/foo verifies that the commit changes only services/foo/skills/b/SKILL.md, the gate and push see the same selected content, and the publish hook receives that exact nested skill directory.

deploy/skill_evolution_job/skill_evolution_job/tools.py
caohy1988: > Resolve the directory to a registry agent and enforce the binding before round accounting, scoring, or file writes, covering both aliases and direct paths.

Fixed in `366379e`. The single-agent tool now resolves the allowed registry agents to canonical paths and refuses any other target before round accounting, hooks, or writes. Regression tests cover agent names, direct paths, and symlinks for both allowed and excluded targets.

deploy/skill_evolution_job/skill_evolution_job/tools.py
caohy1988: > Allocate an artifact destination for single-candidate runs too, or pass the score destination independently of the candidate count.

Fixed in `366379e`. Single-candidate runs now receive a candidate/artifact directory from run_evolution, and evolve records the measured current skill or selected winner rather than the historical report’s baseline. It clears a prior score before baseline measurement, persists a successful baseline immediately, and overwrites it only with the selected measured winner. Real-engine tests verify retained-incumbent 90 and selected-winner 95 records, plus baseline-error and candidate-error recovery. Omitting the artifact directory now allocates one instead of disabling scoring; single-agent results expose run_dir and direct coevolution results expose output_dir. Paired real-engine tests verify that the retained incumbent is recorded there.

deploy/skill_evolution_job/README.md
caohy1988: > `{report}` is advertised here and in the hook-module documentation, but none of the traffic, score, or gate substitution mappings supplies a `report` value.

Fixed in `366379e`. both documentation surfaces now list the exact placeholders for each hook, including gate's version. The unsupported report token is explicitly rejected before execution; hosts can pass an explicit report filename under run_dir. Tests exercise every documented mapping and rejection of report for all three command hooks.

caohy1988:
> Requesting changes: **5 P1 and 9 P2 findings**, detailed in the inline comments.

All 14 findings are addressed in `366379e`, with the branch updated to current main, including #395 and #470. Each original thread contains its fix and regression evidence.

- Scoring measures incumbents and candidates on the same evaluation set, rejects unusable results, restores failed attempts, and records the deployed outcome even for one candidate or omitted artifact directories.
- Report failures remain job failures; target restrictions and the default two-round limit are enforced by the tools.
- Command hooks quote values once, reject unsafe substitution contexts, use the host checkout, and preserve actual exit status.
- Publication gates the selected commit on its actual base, preserves nested registry paths, and restores the original checkout and index after failures.

Validation: **4,924 tests passed, 56 skipped**, including all **213 skill-evolution job tests**, on Python 3.13.5. Formatting, notebook synchronization, and sdist/wheel builds passed. Independent review included shell injection reproductions, real-engine score selection, and real local Git publication failures.

The configured gate still controls dependency acceptance; documented `None` and explicit no-hook `GATE_POLICY=skip` behavior are preserved. Live Cloud Run, IAM, BigQuery, and Vertex AI execution was not repeated for this revision.

GitHub CI also passed on this exact commit: Python **3.10, 3.11, 3.12, 3.13, and 3.14**, formatting, browser smoke, security checks, and package build.

caohy1988 (approved):
Verified the fixes for all 14 findings in 366379e. Every original thread has a submitted reply and is resolved. The full local suite passed (4,924 tests; 56 skipped), and all CI checks, including Python 3.10–3.14 and the package build, passed on this commit. Independent review and regressions cover score comparability/restoration, command injection, job failure signaling, target/round limits, and exact-tree publication.

No remaining blocking findings from this review. Live Cloud Run/BigQuery/Vertex execution was not repeated.

PR #477, opened 2026-09-05
feat: make skill evolution host runs auditable

### LINKED ISSUE #397: skill_evolution host hooks: contract hardening follow-ups (from PR #395 review)

Beverly621:
## Summary

Complete the remaining host-contract hardening tracked in #397 so hosted skill-evolution runs are both auditable and safer to integrate.

## Changes

- Record `source: "builtin" | "host"` on every quality-gated patch in `*_patches.json`.
- Keep patch text and provenance aligned through the quality gate, including mixed builtin/host fleets.
- Preserve provenance through collector wrappers that reorder, filter, deduplicate, or reconstruct patch strings when equal-text provenance is unambiguous.
- Fall back conservatively when equal recorded patch text has conflicting sources.
- Preserve the public `collect_patches() -> list[str]` return contract and compatibility with hosts that replace `collect_patches`.
- Define the supported integration surface with `__all__` and document the additive session-schema policy.
- Document that host analysts must treat `correction_boundaries[*].correct_fact` as an unverified user hypothesis and verify it through tools before producing a patch.
- Add the host-contract stability and artifact-schema change to the changelog.

The non-string result handling and bounded analyst timeout described in #397 are already present on `main`; this change covers the remaining provenance and stability/documentation items.

## Validation

- `pytest -q tests/test_skill_evolution.py tests/test_skill_evo_job_selection_regressions.py` — 150 passed
- Full sandbox suite — 4,933 passed, 74 skipped; the four local permission-sensitive failures (`ps` access and browser smoke port binding) were rerun outside the sandbox and all 4 passed
- `pyink --check scripts/skill_evolution.py tests/test_skill_evolution.py`
- `isort --check-only scripts/skill_evolution.py tests/test_skill_evolution.py`
- `git diff --check`

Closes #397

### DIFF
--- a/scripts/skill_evolution.py
+++ b/scripts/skill_evolution.py
@@ -50,6 +50,7 @@ import argparse
 from collections import Counter
 from concurrent.futures import as_completed
 from concurrent.futures import ThreadPoolExecutor
+from contextvars import ContextVar
 import json
 import logging
 import math
@@ -70,6 +71,15 @@ __all__ = [
     "partition_trajectories",
     "select_candidate",
 ]
+
+# ``collect_patches`` is an established replacement seam, so its signature and
+# exact ``list[str]`` result must stay unchanged. During ``evolve_skill`` only,
+# this context-local identity map carries provenance alongside the returned
+# string objects. Wrappers that reorder, filter, or deduplicate those objects
+# therefore retain correct provenance without receiving a new private keyword.
+_PATCH_PROVENANCE: ContextVar[Optional[dict[int, list[str]]]] = ContextVar(
+    "skill_evolution_patch_provenance", default=None
+)
 
 # Segment outcome icons shared by both sub-trajectory renderers.
 _SEGMENT_OUTCOME_ICONS = {"recovered": "+", "parroted": "~"}
@@ -1040,7 +1050,6 @@ def collect_patches(
     tools=None,
     error_analyst_fn: Optional[ErrorAnalystFn] = None,
     analyst_timeout_s: Optional[float] = None,
-    _patch_sources: Optional[list[str]] = None,
 ):
   """Run the analyst fleet over the report. Returns the list of kept patches.
 
@@ -1210,12 +1219,13 @@ def collect_patches(
     )
 
   kept = []
-  kept_sources = []
+  provenance = _PATCH_PROVENANCE.get()
   for patch, source in patches:
     reason = _quality_gate_reason(patch)
     if reason is None:
       kept.append(patch)
-      kept_sources.append(source)
+      if provenance is not None:
+        provenance.setdefault(id(patch), []).append(source)
     else:
       logger.warning("Quality gate rejected a patch (%s): %.80r", reason, patch)
   logger.info(
@@ -1223,8 +1233,6 @@ def collect_patches(
       len(patches),
       len(kept),
   )
-  if _patch_sources is not None:
-    _patch_sources.extend(kept_sources)
   return kept
 
 
@@ -1425,29 +1433,34 @@ def evolve_skill(
   # applies to standalone collect_patches usage only.
   client = client or _make_client(project, location)
 
-  patch_sources: list[str] = []
-  patches = collect_patches(
-      report,
-      current_skill,
-      client=client,
-      model=model,
-      max_workers=max_workers,
-      max_success_samples=max_success_samples,
-      analyst_mode=analyst_mode,
-      tools=tools,
-      error_analyst_fn=error_analyst_fn,
-      analyst_timeout_s=analyst_timeout_s,
-      _patch_sources=patch_sources,
-  )
+  patch_provenance: dict[int, list[str]] = {}
+  provenance_token = _PATCH_PROVENANCE.set(patch_provenance)
+  try:
+    patches = collect_patches(
+        report,
+        current_skill,
+        client=client,
+        model=model,
+        max_workers=max_workers,
+        max_success_samples=max_success_samples,
+        analyst_mode=analyst_mode,
+        tools=tools,
+        error_analyst_fn=error_analyst_fn,
+        analyst_timeout_s=analyst_timeout_s,
+    )
+  finally:
+    _PATCH_PROVENANCE.reset(provenance_token)
   if not patches:
     logger.warning("No patches to consolidate; returning the current skill.")
     return current_skill
-  # Older hosts and tests may replace ``collect_patches`` with a compatible
-  # callable that returns ``list[str]`` but does not know about the private
-  # provenance sink. Preserve that extension pattern and conservatively label
-  # those patches as builtin-generated.
-  if not patch_sources:
-    patch_sources = ["builtin"] * len(patches)
+  # Plain strings from a legacy replacement have no recorded provenance, so use
+  # the conservative builtin fallback. Identity-retaining wrappers around the
+  # SDK collector resolve through the context map even when they reorder or
+  # filter the result.
+  patch_sources = []
+  for patch in patches:
+    sources = patch_provenance.get(id(patch), [])
+    patch_sources.append(sources.pop(0) if sources else "builtin")
 
   logger.info("Generating %d candidate(s)...", candidates)
   cands = []
--- a/scripts/skill_evolution.py
+++ b/scripts/skill_evolution.py
@@ -36,6 +36,14 @@ Or run as a CLI:
 Auth: uses Vertex AI via the google-genai client. Set GOOGLE_CLOUD_PROJECT and
 GOOGLE_CLOUD_LOCATION (or pass project/location), and authenticate with ADC
 (gcloud auth application-default login).
+
+Host contract: integrations should import only the names listed in ``__all__``.
+``evolve_skill`` and ``collect_patches`` accept the quality-report session
+mapping described in their docstrings; new optional session fields are additive,
+and any incompatible schema change must be called out in ``CHANGELOG.md``.
+``error_analyst_fn`` is the supported extension seam for failure analysis and
+must return patch text or ``None``. The public return value of
+``collect_patches`` remains ``list[str]``.
 """
 
 import argparse
@@ -53,6 +61,15 @@ from typing import Any, Callable, Optional
 # Host analyst signature: fn(client, model, session, current_skill, tools)
 # -> patch text or None (see collect_patches).
 ErrorAnalystFn = Callable[[Any, str, dict, str, Optional[str]], Optional[str]]
+
+__all__ = [
+    "ErrorAnalystFn",
+    "collect_patches",
+    "evolve_skill",
+    "format_trajectory",
+    "partition_trajectories",
+    "select_candidate",
+]
 
 # Segment outcome icons shared by both sub-trajectory renderers.
 _SEGMENT_OUTCOME_ICONS = {"recovered": "+", "parroted": "~"}
@@ -639,12 +656,14 @@ def _write_evolution_artifacts(
     current_skill,
     version_label="v1",
     selection_note=None,
+    patch_sources=None,
 ):
   """Persist the engine's intermediate reasoning for inspection/audit.
 
   When ``evolve_skill`` is given ``artifacts_dir`` it writes, under that dir
   (``<label>`` is ``version_label``, so a V1->V2 round writes ``v2_*``):
-    - ``<label>_patches.json``  -- every analyst patch (root-cause category + text)
+    - ``<label>_patches.json``  -- every analyst patch (source, root-cause
+                                   category, and text)
     - ``<label>_candidates/``   -- each best-of-N consolidation candidate (the
                                    chosen one tagged ``_SELECTED``)
     - ``<label>_prevalence.txt``-- root-cause category tally across the patches
@@ -653,14 +672,26 @@ def _write_evolution_artifacts(
   """
   os.makedirs(artifacts_dir, exist_ok=True)
 
+  if patch_sources is None:
+    patch_sources = ["builtin"] * len(patches)
+  if len(patch_sources) != len(patches):
+    raise ValueError("patch_sources must align one-to-one with patches")
+  unexpected_sources = set(patch_sources) - {"builtin", "host"}
+  if unexpected_sources:
+    raise ValueError(
+        "patch_sources entries must be 'builtin' or 'host'; got"
+        f" {sorted(unexpected_sources)!r}"
+    )
+
   records = []
-  for i, patch in enumerate(patches):
+  for i, (patch, source) in enumerate(zip(patches, patch_sources)):
     match = re.search(r"## Root Cause\s*\n\s*\[?(\w+)\]?", patch) or re.search(
         r"## Pattern\s*\n\s*\[?(\w+)\]?", patch
     )
     records.append(
         {
             "index": i + 1,
+            "source": source,
             "category": match.group(1) if match else None,
             "patch": patch,
         }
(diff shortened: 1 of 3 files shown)


Beverly621:
@caohy1988, this PR completes the remaining host-contract items identified during the review of #395 and tracked in #397.

The CI workflow from my fork is currently waiting for maintainer approval. When you have time, could you please approve the workflow run and review the PR?

The focused skill-evolution test suites pass locally (142 tests), with full validation details included in the PR description. Thank you!

haiyuan-eng-google:
@Beverly621 please help star the repo and thanks for contribution

scripts/skill_evolution.py
```diff
@@ -1397,10 +1437,17 @@ def evolve_skill(
       tools=tools,
       error_analyst_fn=error_analyst_fn,
       analyst_timeout_s=analyst_timeout_s,
+      _patch_sources=patch_sources,
```
caohy1988: [P1] Preserve the previous signature for replacement collectors

`evolve_skill` now passes `_patch_sources` unconditionally. A host replacement declaring the previous `collect_patches(report, current_skill, *, client, model, max_workers=10, max_success_samples=15, analyst_mode="both", tools=None, error_analyst_fn=None, analyst_timeout_s=None)` signature raises `TypeError: unexpected keyword argument '_patch_sources'` before the compatibility fallback can run. I reproduced this against the base and PR commits: the same replacement succeeds on base and fails here, including when artifacts are disabled.

The added compatibility test uses `lambda *args, **kwargs`, which hides this failure. Keep the old keyword set for legacy replacements, and pass the private sink only where its support is known. Add the exact-signature regression; avoid catching and retrying arbitrary `TypeError` from inside the collector.

caohy1988: [P2] Keep provenance aligned through collector wrappers

Accepting `**kwargs` alone does not make a legacy wrapper compatible with this private output list. A wrapper can call the original collector and then return a sorted, filtered, or deduplicated `list[str]`. The original fills `patch_sources` before that transformation. I reproduced both cases with a mixed host/builtin fleet: reversing the returned list silently swaps the source labels; filtering one patch raises `ValueError: patch_sources must align one-to-one with patches` while writing artifacts, after consolidation has already run. Both wrappers complete successfully on base.

Keep provenance coupled to patches through transformations, or use an explicit aligned-provenance adapter and keep legacy wrappers on the former call contract. Add reorder/filter wrapper regressions through `evolve_skill(..., artifacts_dir=...)`.

caohy1988 (changes requested):
Reviewed commit `63d0332` against `f5e03bd`.

The built-in collection path keeps patch text and source together through the quality gate, and the host-contract documentation addresses the remaining #397 documentation items. The supported replacement-collector paths still have two confirmed compatibility/provenance defects detailed inline, so I recommend fixing those before merge.

Validation: 142 focused tests passed locally; pyink/isort and diff checks passed. I approved the pending workflow runs, and [CI passed](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/33941973011) across Python 3.10–3.14, formatting, browser smoke, and package build. Additional mocked reproductions expose the replacement-path failures despite the existing tests passing. No live model calls were needed for these reproductions.

scripts/skill_evolution.py
Beverly621: **Fixed in `1c8cd6d`.**

- **Restored backward compatibility:** removed the private `_patch_sources` keyword from both the `collect_patches` signature and the `evolve_skill` call site.
- **Preserved the exact legacy contract:** replacement collectors now receive only the pre-change keyword set.
- **Added a realistic regression test:** the replacement uses an exact legacy signature without `**kwargs` and is exercised through `evolve_skill(..., artifacts_dir=...)`.
- **No unsafe fallback:** the implementation does not catch and retry arbitrary `TypeError` exceptions.

Beverly621: **Fixed in `1c8cd6d`.**

- **Provenance now stays aligned with each patch** through a context-local identity map.
- Wrappers can **reorder, filter, or deduplicate** returned patches without swapping `host`/`builtin` labels or causing length-mismatch errors.
- The public **`collect_patches() -> list[str]` contract remains unchanged**.
- Added end-to-end mixed `host`/`builtin` regression coverage for **reordering, filtering, and deduplication** through `evolve_skill(..., artifacts_dir=...)`.

The focused suites now pass **145 tests**.

scripts/skill_evolution.py
```diff
@@ -1175,10 +1219,13 @@ def collect_patches(
...
 
   kept = []
-  for patch in patches:
+  provenance = _PATCH_PROVENANCE.get()
+  for patch, source in patches:
     reason = _quality_gate_reason(patch)
     if reason is None:
       kept.append(patch)
+      if provenance is not None:
+        provenance.setdefault(id(patch), []).append(source)
```
caohy1988: [P2] Keep patch objects alive while using their IDs as provenance keys

The map stores only `id(patch)` and source labels, so it does not keep the corresponding string alive. An identity-preserving wrapper can call the SDK collector for one batch, discard/filter those results, then call it for another batch. Python can reuse the freed string IDs; the new patches then inherit the old sources at the lookup in `evolve_skill`.

I reproduced this through public `evolve_skill` with a wrapper that discarded host patches and returned a subsequent SDK builtin collection. With only analyst/consolidator model calls stubbed, 4/10, 30/30, and 40/100 builtin records were incorrectly labeled `host` in three runs. No forced garbage collection or mutation of the private map was needed.

Keep a strong reference to the original string in each provenance entry for the map's lifetime, and verify identity during lookup (for example `id -> (patch_object, sources)`). Add a wrapper regression that discards an earlier collection before returning later patches.

scripts/skill_evolution.py
```diff
@@ -1386,21 +1433,34 @@ def evolve_skill(
...
   if not patches:
     logger.warning("No patches to consolidate; returning the current skill.")
     return current_skill
+  # Plain strings from a legacy replacement have no recorded provenance, so use
+  # the conservative builtin fallback. Identity-retaining wrappers around the
+  # SDK collector resolve through the context map even when they reorder or
+  # filter the result.
+  patch_sources = []
+  for patch in patches:
+    sources = patch_provenance.get(id(patch), [])
```
caohy1988: [P2] Give each accepted patch occurrence its own provenance identity

Two analyst calls can return the same cached string object. If one is host and the other builtin, both source labels are appended to the same identity queue in collection order. A wrapper that keeps only `patches[1:]` still resolves the first source in that queue, so keeping the builtin occurrence writes `source: "host"`. Reversing the two occurrences similarly leaves the source order unchanged.

I reproduced the filter case through public `evolve_skill`; the new tests use distinct host/builtin objects, so they do not exercise this aliasing case. Give each accepted occurrence a distinct string identity (while preserving the public plain-string result contract) before recording provenance, and add a shared-object filter/reorder regression. This occurrence-identity fix is separate from keeping discarded objects alive: both invariants matter for this map.

caohy1988 (changes requested):
Fresh full review of `1c8cd6d` against `f5e03bd`, covering the entire three-file PR and the host-contract requirements in #397.

The previous signature regression and the simple reorder/filter/deduplication cases are fixed; the two earlier threads are resolved. The public collector signature matches the base. Fresh review and public-entry-point probes found two further P2 provenance defects: recycled object IDs can reuse discarded patches' sources, and shared string objects cannot distinguish source occurrences after a wrapper filters or reorders them. An independent validator confirmed both defects; details and targeted regression suggestions are inline.

Validation: 145 focused tests passed; the full local suite passed with 5,049 tests and 57 skips. Pyink/isort and diff checks passed. [CI passed](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/33943603152) on this commit across Python 3.10–3.14, formatting, browser smoke, and package build. Additional checks covered concurrent runs and context restoration. Live model calls were not run; the defects are reproduced with model calls stubbed.

Please fix the two source-integrity cases before merging. The documented fallback for opaque replacement collectors is not included as a change request.

scripts/skill_evolution.py
Beverly621: **Fixed in `d219ed4`.**

- **Added a strong-reference invariant:** each provenance entry now stores `id -> (patch_object, source)` for the lifetime of the `evolve_skill` collection.
- **Added identity verification:** lookup accepts a provenance entry only when the retained object `is` the returned patch.
- This prevents Python from recycling an ID from a discarded earlier batch and assigning its source to a later patch.
- Added a public-entry-point regression where a wrapper discards a host batch and returns a subsequent builtin batch; the resulting artifact remains correctly labeled `builtin`.

scripts/skill_evolution.py
Beverly621: **Fixed in `d219ed4`.**

- **Each accepted patch occurrence now receives a distinct identity** before provenance is recorded, even when host and builtin analysts return the same cached string object.
- The copied values remain exact plain `str` objects, preserving the public `collect_patches() -> list[str]` contract.
- Added shared-object regressions for both **filtering** and **reordering** wrappers.
- The filter case correctly retains the second `builtin` occurrence; the reorder case produces `builtin`, then `host`, matching the returned occurrence order.

The focused suites now pass **148 tests**.

Beverly621:
**Both new provenance-integrity findings are addressed in `d219ed4`.**

The focused suites pass all **148 tests**. The complete local run produced **4,931 passed and 74 skipped**, with the four environment-restricted tests also passing when rerun outside the sandbox. Formatting and diff checks pass as well.

The new CI run is currently waiting for maintainer approval.

scripts/skill_evolution.py
```diff
@@ -1386,21 +1441,39 @@ def evolve_skill(
...
+  # the conservative builtin fallback. Identity-retaining wrappers around the
+  # SDK collector resolve through the context map even when they reorder or
+  # filter the result.
+  patch_sources = []
+  for patch in patches:
+    provenance_entry = patch_provenance.get(id(patch))
+    source = (
+        provenance_entry[1]
+        if provenance_entry is not None and provenance_entry[0] is patch
+        else "builtin"
```
caohy1988: [P2] Preserve known provenance when wrappers copy patch text

An SDK collector wrapper can preserve every patch value while returning fresh plain strings, for example `return [p.encode("utf-8").decode("utf-8") for p in original_collect_patches(...)]`. Those strings miss the identity lookup and reach this `builtin` fallback.

I reproduced this through public `evolve_skill` with one host failure analyst and one builtin success analyst. Both analysts ran and both patch texts were preserved, but the emitted sources were `["builtin", "builtin"]` instead of `["host", "builtin"]`. The SDK has already recorded both sources before the wrapper copies the strings. Serialization or string reconstruction in a wrapper can therefore silently corrupt the new audit field.

Preserve recorded provenance for equal-text copies when its source is unambiguous, or give wrappers a supported way to carry it. Avoid guessing when equal recorded occurrences have conflicting sources. Add a regression that returns fresh equal strings from the wrapper and checks the resulting artifact labels.

caohy1988 (changes requested):
Full fresh review of d219ed48c5a2f279f76425e901a2898b7c24e021 against main f5e03bdc4b6c87f5eae7e8c608eeb480aa7398e5, covering all three changed files and downstream host adapters.

All four findings from the earlier reviews are fixed. I independently reproduced the discarded-batch and shared-string cases successfully and resolved the remaining two threads.

One P2 remains: a wrapper around the SDK collector that reconstructs equal patch strings loses already-recorded host provenance. The inline comment includes the reproduced result. Please preserve that provenance before merging.

Validation: 148 focused tests passed; complete local suite 5,052 passed, 57 skipped; pyink, isort, and git diff --check passed. CI 33944366992 is green, including Python 3.10-3.14, browser smoke, formatting, and package build. The eight review passes completed, and the remaining finding received independent validation.

Other integration limits: moving collection to another thread requires propagating context, and host analysts remain responsible for verifying user corrections. Additional signature/context-cleanup regression tests and splitting the large test module are optional follow-ups. No code changes were made in this review.

scripts/skill_evolution.py
Beverly621: Addressed in 09ff598. 

Identity-based provenance remains the first priority. When a wrapper returns freshly reconstructed strings, evolve_skill now recovers provenance by equal text only when all recorded occurrences agree on a single source. Equal text with conflicting host and builtin sources still uses the conservative builtin fallback.

I added regressions using encode("utf-8").decode("utf-8") copies for both the unambiguous and conflicting-source cases. The focused suites pass all 150 tests; pyink, isort, and git diff --check also pass. The full sandbox run produced 4,933 passed and 74 skipped, and all four permission-sensitive tests passed when rerun outside the sandbox.

caohy1988 (approved):
Full fresh review of 09ff598f0166fc414a2e8b58229e164833a3d500 against main f5e03bdc4b6c87f5eae7e8c608eeb480aa7398e5, covering all three changed files, the complete provenance flow, host adapters, and tests.

No actionable findings remain. All five findings from the earlier reviews are fixed. I reran the public-entry-point copied-string reproduction: the artifact now correctly records ["host", "builtin"]. The earlier shared-string and discarded-batch regressions also pass, and the public signatures remain unchanged. I resolved the last outstanding review thread.

Validation: 150 focused tests passed; full local suite 5,054 passed, 57 skipped; pyink, isort, and git diff --check passed. CI 33978636902 is green, including Python 3.10-3.14, browser smoke, formatting, and package build. Eight independent review passes completed. Additional probes covered concurrent and nested evolution, exception cleanup, Unicode patch text, mixed host failures, and quarantined late results.

Nonblocking integration limits: provenance is context-local. A host wrapper that creates a thread to invoke the collector must propagate context (for example, copy_context().run). The documented callback and same-context wrapper paths retain provenance; opaque replacement collectors and ambiguous copied text retain the deliberate builtin fallback, which is not proof of builtin authorship. Host callbacks remain responsible for verifying user corrections. These limits do not require expanding this PR into a new cross-thread or provenance-registration API.

Optional follow-ups are stronger default-source/context-cleanup assertions and separating the large test module. The requested __all__ stability declaration matches issue #397 and does not break the inspected qualified imports. Approving this head.

haiyuan-eng-google (approved):

Issue #339, opened 2026-07-07
feat: add a free public Grafana dashboard for BQAA


caohy1988:
## Summary

Add a public, free Grafana dashboard option for BigQuery Agent Analytics (BQAA), using the existing `dashboard_v2` experience as the reference information architecture.

`dashboard_v2` already gives users a React/Vercel dashboard for BigQuery-backed BQAA data (self-hosted per customer). A Grafana dashboard is a lower-friction alternative: no custom app hosting, native BigQuery querying, built-in time-range controls, and an easy way to share a redacted public demo link from Grafana Cloud Free.

This issue has been verified against the current SDK: the **actual `agent_events` schema**, the SDK's **per-event-type views** (`ViewManager`), and the **`bqaa seed-events`** synthetic-data CLI are documented below, so implementation can start without schema spelunking.

## Why this matters

Many teams already use Grafana for observability. For BQAA, a Grafana dashboard makes the first evaluation path much simpler:

1. Create or use a free Grafana Cloud account.
2. Connect the Google BigQuery data source.
3. Import a BQAA dashboard JSON file from this repo.
4. Point it at `PROJECT.DATASET.agent_events`.
5. Optionally share a public, redacted demo dashboard link.

This complements, not replaces, `dashboard_v2`.

## Current reference

- Product reference: [`dashboard_v2/`](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/tree/main/dashboard_v2) (React/TS/Vite, Vercel-deployed, per-customer; project/dataset/table selection, FinOps summary, session/audit exploration, trace/event visualization, time-range analytics).
- Grafana docs: [BigQuery data source plugin](https://grafana.com/grafana/plugins/grafana-bigquery-datasource/) · [Grafana Cloud Free](https://grafana.com/pricing/) · [Externally shared dashboards](https://grafana.com/docs/grafana/latest/dashboards/dashboard-public/)

## Verified repo facts (read this before implementing)

**1. The real `agent_events` schema** (from `src/bigquery_agent_analytics/seed_events.py::_EVENT_SCHEMA_FIELDS`):

| Column | Type | Notes |
| --- | --- | --- |
| `timestamp` | TIMESTAMP (REQUIRED) | use for `$__timeFilter` |
[...]

PR #480, opened 2026-09-08
feat(dashboards): add Streamlit agent analytics dashboard

whuang602:
**What**: Python-native alternative to the Grafana dashboard, backed by the same typed views and filter semantics, essentially extending existing Grafana features into Streamlit. Includes scan-cap guardrails and query caching. **Intentionally leaving existing Streamlit content intact.**

**Why**: Addresses the optional dashboard integrations mentioned in #339. This is intentionally paired with PR #481, which is working to consolidate 3P dashboard integration into a dedicated dashboards/ directory.

**Reason this is in draft**: Anti-drift check script & tests are in review.

image

image

image

image


caohy1988 (changes requested):
P0: 1 / P1: 3 / P2: 1 — **REQUEST_CHANGES**

Reviewed HEAD `08d42ac55e3dd8641d16502f0c1aaa24b01296e0`. The PR is still OPEN + DRAFT. The acknowledged anti-drift work remains a readiness requirement; this is not LGTM pending only an undraft.

1. **P0 — Report billed bytes accurately.** `dashboards/streamlit/queries.py:790` returns `job.total_bytes_processed`, which `dashboards/streamlit/app.py:575-580` totals and labels **billed**. These are different BigQuery statistics. A synthetic successful job with 1,024 bytes processed and 10,485,760 bytes billed renders **“1.0 KB billed”**. This silently understates the dashboard's FinOps accounting. Carry `total_bytes_billed` separately for on-demand billing, or label this measure “processed” and remove the billing claim. Preserve zero incremental usage for cache hits. The [BigQuery job contract](https://docs.cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobStatistics2) distinguishes the fields and qualifies billed bytes under capacity pricing.

2. **P1 — Preserve selected filters when options expire or fall outside the cap.** `dashboards/streamlit/app.py:90-91,192-194` removes selections absent from refreshed options; `dashboards/streamlit/models.py:242` then turns an empty selection into `___ALL___`. Streamlit AppTest reproduces selected agent `alpha` / session `s1` becoming **all agents / all sessions** when those options disappear, without an explicit filter-clear action. This can happen after narrowing the window or after newer sessions displace the selected session from the 1,000-option list. Preserve the selected values in the widget's option set so queries retain their scope and return no matching data when appropriate.

3. **P1 — Keep truncated IDs reachable, as Grafana does.** `dashboards/streamlit/app.py:197-201,213-217` uses multiselects without `accept_new_options` or another exact-ID input, while `dashboards/streamlit/queries.py:146-148` caps each option list at 1,000. Streamlit 1.49 defaults custom values to disabled. In a busy window, an older known session/user cannot be selected even by typing its exact ID; the shortest available window can still contain more than 1,000 IDs. `grafana/README.md:192-194` explicitly guarantees custom values for Agent, User ID and Session. Add the equivalent parameterized input path and retain those values across option refreshes.

4. **P1 — Land the promised Streamlit contract tests and CI before readiness.** `dashboards/streamlit/queries.py:82-113,639-669,726-838` introduces filter, cap and cache contracts, but this nine-file diff contains no tests or anti-drift checker. The existing `.github/workflows/grafana-sync-check.yml:20-31` watches Grafana paths only, and the current checker verifies Grafana SQL/JSON, not these Python builders. Land the acknowledged draft follow-up with explicit panel/filter mappings, whole-session rollups, missing-data behavior, bound parameters, real-job byte caps, cache-scope cases and filter-retention regressions; ensure Streamlit changes trigger it.

5. **P2 — Document typed-view setup before the run command.** `dashboards/streamlit/README.md:17-42` goes directly from environment configuration to launch. An existing raw events table alone is insufficient for the overview latency, LLM and tool panels. Include the `views create-all` prerequisite with matching project/dataset/table and view-prefix guidance, plus the query identity's jobUser/dataViewer requirements. The missing-view error handler already provides a recovery command, but setup should establish this dependency first.

Validation: all 15 builders parsed as single SELECT/UNION queries; identifier-injection rejection, bound filter parameters, cache misses for changed filter/project/dataset/window/cap, repeated-query cache hits, and over-cap refusal passed in offline probes. Each real-job configuration retained `maximum_bytes_billed`. A synthetic four-tab AppTest passed on Streamlit 1.49.0 / Plotly 6.0.0. Existing Grafana sync + typed-view tests: **37 passed**; the standalone checker verified **37 Grafana queries**. Wheel metadata and a `.[dashboards]` dependency-resolution dry run passed. No customer BigQuery queries were executed; no source fixes were made.

Pairing with #481: links resolve at this HEAD. When Grafana moves, reconcile `dashboards/README.md:10` and the old `grafana/queries/README.md` references in `app.py:210` / `queries.py:86`, along with the new anti-drift CI paths. This is merge-order coordination, not an additional current-path defect.

caohy1988 (changes requested):
REVIEW_VERDICT=REQUEST_CHANGES
P0=0 / P1=2 / P2=1

Astra re-review of exact HEAD `11b2624560615d3b6dedc0c2097e9155e5b7ef8e`, including the complete delta from `08d42ac55e3dd8641d16502f0c1aaa24b01296e0`. Counts are deduplicated by root cause: the two prior filter findings now share one demonstrated widget-state defect. The PR remains OPEN + DRAFT.

| Prior finding | Disposition at this HEAD | Evidence |
| --- | --- | --- |
| P0: processed bytes labeled billed | **CLOSED** | Actual `fetch()` -> `scan_log` -> footer path reports **10.0 MB billed (1.0 KB processed)** for a synthetic job with those distinct statistics. Repeated local-cache reads add zero bytes; a BigQuery cache hit reports zero billed bytes. |
| P1: expired selections silently become ALL | **PARTIAL; blocker remains** | Explicit pruning is removed and selected values are added to options, but real Streamlit 1.49.0 widgets still reset their selections when the options change. See R1. |
| P1: capped-out exact IDs unreachable | **PARTIAL; blocker remains** | Agent/User/Session now enable custom input, but entering a new ID and clicking Apply resets it to ALL. See R1. |
| P1: missing Streamlit contract tests/CI | **OPEN** | No tests, checker, or workflow changes in the four-file follow-up. See R2. |
| P2: typed-view setup before launch | **PARTIAL** | Setup order, query IAM roles, and the create-all command are added; the custom-prefix flag is invalid. See R3. |

**R1 — P1: Preserve filter values across widget identity changes.** `dashboards/streamlit/app.py:221-252`, particularly the `_merge_widget_options()` calls. On the declared minimum `streamlit==1.49.0`, options participate in widget identity even with a key. Appending the active values keeps them available but does not restore the selection when Streamlit registers the replacement widget.

Actual AppTest reproduction: select agent `alpha`, user `u1`, event `LLM_RESPONSE`, and session `s1`; submit; refresh options so those values disappear. All four returned selections and the resulting BigQuery array parameters become `['___ALL___']`, with no explicit clear. The same reset occurs when an unselected option is merely added, or the list is reordered. Separately, enter `custom-agent`, `custom-user`, and `custom-session` and click Apply: merging these newly entered values changes the widget identities immediately, and all three submitted filters become ALL. These probes exercise the actual sidebar and query-parameter builder, not just the helper in isolation. Preserve submitted selections independently of widget recreation and test this through Apply and subsequent refreshes on the supported minimum version. The [1.49 custom-input contract](https://docs.streamlit.io/1.49.0/develop/api-reference/widgets/st.multiselect) permits new values without inserting them into the options list.

**R2 — P1: Land the promised Streamlit contract coverage and run it in CI.** There are still no Streamlit tests or anti-drift checks in this PR. `.github/workflows/grafana-sync-check.yml:20-31` only watches Grafana/checker paths; general CI installs `.[dev]`, which does not include the Streamlit extra. Add the acknowledged panel/filter mappings, whole-session and missing-data cases, bound parameters, real-job byte caps, billing/cache accounting, and widget retention/custom-ID regressions, with Streamlit dependencies and triggers. The reproduced R1 failure is exactly what the current passing Grafana tests do not exercise.

**R3 — P2: Use the supported custom-prefix CLI flag.** `dashboards/streamlit/README.md:58` names `--view-prefix`, but `views create-all` exposes `--prefix` (`src/bigquery_agent_analytics/cli.py:1848`). CLI parsing of the documented flag exits 2 with “No such option”; the same invocation with `--prefix custom_` reaches the mocked view manager and exits 0. Change the flag and keep `BQ_VIEW_PREFIX` aligned with it.

**Delta assessment:** no independent additional P0/P1 found beyond these unresolved contracts. The newly enabled custom-input path's immediate silent-ALL behavior is included in R1, not counted again. The billing fix is credited as closed.

**Pairing with #481:** it is still OPEN at `47620b8f05926568ba4f94968a248f6935f3fd79`; Grafana has not moved in #480. Current paths resolve. Combining the branches has a verified `.gitignore` three-way conflict. Whichever lands second must preserve both integrations' ignore rules and reconcile the legacy-path protection identified in #481. After the move, update `dashboards/README.md:10` to `grafana/README.md`, plus `app.py:243`, `queries.py:86`, and `models.py:17` to the new repository paths; target the new locations in anti-drift coverage. This remains merge-order coordination, not an extra current-path finding.

Validation: **106 existing Grafana-sync/local-runner/typed-view tests passed**; standalone checker verified **37 Grafana queries**. All **15** Streamlit builders parsed as single read-only queries; synthetic cap/cache/billing probes passed; four-tab AppTest smoke passed with Streamlit **1.49.0** / Plotly **6.0.0**. Four refresh scenarios and custom-ID submission reproduce R1. No live BigQuery jobs or source implementation changes. No undraft or merge; Haiyuan remains the gate.

caohy1988:
@whuang602 Astra re-review comments for HEAD `11b2624` (also filed as [REQUEST_CHANGES](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/480#pullrequestreview-5149456298)):

REVIEW_VERDICT=REQUEST_CHANGES
P0=0 / P1=2 / P2=1

Astra re-review of exact HEAD `11b2624560615d3b6dedc0c2097e9155e5b7ef8e`, including the complete delta from `08d42ac55e3dd8641d16502f0c1aaa24b01296e0`. Counts are deduplicated by root cause: the two prior filter findings now share one demonstrated widget-state defect. The PR remains OPEN + DRAFT.

| Prior finding | Disposition at this HEAD | Evidence |
| --- | --- | --- |
| P0: processed bytes labeled billed | **CLOSED** | Actual `fetch()` -> `scan_log` -> footer path reports **10.0 MB billed (1.0 KB processed)** for a synthetic job with those distinct statistics. Repeated local-cache reads add zero bytes; a BigQuery cache hit reports zero billed bytes. |
| P1: expired selections silently become ALL | **PARTIAL; blocker remains** | Explicit pruning is removed and selected values are added to options, but real Streamlit 1.49.0 widgets still reset their selections when the options change. See R1. |
| P1: capped-out exact IDs unreachable | **PARTIAL; blocker remains** | Agent/User/Session now enable custom input, but entering a new ID and clicking Apply resets it to ALL. See R1. |
| P1: missing Streamlit contract tests/CI | **OPEN** | No tests, checker, or workflow changes in the four-file follow-up. See R2. |
| P2: typed-view setup before launch | **PARTIAL** | Setup order, query IAM roles, and the create-all command are added; the custom-prefix flag is invalid. See R3. |

**R1 — P1: Preserve filter values across widget identity changes.** `dashboards/streamlit/app.py:221-252`, particularly the `_merge_widget_options()` calls. On the declared minimum `streamlit==1.49.0`, options participate in widget identity even with a key. Appending the active values keeps them available but does not restore the selection when Streamlit registers the replacement widget.

Actual AppTest reproduction: select agent `alpha`, user `u1`, event `LLM_RESPONSE`, and session `s1`; submit; refresh options so those values disappear. All four returned selections and the resulting BigQuery array parameters become `['___ALL___']`, with no explicit clear. The same reset occurs when an unselected option is merely added, or the list is reordered. Separately, enter `custom-agent`, `custom-user`, and `custom-session` and click Apply: merging these newly entered values changes the widget identities immediately, and all three submitted filters become ALL. These probes exercise the actual sidebar and query-parameter builder, not just the helper in isolation. Preserve submitted selections independently of widget recreation and test this through Apply and subsequent refreshes on the supported minimum version. The [1.49 custom-input contract](https://docs.streamlit.io/1.49.0/develop/api-reference/widgets/st.multiselect) permits new values without inserting them into the options list.

**R2 — P1: Land the promised Streamlit contract coverage and run it in CI.** There are still no Streamlit tests or anti-drift checks in this PR. `.github/workflows/grafana-sync-check.yml:20-31` only watches Grafana/checker paths; general CI installs `.[dev]`, which does not include the Streamlit extra. Add the acknowledged panel/filter mappings, whole-session and missing-data cases, bound parameters, real-job byte caps, billing/cache accounting, and widget retention/custom-ID regressions, with Streamlit dependencies and triggers. The reproduced R1 failure is exactly what the current passing Grafana tests do not exercise.

**R3 — P2: Use the supported custom-prefix CLI flag.** `dashboards/streamlit/README.md:58` names `--view-prefix`, but `views create-all` exposes `--prefix` (`src/bigquery_agent_analytics/cli.py:1848`). CLI parsing of the documented flag exits 2 with “No such option”; the same invocation with `--prefix custom_` reaches the mocked view manager and exits 0. Change the flag and keep `BQ_VIEW_PREFIX` aligned with it.

**Delta assessment:** no independent additional P0/P1 found beyond these unresolved contracts. The newly enabled custom-input path's immediate silent-ALL behavior is included in R1, not counted again. The billing fix is credited as closed.

**Pairing with #481:** it is still OPEN at `47620b8f05926568ba4f94968a248f6935f3fd79`; Grafana has not moved in #480. Current paths resolve. Combining the branches has a verified `.gitignore` three-way conflict. Whichever lands second must preserve both integrations' ignore rules and reconcile the legacy-path protection identified in #481. After the move, update `dashboards/README.md:10` to `grafana/README.md`, plus `app.py:243`, `queries.py:86`, and `models.py:17` to the new repository paths; target the new locations in anti-drift coverage. This remains merge-order coordination, not an extra current-path finding.

Validation: **106 existing Grafana-sync/local-runner/typed-view tests passed**; standalone checker verified **37 Grafana queries**. All **15** Streamlit builders parsed as single read-only queries; synthetic cap/cache/billing probes passed; four-tab AppTest smoke passed with Streamlit **1.49.0** / Plotly **6.0.0**. Four refresh scenarios and custom-ID submission reproduce R1. No live BigQuery jobs or source implementation changes. No undraft or merge; Haiyuan remains the gate.

dashboards/streamlit/app.py
```diff
@@ -0,0 +1,732 @@
...
+
+  Returns:
+    A tuple of (Filters instance, price_in float, price_out float).
+  """
+  st.sidebar.subheader("Filters")
+
+  with st.sidebar.form("filters"):
+    agents = st.multiselect(
+        "Agent",
+        options=_merge_widget_options("flt_agent", options.get("agent", [])),
```
caohy1988: **P1 — R1: Preserve filter values across widget identity changes**

`_merge_widget_options()` keeps selected values in the options list, but on Streamlit **1.49.0** options still participate in widget identity. When options change (refresh/reorder/new custom IDs), the multiselect recreates and submitted selections become `___ALL___`.

**Reproduced (AppTest):** select agent `alpha` / user `u1` / event `LLM_RESPONSE` / session `s1` → refresh so those values leave the option set → all four filters and BigQuery array params become `['___ALL___']`. Same reset when entering custom IDs and clicking Apply.

Preserve submitted selections independently of widget recreation; test through Apply + refreshes on the supported minimum. [1.49 custom-input](https://docs.streamlit.io/1.49.0/develop/api-reference/widgets/st.multiselect) allows new values without inserting them into options.

Shared root cause for prior “expired → ALL” and “custom ID → ALL” findings.

dashboards/streamlit/README.md
```diff
@@ -0,0 +1,68 @@
...
+Generate the typed analytical views matching your project, dataset, and events table:
+
+```bash
+bq-agent-sdk views create-all \
+  --project-id YOUR_PROJECT \
+  --dataset-id YOUR_DATASET \
+  --table-id YOUR_TABLE
+```
+
+> **Note:** Views default to the `adk_` prefix. If you specify a custom prefix with `--view-prefix`, set `BQ_VIEW_PREFIX` in `.env` (or in the dashboard sidebar) to match.
```
caohy1988: **P2 — R3: wrong CLI flag for custom prefix**

Docs name `--view-prefix`, but `views create-all` exposes `--prefix` (`cli.py`). `--view-prefix` exits 2 (“No such option”); `--prefix custom_` works. Please fix the documented flag and keep `BQ_VIEW_PREFIX` aligned.

pyproject.toml
```diff
@@ -59,8 +59,18 @@ improvement = [
   "pandas>=2.0.0",
   "python-dotenv>=1.0.0",
 ]
+streamlit = [
```
caohy1988, line 62: **P1 — R2: Streamlit contract tests + CI still missing**

No Streamlit tests/checker/workflow landed in this follow-up. Existing Grafana sync CI only watches `grafana/**` + the Grafana checker (not in this PR's path set), and general CI installs `.[dev]` without the Streamlit extra.

Please add the acknowledged contracts (panel/filter mappings, whole-session + missing-data, bound params, byte caps, billing/cache, **widget retention / custom-ID regressions** for R1) and wire CI so Streamlit path changes run them with the Streamlit deps.

caohy1988:
## Rebase note after #481 merge (PE)

`#481` (`dashboards/grafana/` move) landed on `main` first, so this draft is now **CONFLICTING**.

### Conflict (only file)
`.gitignore` three-way:

| Side | Lines |
| --- | --- |
| This PR (`11b2624`) | `dashboards/**/*.json` |
| `main` post-#481 | `dashboards/grafana/*.json` (+ `dashboards/grafana/datasource.yaml`, `dashboards/grafana/.local/`) |

**Recommended resolution:** take `main`’s scoped `dashboards/grafana/*` ignores (do **not** keep `dashboards/**/*.json` — that would also ignore any committed Streamlit JSON under `dashboards/streamlit/` if added later). Keep the legacy `grafana/*.json` lines for older checkouts.

Also ensure these post-#481 lines from `main` remain:

```gitignore
dashboards/grafana/datasource.yaml
dashboards/grafana/.local/
dashboards/grafana/*.json
```

### Path leftovers still on this branch (from prior Astra review)
After resolving `.gitignore`, please also refresh Streamlit doc pointers that still say bare `grafana/...`:

- `dashboards/README.md` → Grafana guide still links `../grafana/README.md` (should be `grafana/README.md` under `dashboards/` or `./grafana/README.md`)
- `dashboards/streamlit/models.py`, `queries.py`, `app.py` comments still cite `grafana/queries...` — prefer `dashboards/grafana/queries...` for clarity

CI path filters in `.github/workflows/grafana-sync-check.yml` already look updated to `dashboards/grafana/**` on the merge tree.

### Suggested author steps
```bash
git fetch upstream main
git rebase upstream/main   # or merge
# resolve .gitignore as above
# fix the three Streamlit/README path strings
git push --force-with-lease   # if rebasing
```

Happy to re-review once the conflict is cleared. `#481` is already on `main` @ `a4b9464`.

whuang602:
Summary of status:

- Rebased & resolved merge conflict
- Anti-drift check & CI workflow implemented (R2)
- Tests for streamlit app & anti-drift check script (R2)
- Addressed R1 & R3
- Addressed minor errors and code polish

caohy1988 (changes requested):
REVIEW_VERDICT: CHANGES
P0=0 / P1=2 / P2=2

Re-reviewed exact HEAD `9c71f08d0923e06ddb834e81f2e05494a76da545`, including the follow-up from `11b2624` and the full Streamlit addition against rebased main. **REQUEST_CHANGES.**

| Prior finding | Disposition | Evidence at this HEAD |
| --- | --- | --- |
| R1 P1: filter values lost across widget identity changes | **PARTIAL** | Applied filters now survive option retirement, addition, reordering, and an empty refresh, including subsequent Apply. First custom-ID Apply still submits ALL; changing/clearing an existing selection on the next Apply is also ignored. Details below. |
| R2 P1: missing Streamlit contract tests/CI | **PARTIAL** | A dedicated workflow, 14-query SQL sync checker, mutation checks and dashboard unit tests are present. Runtime widget/cache contracts are mocked away, including the reproduced R1 failures. |
| R3 P2: README `--view-prefix` | **CLOSED** | `dashboards/streamlit/README.md:58` now uses `--prefix`, matching `views_create_all` in `src/bigquery_agent_analytics/cli.py:1848`. |
| Prior billing P0: processed bytes labeled billed | **CLOSED** | Successful-job path reports **10.0 MB billed (1.0 KB processed)** from distinct statistics; repeated local-cache reads add zero. The legacy tuple fallbacks are removed. R5 below is a separate exception-path defect. |

**R1 — P1 (confidence 10/10): Preserve the newly submitted values before widget recreation.** `dashboards/streamlit/app.py:233–267,296–303`. The new `default=_default_for(applied.agents)` and equivalent arguments use the *previous* applied filters, while `_seed_options()` adds pending custom values before widget registration. On supported Streamlit **1.49.0**, both options and defaults participate in identity.

Real AppTest, production sidebar and BigQuery parameter builder:

- From ALL, enter `custom-agent`, `custom-user`, `custom-session` and click Apply once: all three submitted arrays are `['___ALL___']`; the custom values are in the option lists but unselected.
- Apply `alpha/u1/LLM_RESPONSE/s1`, then change to existing `beta/u2/TOOL_END/s2` and click Apply again: all four arrays retain the first values. Clearing them on that next Apply likewise retains the first values. The changed defaults recreate the widgets before the pending edit is committed.

The applied-state anchor fixes refresh retention, but it must also preserve the current form submission before re-registration can overwrite it. Regress custom first-submit, successive edits and clearing through actual widget-to-query behavior. The [1.49 custom-input contract](https://docs.streamlit.io/1.49.0/develop/api-reference/widgets/st.multiselect) returns new values without adding them to the option menu.

**R2 — P1 (confidence 10/10): Exercise real Streamlit behavior in the new CI tests.** `tests/test_dashboards_streamlit_app.py:37–44,153–160,1153–1195`. Streamlit is unconditionally replaced with a mock, cache decorators become passthrough functions, and the sidebar test replaces `multiselect` with `return state.get(key, [])`. Installing Streamlit in the workflow therefore does not exercise widget identity, reruns, or actual cache keys. As an independent mutation check, removing all four `default=` repair arguments **in memory** still passes all five filter-retention/custom-ID tests. Add real AppTest regressions on the supported minimum, asserting bound arrays after Apply/refresh; retain real cache-key variation coverage. Credit the new SQL mapping/checker and CI wiring as delivered.

**R4 — new P2 (confidence 10/10): Keep the inspected trace selected when recent sessions refresh.** `dashboards/streamlit/app.py:605`: `chosen = st.selectbox("Session", options=ids, index=0)`. Real AppTest of `row_sessions`: select `s2` from `['s1','s2']`; refresh to `['s3','s1','s2']`; the picker and actual bound `session_ids` silently switch to `['s3']` although `s2` remains available. A newly arriving session can switch the trace being investigated. Preserve the chosen session across option changes while it remains eligible, and make any necessary reset explicit. This is the separate trace selector, not the shared sidebar filters.

**R5 — new P2 (confidence 10/10): Retain billing statistics after result-download failure.** `dashboards/streamlit/queries.py:824–830` returns `bytes_processed=0, bytes_billed=0` for every exception, including failure inside `job.to_dataframe()` after successful query execution. An offline probe using an actual SDK `QueryJob` with synthetic DONE/no-error metadata (1,024 processed; 10,485,760 billed), and an injected row-download `ServiceUnavailable`, produces **“1 queries this run · 0 B billed (0 B processed)”** through `fetch → scan_log → footer`. Preserve known completed-job statistics on materialization failure; show unknown when statistics are unavailable. The real-job cap remains set. R4/R5 are newly identified defects in the original addition, not claims that the latest commits introduced them.

**CI evidence:** the six green check results are five GitHub Actions security-scan jobs plus CLA. At this exact SHA, [general CI](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/34437052265) and [Streamlit query sync check](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/34437052259) both report `action_required`; neither establishes passing application CI.

**Validation:** 142 targeted dashboard/Streamlit-sync/Grafana-sync/local-runner tests passed locally; standalone checkers verified 14 Streamlit and 37 Grafana queries. All 15 Streamlit builders parsed as single read-only queries. Bound-parameter, real-job cap, cache-scope and successful billing probes passed; four-tab AppTest smoke passed (8 metrics). Pyink 25.12.0, isort and diff whitespace checks passed for the relevant changed files. Runtime probes used Streamlit 1.49.0 / Plotly 6.0.0 / BigQuery SDK 3.45.0. Synthetic data only; no live BigQuery jobs, source fixes, merge or undraft. #481 is now in the base; the migrated Grafana links/checker paths resolve.

caohy1988 (changes requested):
REVIEW_VERDICT: CHANGES
P0=0 / P1=2 / P2=2

Independent review of exact HEAD `9c71f08d0923e06ddb834e81f2e05494a76da545`. I did not take prior review claims at face value: I reproduced every runtime finding below from scratch with **real Streamlit AppTest** (no mocks of `streamlit`) in throwaway venvs at both the pinned minimum **1.49.0** and latest **1.63.0**, plus unit-level probes with synthetic BigQuery job metadata. Where my repro diverges from the prior review, I say so.

**P1-1 — Filter Apply is broken on the declared minimum Streamlit 1.49.0** (`dashboards/streamlit/app.py:234-303`, `pyproject.toml:63`).
AppTest driving the real `sidebar_filters` on streamlit 1.49.0:
- From ALL, type custom IDs (`custom-agent` / `custom-user` / `custom-session` via `accept_new_options`) and click Apply once → all three bind `('___ALL___',)`; the custom values appear in the option lists but are not submitted.
- Apply `alpha/u1/LLM_RESPONSE/s1` (binds correctly), then change to `beta/u2/TOOL_END/s2` and Apply again → all four arrays still bind the first values.
- Mechanism: `_seed_options` merges the in-progress value into `options`, and `default=_default_for(applied)` changes with each apply; on 1.49.0 both participate in widget identity, so the submit rerun recreates the widget and resets it to the *previous* applied values before the pending edit is read.
- The same script on streamlit 1.63.0 passes every step, so this is version-dependent — but `streamlit>=1.49.0` makes 1.49.0 a supported configuration. Either restructure the seeding/default so options and default are stable across the submit rerun, or raise the version floor to a release where this works.
- One divergence from the prior review, for honesty: in my repro, *clearing* an existing selection and applying DID bind the `___ALL___` sentinel on 1.49.0. I could not reproduce the "clearing is ignored" sub-claim; the custom-ID and successive-edit failures reproduced exactly.
- What works on both versions: retention of applied filters across an empty/failed options refresh (widget and `applied_filters` both keep the selection).

**P1-2 — The new CI tests never exercise real Streamlit, and no application CI has run on this SHA** (`tests/test_dashboards_streamlit_app.py:37-44`, `.github/workflows/streamlit-sync-check.yml`).
- `streamlit` is in `_UNCONDITIONAL_MOCKS`: even though the workflow installs `.[dev,streamlit]`, the suite replaces Streamlit with a MagicMock, turns `cache_data`/`cache_resource` into passthroughs, and swaps `multiselect` for `return state.get(key, [])` — a fake that ignores `default=` entirely. The P1-1 defects are interactions between `default`, `options`, and widget identity, so they are structurally invisible to this suite: the tests assert the app's own merging logic, not widget behavior. Verified locally: the full 62-test suite passes on a machine *without streamlit installed at all*.
- Add real AppTest regressions (minimum supported version) that assert the bound query-parameter arrays after Apply/refresh, and at least one test that exercises real `st.cache_data` key variation.
- CI state verified via the Actions API at this SHA: the `CI` workflow and the new `Streamlit query sync check` workflow both report **`action_required`** — the only green checks are the zizmor security scans and CLA. So no test or sync-check signal exists for this HEAD yet.
- Credit where due: the 14-query SQL sync checker is real and passes locally (`All 14 Streamlit dashboard queries match canonical Grafana SQL`), and the SQL-sync CI wiring itself is sound.

**P2-1 — Session selectbox refresh silently switches the trace under investigation** (`dashboards/streamlit/app.py:605`).
Real AppTest of `row_sessions` (reproduced on **both** 1.49.0 and 1.63.0, so not version-dependent): select `s2` from `[s1, s2]` → trace query binds `('s2',)`; options refresh to `['s3', 's1', 's2']` (a new session arrives) → the picker and the bound `session_ids` silently switch to `('s3',)` even though `s2` is still eligible. Preserve the chosen session while it remains in the options; reset explicitly only when it leaves.

**P2-2 — Billing statistics are dropped on result-download failure** (`dashboards/streamlit/queries.py:771-786,824-830`).
Probe with a synthetic completed job (`total_bytes_processed=1024`, `total_bytes_billed=10,485,760`, `cache_hit=False`) whose `to_dataframe()` raises: `run_query` returns `bytes_processed=0, bytes_billed=0`, so `fetch → scan_log → footer` renders **"1 queries this run · 0 B billed (0 B processed)"** for a query that actually billed 10 MB — contradicting the footer's "what this rerun actually scanned" contract. On the materialization-failure path the job object and its statistics exist; report them (or say "unknown") instead of zero.

**Verified as solid (no action needed):**
- Injection posture: table/view identifiers are regex-validated before interpolation, filter values go out as `ArrayQueryParameter`s, `LIMIT`s are int-cast, timestamp literals are built from owned `datetime`s.
- Success-path billing now reports distinct billed vs processed stats correctly (10.0 MB billed / 1.0 KB processed for a synthetic job), and the run-id dedup keeps cached reruns from double-counting.
- Dry-run preflight + `maximum_bytes_billed` cap is set on the real job; guardrail message path works.
- 62 mock-based tests and the sync checker pass locally; README/`.env.sample`/`.gitignore` (SA JSON keys) are consistent.

**Verdict: REQUEST_CHANGES** — two P1s (broken Apply on the supported minimum Streamlit; mock-only widget/cache test coverage with no CI signal on this SHA) and two P2s (trace-selectbox hijack on refresh; billing stats lost on download failure), each reproduced independently against real Streamlit.

pyproject.toml
```diff
@@ -59,8 +59,18 @@ improvement = [
...
+  "streamlit>=1.49.0",
```
caohy1988: **P1: the declared Streamlit minimum silently broadens refreshed filters**

This lower bound is not compatible with the state-retention implementation at this head. With Python 3.13 and Streamlit 1.49.0, the branch's own `test_filter_widget_options_change_preserves_applied_filters` fails: after `agent-a` disappears from the refreshed BigQuery options, the code re-adds it to the option list, but the widget value is `[]`, not `['agent-a']`. Continuing the real AppTest and clicking Apply changes `applied_filters.agents` from `('agent-a',)` to `('___ALL___',)`, so every panel is silently broadened.

The same 102-test set passes on Streamlit 1.63.0, so current CI resolution masks a broken advertised floor. Please either make the widget/session-state pattern work on 1.49.0 or raise the minimum to the first version you verify, and add a minimum-dependency CI cell so this contract stays exercised.

dashboards/streamlit/queries.py
```diff
@@ -0,0 +1,956 @@
...
+    df, bytes_processed, bytes_billed, cache_hit, run_id = _run_query_cached(
+        sql, filters, project, max_bytes
+    )
+  except Exception as exc:
+    if isinstance(exc, QueryExecutionError):
+      bytes_processed = int(exc.bytes_processed)
+      bytes_billed = int(exc.bytes_billed)
+      cache_hit = bool(exc.cache_hit)
+    else:
+      bytes_processed, bytes_billed, cache_hit = 0, 0, False
```
caohy1988, line 971: **P2: unavailable billing statistics are still presented as zero**

This fallback makes an unknown outcome indistinguishable from a proven zero. A query can be accepted and billed before `to_dataframe()` or the metadata reload fails; this path then records `(0, 0, False)`, `fetch()` appends it to the scan log, and the footer describes the total as what the rerun actually billed/processed.

There is a second instance in `_extract_job_stats`: if `total_bytes_processed` is populated but `total_bytes_billed` is `None`, it does not reload and returns billed bytes as zero. An exact-head probe produced `(2048, 0, False)` with no reload call.

Please preserve unknown explicitly (optional fields or a `stats_known` flag), attempt hydration whenever a required statistic is missing, and have the footer identify failed queries with unavailable cost data rather than summing zero.

.github/workflows/streamlit-sync-check.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
+        with:
+          python-version: "3.12"
+
+      - name: Install package with dev and streamlit dependencies
+        run: pip install -e ".[dev,streamlit]"
+
+      - name: Validate Streamlit queries match canonical Grafana SQL
+        run: python3 scripts/check_streamlit_queries_sync.py
```
caohy1988, line 63: **P3 (non-blocking): dashboard Python bypasses the formatting gate**

This workflow runs the SQL checker and tests, but no formatter. The repository's `autoformat.sh` and the main CI dirty-diff assertion cover only `src/`, `tests/`, `examples/`, and `scripts/`, so the new production Python under `dashboards/streamlit/` is outside both paths. The gap is observable on this head: `pyink --check dashboards/streamlit/charts.py` wants to add the missing final newline.

Please add `dashboards/streamlit` to the shared format command and CI diff list, or run `pyink --check` and `isort --check-only` for it here.

caohy1988 (changes requested):
Fresh review of exact head `295542a64e59eafe7a76b3a251388ff14057f6f3`.

Requesting changes for one P1 compatibility regression and one P2 cost-reporting issue. I also left one non-blocking P3 CI-quality comment.

Prior-review disposition:
- The active custom-value, successive-apply, and clear-selection cases now pass on current Streamlit.
- R3 (`--prefix`) and R4 (session selector retention) are fixed.
- The known-statistics half of R5 is fixed: a completed job whose stats are available preserves them after result-download failure.
- R1 is not fully closed at the declared Streamlit floor, and R5 still conflates unavailable statistics with zero.

Verification:
- Python 3.13 / Streamlit 1.63.0: all 102 focused tests pass.
- Python 3.13 / Streamlit 1.49.0 (the declared minimum): 101 pass, 1 fails. The failing committed AppTest is `test_filter_widget_options_change_preserves_applied_filters`.
- Manual AppTest continuation on 1.49.0: after option refresh, the widget value becomes empty while `applied_filters` still contains `agent-a`; the next Apply changes it to `___ALL___`.
- All 14 Streamlit queries pass the canonical Grafana SQL anti-drift checker.
- `compileall`, `git diff --check`, and `isort --check-only` pass. `pyink --check` reports `dashboards/streamlit/charts.py` would be reformatted (missing final newline).
- The PR's CI and Streamlit workflows are currently `action_required` for this fork, so there is no application CI result on this head.

I did not run a live BigQuery integration query in this pass.

caohy1988 (approved):
REVIEW_VERDICT: APPROVE
P0=0 / P1=0 / P2=0

Independent re-review of exact HEAD `b8af9da338d96ef392d6434a20a07bb712b35ea9`. Not a rubber-stamp: every disposition below was verified from scratch in a throwaway venv at the declared floor (**Streamlit 1.63.0**, Python 3.13), using real AppTest probes of the production `sidebar_filters` / `row_sessions` / `footer` code paths plus unit probes with synthetic BigQuery job metadata.

**Prior-finding dispositions (all CLOSED):**

1. **P1 widget retention / Apply-on-floor (R1) → CLOSED.** Floor bumped to `streamlit>=1.63.0` (pyproject.toml:63) and the Apply flow redesigned (`on_click=_commit_filters` reading widget state, no more `default=` re-registration). My AppTest probes on 1.63.0 all pass: first custom-ID Apply binds the custom values (not `___ALL___`); successive Apply A→B binds B; clearing binds the ALL sentinel; applied filters and widget values survive an empty/failed options refresh. Astra's @295542a failing case was on 1.49.0, which is now below the declared floor; CI adds a floor-pinned step (`pip install streamlit==1.63.0` + AppTest suite) so the floor is enforced, not just documented.
2. **Mock-only Streamlit tests (R2) → CLOSED.** New `tests/test_dashboards_streamlit_apptest.py` (12 tests) runs `AppTest.from_file` on the real app with only the BigQuery layer stubbed — including `test_run_query_cached_keys_on_every_argument`, which exercises the *real* `st.cache_data` keying (cache clear + argument-permutation misses). The sync-check workflow now runs the AppTest and charts suites, and ci.yml installs `.[dev,streamlit]` so the general suite collects them too.
3. **P2 billing stats conflated with zero (R5) → CLOSED.** Verified both halves: (a) a completed job (1,024 processed / 10,485,760 billed) whose `to_dataframe()` fails now yields `bytes_billed=10485760, billed_known=True` through `run_query` — stats preserved via `QueryExecutionError`/`_extract_job_stats` with a metadata-only `job.reload()` fallback; (b) when stats are genuinely unavailable (None + reload fails), the result is flagged `*_known=False` and the footer prints "1 with unavailable billing data / 1 with unavailable scan data" instead of silently reporting zero.
4. **P2 session selectbox refresh (R4) → CLOSED.** Real AppTest: select `s2` from `[s1, s2]`, refresh options to `['s3','s1','s2']` — picker and bound trace `session_ids` stay `s2` (`_selected_session_id` + computed `default_idx`). A committed AppTest (`test_session_selectbox_preservation`) covers the same flow.
5. **P3 pyink formatting (charts.py) → CLOSED.** `bash autoformat.sh` produces a clean tree across `src/ tests/ examples/ scripts/ dashboards/`; ci.yml now diffs `dashboards/` as well.
6. **README R3 → remains CLOSED**; new README sections (lazy tabs, loopback binding, access-control note) match the code (`STREAMLIT_LAZY_TABS` default `true`, `.streamlit/config.toml` binds 127.0.0.1).

**New-HEAD claims verified (honesty check):**
- *Lazy tabs*: real AppTest of the full app — initial load fires only the 5 Overview queries; switching the segmented control fires only that tab's queries. Confirmed.
- *Cost guardrails*: dry-run preflight + `maximum_bytes_billed` intact; `create_bqstorage_client=False` avoids the BQ Storage API dependency; scan-cap tests pass.
- *Security posture improved*: `validate_refs` now strips whitespace and anchors with `\Z` (closes the trailing-newline bypass the `$` anchor allowed); filter values remain parameterized; identifier interpolation unchanged and safe.
- Local suite: **110 passed** (apptest + charts + app + sync tests) on Streamlit 1.63.0; anti-drift checker passes (14/14 queries match canonical Grafana SQL).

**Caveats (non-blocking):**
- All four app workflows (`CI`, `Streamlit query sync check`, `Grafana query sync check`, security scan) report `action_required` at this SHA — no application CI has run on this HEAD (fork approval gate). The workflow wiring itself is correct; a maintainer approving the runs should turn them green, but that signal does not exist yet. My local floor-version runs substitute partially.
- PR is still DRAFT — undrafting is the author's call.
- Nits (no action needed): the mock-based suite remains alongside the AppTest suite (fine — different layers); footer "scan avoided via cache" counts both BigQuery cache hits and local reruns (reasonably labelled).

**Ball with author:** no code changes requested. Remaining steps are undraft + maintainer approval of the gated workflow runs. Approving on the strength of the local floor-version verification above.

caohy1988 (approved):
REVIEW_VERDICT: APPROVE
P0=0 / P1=0 / P2=2

Astra re-review of exact HEAD `b8af9da338d96ef392d6434a20a07bb712b35ea9`, including the new delta from `295542a` and the affected dashboard contracts. Local and GitHub head SHAs match. Approving under the requested P0/P1 blocking bar, with two non-blocking P2 follow-ups below. This is not a passing application-CI result.

| Prior finding | Disposition | Evidence at this HEAD |
| --- | --- | --- |
| R1 P1: state retention fails on declared Streamlit minimum | **CLOSED** | `pyproject.toml:63` now declares `streamlit>=1.63.0`; the sync workflow explicitly tests 1.63.0. Real AppTest at that floor passes option refresh, retirement/reapply, first custom-ID submission, successive Apply and clear, checking all four filters and bound query arrays. The previously failing 1.49.0 configuration is outside the declared support range. |
| R2: mock-only widget/cache tests and missing Streamlit CI coverage | **CLOSED** for implementation/coverage | The real AppTest suite exercises production widgets and actual cache-key variation; both general CI and the dedicated workflow install the Streamlit extra. Execution of the fork workflows remains gated, as detailed below. |
| R5 P2: unavailable statistics conflated with zero | **CLOSED** | Independent processed/billed known flags survive success and failure paths into the footer, which identifies unavailable fields and explains their exclusion from totals. Six SDK-backed synthetic cases cover unknown, partially known and known metadata, including successful download and materialization failure. Original billing P0 remains **CLOSED**. |
| R4 P2: session selector retention | **PARTIAL** | The previously reported options-refresh case passes. Consecutive user selections expose a separate residual identity bug; see P2-1. |
| R3: README prefix flag | **CLOSED** | README uses `--prefix`, matching the CLI. |
| P3: charts.py final newline / formatting gate | **CLOSED** | Newline is restored, dashboards are included in formatting and CI diff checks, and local Pyink/isort checks pass. |

**P2-1 — Preserve consecutive trace selections.** [app.py:655–658](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/b8af9da338d96ef392d6434a20a07bb712b35ea9/dashboards/streamlit/app.py#L655-L658). The unkeyed selectbox derives `index` from the previous selection. Real full-app AppTest on Streamlit 1.63.0, with ALL sidebar filters and sessions `[s1,s2,s3]`: selecting `s2` displays and binds `s2`; immediately selecting `s3` still displays and binds `s2`; repeating `s3` finally works. The changed default index recreates the widget and loses the pending choice. Use stable widget identity/state and add a consecutive-selection regression. The committed test covers a single choice followed by options refresh and misses this sequence. This is a residual defect in the prior selector repair, not a claim that b8af9da introduced it.

**P2-2 — Bound retries for optional metadata retrieval after a successful query.** [queries.py:900–906](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/b8af9da338d96ef392d6434a20a07bb712b35ea9/dashboards/streamlit/queries.py#L900-L906). The new success path calls `job.reload(timeout=5)` after the DataFrame has downloaded. In BigQuery SDK 3.45.0 this timeout applies per HTTP attempt; the default retry budget remains 600 seconds. An offline probe using the actual QueryJob/RowIterator materialization and SDK reload/retry implementation, with missing billing metadata and simulated metadata GET 503s, delayed an already successful one-row result through **15 GET attempts and 543 virtual seconds**. This was a virtual-clock reproduction, not a live outage or wall-clock measurement. Bound or disable the optional lookup's retries and retain the existing unknown-statistics result when it fails. The older failure path has the same retry concern; b8af9da adds this exposure to otherwise successful queries.

**New-delta verification:** lazy navigation executes only the active panel's query calls (Overview 5, LLM 4, Tools 3, Sessions 2), with the shared filter-options lookup outside that count; eager mode executes all 14 panel calls. This supports reduced query execution, not a measured dollar-saving claim. All 15 SQL builders parse as single read-only queries; serialized real-job configurations retain parameter arrays, standard SQL and `maximumBytesBilled`. `create_bqstorage_client=False` disables automatic Storage API client creation. Identifier validation rejects internal newline/backtick payloads and returns normalized identifiers. Both documented launch modes resolve to loopback; shared-credential/access-control guidance matches the implementation. Grafana runner changes are formatting-only by AST comparison.

**Validation:** 190 targeted tests passed on Python 3.13.5 / Streamlit **1.63.0** / BigQuery SDK 3.45.0, including real AppTest. Standalone SQL checks passed for 14 Streamlit and 37 Grafana queries. Pyink 25.12.0, isort and diff whitespace checks passed. All reproduction processes completed. Synthetic data only; no live BigQuery integration or billed-scan measurement.

**CI and ownership:** [general CI](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/34924992566) and [Streamlit query sync](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/34924992563) still report `action_required` at this SHA, as does Grafana sync. The green GitHub Actions security-scan/CLA checks do not establish application CI success. The PR remains OPEN and DRAFT. Ball stays with the author for the two P2 follow-ups and readiness/undraft, and with a maintainer for approval of the gated workflows. No source fixes, merge or undraft performed.

caohy1988:
## Status after dual re-review (`b8af9da`)

Marked **ready for review** (was draft). Dual agent LGTM already on this HEAD:
- Astra APPROVE (0 P0 / 0 P1 / 2 nonblocking P2): https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/480#pullrequestreview-5205250879
- Kimi APPROVE: https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/480#pullrequestreview-5205241809

Optional residual P2s (nonblocking): consecutive trace selectbox stickiness; `job.reload()` retry budget on metadata 503s.

App CI workflows were `action_required` (fork gate) — approving those runs now if permissions allow.

caohy1988:
## CI note (after undraft + workflow approve)

App CI re-ran failed jobs on HEAD `b8af9da` — still red on **Python 3.10 / 3.11 / 3.12** with `MemoryError` in BKA/extractor compile smoke (`test_bka_source_compiles_end_to_end` and related). **Python 3.13 / 3.14**, Streamlit sync, Grafana sync, format, and configurator smoke are green.

This does not look Streamlit-dashboard-related. Merge held until the flake is waived/retried on larger runners or fixed in the extractor tests on `main`.

haiyuan-eng-google (approved):

caohy1988 (approved):
REVIEW_VERDICT: APPROVE
P0=0 / P1=0 / P2=0

Independent review of exact HEAD `c4c9b2558918381a44a589b600585137cbd625f3` (delta over my approved `b8af9da`): the merge-main `2a6b9e6` plus the extractor-compilation memory-cap fix `c4c9b25`.

**Root cause — corroborated by CI, not just claimed.** The parent merge commit `2a6b9e6` failed CI on all five Python versions with exactly the claimed signature, e.g. Test (Python 3.10): `AssertionError: compile failed: ast=() smoke=(('SubprocessFailure: MemoryError: ', ...))` (https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/35002547530/job/104494267672). The mechanism is plausible and matches the code history: the child set an absolute `RLIMIT_AS` before importing the SDK harness, and the new `.[dev,streamlit]` install pulls pandas/pyarrow/numpy via `google-cloud-bigquery`, whose import-time virtual reservations alone exceed a 512 MB absolute cap.

**The fix is real, not masking.** In `subprocess_runner.py` the trusted harness imports now happen before `_set_memory_limit`, the cap is sized as post-import VmSize (`/proc/self/statm` field 0 — the correct metric for `RLIMIT_AS`) + `memory_limit_mb` headroom, and the candidate extractor is still imported and executed strictly *after* the cap is set, so untrusted code stays bounded by the same megabyte budget as before. Fallback when the baseline is unreadable (non-Linux) is the old absolute cap — unchanged behavior there. Edge cases check out: `None`/0 still disables the cap; negative values are rejected in the parent before spawn; `setrlimit` rejection stays best-effort.

**Tests genuinely guard the regression.** `test_subprocess_small_memory_limit_clears_import_baseline` (64 MB headroom runs a clean extractor) fails under the old absolute-cap behavior on any Linux box, and `test_subprocess_memory_limit_still_enforced` (1 GiB allocation under 128 MB headroom → per-event `MemoryError`) proves the cap still bites. Both are Linux-gated and ran in CI. The two monkeypatched unit tests pin the headroom arithmetic and the absolute fallback. Local run (macOS): 137 passed, 2 skipped (the Linux pair, by design).

**CI at this exact SHA — all green, completed (not still-running):**
- CI: https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/35003728068 — Test 3.10 / 3.11 / 3.12 / 3.13 / 3.14 all success (3.12 job: 5178 passed, 80 skipped), plus Format check, Build package, Configurator browser smoke.
- Streamlit query sync check: https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/35003727848 — success.
- Grafana query sync check: https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/35003727996 — success.
- Security scans — success.

**Merge-main sanity.** `2a6b9e6` carries only the already-merged #483 skill_evolution_job content plus CHANGELOG; no conflicts bleeding into dashboard code. `c4c9b25` touches only `extractor_compilation/`, its tests, and its doc — no Streamlit scope creep; the dashboard files are byte-identical to the approved `b8af9da`.

**Nit (P3, non-blocking):** `docs/extractor_compilation_scaffolding.md` updates TestSmokeTest to 19 cases but leaves the summary line at "135 tests total"; the file now collects 139. A one-line doc touch-up whenever convenient.

Astra's two open P2s (session selectbox successive selection; metadata-reload retry deadline) are unchanged by this delta and remain out of scope here.

Ball with author: no — no code changes requested on this delta.

caohy1988 (approved):
REVIEW_VERDICT: APPROVE
P0=0 / P1=0 / P2=2 (both carried forward; no new findings)

Reviewed exact HEAD `c4c9b2558918381a44a589b600585137cbd625f3`, verified against local Git and the PR head. Scope: the four-file memory-cap fix and a sanity check of the main merge. The Streamlit tree is byte-for-byte unchanged from the previously approved `b8af9da`; I did not re-review it.

| Item | Disposition | Evidence |
| --- | --- | --- |
| CI smoke `MemoryError` / absolute memory cap | **CLOSED** | Cap is now post-import virtual address space plus configured headroom; candidate loading follows it. Exact-head application CI passes on Python 3.10–3.14. |
| Prior P2-1: immediate second Session selection dropped | **OPEN, non-blocking** | `dashboards/streamlit/app.py:655–658` is unchanged from the prior reproduction. |
| Prior P2-2: metadata reload inherits SDK retry deadline | **OPEN, non-blocking** | `dashboards/streamlit/queries.py:900–906` is unchanged; `timeout=5` still does not bound the default retry budget. |
| Merge `2a6b9e6` | **No regression found** | Its parents change disjoint files. Every imported main-side file matches main's blob, and every PR-side file matches b8af9da's blob; no manual conflict resolution. The affected skill-evolution tests pass locally. |

The previous P2 details remain in [the b8af9da review](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/480#pullrequestreview-5205250879).

**Root cause and trade-off.** The address-space mismatch is real, with one qualification to the stated import-order explanation: `python -m` imports the containing packages before invoking `main()`. Those package initializers already pull in the harness and BigQuery/native dependencies before the *old* cap; a fresh-process import check confirmed BigQuery, pandas, pyarrow, numpy, db_dtypes and smoke_test are loaded at that boundary. Moving the two explicit imports is therefore largely redundant. The substantive repair is replacing the absolute ceiling with baseline plus budget. The old CI logs show the reported smoke failures on 3.10/3.11/3.12, but do not identify the precise failing allocation.

`RLIMIT_AS` constrains virtual address-space growth rather than RSS; it is process-local, and resource limits survive exec. Lowering a limit does not reclaim existing mappings, so later allocations can fail even after imports succeeded. [Linux getrlimit documentation](https://man7.org/linux/man-pages/man2/getrlimit.2.html)

At [subprocess_runner.py:79–116](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/c4c9b2558918381a44a589b600585137cbd625f3/src/bigquery_agent_analytics/extractor_compilation/subprocess_runner.py#L79-L116), the cap is set before `load_callable_from_source` executes the candidate module and before event calls. The default remains 512 MiB, now additional virtual-address headroom. This deliberately permits a higher total ceiling even in lean environments; a huge trusted baseline likewise produces a large total ceiling. It is not a 512 MiB total-process/RSS guarantee. The changed docs describe the headroom semantics.

If the requested ceiling exceeds an inherited lower hard limit and `setrlimit` rejects it, the existing soft/hard limits remain: they are tighter than the requested ceiling, not an unlimited fallback. Reading the first statm field and multiplying by page size is correct; malformed/unavailable input falls back to the absolute budget. I exercised valid pages, empty/non-numeric data and read failures. Best-effort unsupported-platform behavior and pre-cap input unpickling predate this change. No new candidate-execution-before-cap path was found.

**Test quality and verification limits.** Local Python 3.13: **176 extractor/template tests passed, 2 Linux-only tests skipped**, plus **100 merge-related tests passed**. Formatting and diff checks passed. An in-memory mutation restoring the pre-fix `_set_memory_limit` made the committed headroom assertion fail on the wrong numeric ceiling, not merely a missing helper. The absolute-fallback test correctly still passes with old semantics. The Linux 64 MiB clean-run test covers the original failure mode; the 1 GiB/128 MiB negative test protects cap enforcement. Its `MemoryError` string assertion alone does not distinguish candidate allocation failure from a harness failure, so it should not be described as proving the failure phase. I did not run pre-fix Linux mutations locally; Linux execution evidence here is the exact-head Ubuntu CI below.

**Application CI at this exact SHA — all completed SUCCESS:**

| Job | Conclusion |
| --- | --- |
| [Test Python 3.10](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/35003728068/job/104498583977) | SUCCESS |
| [Test Python 3.11](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/35003728068/job/104498583981) | SUCCESS |
| [Test Python 3.12](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/35003728068/job/104498584089) | SUCCESS |
| [Test Python 3.13](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/35003728068/job/104498583961) | SUCCESS |
| [Test Python 3.14](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/35003728068/job/104498584191) | SUCCESS |

The [application CI run](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/35003728068), [Streamlit sync](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/35003727848) and [Grafana sync](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/35003727996) are green. These are application/test results, separate from the security scan. The Python 3.10 log reports 5,178 passed / 80 skipped.

GitHub currently shows OPEN and **ready**, rather than draft. No code fixes, merge or readiness changes were performed in this review. No blocking changes requested; the two P2 follow-ups remain with the author, and the merge decision rests with maintainers.

PR #483, opened 2026-09-09
skill_evolution_job: wire ANALYST_TIMEOUT_S; drop pre-#395 engine compat scaffolding

evekhm:
Follow-up to #472/#395. Two things:

**Wire `analyst_timeout_s` (the one #395 capability the job left on the table).** New `ANALYST_TIMEOUT_S` env binding (default 600s, `0` = unbounded), passed through to the engine so a host `error_analyst` hook that hangs on its own tool calls cannot stall the fleet past the Cloud Run task timeout. A malformed value fails the job at `get_config()` rather than silently running unbounded, matching the job's binding-env convention.

**Remove the pre-#395 compat scaffolding.** `supported_kwargs()`/`evolve_skill_compat` feature detection, the secondary `error_analyst` capability gate in `_resolve_error_analyst`, and `deploy.sh --scripts-dir` all existed so the job could run against an engine that predated #395. The image bakes `scripts/` from the current checkout, which always post-dates #395 now, so the detection can never trigger and the README's "single-pass fallback when the engine lacks the hooks" description was already false. `engine.py` is a pure locator again.

Tests: the compat tests are replaced by a keyword-contract test against the real engine (asserting `error_analyst_fn`/`incumbent_score`/`analyst_timeout_s` exist) plus forwarding tests for 600.0/30.0/None and config coverage for default/0/fractional/rejected values. `tests/ -k skill_evo`: 314 passed, 3 skipped (pre-existing). `bash -n deploy.sh` and `main.py --test` (`SELF-TEST PASS`) verified.


caohy1988 (changes requested):
**REQUEST_CHANGES — one P2; no P0/P1 findings.**

Full fresh review of exact HEAD `6848228bd9adce725fac51fa07b48de9db6e9392`, against base `716aa0eb0007d66e2fa8f08619d30e64e5f2bb7c`. I read the entire 10-file diff, the relevant existing engine implementation, job callers, deployment/Dockerfile, and tests. No prior #483 review was used.

**1. [P2] Qualify the timeout and Cloud Run guarantees — [README.md:250](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/6848228bd9adce725fac51fa07b48de9db6e9392/deploy/skill_evolution_job/README.md#L250).**

The new environment reference promises a “Per-analyst wall-clock bound,” and the hook table says each call is bounded. The existing engine actually applies a fresh `call.wait(analyst_timeout_s, slots)` in submission order. It neither measures each hook's deadline from invocation nor stops a running hook when the wait expires. Its own contract explicitly documents a worst-case total wait of `N * analyst_timeout_s`. See [the wait implementation](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/6848228bd9adce725fac51fa07b48de9db6e9392/scripts/skill_evolution.py#L983), [the engine contract](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/6848228bd9adce725fac51fa07b48de9db6e9392/scripts/skill_evolution.py#L1078), and [the collection loop](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/blob/6848228bd9adce725fac51fa07b48de9db6e9392/scripts/skill_evolution.py#L1149).

Two offline reproductions against the real exact-HEAD implementation:

- With two concurrent host hooks and a 0.1-second timeout, the first hook blocked and the second ran for approximately 0.155 seconds. The second hook's patch was still accepted, and the first hook was still running when collection returned.
- Through the actual job adapter, 25 blocked failure analysts with `ANALYST_TIMEOUT_S=0.03` and the default 10 workers took approximately 0.911 seconds before the all-host-failures error. Twenty hooks were still running at return; five had been cancelled before starting. I released all test hooks and verified completion afterward. At the production defaults, the engine's documented worst-case wait alone is `25 * 600 = 15,000` seconds, exceeding deploy.sh's 14,400-second task timeout before other job phases are counted.

Thus this setting improves progress after a hang but does not guarantee that the fleet finishes before the platform deadline. Cloud Run independently stops a task attempt that exceeds its configured duration. [Cloud Run task-timeout documentation](https://docs.cloud.google.com/run/docs/configuring/task-timeout).

Please align the new README hook table/environment row, `evolve()` docstring, CHANGELOG entry, and PR description with the actual contract: a timeout on each submission-order wait; running hooks are quarantined rather than cancelled; slot donation is bounded; total waiting can accumulate across analysts and must be budgeted separately from Cloud Run's task timeout. This finding concerns the newly introduced operator-facing promises; the engine's sequential-wait design predates this PR.

The remaining changes checked out:

- **Configuration:** default `600.0`, positive/fractional values, surrounding whitespace, and `0 -> None` reach the engine correctly. Ordinary malformed strings, negatives, NaN, and infinities fail in `get_config()`. Unset/blank values follow the existing `_env` defaulting convention. The CLI also failed with exit 1 for `abc`, `-1`, `nan`, and `inf`, naming `ANALYST_TIMEOUT_S`.
- **Engine integration and compatibility:** single-agent and co-evolution paths converge on the updated adapter. The current engine accepts the forwarded hooks, incumbent score, and timeout. Removed helper references are gone from executable callers; missing host hooks retain their documented fallbacks. Dropping pre-#395 engine support and `--scripts-dir` is intentional. An explicit `SDK_SCRIPTS_DIR` override must supply the current engine contract.
- **Deployment, security, and cleanup:** the staged engine/report files come from this checkout; source existence checks and quoted temporary-directory cleanup remain intact. The diff adds no auth/IAM expansion or new shell execution path. The removed deployment flag fails during argument parsing, before cloud operations.

Validation completed:

- `python -m pytest tests/test_skill_evo*.py -q --tb=short -rs -p no:cacheprovider`: **314 passed** on Python 3.13 and **314 passed** in an isolated Python 3.11 environment satisfying the job requirements (`google-adk 1.35.2`, `google-genai 1.75.0`).
- Isolated `python deploy/skill_evolution_job/main.py --test`: **SELF-TEST PASS**, with the real engine and 16 registered tools. Registry and host hooks were unconfigured for this smoke check.
- `bash -n`, deployment help/removed-flag checks, and `git diff --check`: passed.
- Exact-head GitHub CI is green, including Python 3.10–3.14, formatting, package build, and configurator smoke. [CI run](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/34404227800).

The review included independent correctness, testing, reliability, maintainability, and adversarial passes, plus separate validation of finding #1. No live Cloud Run execution, container build, or host/model calls were performed. The worktree remains unchanged at the reviewed commit.

**Requested change: correct the P2 timeout contract above before approval.**

caohy1988 (approved):
REVIEW_VERDICT: APPROVE

Independent full review at HEAD `6848228bd9adce725fac51fa07b48de9db6e9392` (matches PR head). Read the complete diff plus surrounding code in `config.py`, `engine.py`, `evolve.py`, `main.py`, `deploy.sh`, `scripts/skill_evolution.py` (the engine's timeout machinery), and all touched tests. Verified claims by execution, not by reading the PR description.

## Verified by running (not trusting the PR body)

- **Env contract** (`config.py:88-110`): `''`→`600.0`, `'0'`→`None`, `'12.5'`→`12.5`, `'-1'`/`'nan'`/`'inf'`/`'abc'`→`ValueError`. Blank string falls back to the default via `_env`'s strip semantics — correct.
- **Failure mode**: `ANALYST_TIMEOUT_S=abc python -m skill_evolution_job.main --test` exits 1 with `ValueError: ANALYST_TIMEOUT_S must be a non-negative number of seconds...` — fails fast at `get_config()` as documented, not silently unbounded.
- **Self-test**: `main.py --test` → `SELF-TEST PASS` (with `google-adk` installed). The new engine section (`load_engine()` + `assert callable(module.evolve_skill)`) works.
- **`bash -n deploy.sh`**: clean.
- **Tests** (py3.13 venv, dev+llm+improvement deps): all PR-touched test files pass — `test_skill_evo_job_engine.py` + `test_skill_evo_job_orchestrator.py` = 44/44, including the new keyword-contract test and the 600.0/30.0/None forwarding tests. Across the full `-k skill_evo` set I get 17 failures, but I ran the **same suite on the parent commit `716aa0e` in the same environment: identical failure set** — all pre-existing environmental issues in engine tests this PR doesn't touch, zero regressions introduced.
- **Engine contract is real**: `scripts/skill_evolution.py:1379` accepts `analyst_timeout_s` and the fleet implements quarantine + budgeted slot handoff (`collect_patches`, `_FleetSlots`), so a hung host `error_analyst` genuinely hands its slot to the next analyst — the PR's core claim holds against the actual engine code, and the new `test_real_engine_keyword_contract` correctly converts a future missing-kwarg into a build break rather than a runtime degrade.

## Compat removal — checked for stragglers

- Repo-wide grep: no remaining references to `evolve_skill_compat`, `supported_kwargs`, or `--scripts-dir` anywhere except the CHANGELOG entry describing their removal.
- `deploy.sh`: flag, `SCRIPTS_DIR` var, help text, and the conditional staging/logging all removed; the staging loop still validates `skill_evolution.py` / `quality_report.py` / `eval/eval_config.json` with a clear error. Simplified `SCRIPTS_SRC` resolution is correct.
- `_resolve_error_analyst` (`evolve.py:329`) is coherent as a single-gate resolver; the host-side degradation story (no hook → single-pass analysts, no score hook → size-based selection) matches the README rewrite, which is now truthful where the old "single-pass fallback when the engine lacks the hooks" text was not.
- The selection-regression reparameterization (`older_engine_contract` → `engine_rescores_incumbent`) preserves the scenario's semantics: pop `incumbent_score`, engine re-scores V0 itself, same-baseline invariant still asserted. This is a rename of the test's *meaning*, not a weakening.
- Timeout flows through the coevolve path too (`coevolve.py` calls `evolve.evolve`), so `ANALYST_TIMEOUT_S` covers every mode the job runs.

## Findings

**P0:** none. **P1:** none.

**P2 (optional, non-blocking):**
1. `deploy.sh` does not forward `ANALYST_TIMEOUT_S` into the Cloud Run job's `--set-env-vars`, so operators tune it via `gcloud run jobs update` after deploy. This is consistent with every other tuning knob in the README table (`EVOLUTION_TOOLBOX`, `GATE_POLICY`, etc. are also not forwarded), so it's a convention note, not a defect — but a one-line README remark ("set via `gcloud run jobs update --set-env-vars`") would save an operator a debugging round-trip.
2. `main.py --test` no longer prints the engine's accepted kwargs (old behavior listed them); the keyword-contract test covers the contract, so this is only a slight loss of self-test observability.
3. Micro-nit: `_env_timeout`'s `default or None` would silently turn a hypothetical future `default=0.0` into `None` — harmless today, and arguably the intended semantics.

## Conclusion

The wiring is correct end-to-end (env → config → `evolve_skill` → fleet quarantine), the removal rationale (image and engine ship together) is sound and the cleanup is complete with no dangling references, and test coverage actually improved: the contract is now asserted against the real engine instead of exercised against simulated old engines. Fail-fast-on-malformed is the right call for a timeout an operator explicitly set. Approve.

PR #484, opened 2026-09-18
chore: Update quickstart sample

sararob:


haiyuan-eng-google:
thanks @sararob for the contribution, can you help star the repo first and would take a look for the PR soon

README.md
```diff
@@ -94,11 +94,18 @@ pip install bigquery-agent-analytics[langsmith]
...
-from bigquery_agent_analytics import Client
+from bigquery_agent_analytics import Client, TraceFilter
 
-client = Client(project_id="my-project", dataset_id="analytics")
-trace = client.get_trace("trace-abc-123")
-trace.render()
+client = Client(
+  project_id="my-project",
+  dataset_id="analytics",
+  table_id="my_table",
```
caohy1988: `table_id` defaults to `"agent_events"`, which is also what the Prerequisites (ADK exporter) and the LangSmith example below (`my-project.analytics.agent_events`) assume. With `"my_table"`, `Client()` only logs a schema warning (`_verify_schema` in client.py) and `list_traces()` then raises `NotFound` for `my-project.analytics.my_table`. Suggest dropping the kwarg, or using `"agent_events"` as SDK.md does.

README.md
```diff
@@ -94,11 +94,18 @@ pip install bigquery-agent-analytics[langsmith]
...
+  location="my-location",
```
caohy1988: `location` is optional (`None` falls through to the BigQuery client default), and `"my-location"` is not a valid BigQuery location, so a copy-paste that keeps the placeholder fails on the first query with an invalid-location error. Suggest omitting it, or using a real example like `"US"` to match SDK.md.

README.md
```diff
@@ -94,11 +94,18 @@ pip install bigquery-agent-analytics[langsmith]
...
+)
+
+traces = client.list_traces(TraceFilter(limit=1))
+if traces:
```
caohy1988, line 108: Nit: with no `else`, an empty or brand-new table makes the quickstart exit with no output, so a first-time user gets no signal that credentials and the table are wired correctly. An `else: print("No traces found")` covers it.

caohy1988 (commented):
Thanks for replacing the made-up trace id with a runnable sample. I ran the new snippet's pure-Python parts against the SDK: every call is API-correct (`Client(project_id, dataset_id, table_id, location)`, `TraceFilter(limit=1)`, `list_traces()`, `Trace.render()`). Three inline suggestions on the placeholder values, plus one follow-up outside the diff:

The package docstring's `Quick start::` block in `src/bigquery_agent_analytics/__init__.py` (around line 30) still shows the old `client.get_trace("trace-123")` sample this PR replaces, so `help(bigquery_agent_analytics)` and the README now disagree. Worth mirroring the new `list_traces` snippet there in this PR or a follow-up.

caohy1988:
Pushed `fb75a4f` onto this branch (maintainer edits were enabled) with the four items from my review: dropped the `table_id` and `location` placeholders in favour of the defaults with a comment on when to set each, added the `else` branch for an empty table, and mirrored the new `list_traces` sample in the package docstring. Verified the README snippet parses and that `Client`, `TraceFilter`, `list_traces` and `insights` resolve with `PYTHONPATH=src`. Feel free to amend or revert if you'd rather keep the explicit kwargs.

README.md
```diff
@@ -94,11 +94,21 @@ pip install bigquery-agent-analytics[langsmith]
...
-from bigquery_agent_analytics import Client
+from bigquery_agent_analytics import Client, TraceFilter
+
+client = Client(
+  project_id="my-project",
+  dataset_id="analytics",
+  # table_id defaults to "agent_events", the table the ADK exporter writes.
+  # location defaults to the BigQuery client's own; pass one such as "US"
+  # only if your dataset lives elsewhere.
+)
```
caohy1988, line 106: **P2 — Update the documentation tests for this new quickstart.** Both `TestGetTraceDocsUseTraceId.test_readme_uses_trace_id` and `TestDocsConsistency.test_init_py_quick_start_uses_trace_id` in `tests/test_pr16_fixes.py` still assert that these files contain `get_trace("trace-`. Replacing those examples makes both tests fail deterministically; I reproduced this locally, and the [Python 3.11 CI log](https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/actions/runs/35397173991/job/105768574355) shows the same failures. Please update the tests in this PR to validate the new discover-and-render examples, retaining the trace-ID-vs-session-ID checks wherever `get_trace()` remains documented. A small smoke test that extracts the actual snippets and exercises empty/populated results with cloud I/O mocked would also catch API drift without locking the docs to a fabricated ID.

caohy1988 (commented):
Reviewed both changed quickstarts at `fb75a4f`. The README example works for both empty and populated results using the real SDK with cloud boundaries mocked, and the earlier table/location/empty-result suggestions are addressed.

One fix is needed before this is ready: update the two documentation regression tests that still require the removed `get_trace("trace-...")` examples. I also left an optional cleanup for an existing invalid `insights` call in the adjacent module example.

Validation: targeted documentation tests **2 failed / 4 passed**; Client initialization/list tests **4 passed**. The Python 3.11 CI job independently reports **2 failed / 5,176 passed / 80 skipped** with those same two assertions; all five Python matrix jobs are currently red. No live BigQuery calls were made.

src/bigquery_agent_analytics/__init__.py
```diff
@@ -29,11 +29,12 @@
...
 
-    from bigquery_agent_analytics import Client
+    from bigquery_agent_analytics import Client, TraceFilter
 
     client = Client(project_id="my-project", dataset_id="analytics")
-    trace = client.get_trace("trace-123")
-    trace.render()
+    traces = client.list_traces(TraceFilter(limit=1))
+    if traces:
+        traces[0].render()
```
caohy1988, line 37: **Optional adjacent cleanup (pre-existing, not introduced by this PR):** running this entire `Quick start::` block reaches `client.insights(max_sessions=50)` below and raises `TypeError: Client.insights() got an unexpected keyword argument 'max_sessions'`. I reproduced this with both empty and populated trace results. `max_sessions` belongs to `InsightsConfig`, while `Client.insights()` accepts `config`. Since the default is already 50, the smallest repair is `report = client.insights()`; alternatively import `InsightsConfig` and pass `config=InsightsConfig(max_sessions=50)`. Fixing it while updating this block would make the complete copy-paste example API-correct.

README.md
caohy1988: Fixed in 25e515c. `test_readme_uses_trace_id` and `test_init_py_quick_start_uses_trace_id` now assert the `list_traces(TraceFilter(limit=1))` sample and that no fabricated trace or session id remains; the SDK.md checks are untouched since SDK.md still documents `get_trace`. Added `TestQuickStartSnippetsRun`, which extracts both snippets from the files and executes them as written with `make_bq_client`, `_verify_schema`, `list_traces` and `insights` patched at the SDK boundary: an empty table prints "No traces found" (README), a populated one renders the first trace (both), and the filter asks for exactly one trace (both). `tests/test_pr16_fixes.py` is 48 passed locally; the one remaining local failure (`test_compute_drift_with_embedding_model`) fails identically on the untouched tree and is not in the CI failure list, so it is my environment.

src/bigquery_agent_analytics/__init__.py
caohy1988: Fixed in 25e515c: the docstring now calls `client.insights()` with a comment that `max_sessions` defaults to 50 and lives on `InsightsConfig`. The new docs test also asserts `insights(max_sessions=` is absent so it cannot come back, and the snippet smoke test executes the whole block including the `insights()` call.

tests/test_pr16_fixes.py
```diff
@@ -781,6 +794,68 @@ def test_default_table_id_is_agent_events(self):
...
+    report = MagicMock()
+    report.summary.return_value = "insights summary"
+    with (
+        patch(
+            "bigquery_agent_analytics.client.make_bq_client",
+            return_value=MagicMock(),
+        ),
+        patch.object(Client, "_verify_schema", return_value=None),
+        patch.object(Client, "list_traces", return_value=traces) as listed,
+        patch.object(Client, "insights", return_value=report),
```
caohy1988: **P2 — Preserve the public method signatures in these mocks.** Without `autospec=True`, these replacements accept arguments that the SDK rejects, so the new smoke tests miss the API regression they are meant to catch. I changed only the extracted package snippet in memory to `client.insights(max_sessions = 50)`: all seven changed/new documentation checks still passed (the spacing also bypasses the literal-string guard), while the real `Client.insights` raised `TypeError`. Please add `autospec=True` to the `insights` and `list_traces` patches and update the listing assertion to account for the bound `self`; a negative check for this unsupported keyword would verify the protection. The current documented snippets are valid; this is a gap in the new regression tests.

caohy1988 (commented):
Fresh review of the complete diff at `25e515c`: the earlier documentation-test failures and invalid `insights` call are fixed. Both current quickstarts execute successfully with the real public SDK methods and mocked cloud I/O.

I found one remaining, nonblocking test issue: the new smoke tests accept unsupported API arguments because the public methods are replaced with unrestricted mocks. The inline comment includes a reproduced example and a small fix.

Validation: **127 passed** across `tests/test_pr16_fixes.py` and `tests/test_sdk_client.py`; the current Python 3.10–3.14 CI matrix, formatting, browser smoke test, and package build are green. The Python 3.11 CI job reports **5,183 passed / 80 skipped**. No live BigQuery or model calls were made.

tests/test_pr16_fixes.py
caohy1988: Fixed in a72234d. `list_traces`, `insights` and `_verify_schema` are patched with `autospec=True`, the listing assertion now unpacks the bound instance autospec passes through, and a new `test_mocks_keep_the_public_signatures` rewrites the docstring snippet to `client.insights(max_sessions = 50)`, the spacing the literal guard misses, and requires `TypeError`. I watched that test fail against the permissive mocks (`DID NOT RAISE TypeError`) before adding autospec. `tests/test_pr16_fixes.py` is 49 passed locally.

caohy1988 (approved):
LGTM at `a72234d`. The latest commit closes the remaining test-fidelity finding: both public method mocks now preserve their signatures, the assertions handle the bound instance, and the negative regression rejects the unsupported `max_sessions` keyword. Earlier review findings are also addressed.

Validation: **128 passed** across `tests/test_pr16_fixes.py` and `tests/test_sdk_client.py` on this exact head, including the new negative regression. Current CI is green. No blocking findings remain. No live cloud or model calls were made.

PR #486, opened 2026-09-18
fix: README onboarding and EventsTableNotFoundError (#485)

### LINKED ISSUE #485: docs: README Prerequisites link is 404 and Quick Start fails on non-default table_id/location
## Summary

A new user following `README.md` top-to-bottom hits two blockers before the first successful `trace.render()`:

1. **Prerequisites link is dead.** README "Prerequisites" points to
   `https://github.com/google/adk-python/tree/main/contributing/extensions/bigquery_trace_exporter`,
   which returns **404** (the whole `contributing/extensions/` directory no longer exists in adk-python).
   The plugin now lives in core ADK and is documented at
   https://adk.dev/observability/bigquery-agent-analytics/ — which the README already links in its first paragraph.

2. **Quick Start silently assumes `table_id="agent_events"` and the client's default location.**
   `Client(project_id=..., dataset_id=...)` → `client.get_trace("...")` fails for any user whose events table has a
   different name or whose dataset is outside the default location. The only visible output before the traceback is
   the unrelated, swallowed warning
   `Schema verification failed: ... Continuing without verification.`
   (`Client._verify_schema` catches and logs a bare `Exception`), which is easy to misread as an auth problem.
   Passing `table_id=` and `location=` explicitly — as `SDK.md` §1 already does — fixes it.

Additionally, the Quick Start hard-codes a placeholder `trace_id` (`"trace-abc-123"`); users are not shown an in-SDK
way to discover one even though `Client.list_traces()` exists (`client.py` L1906, documented in `SDK.md` §2).
PR #484 (open) addresses this last point but not (1) or (2).

## Repro

```python

client = Client(project_id="my-project", dataset_id="analytics")  # table is not named agent_events
trace = client.get_trace("<real trace_id copied from BQ>")
# WARNING  Schema verification failed: <google-auth metadata-server error>. Continuing without verification.
# Traceback ... in get_trace -> self.bq_client.query(query, job_config=job_config).result()
```

Adding `table_id="<my_table>", location="<my_location>"` to the constructor makes the same call succeed.

## Proposed fixes

- [ ] **README Prerequisites** — replace the 404 URL with `https://adk.dev/observability/bigquery-agent-analytics/` (one-line fix).
- [ ] **README Quick Start** — show `table_id=` and `location=` (or `table_id="auto"`) in the constructor, and discover a trace via `list_traces(TraceFilter(limit=1))` instead of a placeholder id. Could land as an extension of #484.
- [ ] **`Client._verify_schema`** — either demote the catch-all warning to `debug`, or have `get_trace` raise a typed error that names `table_id`/`location` when the underlying query fails with `NotFound`, so the swallowed warning is not the only signal a user sees.

## References

- `README.md` — Prerequisites (broken link), Quick Start snippet
- `src/bigquery_agent_analytics/client.py` — `Client.__init__` (`table_id: str = "agent_events"`, `location: Optional[str] = None`), `_verify_schema`, `get_trace`, `list_traces` (L1906)
- `SDK.md` §1 (constructor with `table_id`/`location`), §2 (`list_traces` with `TraceFilter`)
- #484

caohy1988:
Closes #485.

## What

A new user following `README.md` top to bottom hit two blockers before the first `render()`: the Prerequisites link returned 404, and the Quick Start silently assumed `table_id="agent_events"` and the client's default location, so a read against any other table or region failed with a bare `NotFound` whose only earlier signal was the swallowed schema-verification warning.

**README**
- A **"How this fits together"** block replaces the dead link: producers (the ADK `BigQueryAgentAnalyticsPlugin` and the LangChain/LangGraph `BigQueryCallbackHandler`, both writing the same `agent_events` schema) and this SDK as the read-only consumer. This is item C from the issue's follow-up comment as well as the broken-link item.
- The Quick Start passes `table_id="agent_events"` and `location="US"` explicitly, each with a one-line comment on when to change it, and says what happens when they are wrong.

**Client**
- `get_trace()` and `list_traces()` route reads through `_read_rows()`, which re-raises `google.api_core.exceptions.NotFound` as **`EventsTableNotFoundError`**: a `NotFound` subclass (existing `except NotFound` handlers keep working) whose message names the fully qualified table and the location and says to pass `table_id=` and `location=`. Other errors pass through unchanged. Exported at the package top level.
- `_verify_schema`'s catch-all warning now names the table, the exception type and the same fix, so it no longer reads like an auth problem.

**CHANGELOG**: Unreleased / Fixed entry.

## Not in scope

Items A and B in the issue's follow-up comment belong to `google/adk-python` and the LangChain docs respectively.

## Test plan

- [x] Tests written first and watched fail: `TestEventsTableNotFound` (5) failed on the missing import; `TestReadmeOnboarding` (4) failed on the README content.
- [x] `pytest tests/test_sdk_client.py tests/test_pr16_fixes.py` — 134 passed. Three failures are environment-only and reproduce identically on the untouched tree (`test_compute_drift_with_embedding_model`, two `TestFetchSessionMetadata` cases); CI on the same head is green for them.
- [x] `pyink --check` clean on the four Python files; `git diff --check` clean.
- [x] Rebased onto `main` after #484 merged; the PR is one commit.


src/bigquery_agent_analytics/client.py
```diff
@@ -2026,9 +2061,7 @@ def _fetch_filtered_traces(
       ]
       job_config = bigquery.QueryJobConfig(query_parameters=attempt_params)
       job_config = with_sdk_labels(job_config, feature=feature)
-      results = list(
-          self.bq_client.query(query, job_config=job_config).result()
-      )
+      results = self._read_rows(query, job_config)
```
caohy1988: **P2 — Pass the actual queried table into `_read_rows()`.** This shared fetch also serves performance evaluation, the LLM-judge fallback, and contextual categorical evaluation, which accept a `dataset=` table override. I reproduced `client.evaluate(PerformanceEvaluator(), dataset="golden_events")` with `table_id="agent_events"`: the SQL queries `proj.ds.golden_events`, but the new exception sets `table_ref` to `proj.ds.agent_events` and tells the user to change the constructor. The default table may exist and be unrelated to the failure. Please pass the effective source table through the helper, use it in the diagnostic, and add a regression with a non-default evaluation table.

src/bigquery_agent_analytics/client.py
```diff
@@ -640,10 +663,22 @@ def _verify_schema(self) -> None:
...
+          self._table_ref,
+          type(e).__name__,
           e,
       )
 
+  def _read_rows(self, query: str, job_config: bigquery.QueryJobConfig) -> list:
+    """Runs a read, naming the table and location if it is not found."""
+    try:
+      return list(self.bq_client.query(query, job_config=job_config).result())
+    except gcp_exceptions.NotFound as e:
```
caohy1988, line 703: **P2 — Do not classify every BigQuery `NotFound` as a missing events table.** This catch covers job polling and result retrieval as well as query submission. BigQuery also returns `NotFound` for a missing job or a job-location mismatch ([error reference](https://docs.cloud.google.com/bigquery/docs/error-messages#errortable)). With a real `QueryJob.result()` and only its reload boundary mocked to raise `NotFound("Not found: Job proj:US.query_job_123")`, `list_traces()` now raises `EventsTableNotFoundError` asserting that `proj.ds.agent_events` is missing. Changing `table_id` cannot fix that failure, and callers handling the new specific exception receive the wrong category. Translate only identified source-table/location failures and preserve unrelated `NotFound` exceptions, or use a generic read-failure type/message that does not assert that the events table is absent. Please cover job-not-found separately from table-not-found.

src/bigquery_agent_analytics/client.py
```diff
@@ -478,6 +479,28 @@ def _run_sync(coro):
...
+  match what the producer wrote, so the message names both and says
+  what to pass. It is still a ``NotFound``, so existing handlers that
+  catch that keep working.
+  """
+
+  def __init__(self, table_ref: str, location: Optional[str], cause: Exception):
+    self.table_ref = table_ref
+    self.location = location
+    where = f"location={location}" if location else "location=client default"
+    super().__init__(
```
caohy1988, line 503: **P2 — Preserve the original structured exception metadata.** Calling the superclass with only a message discards the original `NotFound.errors`, `response`, and `details` (and structured reason information). I compared the same populated exception on base and head: those fields survive on base but become `[]`, `None`, and `[]` after this wrapper. Existing `except NotFound as e` handlers still match, but handlers inspecting the standard fields lose the data they previously received. The chained cause remains available, but existing callers would need rewriting to use it. Please forward or delegate the original structured metadata while adding the guidance, retain `raise ... from e`, and add a test using a populated exception rather than a message-only mock.

caohy1988 (commented):
Fresh solo review of the full diff at `db20516`. I found three P2 regressions in the new exception handling and recommend addressing them before merge; each inline comment includes a reproduction and a suggested fix.

Validation: **347 local tests passed** (137 in the two changed test files and 210 related identity/client/telemetry regressions). Current CI is green; Python 3.11 reports **5,193 passed / 80 skipped**. The actual README quickstart works for empty and populated results using real SDK methods with cloud boundaries mocked, and the documentation links resolve. No live BigQuery or model calls were made.

src/bigquery_agent_analytics/client.py
caohy1988: Fixed in e0c04f9. `_read_rows()` now takes the table the query actually reads (`table=table` from `_fetch_filtered_traces`, so every `dataset=` override flows through) and builds the diagnostic from that, not from the constructor default. Regression `test_override_table_is_the_one_named` reproduces your case, `evaluate(PerformanceEvaluator(), dataset="golden_events")` with `table_id="agent_events"`, and asserts the message and `table_ref` name `proj.ds.golden_events` and not `proj.ds.agent_events`.

src/bigquery_agent_analytics/client.py
caohy1988: Fixed in e0c04f9. Only a `NotFound` whose message names the queried table or its dataset is translated (`Table proj:ds.table` or `Dataset proj:ds`, colon or dotted form, case-insensitive, via `_not_found_names_source`); anything else is re-raised untouched. `test_job_not_found_is_not_translated` raises `NotFound("Not found: Job proj:US.query_job_123")` from the read and asserts the caller receives the identical original object, not the typed error; `test_missing_dataset_is_translated` covers the dataset case separately from the table case.

src/bigquery_agent_analytics/client.py
caohy1988: Fixed in e0c04f9. The wrapper now forwards the original `errors`, `details`, `response` and error info to the `NotFound` constructor (guarded so a non-API cause still works) and keeps `raise ... from e`. `test_structured_metadata_is_preserved` builds a populated `NotFound` with `errors`, `details` and a response object rather than a message-only mock, and asserts `errors`, `details`, `response` identity, `code` and `__cause__` all survive on the wrapped exception.

src/bigquery_agent_analytics/client.py
```diff
@@ -640,10 +672,57 @@ def _verify_schema(self) -> None:
...
+    """Whether a NotFound is about the queried table or its dataset.
+
+    BigQuery spells the resource as ``Table project:dataset.table`` or
+    ``Dataset project:dataset`` in the message; the dotted form is
+    accepted too.
+    """
+    message = str(error).lower()
+    for sep in (":", "."):
+      dataset_ref = f"{self.project_id}{sep}{self.dataset_id}".lower()
+      if f"table {dataset_ref}.{table}".lower() in message:
```
caohy1988: **P2 — Match complete resource identifiers before translating `NotFound`.** The substring checks also match different resources whose names begin with the configured name. With `table_id="agent_events"`, I reproduced `list_traces()` wrapping `NotFound("Not found: Table proj:ds.agent_events_archive was not found in location US")` as `EventsTableNotFoundError(table_ref="proj.ds.agent_events")`; `Dataset proj:ds_archive` is likewise mistaken for `proj.ds`. This can misdiagnose a missing dependency when the configured source is a view: the source exists, but the new exception says it is absent and recommends changing its configuration.

Lowercasing the entire identifier also conflates `agent_events` with `AGENT_EVENTS`, which can be different tables in a [case-sensitive BigQuery dataset](https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/lexical#case_sensitivity). Please extract and compare the complete Table/Dataset resource token, respecting identifier case, and pass through errors naming other resources. Add negative cases for table/dataset prefix collisions and case-distinct names alongside the existing exact-name and missing-job controls. The reproductions above used the real public SDK read method with the cloud failure mocked; no live BigQuery call was made.

caohy1988 (commented):
Fresh solo review of the complete diff at `e0c04f98148e64df9763eb3fb1984065bb457d4b` against `71c6ea37c0c8493e676c415008aa79de90fc262f`.

The original three reproductions now pass: evaluation overrides identify the actual table, missing-job errors retain the original exception, and structured error metadata survives. One narrower P2 remains in the newly added source matcher, detailed inline; I recommend fixing it before merge. No P0/P1 findings.

Validation: 351 focused tests passed across the client, README regressions, trace identity, labels, PR123 regressions and telemetry suites. Additional probes covered query/result/pagination failures, overrides, structured ErrorInfo, Forbidden passthrough, and the actual QueryJob.result() missing-job path with its reload boundary mocked. Current CI is green; the Python 3.11 full-suite log reports 5,197 passed and 80 skipped. No live BigQuery or model calls, and no repository changes.

src/bigquery_agent_analytics/client.py
caohy1988: Fixed in `340eca0`.

`_not_found_names_source` now extracts complete `Table` / `Dataset` resource tokens and compares them with case-preserving exact equality (no substring / no `.lower()` on identifiers). Negative tests cover `agent_events_archive`, `ds_archive`, and `AGENT_EVENTS`; job NotFound still passes through.

Intent/spec/plan for this residual slice: `intent/pr486-not-found-names/`.

caohy1988 (commented):
Second-seat verdict: APPROVE (posted as COMMENT because this identity cannot approve its own PR; blind review, Fable off review as author).

Verified residual P2 from discussion_r4052121395 against intent/pr486-not-found-names/{intent,spec,plan}.md at cc78a6b (format-only delta over 340eca0, semantics identical):

- Complete token match: Client._not_found_names_source extracts the full token after Table/Dataset via regex and compares by exact set membership against both colon and dotted expected forms. No .lower(), no substring in.
- Case respect: message kept as-is; (?i) applies only to the Table/Dataset keyword, token comparison is exact. New test_case_distinct_table_name_is_not_translated (AGENT_EVENTS) asserts identity passthrough.
- Prefix-collision negatives: new test_table_prefix_collision_is_not_translated (agent_events_archive) and test_dataset_prefix_collision_is_not_translated (ds_archive) both assert identity passthrough (excinfo.value is original, not EventsTableNotFoundError).
- Job passthrough: existing test_job_not_found_is_not_translated green; Job messages match neither regex.

Verification: python3 -m pytest tests/test_sdk_client.py::TestEventsTableNotFound -q -> 12 passed.

Non-blocking nit: no positive test exercises the dotted project.dataset.table form (existing positives use colon form); implementation handles it via the expected sets. Consider adding one. No merge performed.

src/bigquery_agent_analytics/client.py
```diff
@@ -640,10 +673,71 @@ def _verify_schema(self) -> None:
...
+        f"{self.project_id}:{self.dataset_id}.{table}",
+        f"{self.project_id}.{self.dataset_id}.{table}",
+    }
+    expected_datasets = {
+        f"{self.project_id}:{self.dataset_id}",
+        f"{self.project_id}.{self.dataset_id}",
+    }
+    # Capture the complete resource token after Table/Dataset; stop at
+    # whitespace or common trailing punctuation in BQ error text.
+    for match in re.finditer(r"(?i)(?:^|[\s])Table\s+([^\s,;]+)", message):
```
caohy1988: [P2] Preserve spaces inside the complete table identifier

`[^\s,;]+` still truncates valid table identifiers: BigQuery permits spaces in table names ([table naming rules](https://docs.cloud.google.com/bigquery/docs/tables#table_naming)). With the configured source `proj.ds.agent_events`, I reproduced that a mocked `NotFound("Not found: Table proj:ds.agent_events archive was not found in location US")` becomes `EventsTableNotFoundError(table_ref="proj.ds.agent_events")`, rather than preserving the original error for the different missing resource. This retains the wrong-source diagnostic when, for example, the configured view depends on `agent_events archive`. Conversely, configuring `table_id="agent_events archive"` leaves its exact missing-table error untranslated.

Please extract the full resource name using the surrounding BigQuery error structure (or conservatively pass through ambiguous formats), and add both a space-containing prefix-collision negative and an exact space-containing-name positive. The underscore-prefix and case-distinct fixes already pass.

caohy1988 (commented):
Astra blind review: COMMENT — one residual P2; no P0/P1 found.

Reviewed 340eca0 and the formatting-only follow-up cc78a6b (current head; identical client.py AST). The requested underscore-prefix and case-distinct negatives are fixed, and job NotFound identity is preserved. A remaining table-name-with-spaces collision is described inline, so complete resource-name matching is not fully closed.

Validation:
- 16 focused NotFound/onboarding tests passed.
- Broader client/onboarding run: 142 passed, 2 failed. Both failures reproduce on base 71c6ea3 and are local ADK/OpenTelemetry import incompatibilities (missing GEN_AI_TOOL_DEFINITIONS), not PR regressions.
- 48 independent mocked checks passed across colon/dotted references, exact/prefix/case variants, job passthrough, query/result/iteration error stages, and structured metadata/error_info preservation.
- Skimmed the README, exported exception, schema warning, and evaluation table override. Producer documentation links resolve. Format check is now green on cc78a6b; the CI test matrix was still running when checked.

No implementation changes or merge performed.

src/bigquery_agent_analytics/client.py
caohy1988: Fixed in `518fb5c`.

`_not_found_names_source` now captures the full `Table` / `Dataset` resource through the BQ terminator ` was not found` (case-insensitive), so spaces inside identifiers are preserved and compared with exact equality. Negative `test_table_name_with_spaces_suffix_is_not_translated` covers `agent_events archive` vs configured `agent_events` (identity-preserving re-raise); positive `test_table_name_with_spaces_exact_is_translated` covers configuring `table_id` with spaces. Prefix / case-distinct / job NotFound controls remain green.

Intent/spec/plan updated under `intent/pr486-not-found-names/`.

caohy1988 (commented):
Astra blind re-review — **APPROVE for the residual space-token P2 only** at `518fb5ce83f82e0ca04cfd7cfefa91ee6580a5ef`.

The finding in https://github.com/GoogleCloudPlatform/BigQuery-Agent-Analytics-SDK/pull/486#discussion_r4057920520 is closed within the requested scope. `_not_found_names_source` now captures Table/Dataset identifiers through the `was not found` terminator, preserving embedded spaces and case before comparison with the expected resource names.

Verified:
- Configured `agent_events` + missing `agent_events archive` re-raises the identical original `NotFound`.
- Configured `agent_events archive` + that exact missing name translates to `EventsTableNotFoundError`.
- Prior underscore-prefix, case-distinct, and job-NotFound regressions remain green.

Validation: `python3 -m pytest tests/test_sdk_client.py::TestEventsTableNotFound -q -p no:cacheprovider` — **14 passed**. An independent hermetic matrix passed **56 cases** across colon/dotted forms, query/result failure stages, Table/Dataset parsing (including synthetic dataset-space cases), missing terminators, and case-preserved identity checks. The predecessor parser reproduced both space-name failures.

No remaining finding in this P2. Reviewed independently against the code and intent/spec/plan; no other review was read and no other reviewer was summoned. This is scoped approval, not a full-PR re-review. Submitted as COMMENT because the authenticated GitHub account is the PR author. No merge performed.

caohy1988 (commented):
Blind re-review (second dual-LGTM seat) of HEAD 518fb5c against intent/pr486-not-found-names/{intent,spec,plan}.md. Scope: Client._not_found_names_source + TestEventsTableNotFound only. (Tried APPROVE; GitHub blocks self-approve on this account, so recording this as COMMENT with explicit APPROVE intent.)

1. Complete token match, case respected: impl builds expected colon/dotted table+dataset tokens and compares with exact equality (no .lower(), no substring). Prefix negatives (agent_events_archive, ds_archive) and case-distinct negative (AGENT_EVENTS) all miss the expected sets -> passthrough. Holds.

2. Spaces preserved via BQ terminator: Table/Dataset tokens captured with (.+?) through case-insensitive ' was not found', replacing the old [^\s,;]+ truncation. Space-suffix negative ('agent_events archive' vs configured 'agent_events') misses -> passthrough; exact-with-spaces positive (configured 'agent_events archive') hits -> translates. Holds.

3. Job NotFound passthrough: 'Not found: Job ...' matches neither Table nor Dataset pattern -> re-raised as the identical original exception (existing test_job_not_found_is_not_translated asserts identity). Holds.

Verification: repo suite could not execute here (only pytest available is on Python 3.8, which cannot parse pre-existing 3.9+ syntax at tests/test_sdk_client.py:1777, untouched by this diff). Instead ran the method body verbatim in a standalone harness: 11/11 spec cases pass (exact colon/dotted, both prefix negatives, case negative, space-suffix negative, exact-with-spaces positive, job passthrough, dataset positive, override-table positive, terminator-less ambiguous negative).

No scope drift: diff touches only client.py matching logic (+re import, docstring) and the five specified negative/positive tests. Verdict: APPROVE (intent; recorded as comment due to self-approve block). No merge performed.

haiyuan-eng-google (approved):

